PR #2, opened 2019-08-13
Implement project sub navigation

xla:
As an exercise and we have to support multiple functional contexts within a project this change explores how a sub navigation with associated routes can look like. It leaves the main composition to the App component and enabls differnt sub pages with a prop passed to a page. Along the way it entirely hides the internal use of ReasonReactRouter in the Router module by exposing `currentPage` and instead of returning a raw string, that can be used as a link, it returns a function that can be invoked when the new pushState should take effect.

* implement Project component
* use props to enable Project sub page routing
* show simple project navigation which shows active "tab"
* use `currentPage` as the main routing interface
* move from raw strings to `navigate` functions to be used by components
which want to build links

### DIFF
--- a/app/src/lib/Router.rei
+++ b/app/src/lib/Router.rei
@@ -1,8 +1,24 @@
+/** Routable pages which relate to larger context of features. **/
 type page =
   | Root
   | Projects
+  | Project(string)
+  | ProjectCode(string)
+  | ProjectFunds(string)
   | NotFound(list(string));
 
-let linkOfPage: page => string;
+/** Reads the current url and return a matching page, or NotFound. **/
+let currentPage: unit => page;
+
+/** Given a page returns a function which nvigates to it by pushing a new url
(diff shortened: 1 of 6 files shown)


app/src/lib/Router.rei
```diff
@@ -1,8 +1,24 @@
...
+  | Project(string)
+  | ProjectCode(string)
+  | ProjectFunds(string)
   | NotFound(list(string));
 
-let linkOfPage: page => string;
+/** Reads the current url and return a matching page, or NotFound. **/
+let currentPage: unit => page;
+
+/** Given a page returns a function which nvigates to it by pushing a new url
```
rudolfs: nvigates -> navigates

xla:
```suggestion
/** Given a page returns a function which navigates to it by pushing a new url
```

PR #4, opened 2019-08-13
Set up basic styling

rudolfs:
First attempt to implement a basic styling system. We'll add more primitives and functionality in an iterative manner, so the interfaces are still bound to change.

* add color primitives
* add font primitives
* add button atom
* configure webpack to serve static assets in dev mode
* add normalize.css to clean the slate for cross-browser styling


.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
xla: We have a `.gitignore` in the app directory which should avoid node_modules being tracked. Do we need this one on top of it?

app/src/Index.re
```diff
@@ -1 +1,5 @@
+open Css;
+
+global("body", [color(Particle.Color.black()), ...Particle.Font.text]);
```
xla: I'd move this to App as we need those styles no matter where the app is applied. For me `Index.re` is the place which determines that we mount the app for a website served on some url and could conceive a different entry point for something like a desktop native deployment.

app/src/Topbar.re
```diff
@@ -1,6 +1,8 @@
+open Atom;
+
 module JoinNetwork = {
   [@react.component]
-  let make = () => <button> {React.string("Join Network")} </button>;
+  let make = () => <Button style=Button.Styles.primary text="Join Network" />;
```
xla: Instead of having a text prop could we pass a children? Or could that potentially break the button styles/implementation?

.gitignore
rudolfs: I added it because for some reason my language server keeps creating it at the top level (mentioned it in the commit message).

I'll remove it if bothers you and look into configuring my language server to not write the debug.log in there.

app/src/Index.re
rudolfs: Good point, I haven't yet figured out what's the best way to handle global stuff. Will move it into App for now.

app/src/Topbar.re
rudolfs: You're right, I think that's the proper react way to do it.

PR #6, opened 2019-08-15
Add more styling primitives

rudolfs:
We want to be able to apply all of the styling that @juliendonck provided.
* add title atom
* restructure files
  * introduce molecule namespace and add `topbar` and `footer` to it
  * move `link` to atoms
* fix font styling
* fix button styling and make its interface similar to title elements
* remove unneeded navigation elements


app/src/App.re
```diff
@@ -1,11 +1,26 @@
-open Css;
+module Styles = {
+  open Css;
 
-global("body", [color(Particle.Color.black()), ...Particle.Font.text]);
+  global(
+    "body",
+    [
+      color(Particle.Color.black()),
+      unsafe("-webkit-font-smoothing", "antialiased"),
```
xla: The dark side is strong in this one!

app/src/App.re
```diff
@@ -1,11 +1,26 @@
...
+      unsafe("-moz-osx-font-smoothing:", "grayscale"),
+      ...Particle.Font.text,
+    ],
+  );
+
+  global("a", [color(Particle.Color.black()), textDecoration(none)]);
```
xla: Can we add `cursor: pointer` to indicate that they are clickable, trips me up even in local debug. :>

app/src/App.re
```diff
@@ -1,11 +1,26 @@
...
+
+  let app = style([]);
```
xla: Is this intentionally empty?

app/src/App.re
rudolfs: 👹

app/src/App.re
rudolfs: Yeah, forgot to remove it.

PR #10, opened 2019-08-19
Add layout primitives

rudolfs:
Start extracting repeated layout patterns into design system primitives. We'll use these to position and align components on a page as well as apply different sizing to containers.

### DIFF
--- a/app/assets/reset.css
+++ b/app/assets/reset.css
@@ -92,6 +92,7 @@ textarea {
   border: 0;
   font-family: inherit;
   font-size: 100%;
+  font-weight: unset;
   color: inherit;
   font-kerning: normal;
   text-decoration: none;
--- a/app/src/Footer.re
+++ b/app/src/Footer.re
@@ -3,12 +3,12 @@ module Styles = {
 
   let footer =
     style([
+      display(grid),
       gridColumnStart(2),
       gridColumnEnd(8),
       gridRowStart(-1),
       height(px(64)),
       justifySelf(center),
-      display(grid),
     ]);
 
   let ul =
--- a/app/src/Footer.re
+++ b/app/src/Footer.re
@@ -1,6 +1,3 @@
-open Router;
-open Atom;
-
 module Styles = {
   open Css;
 
@@ -10,36 +7,26 @@ module Styles = {
       gridColumnEnd(8),
       gridRowStart(-1),
       height(px(64)),
-      display(`flex),
-      justifyContent(`center),
-      unsafe("placeSelf", "end stretch"),
+      justifySelf(center),
+      display(grid),
     ]);
 
-  let li = style([marginLeft(px(24)), display(inline)]);
+  let ul =
+    style([
+      children([marginRight(px(24)), display(inline), alignSelf(center)]),
+      alignSelf(center),
+    ]);
 };
 
 [@react.component]
 let make = () =>
   <footer className=Styles.footer>
-    <Link page=Root> <Atom.Icon.Logo /> </Link>
-    <ul>
-      <li className=Styles.li>
-        <a href="/"> {React.string("What is oscoin?")} </a>
-      </li>
-      <li className=Styles.li>
-        <a href="/"> {React.string("Maintainers")} </a>
-      </li>
-      <li className=Styles.li>
-        <a href="/"> {React.string("Contributors")} </a>
-      </li>
-      <li className=Styles.li>
-        <a href="/"> {React.string("Supporters")} </a>
-      </li>
-      <li className=Styles.li>
-        <a href="/"> {React.string("Security")} </a>
-      </li>
-      <li className=Styles.li>
-        <a href="/"> {React.string("Privacy")} </a>
-      </li>
+    <ul className=Styles.ul>
+      <li> <a href="/"> {React.string("What is oscoin?")} </a> </li>
+      <li> <a href="/"> {React.string("Maintainers")} </a> </li>
+      <li> <a href="/"> {React.string("Contributors")} </a> </li>
+      <li> <a href="/"> {React.string("Supporters")} </a> </li>
+      <li> <a href="/"> {React.string("Security")} </a> </li>
+      <li> <a href="/"> {React.string("Privacy")} </a> </li>
     </ul>
   </footer>;
--- a/app/src/Footer.re
+++ b/app/src/Footer.re
@@ -12,7 +12,7 @@ module Styles = {
       height(px(64)),
       display(`flex),
       justifyContent(`center),
-      unsafe("place-self", "end stretch"),
+      unsafe("placeSelf", "end stretch"),
     ]);
 
   let li = style([marginLeft(px(24)), display(inline)]);
--- a/app/src/Footer.re
+++ b/app/src/Footer.re
@@ -6,10 +6,13 @@ module Styles = {
 
   let footer =
     style([
+      gridColumnStart(2),
+      gridColumnEnd(8),
+      gridRowStart(-1),
+      height(px(64)),
       display(`flex),
-      justifyContent(`spaceAround),
-      alignItems(`center),
-      height(px(118)),
+      justifyContent(`center),
+      unsafe("place-self", "end stretch"),
     ]);
 
   let li = style([marginLeft(px(24)), display(inline)]);
--- a/app/src/Molecule.re
+++ b/app/src/Molecule.re
@@ -13,6 +13,7 @@ module Styles = {
       padding(px(13)),
       borderBottom(px(1), solid, Color.lightGray()),
       lastChild([borderBottomWidth(px(0))]),
+      hover([backgroundColor(Color.almostWhite())]),
     ]);
 
   let description =
--- a/app/src/Molecule.re
+++ b/app/src/Molecule.re
@@ -1,4 +1,3 @@
-open Source.Project;
 open Atom;
 open Particle;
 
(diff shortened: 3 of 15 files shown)


app/src/Footer.re
```diff
@@ -6,10 +6,13 @@ module Styles = {
...
+      gridColumnStart(2),
+      gridColumnEnd(8),
+      gridRowStart(-1),
+      height(px(64)),
       display(`flex),
-      justifyContent(`spaceAround),
-      alignItems(`center),
-      height(px(118)),
+      justifyContent(`center),
+      unsafe("placeSelf", "end stretch"),
```
xla: This could be done safely with `justifySelf` and `alignSelf`.

app/src/Molecule.re
```diff
@@ -0,0 +1,43 @@
+open Source.Project;
```
xla: Not immediate action needed, but given we want to have molecules part of the design system as library - we should remove this dependency. One way is could be to be explicit about the props the `ProjectCard` needs and just enumerate those explicitly. Examples of how to use `makeProps` for it, can be found in the [componets docs](https://reasonml.github.io/reason-react/docs/en/components#reactcomponent) and [this article](https://khoanguyen.me/writing-reason-react-bindings-the-right-way/). Which will only be needed once the props grow more, for now we can just put it in the make signature.

app/src/Topbar.re
```diff
@@ -32,9 +34,12 @@ module Navigation = {
...
   Router.(
-    <header className=Styles.topBar>
-      <Link page=Root> <Atom.Icon.Logo /> </Link>
-      <Navigation />
-      <JoinNetwork />
+    <header className=Styles.header>
+      <Container.TwoColumns>
+        ...(
+             <> <Link page=Root> <Atom.Icon.Logo /> </Link> <Navigation /> </>,
+             <JoinNetwork />,
```
xla: Is this different from?

``` jsx
<Link page=Root><Atom.Icon.Logo /></Link>
<Navigation />
<JoinNetwork />
```

app/src/atoms/AtomLink.re
```diff
@@ -1,12 +1,23 @@
...
   let content =
     switch (children) {
     | Some(child) => child
     | None => React.string(nameOfPage(page))
     };
 
-  <a onClick={navigateOfPage(page)}> content </a>;
+  let style =
+    switch (style) {
+    | Some(style) => merge([Styles.link, style])
```
xla: Just for my own clarification, is the order of styles passed to merge significant?

app/src/pages/PageProjects.re
```diff
@@ -52,9 +48,13 @@ let make = () => {
   });
 
   <>
-    <div>
-      <Title.Huge> {React.string("Explore")} </Title.Huge>
-      <Button> {React.string("Register project")} </Button>
+    <div className=Styles.projectHeading>
+      <Container.TwoColumns>
+        ...(
```
xla: What does this spread do here?

app/assets/reset.css
```diff
@@ -92,6 +92,7 @@ textarea {
   border: 0;
   font-family: inherit;
   font-size: 100%;
+  font-weight: unset;
```
xla: Is this a custom addition from us to the reset?

app/src/Molecule.re
```diff
@@ -0,0 +1,43 @@
...
+
+  let image =
+    style([marginRight(px(21)), width(px(64)), height(px(64))]);
+};
+
+module ProjectCard = {
+  [@react.component]
+  let make = (~project) =>
+    <li className=Styles.item key={project.address}>
+      <Link style=Styles.link page={Router.Project(project.address)}>
```
xla: I'd assume that a `ProjectCard` is a link is optional. And should be done on the caller side.

app/src/Molecule.re
```diff
@@ -0,0 +1,43 @@
...
+  let link = style([display(`flex)]);
+
+  let image =
+    style([marginRight(px(21)), width(px(64)), height(px(64))]);
+};
+
+module ProjectCard = {
+  [@react.component]
+  let make = (~project) =>
+    <li className=Styles.item key={project.address}>
```
xla: Should we make this a `div` or other container to not assume that it is rendered in a list?

app/src/atoms/AtomLayout.re
```diff
@@ -0,0 +1,21 @@
+module Styles = {
+  open Css;
+
+  let left = style([flex(`num(1.0)), display(`flex)]);
```
xla: Totally valid approach, but can we try to use grids wherever possible to stick to one layout principle. Unless there is strong reason to use flex here.

app/src/atoms/AtomLayout.re
```diff
@@ -0,0 +1,21 @@
...
+  let right =
+    style([flex(`num(1.0)), display(`flex), justifyContent(flexEnd)]);
+
+  let twoColumns = style([display(`flex), flex(`num(1.0))]);
+};
+
+module Container = {
+  module TwoColumns = {
+    [@react.component]
+    let make = (~children) =>
```
xla: This excpects a tuple of children if I understand correctly. Maybe we can reflect that in the type system to know at a glance that this is not just a normal `React.element`.

app/src/atoms/AtomLayout.re
```diff
@@ -0,0 +1,21 @@
...
+      <div className=Styles.twoColumns>
+        <div className=Styles.left> {fst(children)} </div>
+        <div className=Styles.right> {snd(children)} </div>
```
xla: Do I assume orrectly that `fst` and `snd` is tuple accessors?

app/src/pages/PageProjects.re
```diff
@@ -1,22 +1,18 @@
...
   [@react.component]
   let make = (~projects: array(project)) => {
-    let ps =
-      Array.map(
-        project =>
-          <li key={project.address}>
-            <Link page={Project(project.address)}>
-              <Title> {React.string(project.name)} </Title>
-              <p> {React.string(project.description)} </p>
-              <img src={project.imgUrl} />
```
xla: As commented above I'd see the `ProjectCard` go inside here, so that it just depends on the three things (`name`, `description`, `imgUrl`). Makes it more dumb and only care about presenting.

app/assets/reset.css
rudolfs: Yes, I used the same CSS reset which @juliendonck used for other projects. But since we're using `<h1>`, `<h2>` and `<h3>` for our headings we had to add this custom rule.

Otherwise the html elements have different font weights than the ones in our style guide.

app/src/Topbar.re
rudolfs: Yes, `Container.TwoColumns` expects a tuple of two elements, one for the left side and another one for the right side. Also reason-react expects children to always be an array, hence the spread operator.

You can find out more about it [here](https://github.com/reasonml/reason-react/blob/master/docs/children.md#children-spread).

app/src/atoms/AtomLayout.re
rudolfs: Correct, provided by the standard library as mentioned [here](https://reasonml.github.io/docs/en/tuple#usage).

app/src/atoms/AtomLink.re
rudolfs: Yes, it is significant. The latter overrides the former if they contain the same attributes.

My thought behind this is that a component has a default style which can be made more specific (or altered) from outside. This way we'll have to make less variations of our basic elements.

app/src/pages/PageProjects.re
rudolfs: Same as [above](https://github.com/oscoin/mvp/pull/10/files#r315655987).

app/src/Topbar.re
xla: Ah sorry, missed the that's a tuple, sneaky semis.

app/src/atoms/AtomLink.re
xla: Good to know the presidence for that. Thanks!

app/src/atoms/AtomLayout.re
xla: :pray:

app/src/Molecule.re
rudolfs: Good point.

app/src/Molecule.re
rudolfs: Removed the dependency and added named props for now.

app/src/atoms/AtomLayout.re
rudolfs: My understanding of the "flexbox vs grid" topic is that it's easier to lay out stuff along one axis (1d) with grid and multiple-axes (2d) with flexbox. And that they're commonly used together, as described [here](https://css-tricks.com/css-grid-replace-flexbox/).

Does @juliendonck have an opinion on this?

juliendonck: Yeah, how I understood it they are meant to be used interchangeably. Grid for page structure, flexbox for layout.

xla: Great, thanks for clarification and pointers!

app/src/Molecule.re
```diff
@@ -0,0 +1,39 @@
+open Atom;
+open Particle;
+
+module Styles = {
```
xla: Shouldn't this be inside of the `ProjectCard` module?

app/src/atoms/AtomLayout.re
```diff
@@ -0,0 +1,23 @@
+module Styles = {
```
xla: Same as above shouldn't associated Styles be inside the module they belong to?

app/src/pages/PageProjects.re
```diff
@@ -1,20 +1,28 @@
...
-open Router;
 open Atom;
+open Layout;
+open Molecule;
+
+module Styles = {
+  open Css;
+
+  let projectHeading = style([marginTop(px(94)), marginBottom(px(48))]);
+  let link = style([]);
```
xla: Can we delete it as it's empty?

app/src/Molecule.re
rudolfs: You're right, especially so when there's going to be more molecules.

PR #14, opened 2019-08-22
Add "Join the network" modal

rudolfs:
This PR implements the "Join the network" modal. It is also an attempt in finding the right abstraction for our design system layout.


app/src/Util.re
```diff
@@ -0,0 +1,12 @@
+module CssHelper = {
+  open Css;
+  let (<<) = (a, b) => merge([a, b]);
```
xla: Cute little operator!

Generally I have an inert resistance towards packages/modules that are called common, util, shared, etc. On the surface the naming is not strong indication for the functionality it provides as everything in a project should serve some form of utility. It opens up a catch-all bucket for everything that hasn't been placed in the right place.

If we need operators for our styles, let's put them in an equally named module.

app/src/Util.re
```diff
@@ -0,0 +1,12 @@
...
+
+  let margin = (top, right, bottom, left) =>
```
xla: Would it make sense to have them named and with `0` defaults so that the caller can set them selectively?

app/src/atoms/Atom.re
```diff
@@ -1,6 +1,7 @@
-module Icon = AtomIcon;
 module Button = AtomButton;
+module Icon = AtomIcon;
+module Input = AtomInput;
+module Layout = AtomLayout;
+module Link = AtomLink;
```
xla: This makes me very happy.

app/src/atoms/AtomButton.re
```diff
@@ -1,87 +1,100 @@
...
+      borderColor(Color.lightGray()),
+      color(Color.darkGray()),
+      cursor(`pointer),
+      height(px(48)),
+      outlineStyle(`none),
+      padding4(~top=px(13), ~right=px(24), ~bottom=px(14), ~left=px(24)),
+      hover([backgroundColor(Color.almostWhite(~alpha=0.85, ()))]),
+      active([backgroundColor(Color.almostWhite(~alpha=0.2, ()))]),
+    ])
+    << style(Font.title);
```
xla: Out of curiosity can you chain them? a la `style(...) << style(...) << style(...)`?

app/src/atoms/AtomButton.re
```diff
@@ -1,87 +1,100 @@
...
+      children
+    </button>;
 };
 
 [@react.component]
-let make = (~children, ~disabled=false) =>
-  <button className={disabled ? Styles.disabled : Styles.button} disabled>
-    children
-  </button>;
+let make = (~children, ~disabled=false, ~onClick=_ => ()) =>
```
xla: Would be cool if we can type our onClick props according to the [docs](https://github.com/reasonml/reason-react/blob/521295715e21a510442f8174e3cea2323f3ec80b/src/ReactEvent.rei#L4-L11).

app/src/atoms/AtomInput.re
```diff
@@ -0,0 +1,18 @@
...
+      display(block),
+      padding(px(8)),
+      borderRadius(px(4)),
+      width(pct(100.0)),
+      height(px(48)),
+    ]);
+};
+
+[@react.component]
+let make = (~placeholder="", ~onChange=_ => ()) =>
```
xla: Proper type the `onChange` similar to `onClick` would be great.

app/src/index.html
```diff
@@ -6,6 +6,7 @@
   <link rel="stylesheet" href="/reset.css" type="text/css" media="screen">
 </head>
 <body>
+  <div id="portal"></div>
```
xla: Is it significant that this one comes before app?

app/src/pages/PageJoinNetwork.re
```diff
@@ -0,0 +1,52 @@
+open Atom;
```
xla: The good thing about our pages so far was that they didn't assume anything about the outer structure they were rendered into and I think we should preserve this property. So maybe we can extract the button component which fires up the modal and keep the join network page purely concerned with the form and logic of capturing the data we like to get from a user.

app/src/atoms/AtomButton.re
rudolfs: Yes, my idea was that this would allow us to write more concise inline styles.

app/src/index.html
rudolfs: Yes, because otherwise it would be beneath of whatever "app" renders and we would have to force it above manually by setting `zIndex`. I rather prefer the order of the markup to reflect what we see on screen.

app/src/Util.re
rudolfs: Good point.

How about putting everything design system related in a `DesignSystem.re` file? I feel like having a separate module called `Styles` could be confusing, because our components and pages have modules in them which are already called `Styles`:
```
module Styles = {
  open Css;
  open Particle;

  let regular =
    merge([style([color(Color.darkGray())]), style(Particle.Font.text)]);
};

[@react.component]
let make = (~children) => <p className=Styles.regular> children </p>;
```

What do you think?

app/src/Util.re
rudolfs: You mean like so?
```
let margin = (~top=0, ~right=0, ~bottom=0, ~left=0, ()) =>
  style([
    marginTop(px(top)),
    marginRight(px(right)),
    marginBottom(px(bottom)),
    marginLeft(px(left)),
  ]);
```
```
<Container style={margin(~top=48, ~bottom=24, ())}>
  <Container style={margin(~bottom=16, ())}>
    <Input placeholder="Enter your name" />
  </Container>
  <Input placeholder="Enter an avatar URL" />
</Container>
```

This could also work, I personally prefer the more concise form as it looks like raw css. I don't like that you have to add the empty unit `()` at the end for Reason to work.

I also think that the shorter form would lend itself better for composing styles inline, like so:
```
<Container style={ Styles.someBaseStyle << margin(0, 8, 0, 0) << Styles.moreOverrides}>
```

In the end I think it's a matter of taste. If you prefer the named style, we can do it.

app/src/atoms/AtomButton.re
rudolfs: Ah, thanks for reminding me, I remember we talked about this. On it.

app/src/pages/PageJoinNetwork.re
rudolfs: Also good point, I took it verbatim from our DAPP exploration. Will have a look on how to do it better.

app/src/Util.re
xla: Makes perfect sense. `open DesignSystem` or `open DesignSystem.Operators` reads nicely.

app/src/index.html
xla: Great, maybe inline a comment about it.

app/src/Util.re
xla: It's not purely a matter of taste as a set of 4 integers without any further explanation make it hard to deduce what each param is meant for. So in terms of ergonomics named params add a lot. The strictness of non-optional trailing parameters ruin this a bit for us for sure and as the order - top, right, bottom, left - is well known from CSS - I think we can get away with it for now.

rudolfs: Cool, let's stick with the original version then.

app/src/atoms/AtomButton.re
rudolfs: As we just discussed this IRL, the Reason type system is clever enough to infer the types for us and adding the type annotations by hand just makes things more verbose without added value.

app/src/atoms/AtomInput.re
rudolfs: Ditto. https://github.com/oscoin/mvp/pull/14#discussion_r318096033

app/src/pages/PageJoinNetwork.re
```diff
@@ -0,0 +1,35 @@
...
+
+module Styles = {
+  open Css;
+
+  let buttonContainer = style([display(`flex), justifyContent(flexEnd)]);
+};
+
+[@react.component]
+let make = (~closeButtonCallback) =>
+  <Modal closeButtonCallback>
```
xla: Modal as a render target should not be known to the page. Similar to how the App controls pages are rendered into the section between topbar and footer.  If the Modal is moved to the caller side then we have full flexibility as to where to render this page.

rudolfs: That makes sense.

PR #16, opened 2019-08-27
Expand design system

rudolfs:
Add all missing components from our MVP mockup:
- [x]  members list
- [x]  person/member/add member component
- [x]  project back "<" button
- [x] get rid of footer (not part of design)
- [x] fix layout on larger screens content is jumping around when loading
- [x] project card variant with bigger title
- [x] design system improvements
  - [x] clean up separation between layout and content
  - [x] optional style merge operator


app/src/DesignSystem.re
```diff
@@ -1,8 +1,12 @@
 open Css;
 
-module Operators = {
-  let (<<) = (a, b) => merge([a, b]);
-};
+let (<<) = (a, b) => merge([a, b]);
+
+let (<<?) = (a, b) =>
```
xla: With the `DesignSystem` growing it starts to make sense to introduce a `.rei` with documentation on what the different parts do and how to use them.

app/src/DesignSystem.re
```diff
@@ -11,3 +15,44 @@ let margin = (top, right, bottom, left) =>
...
+};
+
+module Positioning = {
+  let fullWidth = style([gridColumnStart(1), gridColumnEnd(9)]);
+  let wideWidthCentered = style([gridColumnStart(2), gridColumnEnd(8)]);
+  let mediumWidthCentered = style([gridColumnStart(3), gridColumnEnd(7)]);
+  let narrowWidthCentered = style([gridColumnStart(2), gridColumnEnd(8)]);
+};
+
+module Container = {
```
xla: As this is one of the most used component can we shorten it to something like `El`, it otherwise takes a lot of visual significance and distracts from the essential elements. I liked the [abstraction](https://package.elm-lang.org/packages/mdgriffith/elm-ui/latest/Element#el) in [elm-ui](https://github.com/mdgriffith/elm-ui).

app/src/Router.re
```diff
@@ -9,7 +9,7 @@ type page =
   | ProjectFunds(string)
   | NotFound(list(string));
 
-let navigateOfPage = (p: page) => {
+let navigateToPage = (p: page) => {
```
xla: Thanka you for fixing my mess.

app/src/molecules/MoleculeBreadcrumb.re
```diff
@@ -0,0 +1,23 @@
...
+
+  let text = style([color(Color.gray())]);
+
+  let container = style([display(`flex), alignItems(center)]);
+};
+
+[@react.component]
+let make = (~page) =>
+  <Container style=Styles.container>
+    <Icon.Back />
```
xla: This renders a tad to high on my machine with firefox dev edition.

app/src/molecules/MoleculeBreadcrumb.re
```diff
@@ -0,0 +1,23 @@
+open Atom;
+open DesignSystem;
+open Particle;
+open Router;
```
xla: It depending on Router might indicate it was extracted out to early in a molecule.

app/src/molecules/MoleculeProjectCard.re
```diff
@@ -24,14 +15,45 @@ module Styles = {
...
+let make = (~imgUrl=?, ~name, ~description) => {
+  let title = <Title> {React.string(name)} </Title>;
+  <Template ?imgUrl description title />;
+};
+
+module Alternate = {
+  [@react.component]
+  let make = (~imgUrl=?, ~name, ~description) => {
+    let title =
+      <Title.Big style={margin(0, 0, 6, 0)}>
```
xla: Can't `title` be passed as  a children to `Template`?

app/src/pages/PageJoinNetwork.re
```diff
@@ -1,34 +1,38 @@
 open Atom;
-open Layout;
 open DesignSystem;
 
 module Styles = {
   open Css;
 
+  let content =
```
xla: Could this be moved out to the Layout in the DesignSystem?

rudolfs: For sure, I forgot to do it in the rest of the codebase.

app/src/DesignSystem.re
rudolfs: Good idea. I also converted the TwoColumn container to a bunch of Els, as it didn't feel like a good abstraction.

app/src/DesignSystem.re
rudolfs: That's one possibility.

I was thinking of doing the usage documentation on the living-styleguide end with examples where you can see how to use it and also how it's rendered. It might be harder to maintain if I have to keep the documentation in two places: style guide + `.rei`. What do you think?

app/src/molecules/MoleculeProjectCard.re
rudolfs: Indeed, good point.

app/src/DesignSystem.re
xla: I think `.rei` files are non-avoidable and we should get in the habit of having them around by default. Do we want low-level code documentation in the styleguide? Thought of it more as a high-level showcase of what elements exist and how they look. Maybe there is a way to generate part of the styleguide from the `.rei` file, but haven't looked deeply into that part of reason yet.

app/src/DesignSystem.re
```diff
@@ -11,3 +15,35 @@ let margin = (top, right, bottom, left) =>
     marginBottom(px(bottom)),
     marginLeft(px(left)),
   ]);
+
+module Flex = {
```
xla: Is it significant that this module is distinct from `Layout`, or can we treat everything in here as layout primitives as well?

app/src/DesignSystem.re
```diff
@@ -11,3 +15,35 @@ let margin = (top, right, bottom, left) =>
...
+};
+
+module Positioning = {
+  let fullWidth = style([gridColumnStart(1), gridColumnEnd(9)]);
+  let wideWidthCentered = style([gridColumnStart(2), gridColumnEnd(8)]);
+  let mediumWidthCentered = style([gridColumnStart(3), gridColumnEnd(7)]);
+  let narrowWidthCentered = style([gridColumnStart(4), gridColumnEnd(6)]);
+};
+
+module El = {
```
xla: :black_heart:

app/src/DesignSystem.re
rudolfs: No, you're right, actually everything falls into place if I move it into layout.

app/src/DesignSystem.re
rudolfs: Added the `.rei` file + interface documentation.

Even though I think it's a good idea to have the interface documented, I think it might be a little too early to cement it in with comments and explicit typing.

I'm still figuring out what the interface should be and there's a lot of trial-and-error and re-writes when I find a better abstraction. I think at this stage it would be better to use the momentum to get as much elements done as possible, then refactor out the duplication and a bit later, when it all seems somewhat usable and doesn't change as much, add the docs/types.

On the other hand, maybe writing the docs upfront would make me think harder about the interface. I'll try to work on the `.rei` file as I go for now, perhaps we can re-evaluate the decision later if it's getting too much in the way.

app/src/molecules/MoleculeBreadcrumb.re
rudolfs: We have the same problem in the `Link` atom. I'd rather put it in my backlog and work on it in a separate PR, as this one has already gotten quite big.

rudolfs:
Thanks for the suggestions, appreciate it!

> One thing to make reviewing faster and easier for the future is to not conflate structural refactors and actual additions/improvements on pages and visual appearance.

Will give my best to have smaller PRs with better isolated logical changes from now on. As we already discussed IRL, this might be possible with stacked branches which might be a bit harder to manage on my end.

rudolfs:
One more thought on the structural refactor/additions conflation: I usually try to make each commit a single logical change.

Unless you're already doing it, maybe it could be worth briefly skimming over each commit first, before diving into reviewing all the changes together?

app/src/molecules/MoleculeBreadcrumb.re
xla: Safe.

app/src/DesignSystem.re
xla: Wholeheartly agreed on all points, mostly trying to see if we can make it work.

app/src/DesignSystem.rei
```diff
@@ -0,0 +1,91 @@
...
+/** Shorthand for setting margins, for ease of use:
+ **
+ **   <El style=margin(0, 10, 0, 0)> </El>
+ **/
+let margin: (int, int, int, int) => string;
+
+module Layout: {
+  /** CSS grid style, to be used for container elements in combination with
+   ** the respective grid* Positioning styles for child elements:
+
```
xla: Guess a `**` got lost. :>

PR #20, opened 2019-09-03
Guard project registration with account overlay

xla:
This changeset introduces a guarded project registration flow where it forces the join network overlay if a session is not present. To encode the necessary information it uses the the search part of the URL for which overlay should be rendered and what page to go to once the overlay has its flow completed. Additionally there is a little guard in the App now which redirects if the page is not accessible without a session.


rudolfs:
> To encode the necessary information it uses the the search part of the URL for which overlay should be rendered and what page to go to once the overlay has its flow completed.

Is there a reason why we don't have this information in the store of the app?

xla:
> Is there a reason why we don't have this information in the store of the app?

Could you expand what you mean and what it could look like?

rudolfs:
> Could you expand what you mean and what it could look like?

Yes. Sorry for not being clear enough the first time around.

So if I understand the implementation correctly, we're storing the overlay state as URL params, like so: `?overlay=join-network&next=projects/register`. I'm curious as to why we would want to encode it in the URL and expose it to the user. I see the overlay state as an internal implementation detail.

A possible alternative would be to have a new store, for example `StoreOverlay`, which would contain the name of the overlay that is currently open, as well as the next page to be loaded after the overlay is finished.

This way the state is hidden from the user, yet is accessible to any component that might need this state without relying on parsing URL strings and making it more consistent with the overall architecture. As a bonus all the logic for dealing with the overlay state would be in one place.

What do you think?

xla:
@rudolfs That is a great alternative indeed. Up until I had a hard time deciding who would own this information and therefore write to the storage. Yet we introduce the `<Overlay />` component which could be the perfect candidate.

Might be that following your approach things will fall in place naturaly, I take it for a spin. Thanks for putting so much thought into it.

PR #22, opened 2019-09-09
Implement GraphQL interactions with hooks

xla:
Alternative GraphQL integration with hooks API.


app/graphql_schema.json
```diff
@@ -0,0 +1,1143 @@
+{
```
rudolfs: Forgot to mention: this was needed to get the build green.

PR #23, opened 2019-09-09
Implement AppProxy

xla:
Start of the binary we going to use as the backend for the MVP frontned. It serves a GraphQL API which exposes the main interactions we currently need backed by an in-memory datastore.


rudolfs:
This is probably out of scope for this issue, but maybe add CI integration somewhere to your backlog.

proxy/src/schema.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+#[test]
+fn test_schema_all_projects() {
+    use juniper::Variables;
+
+    let ctx = Context::new(Local::new());
+    let (res, _errors) = juniper::execute(
+        "query { allProjects { name } }",
```
rudolfs: Could we maybe rename this to just `projects`? I know I had this in the schema initially, but now after looking at other examples it seems like it would be more elegant if we just use the plural form for collection endpoints.

proxy/src/server_warp.rs
```diff
@@ -0,0 +1,71 @@
...
+        Response::builder()
+            .header("content-type", "text/html")
+            .body(juniper::graphiql::graphiql_source("/graphql"))
+    });
+
+    let routes = index
+        .or(make_graphql_filter("graphql", schema, context))
+        .with(
+            warp::cors()
+                .allow_origins(vec!["http://localhost:8000", "http://localhost:8080"])
```
rudolfs:
```suggestion
                .allow_origins(vec!["http://localhost:8000", "http://localhost:8080", "http://127.0.0.1:8080"])
```

PR #24, opened 2019-09-09
Experiment to use a GraphQL Source without React

xla:
This is an experiment to use GraphQL without React with the help of ReasonQL. It's not complete but fetches data correctly. There is still some unsolved issue with conversion of records.


rudolfs:
We can probably close this for now as we decided to go with the hooks approach: https://github.com/oscoin/mvp/pull/22.

rudolfs:
This isn't relevant anymore as we're switching to JS.

PR #27, opened 2019-09-10
Implement alert component

rudolfs:
This implements a very simple alert notification. To test it run the app with the proxy not running, it should show the network error message in an alert widget. The styleguide has also been updated with the new element.


app/src/molecules/MoleculeAlert.re
```diff
@@ -0,0 +1,105 @@
+open Atom;
+open Particle;
+
+type severity =
```
xla: Uncertain about this and `t` below as alluded to in another comment down below. My initial reaction was that it should be in a place unrelated to the presentation layer, but then again it's used in here mostly.

app/src/molecules/MoleculeAlert.re
```diff
@@ -0,0 +1,105 @@
...
+
+  let icon = style([height(px(24)), Css.margin(px(8))]);
+};
+
+module Template = {
+  [@react.component]
+  let make = (~children, ~severity, ~onClick=_ => ()) =>
+    <div className={Styles.alert(severity)}>
+      <div className=Styles.icon>
+        {
```
xla: Not specific to this changeset, but one thing about JSX that I find unergonomic is that it invites to freely interleave markup and normal logic. That's why I usually try to avoid more complex inline code and rather assign the result of such switch statements to variables and use them in the markup.

app/src/store/StoreAlerts.re
```diff
@@ -0,0 +1,13 @@
+type action =
+  | Show(Molecule.Alert.t)
```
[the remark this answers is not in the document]
rudolfs: You're right, this belongs to `StoreAlerts`.

app/src/molecules/MoleculeAlert.re
rudolfs: I moved the `alert` type to the store, but will leave the `severity` type here, as it's mostly used in this module.

app/src/molecules/MoleculeAlert.re
rudolfs: I was wondering about the way you decide what to leave in JSX and what not, this clears things up. Will adapt my code.

app/src/pages/PageProjects.re
```diff
@@ -76,9 +78,14 @@ let make = () => {
     {
       switch (simple) {
       | Error(err) =>
-        <div className="error">
-          {"ERROR: " ++ err##message |> React.string}
-        </div>
+        dispatch(
```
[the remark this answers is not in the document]
rudolfs: It would just overwrite the message in this case.
I committed some changes that allow showing multiple alerts.

app/src/pages/PageProjects.re
```diff
@@ -76,9 +78,14 @@ let make = () => {
...
+          AlertsAction(
+            StoreAlerts.Show(
+              Alert.{severity: Alert.Error, message: err##message},
```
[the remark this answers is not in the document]
rudolfs: I agree, we should invest some time to figure out how we guide the user in case there are some problems. We should provide the user with a good explanation of what happened and what to do next to resolve the issue. Additionally we should think about what tone we use in the messaging (formal vs informal/friendly).

But I think this should happen at a later stage when we already have nailed down the user flows then we'll be able to see what the actual error cases are and provide the necessary guidance.

app/src/store/StoreAlerts.re
```diff
@@ -0,0 +1,13 @@
...
+  | Remove;
+
+type state = {latest: option(Molecule.Alert.t)};
```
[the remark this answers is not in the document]
rudolfs: Added code for multiple simultaneous alerts.

rudolfs:
>Outcome looks awesome.
Yay, thanks for the feedback!

> Currentoy the user has to explicit close them and they don't resolve automatically.
Added support for multiple simultaneous alerts. They're automatically dismissed after 3 seconds now.

> Additionally what happens when multiple errors need to appear and as commented inline can we improve the ergonomics of our errors and how actionable they are.

The designs that @juliendonck provided don't really show how to implement this visually yet, but I added a basic implementation of what multiple simultaneous alerts might look like. I think we can tweak the UX in another PR as this would be a bit out of scope here.

I also added a notification for when the user joins the network.

app/src/store/StoreAlerts.re
```diff
@@ -0,0 +1,30 @@
+open Molecule.Alert;
```
xla: The reason why we have to import this is for severity only, right? For how I understand the dependency graph the store shouldn't know about the high-up layers (anything visual here).

app/src/store/StoreAlerts.re
```diff
@@ -0,0 +1,30 @@
...
+  severity,
+  message: string,
+  id: int,
+};
+
+type action =
+  | Show(alert)
+  | Remove(alert);
+
+type state = {
```
xla: If we get rid of the id handling here, see comment below, we can have state be as simple as: `type state = array(alert)`.

app/src/store/StoreAlerts.re
```diff
@@ -0,0 +1,30 @@
+open Molecule.Alert;
+
+type alert = {
+  severity,
+  message: string,
+  id: int,
+};
+
+type action =
+  | Show(alert)
```
xla: Would it be more symmetrically to call it `Add`?

app/src/store/ThunkSession.re
```diff
@@ -38,7 +40,12 @@ let createAccount =
          switch (result) {
          | Belt.Result.Ok(account) =>
            Router.navigateToPage(next, ());
-           SessionAction(Created(account)) |> dispatch |> resolve;
+           SessionAction(Created(account)) |> dispatch;
+           StoreMiddleware.Thunk(
+             ThunkAlerts.showAlert(Alert.Success, "Welcome " ++ keyName),
```
xla: Cute!

app/src/store/ThunkAlerts.re
```diff
@@ -0,0 +1,19 @@
+open AppStore;
+open StoreAlerts;
+open Source;
+
+let showAlert = (severity, message, dispatch, state, _source: source) => {
+  let alert = {severity, message, id: state.alerts.idCounter};
+
+  dispatch(AlertsAction(Show(alert)));
+
+  Js.Promise.make((~resolve, ~reject as _) =>
```
xla: We don't need this to be a promise, as their is not asynchronous interaction after the alert has been set.

app/src/molecules/MoleculeAlert.re
xla: See comment in the store.

app/src/store/StoreAlerts.re
```diff
@@ -0,0 +1,30 @@
...
+  | Remove(alert);
```
xla: This only depends on the `id` not the full alert, would make the surface smaller.

app/src/store/ThunkAlerts.re
```diff
@@ -0,0 +1,19 @@
+open AppStore;
+open StoreAlerts;
+open Source;
+
+let showAlert = (severity, message, dispatch, state, _source: source) => {
+  let alert = {severity, message, id: state.alerts.idCounter};
```
xla: Instead of depending on the store keep the id handling, we can use the Random module from Reason and create it locally. This would simplify things.

app/src/store/StoreAlerts.re
rudolfs: Ha, was thinking of this before, but wanted the alert system to be 1000% robust, you know, in case the random ids clash ; P

Will try it out!

app/src/store/ThunkAlerts.re
rudolfs: Ha, was thinking of this before, but wanted the alert system to be 1000% robust, you know, in case the random ids clash ; P

Will try it out!

app/src/store/ThunkAlerts.re
rudolfs: True, copy/pasta accident.

app/src/store/StoreAlerts.re
rudolfs: I like that the interface is the same for all the actions. I can change it to reduce the surface area if you insist.

xla: I don't. :> Smaller interfaces just make it easier to reason about and the added benefit is that the caller side needs to hold on to less data, needs less to fullfill the call.

PR #29, opened 2019-09-11
Ledger integration

xla:
*intentionally left blank*


proxy/Cargo.toml
```diff
@@ -12,6 +12,7 @@ futures = "0.1"
 juniper = "0.13"
 juniper_warp = "0.4"
 log = "0.4"
+oscoin_client = { path = "/home/xla/dev/src/github.com/oscoin/oscoin-parity-wasm-prototype/client" }
```
cloudhead: :eyes:

proxy/src/source.rs
```diff
@@ -137,6 +137,17 @@ impl Source for Local {
...
+        let client = oscoin_client::Client::new_from_file().unwrap();
+        let sender = client.new_account().wait().unwrap();
+        let project_address = oscoin_client::Address::zero();
+        let url = "https://example.com";
+
+        client
+            .register_project(sender, project_address, url.to_string())
+            .wait()
+            .unwrap();
+
```
cloudhead: Lots of unwraps, can we instead return a `Result` from this function?

proxy/Cargo.toml
xla: @cloudhead Don't we all have a `/home/xla`. Needs some changes upstream, this is a draft and just wanted to commit before going to bed.

proxy/src/source.rs
xla: Sure thang, this is just to show that we can do an end-to-end registration. Np proper care has been taken in rgards to ergonomics.

proxy/src/main.rs
```diff
@@ -1,8 +1,16 @@
 //! Proxy to serve a specialised HTTP to the oscoin MVP.
 
-#![deny(clippy::all, clippy::pedantic)]
 #![deny(missing_docs)]
 #![deny(warnings)]
+#![forbid(
+    clippy::all,
+    // clippy::cargo,
```
rudolfs: Does it make sense to commit commented out code into the repo?

xla: Good point, tried out hardmode with all clippy checks. But the `cargo` one only makes sense for public crates consumed as libraries.

PR #30, opened 2019-09-13
Testdrive Cypress e2e testing

rudolfs:
This PR explores e2e integration testing with cypress.


rudolfs:
Closing this as we'll need to re-think how to do it with an electron app.

PR #31, opened 2019-09-16
Cleanup graphql

rudolfs:
Noticed that we had a lot of dead code while working on the localstorage session implementation. Will base my further work on this cleanup PR.


rudolfs:
> Deletions! \0/

Yeah, I hope you weren't too attached to the code :grimacing:

> How would look ap testing look, without depending on the proxy to run?

For frontend stuff I'm leaning towards integration tests as they cover a wide base and UI stuff is mostly glue anyway. For this we could add a mock GraphQL endpoint to the codebase, which would serve test data when tests are being run. I'm currently investigating if we can use cypress for e2e needs. Have a look: https://github.com/oscoin/mvp/pull/30 (WIP).

xla:
> For frontend stuff I'm leaning towards integration tests as they cover a wide base and UI stuff is mostly glue anyway. For this we could add a mock GraphQL endpoint to the codebase, which would serve test data when tests are being run. I'm currently investigating if we can use cypress for e2e needs. Have a look: #30 (WIP).

Doesn't answer the question how we would approach local testing, withe local source gone, and not having to depend on a running proxy.

rudolfs:
> Doesn't answer the question how we would approach local testing, withe local source gone, and not having to depend on a running proxy.

I don't know if it's feasible, but maybe by mocking the GraphQL lib?

rudolfs:
Can you maybe elaborate a bit more what you mean by "local testing"?

xla:
> Can you maybe elaborate a bit more what you mean by "local testing"?

Testing the app in isolation and components in it.

rudolfs:
> Testing the app in isolation and components in it.

I don't know yet, but I'm looking into it right now.

rudolfs:
To recap our face-to-face conversation: we can do frontend-only testing with cypress as well, we'll just provide a local fake graphql endpoint together with the tests. I'm still investigating how we could also do regular unit tests with Jest.

PR #33, opened 2019-09-16
Store mock key in browser's localstorage

rudolfs:
Until now a page refresh would lose our mock key and the "session" would be lost. This stores the key in localstorage so we can play around with the overall UX of the app until we figure out the real requirements for handling private keys.

### DIFF
--- a/app/src/Source.re
+++ b/app/src/Source.re
@@ -12,11 +12,18 @@ type source = {
   fetchAccount: unit => Js.Promise.t(fetchAccountResult),
 };
 
+module StorageLabel = {
+  let keyName = "keyName";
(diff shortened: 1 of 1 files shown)


app/src/Source.re
```diff
@@ -1,158 +1,44 @@
...
-          },
-        |],
-      },
-    |]);
-
   let createAccount = (keyName, avatarUrl) =>
     Js.Promise.make((~resolve, ~reject as _) => {
-      let account = {keyName, avatarUrl};
-      localAccount := Some(account);
+      Dom.Storage.(localStorage |> setItem("keyName", keyName));
```
xla: We use these strings in two places, maybe we can make them into constants.

PR #34, opened 2019-09-19
Show a friendlier error message when proxy is down

rudolfs:


app/src/atoms/AtomIcon.re
```diff
@@ -225,6 +225,47 @@ module Back = {
     </svg>;
 };
 
+module SadFace = {
```
xla: Loool!

app/src/pages/PageProjects.re
```diff
@@ -76,16 +97,17 @@ module GetProjectsQuery = ReasonApolloHooks.Query.Make(GetProjectsConfig);
...
-        ThunkAlerts.showAlert(StoreAlerts.Error, err##message),
-      )
-      |> dispatch;
-      React.null;
+    | Error(_error) =>
+      <ErrorMessage>
+        <Text> {React.string("Coundn't load projects.")} </Text>
+        <br />
+        <Text>
+          {React.string("Please make sure the proxy is running.")}
```
xla: Maybe this one is better suited as a log message and the end-user sees something like backend not reachable. What you think?

rudolfs: Good point! Changed messaging. Also moved the card to a molecule and added the molecule to the styleguide.

PR #37, opened 2019-10-09
Electron app via ReasonML

rudolfs:
An experiment to get our existing ReasonML stack working in electron.


rudolfs:
And now packaging also works via `yarn dist`. Also added extensive documentation.

app/src/main/Electron.re
```diff
@@ -0,0 +1,58 @@
+%raw
```
rudolfs: TBD: convert this to actual ReasonML.

proxy/src/server_warp.rs
```diff
@@ -16,7 +16,7 @@ pub fn run(schema: Schema, context: Context) {
         .or(make_graphql_filter("graphql", schema, context))
         .with(
             warp::cors()
-                .allow_origins(vec!["http://localhost:8000", "http://localhost:8080", "http://127.0.0.1:8080"])
+                .allow_any_origin()
```
rudolfs: Have to figure out how to properly just allow `file://`.

PR #38, opened 2019-10-17
Fresh start with Js

rudolfs:

PR #39, opened 2019-10-18
Switch to javascript and react

xla:
This proposes an alternative way to #38 for a fresh bootstrap using [electron-userland/electron-forge](https://github.com/electron-userland/electron-forge) as the basis to execute and package a native app container.

* [x] bootstrap electron app
* [x] support for es6+ and react
* [ ] liniting / Q&A
* [ ] testing
* [x] CI
* [ ] package proxy
* [ ] lifecycle management


app/native/webpack.main.config.js
```diff
@@ -3,9 +3,9 @@ module.exports = {
    * This is the main entry point for your application, it's the first file
    * that runs in the main process.
    */
-  entry: './src/main.js',
+  entry: './native/main.js',
```
rudolfs: From what I've seen in the community, the convention is:
```
src/main
src/renderer
```
I know it's just a convention, but is there a reason to diverge from that?

xla: From where I stand our app is the functional core providing the feature set according to our product vision. Where electron is a deployment/container hosting/running the app. By separating like this it becomes simpler to understand responsibilities.

The core app interface here is a React component called `App`, which is a managable surface and could be further extended by using props for example. That in turn means that the any container needs to be able to run this React component.

rudolfs:
We decided to go with Svelte.

PR #41, opened 2019-10-21
Bootstrap electron/svelte scaffolding

xla:
New stack based on Svelte as view library.


app/src/pages/Project.svelte
```diff
@@ -0,0 +1,34 @@
+<script>
+  export let params = {}
+
+  console.log(params)
```
rudolfs: Stray `console.log`.

app/native/main.js
```diff
@@ -0,0 +1,58 @@
...
+let mainWindow;
+
+function createWindow() {
+  const mode = process.env.NODE_ENV;
+  mainWindow = new BrowserWindow({
+    width: 900,
+    height: 680,
+  });
+
+  console.log(mode)
```
rudolfs: Stray `console.log`.

app/public/global.css
```diff
@@ -0,0 +1,62 @@
...
+	outline: none;
+}
+
+button:active {
+	background-color: #ddd;
+}
+
+button:focus {
+	border-color: #666;
+}
```
rudolfs: We should fix the indentation here (4 vs 2 spaces).

app/src/pages/Home.svelte
```diff
@@ -0,0 +1,12 @@
+<script>
+  import Projects from '../components/Projects.svelte'
+</script>
+
+<style>
+	h1 {
+		color: purple;
+	}
```
rudolfs: More indentation. Probably should add `prettify` so we don't have to think about this anymore.

PR #44, opened 2019-10-24
Add basic design system

rudolfs:
Just add colors and basic typography.


app/src/DesignSystem/Caption.svelte
```diff
@@ -0,0 +1,13 @@
+<style>
+  p {
+    font-family: "GT America Medium";
```
xla: Wouldn't we set the main font face globally for the entire app?

app/src/pages/DesignSystem.svelte
```diff
@@ -0,0 +1,51 @@
+<script>
+  import { Text, Title, Numeric, Caption } from "../DesignSystem";
+  import { link } from "svelte-spa-router";
+
+  let colors = Array.from(document.styleSheets)
```
xla: Pretty sick way of extracting them, very dry.

app/src/components/Projects.svelte
```diff
@@ -19,21 +20,20 @@
...
+
 </style>
 
-<h2>Projects</h2>
+<a href="/design-system" use:link>design system</a>
+<Title.Big>Projects</Title.Big>
+
 {#await $projects}
-  <p>Loading projects...</p>
+  <Text.Regular>Loading projects...</Text.Regular>
```
xla: The consistent use `index.js` in the design system makes this read very nicely.

app/src/pages/DesignSystem.svelte
```diff
@@ -0,0 +1,51 @@
...
+      []
+    );
+</script>
+
+<a href="/" use:link>back</a>
+
+{#each colors as color}
+  <Text.Regular>
+    <div style="background-color: var({color})">{color}</div>
+  </Text.Regular>
```
xla: What would happen if we compress this?

``` svelte
<Text.Regular style="background-color: var({color})">{color}</Text.Regular>
```

app/src/DesignSystem/Caption.svelte
rudolfs: No, because the font face includes the weight of the font.

app/src/pages/DesignSystem.svelte
rudolfs: That's a good idea!

PR #53, opened 2019-11-01
Port components from ReasonML

rudolfs:


rudolfs:
Perhaps @juliendonck can have a look whether it's fine visually?

juliendonck:
a few comments:
- input focus is too intense. I think we can just use the color-purple outline 2px
- grey's shouldn't have tints

rudolfs:
Cool, thanks for the feedback @juliendonck.

> input focus is too intense. I think we can just use the color-purple outline 2px

Fixed. Couldn't do with just a CSS outline, because you can't round outline corners. But box-shadow + border worked.

> grey's shouldn't have tints

I'm using them for the Vanilla, Transparent and Outline button `:hover` and `:active` states. Will remove them once we have colors for these states.

app/public/index.html
```diff
@@ -8,6 +8,7 @@
 
     <link rel="icon" type="image/png" href="favicon.png" />
     <link rel="stylesheet" href="reset.css" />
+    <link rel="stylesheet" href="global-colors.css" />
```
xla: Reads just as well with just `colors.css`.

app/src/components/Sidebar.svelte
```diff
@@ -32,10 +35,16 @@
 
   :global(li .show-on-hover) {
     opacity: 0;
+    display: none;
+  }
+
+  :global(li:hover svg) {
```
xla: Doesn't `:global` makeit apply to any li in the entire app?

rudolfs: It applies to any svg that lives under a `li:hover`. This is what I was talking about regarding svelte not supporting HOC or style overrides: https://github.com/sveltejs/svelte/issues/2870.

So far the only way to prevent svelte from GCing styles is to make them global.
Let me know if you know of a better workaround.

app/public/index.html
rudolfs: You're right 👍

app/src/components/Sidebar.svelte
xla: In this specific case can't we move the `li` to `SidebarItem` and move the style with it?

rudolfs: Not really, because `SidebarAvatar` also needs the same functionality. What I could do is move everything back into sidebar, because `SidebarAvatar` and `SidebarItem` don't really make sense as separate components. Should I do that?

xla: Another way to think about it is that this hover behaviour should be consistent in the app, unless overriden locally bei a component. So it might be worth having it in the global styles.

rudolfs: I think I'll leave it as it is for now to not lose more time on it. We can re-think it if we run into problems or if there is a new svelte release which addresses this functionality.

PR #55, opened 2019-11-05
Code browsing

xla:
*work in postgres*


app/schema.graphql
```diff
@@ -54,3 +54,47 @@ type Info {
...
+  name: String!
+}
+
+# Representation of gitrevisions: https://git-scm.com/docs/gitrevisions/
+union Revision = ObjectName | RefName
+
+type Query {
+  # Source browsing.
+  blob(projectId: ID!, revision: Revision!, path: String!): [BlobEntry]!
+  tree(projectId: ID!, revision: Revision!, prefix: String!): [TreeEntry]!
```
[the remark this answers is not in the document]
xla: Absolutely correctly observed. The problem is that you can't express non-emptiness in GraphQL as the `!` is non-nullable: https://stackoverflow.com/a/46771890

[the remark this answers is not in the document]
rudolfs: Ah, didn't realise that you can't express non-emptiness.

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
+scalar Datetime
+
+type Author {
```
FintanH: Going to drop some links here to help with discovering how to get info from `git2`.

[Commit.author](https://docs.rs/git2/0.10.1/git2/struct.Commit.html#method.author), returns a `Signature` where you can get [`Signature.name`](https://docs.rs/git2/0.10.1/git2/struct.Signature.html#method.name) and [`Signature.email`](https://docs.rs/git2/0.10.1/git2/struct.Signature.html#method.email).

You can also get the time with [`Signature.when`](https://docs.rs/git2/0.10.1/git2/struct.Signature.html#method.when)

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
...
+  name: String
+  email: String
+}
+
+# git commit data.
+type Commit {
+  sha1: String!
```
FintanH: This can be calculated via [`Commit.id`](https://docs.rs/git2/0.10.1/git2/struct.Commit.html#method.id) to get an `Oid`. Then [`Oid.as_bytes`](https://docs.rs/git2/0.10.1/git2/struct.Oid.html#method.as_bytes) to get the byte slice.

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
...
+  author: Author!
+  date: Datetime!
```
FintanH: [`Commit.time`](https://docs.rs/git2/0.10.1/git2/struct.Commit.html#method.time)

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
...
+  message: String!
+  refs: [Ref]
+  objectName: ObjectName!
```
FintanH: [`Commit.as_object`](https://docs.rs/git2/0.10.1/git2/struct.Commit.html#method.as_object) followed by [`Object.kind`](https://docs.rs/git2/0.10.1/git2/struct.Object.html#method.kind)

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
...
+# file info and metadata.
+type Info {
+  # base name of the file.
+  name: String!
+  # length in bytes.
+  size: Int!
+  # if it's a directory.
+  isdirectory: Boolean!
+  # any flags git stores, currently only executable.
+  flags: [Int]!
```
FintanH: What's this?

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
...
+  refs: [Ref]
+  objectName: ObjectName!
+}
+
+# file info and metadata.
+type Info {
+  # base name of the file.
+  name: String!
+  # length in bytes.
+  size: Int!
```
FintanH: For a directory is this the `size` of all files beneath it?

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
...
+  # if it's a directory.
+  isdirectory: Boolean!
+  # any flags git stores, currently only executable.
+  flags: [Int]!
+  # last commit which affected this path.
+  lastcommit: Commit!
```
FintanH: This will be more interesting. Not sure how to model this currently, but if you could make an issue I'll start thinking about this as a sole feature. At the moment we could hardcode it as the head commit of the `History` :grimacing:

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
...
+
+# Representation of a refname. Special refs like HEAD are omitted on purpose as
+# of now.
+#
+# master
+# v1.0.0
+# heads/master
+# tags/v1.0.0
+# remotes/origin/master
+type Ref = {
```
FintanH: I think I need a better story for refs vs branches vs tags. At the moment I just refs for _everything_ and haven't tested it with a commit SHA1 to see if that works.

app/schema.graphql
```diff
@@ -0,0 +1,103 @@
...
+
+enum ObjectType {
+  Commit
+  Tree
+  Blob
+  Tag
+}
+
+# git tags.
+type LightweightTag {
```
FintanH: Again, need to flesh out how to present these to you via `radicle-surf` :thought_balloon:

app/schema.graphql
xla: radicle-dev/radicle-surf#11

app/schema.graphql
xla: Mode flags, currently the only one git preserves afaik is the executable flag on a file.

app/schema.graphql
xla: I'd assume it to be `-1` when `isDIrectory` is true.

app/schema.graphql
xla: We used to have `union Ref = Head | Remote | Tag`. But when you ignore special refs like `HEAD` and friends this type works quite well. Was pretty much working along this: https://git-scm.com/book/en/v2/Git-Internals-Git-References

app/schema.graphql
xla: Not at all urgent, this is still from times when I went down the git object types to extract their shapes. We don't currently have any tag objects surfaced, besides the refnames. It will become more interesting down the road and esp. the power of annotated tags and the ability to stick tree objects in there.

app/schema.graphql
FintanH: Ah, ya `Blob` has this https://docs.rs/git2/0.10.1/git2/struct.Blob.html#method.is_binary. I think Pijul actually has the whole set of modes :)

app/schema.graphql
FintanH: Why not move this to `Blob` then?

app/schema.graphql
FintanH: Cool! I'll take a look :)

app/schema.graphql
xla: Not sure if this is the same, as git supports this: `git update-index --chmod=+x foo.sh`. In general executbales dont need to be binary.

app/schema.graphql
xla: A possibility, but info is used in directory listing. We don't use it anywhere yet, would wonder if you'd be interested in the size without the file content.

app/schema.graphql
rudolfs: Would it make sense to use `mode` similar to what `git-tree` returns?
```
mvp/app » git ls-tree --long code-browsing /Users/rudolfs/work/mvp/
040000 tree e8023e1209a6022568f04962f0cdf48f815b48d3       -	../.buildkite
100644 blob 3c3629e647f5ddf82548912e337bea9826b434af      13	../.gitignore
040000 tree fe3470499dcdf36bc6c048c8d43cae67554e27db       -	./
040000 tree 57f6fc0dced5a21b9af6721340920c3f0eae899d       -	../proxy
```

app/schema.graphql
rudolfs: I could see how getting the size of a directory would be valuable for future UX. However at the moment we would only be interested in file sizes.

app/schema.graphql
FintanH: I wonder where this information is stored then? Basically need to figure out if you can get it from the `Commit` or will I have to store some information alongside `File` and `Directory` in the `surf` model.

app/schema.graphql
FintanH: Well I don't see it as being hard to calculate the size of a `Directory` in my model. It's basically `directory.files.map(|size|).sum + directory.sub_directories.map(|size|).sum` :)

FintanH: So if it's useful, it can be added!

rudolfs: I think it would be useful!

FintanH: You got it, dear customer!

app/schema.graphql
FintanH: Oh this is probably it: https://docs.rs/git2/0.10.1/git2/struct.TreeEntry.html#method.filemode

app/schema.graphql
```diff
@@ -1,42 +1,46 @@
...
   info: Info!
 }
 
-type TreeEntry {
-  path: String!
-  info: Info!
+type Tree {
+  path: String
+  info: Info
+  leaves: [Leaf!]!
```
xla: @rudolfs If we establish `Blob` and `Tree` already, then this fields is technically `entries` and can be either `Blob` or a `Tree`.

app/src/components/CommitTeaser.svelte
```diff
@@ -0,0 +1,60 @@
+<script>
+  import UserCard from "../components/UserCard.svelte";
```
xla: The imported component lives in the same directory. Do we need the exta `../componentes` prefix?

app/src/components/FileList.svelte
```diff
@@ -0,0 +1,172 @@
...
+  import { getClient, query } from "svelte-apollo";
+
+  import { format } from "timeago.js";
+
+  import { getContext } from "svelte";
+  import { revision, objectPath } from "../stores.js";
+  import * as path from "../path.js";
+
+  import { Caption, Icon, Text } from "../DesignSystem";
+  import CommitTeaser from "../components/CommitTeaser.svelte";
```
xla: Same as prefix question above.

app/src/components/FileList.svelte
```diff
@@ -0,0 +1,172 @@
...
+  import { link } from "svelte-spa-router";
+
+  const TREE = gql`
```
xla: These start to pop up more and more and live inline, would it make sense to centralise our queries somewhere? And as a follow-up start validating.

app/src/components/FileSource.svelte
```diff
@@ -0,0 +1,114 @@
...
+  import { gql } from "apollo-boost";
+  import { getClient, query } from "svelte-apollo";
+
+  import { format } from "timeago.js";
+
+  import { getContext } from "svelte";
+  import { revision, objectPath } from "../stores.js";
+
+  import { Icon } from "../DesignSystem";
+  import CommitTeaser from "../components/CommitTeaser.svelte";
```
xla: Same component question like above.

app/src/pages/Project.svelte
```diff
@@ -0,0 +1,97 @@
...
+  const client = getClient();
+  const project = query(client, {
+    query: GET_PROJECT,
+    variables: { id: params.id }
+  });
+
+  const PATH_MATCH = /\/source\/(.*)\/(blob|tree)(.*)/;
+
+  $: rev = $location.match(PATH_MATCH);
+  $: rev = rev === null ? "refs/heads/master" : rev[1];
```
xla: The default ref would be a good target to be a constant in a central place.

xla: Might even be worse to bubble it up from the CoCo defaults through the API.

app/src/components/FileList.svelte
rudolfs: As discussed IRL: we'll leave them where they are for now, because they're very specific to the views.

PR #62, opened 2019-12-03
Add tests for code browsing and navigation

rudolfs:
"Write tests. Not too many. Mostly integration."

~~**Note:** screenshot diffs are somewhat experimental and might be flaky, let's try them out for a while. If they get too much in the way, we can remove them again.~~

Resolver MON-9


app/README.md
```diff
@@ -27,7 +27,42 @@ Build and package the app:
 2. Get the generated package from: `dist/radicle-upstream-0.0.1.dmg`
 
 
-#### Buildkite
+#### Running tests locally
+
+Before running tests locally you'll need to set up a test fixture repository
+via: `cd ../../ && git clone https://github.com/radicle-dev/git-platinum`.
```
xla: Wouldn't it be better to have the repo be set up inside of radicle-upstream? Otherwise you have to make assumptions for any state outside of the current repo.

Additionally a simple clone will not pin it to the right version that is known to work for the current test suite. A way to pin is for example git submodules.

app/src/DesignSystem/Select.svelte
```diff
@@ -30,7 +31,7 @@
   }
 </style>
 
-<select {style} bind:value>
+<select data-cy={dataCy} {style} bind:value>
```
xla: Are all these `data-cy` attributes to make cypress work?

rudolfs: Yes, these are test harnesses for cypress: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

app/README.md
rudolfs: I wanted to have it as a git submodule initially, but then I realised that we depend on the state of the test repository as a whole.

We don't need it to be pinned to a specific version, because we're querying all kinds of data that don't depend on the current checkout, for example:
  - which files exist at a specific revision
  - all tags/branches

rudolfs:
> Are we making sure that cypress uses the same chromium version as we currently use inside of Electron for parity?

To my understanding, we have to use the electron version that is bundled with a specific cypress version. Cypress lags a bit behind the latest electron releases. As I mentioned previously in https://github.com/radicle-dev/radicle-upstream/pull/47: there is some experimental work going on with testing actual electron apps (together with the electron container). But I want to wait until they have something stable before investing a lot of energy into it.

> Is there a way to have these test run continously during development?

I think it wouldn't make sense to run them continuously, as e2e tests tend to take longer to run. My current workflow is to run them manually before doing a commit.

> Can we reduce the time it takes (currently ~30s)?

We could shave off 5sec by deleting the visual snapshot tests. But then we'd have to rely on testing css properties on DOM elements.

An alternative would be to write unit-tests instead of doing e2e, and mock away a lot of the environment, but in my experience the speed they give don't make up for the confidence I lose by not doing e2e tests.

app/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
geigerzaehler: Is this file needed?

app/cypress/integration/navigation.spec.js
```diff
@@ -0,0 +1,269 @@
...
+        .blur();
+      cy.contains("Projects").should("exist");
+
+      // check whether the active icon is highlighted
+      cy.get('[data-cy="sidebar"]').toMatchImageSnapshot({
+        name: "sidebar-projects-active"
+      });
+
+      cy.get('[data-cy="sidebar"]')
+        .get('a[title="Create new project"]')
```
geigerzaehler: You could also use the `data-cy` attribute here to select the “create new project” link. This decouples the selector from the title and the element type and makes identifying elements consistent.

app/cypress/integration/navigation.spec.js
```diff
@@ -0,0 +1,269 @@
...
+        .click()
+        .blur();
+      cy.contains("Design System").should("exist");
```
geigerzaehler: This could be rewritten as
```suggestion
      cy.get("[data-cy=some-value]").should("contain", "Design System");
```
This allows for more consistent and strict assertions.

app/cypress/integration/navigation.spec.js
```diff
@@ -0,0 +1,269 @@
...
+      cy.get('[data-cy="sidebar"]')
+        .get('a[title="Projects"]')
+        .click()
+        .blur();
+      cy.contains("Projects").should("exist");
+
+      // check whether the active icon is highlighted
+      cy.get('[data-cy="sidebar"]').toMatchImageSnapshot({
+        name: "sidebar-projects-active"
+      });
```
geigerzaehler: It might be a bit of overkill to use an screenshot to check whether an icon is highlighted. In particular since screenshots are quite expensive.

app/src/DesignSystem/Select.svelte
xla: It seems odd to have this leak into our markup. Maybe a function of getting used to. Just wonder how leaky it is.

xla:
> To my understanding, we have to use the electron version that is bundled with a specific cypress version. Cypress lags a bit behind the latest electron releases. As I mentioned previously in #47: there is some experimental work going on with testing actual electron apps (together with the electron container). But I want to wait until they have something stable before investing a lot of energy into it.

@rudolfs Think it's worth noting explicitely somewhere as this could be easily a source of sublte errors.

> I think it wouldn't make sense to run them continuously, as e2e tests tend to take longer to run. My current workflow is to run them manually before doing a commit.

Could we make it required before a successful commit like our linting?

app/src/DesignSystem/Select.svelte
rudolfs: I agree, it is a bit weird to add this stuff only for tests. But apparently it's a best practice.

We could potentially strip them out from a production build with some rollup magic if they're getting too much in the way.

app/cypress/fixtures/example.json
rudolfs: Nope.

app/src/DesignSystem/Select.svelte
xla: :black_heart:

app/README.md
rudolfs: @garbados that's exactly what we need, it should be outside of this repository, as we don't need it to be "mounted" as a git-submodule this repository. We need the layout to be like this:
```
workdir/
.
├── git-platinum
└── radicle-upstream
```

xla: I think that's the point, that it makes an assumption about the state outside of the repository directory boundaries.

rudolfs:
> Could we make it required before a successful commit like our linting?

I would leave the e2e tests for CI, because they'll be long running anyway and we block PRs where the tests on CI fail.

I think a per-commit git-hook for running unit tests that are affected by the staged changes could work. But I feel like that might be out of scope for this issue.

app/README.md
rudolfs: I updated the README to make it more clear.

app/cypress/integration/navigation.spec.js
rudolfs: Removed the screenshot tests.

xla:
> I would leave the e2e tests for CI, because they'll be long running anyway and we block PRs where the tests on CI fail.
> 
> I think a per-commit git-hook for running unit tests that are affected by the staged changes could work. But I feel like that might be out of scope for this issue.

Good points all around, we can follow up on this outside of this PR.

app/README.md
rudolfs: Y'all were right, I was wrong. I got it working as a submodule, much cleaner now.

rudolfs:
> Think it's worth noting explicitely somewhere as this could be easily a source of sublte errors.

Added it to the README.

app/README.md
xla: :heart:

PR #64, opened 2019-12-09
Integrate radicle-surf to satisfy CoCo API endpoints

xla:
This change introduces an implementation of the API drafted for #55 powered by radicle-dev/radicle-surf. A couple of adjustments had to be made to surf along the way. Most notably we needed to reexport git2 to have access to some lower level types. We track types and their info we depend on in radicle-dev/radicle-surf#41 to remedy that. We also settled on git-platinum as the body of fixtures to be used in our tests. The infrastructure to test GraphQL queries has been drastically improved in this patch and is expected to be more exhaustive with additions, as well as the discovery of edge cases.

At the time of this writing there is a incosnistency in how `last_commit` behaves, it is tracked in radicle-dev/radicle-surf#40. This means the API will always return the latest commit for the current history, indiscriminate if it touched the path in question - Resulting in errors in the ui integration tests.

* implement branches query
* implement tags query
* implement commit query
* implement blob query
* implement tree query
* integrate API in the UI
* use radicle-dev/git-platinum for test fixtures
* use latest compatible toolchain (dictated by radicle-dev/radicle-registry)
* pull in radicle-dev/radicle-surf as dependency

Resolver MON-10


garbados:
For reference, buildkite is failing because it lacks credentials to download git-golden:

```
Submodule 'data/git-golden' (git@github.com:radicle-dev/git-golden.git) registered for path 'data/git-golden'
Cloning into '/mnt/builds/tete-de-moine-4/monadic/radicle-upstream/data/git-golden'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

xla:
@garbados Probably needs to be changed to an https url. Want to get the rust tests running in CI in an independent PR, but good find. Thanks.

proxy/src/schema.rs
```diff
@@ -53,6 +73,52 @@ impl Query {
...
 
+    fn commit(ctx: &Context, id: IdInput, sha1: String) -> FieldResult<Option<Commit>> {
+        let repo = GitRepository::new(&ctx.dummy_repo_path).expect("setting up repo failed");
+        let mut browser = GitBrowser::new(&repo).expect("setting up browser for repo failed");
+        browser
+            .commit(radicle_surf::vcs::git::Sha1::new(&sha1))
+            .expect("setting commit failed");
+
+        let history = browser.get_history();
+        let maybe_commit = history.iter().last();
```
FintanH: If you use `commit` the `History` is just going to be the singleton list of that commit. So in this case you can just do `browser.get_history().0.first()` and be guaranteed to get the commit.

xla: It's guaranteed to be a commit as it would already fail on the `browser.commit` call?

FintanH: Yup, exactly :)

proxy/src/schema.rs
```diff
@@ -366,8 +445,38 @@ mod tests {
             graphql_value!({
                 "tree": {
                     "path": "src",
+                    "info": {
+                        "name": "src",
+                        "objectType": "TREE",
+                        "lastCommit": {
+                            "sha1": "74ba370ee5643f310873fb288af1c99d639da8ca",
+                            "author": {
+                                "name": "Fintan Halpenny",
```
xla: @FintanH He's back!

proxy/src/schema.rs
```diff
@@ -44,6 +52,82 @@ impl Into<ProjectId> for IdInput {
...
+    sha1: String,
+    author: Person,
+    summary: String,
+    message: String,
+    time: String,
+}
+
+// FIXME(xla): This should be a `std::convert::TryFrom` and needs to be addressed together with
+//             consistent error handling.
+impl From<&radicle_surf::vcs::git::git2::Commit<'_>> for Commit {
```
FintanH: I'm thinking that `Commit` being exposed from `radicle-surf` due to it being the `A` in our model justifies it being re-exported. I think it'd be good to start tracking what we would like exposed, and then limiting the re-export

proxy/src/schema.rs
```diff
@@ -53,6 +137,170 @@ impl Query {
...
+                .to_string(),
+            info: Info {
+                name: last.label,
+                object_type: ObjectType::Blob,
+                last_commit: Commit::from(&last_commit),
+            },
+        })
+    }
+
+    fn commit(ctx: &Context, id: IdInput, sha1: String) -> FieldResult<Option<Commit>> {
```
FintanH: Does it need to be `Option` if we're guaranteed to get it as long as the setting works?

xla: Probably not as we will express non-existent errors through the error of `FieldResult`.

proxy/src/schema.rs
xla: @FintanH Do you have a preferred way of tracking these?

FintanH: I think an issue with a checklist would be good. I think at the moment want:

- Commit
- BranchType
- Oid

Think that may be it for now.

proxy/src/schema.rs
FintanH: :+1:

proxy/src/schema.rs
xla: Something lke: radicle-dev/radicle-surf#41 ?

garbados:
Getting test failures due to timeouts, ex: ` CypressError: Timed out retrying: Expected to find content: '1 week ago' but never did.`

app/package.json
```diff
@@ -47,28 +46,26 @@
...
     "dist": "rm -rf ./dist && mkdir ./dist && yarn svelte:clean && yarn svelte:build && yarn proxy:build && cp ../proxy/target/release/proxy dist && electron-builder",
     "ci:dist": "rm -rf ./dist && mkdir ./dist && cp ../proxy/target/release/proxy dist && electron-builder",
     "electron:start": "wait-on ./public/bundle.js && wait-on ./native/main.comp.js && wait-on http://localhost:8080 && NODE_ENV=development electron .",
     "svelte:clean": "rm -rf public/bundle.*",
     "svelte:build": "yarn svelte:clean && rollup -c",
     "svelte:watch": "yarn svelte:clean && rollup -c -w",
     "proxy:build": "cd ../proxy && cargo build --release",
-    "proxy:start": "cd ../proxy && cargo run --release -- memory",
+    "proxy:start": "cd ../proxy && cargo run --release -- --source=memory",
+    "proxy:start:test:": "cd ../proxy && cargo run --release -- --source=memory --test",
```
rudolfs:
```suggestion
"proxy:start:test": "cd ../proxy && cargo run --release -- --source=memory --test"
```

proxy/src/schema.rs
```diff
@@ -44,24 +77,122 @@ impl Into<ProjectId> for IdInput {
...
+    /// Identifier of the commit in the form of a sha1 hash. Often referred to as oid or object id.
+    sha1: String,
+    /// The author of the commit.
+    author: Person,
+    /// The summary of the commit message body.
+    summary: String,
+    /// The entire commit message body.
+    message: String,
+    /// The recorded time of the commit.
+    time: String,
```
rudolfs: I think we should be a bit more explicit whether this is the [author or commit time](https://stackoverflow.com/questions/11856983/why-git-authordate-is-different-from-commitdate#11857467). I also noted it in the [schema](https://github.com/radicle-dev/radicle-upstream/blob/xla/surf-integration/app/schema.graphql#L13).

proxy/src/schema.rs
```diff
@@ -44,24 +77,122 @@ impl Into<ProjectId> for IdInput {
...
+        let signature = commit.author();
+
+        Self {
+            sha1: commit.id().to_string(),
+            author: Person {
+                name: signature.name().unwrap_or("invalid name").into(),
+                email: signature.email().unwrap_or("invalid email").into(),
+                avatar: "https://avatars.dicebear.com/v2/human/foo.svg".into(),
+            },
+            summary: commit.summary().unwrap_or("invalid subject").into(),
```
rudolfs: I think it's called `subject` as per the [official git documentation](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration). I would prefer sticking to that.

Screenshot 2019-12-16 at 18 08 31

proxy/src/schema.rs
FintanH: YES muchos love ❤️

app/src/components/FileSource.svelte
```diff
@@ -88,26 +84,30 @@
...
-        {@html result.data.blob.content
-          .split('\n')
-          .slice(0, -1)
-          .map((_, index) => {
-            return `${index + 1}`;
-          })
-          .join('\n')}
-      </pre>
+      {#if result.data.blob.binary}
+        ఠ ͟ಠ Binary content.
```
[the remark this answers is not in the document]
xla: Not necessarily, this is what we present to when the file is binary and can't be meaningfully presented to the user without further effort returning proper mimed files and have them shown in the app.

proxy/src/main.rs
```diff
@@ -28,7 +27,7 @@ mod source;
 
 fn main() {
     std::env::set_var("RUST_BACKTRACE", "full");
-    std::env::set_var("RUST_LOG", "info");
+    std::env::set_var("RUST_LOG", "debug");
```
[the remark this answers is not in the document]
xla: This is on an outdated version of the PR, master has been merged back and your changes landed in this PR. :)

.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "app/cypress/git-platinum"]
-	path = app/cypress/git-platinum
-	url = https://github.com/radicle-dev/git-platinum.git
+[submodule "fixtures/git-platinum"]
+	path = fixtures/git-platinum
+	url = https://github.com/radicle-dev/git-platinum
```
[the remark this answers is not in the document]
xla: :1st_place_medal:

app/src/components/FileSource.svelte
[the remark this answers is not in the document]
rudolfs: It's a TODO for @juliendonck ; P

proxy/src/schema.rs
```diff
@@ -77,59 +208,724 @@ impl Mutation {
...
+            .into_iter()
+            .map(|tag_name| Tag {
+                name: tag_name.name(),
+            })
+            .collect();
+
+        Ok(tags)
+    }
+
+    fn tree(ctx: &Context, id: IdInput, revision: String, prefix: String) -> FieldResult<Tree> {
```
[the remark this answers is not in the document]
xla: Addressed in ccee904

proxy/src/schema.rs
```diff
@@ -44,24 +77,122 @@ impl Into<ProjectId> for IdInput {
...
+impl From<&git2::Commit<'_>> for Commit {
+    fn from(commit: &git2::Commit) -> Self {
+        let signature = commit.author();
+
+        Self {
+            sha1: commit.id().to_string(),
+            author: Person {
+                name: signature.name().unwrap_or("invalid name").into(),
+                email: signature.email().unwrap_or("invalid email").into(),
+                avatar: "https://avatars.dicebear.com/v2/human/foo.svg".into(),
```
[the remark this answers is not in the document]
xla: I was lazy. Thanks for the follow-up, addressed in 397695e.

proxy/src/schema.rs
```diff
@@ -44,24 +77,132 @@ impl Into<ProjectId> for IdInput {
     }
 }
 
-/// Encapsulates read paths in API.
-pub struct Query;
+/// Object representing a git branch.
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd, GraphQLObject)]
+struct Branch {
```
FintanH: Why not a newtype?

proxy/src/schema.rs
```diff
@@ -44,24 +77,132 @@ impl Into<ProjectId> for IdInput {
...
-#[juniper::object(Context = Context)]
-impl Query {
-    fn apiVersion() -> &str {
-        "1.0"
-    }
+/// Object representing a git tag.
+///
+/// We still need full tag support.
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd, GraphQLObject)]
+struct Tag {
```
FintanH: Same here

proxy/src/schema.rs
```diff
@@ -44,24 +77,132 @@ impl Into<ProjectId> for IdInput {
...
-        Ok(ctx.source.get_all_projects())
-    }
+/// Representation of a person (e.g. committer, author, signer) from a repository. Usually
+/// extracted from a signature.
+#[derive(GraphQLObject)]
+struct Person {
+    /// Name part of the commit signature commit.
+    name: String,
+    /// Email part of the commit signature commit.
+    email: String,
```
FintanH: Will we ever need to validate this? Maybe newtyping again?

proxy/src/schema.rs
```diff
@@ -44,24 +77,132 @@ impl Into<ProjectId> for IdInput {
...
+    /// Reference (url/uri) to a persons avatar image.
+    avatar: String,
+}
 
-    fn project(ctx: &Context, id: IdInput) -> FieldResult<Option<Project>> {
-        Ok(ctx.source.get_project(id.into()))
+/// Representation of a code commit.
+#[derive(GraphQLObject)]
+struct Commit {
```
FintanH: So actually, this is making think that this module should really be `schema::git` or `git::schema`. Because it's just git related. But there might be things we can extract out such as the user profile, and these things would go under `schema::mod`. What do you think? Obviously that can be done in another PR :)

xla: This module is the surface of the GraphQL API the proxy provides. While it could be broken apart potentially, I want to defer that to the latest possible point and until then keep it centralised so a reader can get an effortless overview of what to expect of the endpoint. We haven't found a way yet ot surface that more ellegantly (i.e. a generated schema file from an introspection call).

xla: There is definitely room to isolate parts of it, do you see an obvious drawback of having it organised as is?

proxy/src/schema.rs
xla: It gets tricky here. This template is not matching how these parts of the commit are called in other places. Most notably the reference implementation https://github.com/libgit2/libgit2/blob/d5017a14284447c289dd04cae1a310dc34aa562c/src/commit.c#L28-L39

@FintanH Also mentioned to not tie these concepts to much to any particular vcs, which would force us to look at it fresh and define our nomenclature.

proxy/src/schema.rs
xla: We need both you are right. There are actualy organised under two fields `author` and `commiter` which are what's known as a `signature`.

So we can have two fields on the commit `authorDate` and `committerDate` or we expand to proper signature data for each. What you think?

proxy/src/schema.rs
xla: Simplified in 75ec3c0.

proxy/src/schema.rs
xla: Addressed in 75ec3c0.

proxy/src/schema.rs
FintanH: It's more just a "housekeeping" thing. If you separate boundaries early you can notice where things need to overlap and where things can be extended easily. But ya this sort of thing can be done as other factors come into play :)

proxy/src/schema.rs
rudolfs: I'm happy with 2 extra fields on the commit.

I feel like the term "signature" is a bit overloaded already (think commit signing), so if you prefer on extracting the fields, I'd urge to choose a different name for it.

xla: The signature here is referring to the "commit signing" one from each party an author signature and a commiter signature, they refer to the same concept. :>

proxy/src/schema.rs
xla: Good point, I gonna attempt this and see what the rippling effects are as the main endpoint `Query` type is an integration point for all these other types. Also curios if it makes the massive test suite that currently starts to grow at the bottom of `schema.rs` becomes more managable.

proxy/src/schema.rs
xla: Which foreign type would you newtype here?

proxy/src/schema.rs
rudolfs: Just to nit-pick some more, the actual git source (as opposed to libgit2) uses "subject":
https://github.com/git/git/blob/master/commit.c#L530.

But good point about the abstraction though. Had a look at pijul, they don't seem to have the concept of annotating patches with messages at all.

So let's leave it as is.

proxy/src/schema.rs
rudolfs: I see, in that case, extracting it probably makes sense.

proxy/src/schema.rs
FintanH: Modularity is King :crown:

app/cypress/integration/source_browsing.spec.js
```diff
@@ -128,29 +128,31 @@ context("source code browsing", () => {
...
 
       context("when the timeframe is less than a week", () => {
         it("shows timeframe in days", () => {
-          cy.clock(Date.parse("4 dec 2019"));
+          cy.clock(Date.parse("14 dec 2019"));
           cy.contains("1 day ago").should("exist");
         });
       });
 
       context("when the timeframe is less than a week", () => {
```
rudolfs: This should be "when the timeframe is **more** than a week"

app/cypress/integration/source_browsing.spec.js
```diff
@@ -128,29 +128,31 @@ context("source code browsing", () => {
...
         it("shows timeframe in weeks", () => {
-          cy.clock(Date.parse("15 dec 2019"));
-          cy.contains("1 week ago").should("exist");
+          cy.clock(Date.parse("17 dec 2019"));
```
rudolfs:
```suggestion
          cy.clock(Date.parse("22 dec 2019"));
```

app/cypress/integration/source_browsing.spec.js
```diff
@@ -128,29 +128,31 @@ context("source code browsing", () => {
...
+          cy.contains("4 days ago").should("exist");
```
rudolfs:
```suggestion
          cy.contains("4 days ago").should("exist");
```

proxy/src/schema.rs
xla: I went for the intermediate solution and ultimately what you proposed to begin with. The field is now explcitiley stating that it is the committer time. This is a shorthand until we expose `author` and `comitter` signatures propr. Check 0bb3a34 for the actual change.

PR #66, opened 2019-12-09
Add a convenience function to set ENV vars if not present

garbados:
I noticed the `set_var` calls were changing between versions, implying that folks are hardcoding values as needed. This PR adds a `set_if_unset` function which only sets an env var if it is unset, allowing you to set log levels and backtrace behavior as necessary while retaining useful defaults.

### DIFF
--- a/proxy/src/main.rs
+++ b/proxy/src/main.rs
@@ -25,10 +25,12 @@ mod schema;
 mod server_warp;
 /// Origin of data required like the on-chain Registry.
 mod source;
+/// lib functions
+mod env;
 
 fn main() {
-    std::env::set_var("RUST_BACKTRACE", "full");
-    std::env::set_var("RUST_LOG", "info");
+    env::set_if_unset("RUST_BACKTRACE", "full");
+    env::set_if_unset("RUST_LOG", "debug");
     pretty_env_logger::init();
 
     let source_type = std::env::args().nth(1).expect("no source was given");
--- a/proxy/src/env.rs
+++ b/proxy/src/env.rs
@@ -1,0 +1,24 @@
+/// Convenience function for setting env vars only if they are unset.
+pub fn set_if_unset (key: &str, val: &str) {
+    use std::env::{var,set_var};
+    var(key).unwrap_or_else(|_| { set_var(key, val); var(key).unwrap() });
+}
+
+#[test]
+fn test_set_if_unset() {
+    match std::env::var("DUMMY_VALUE") {
+        Ok(_val) => {
+            // should be unset
+            panic!("$DUMMY_VALUE should be unset.")
+        },
+        Err(_e) => {
+            set_if_unset("DUMMY_VALUE", "hello world")
+        }
+    }
+    match std::env::var("DUMMY_VALUE") {
+        Ok(val) => { assert_eq!(val, "hello world") },
+        Err(_e) => {
+            panic!("$DUMMY_VALUE is still unset.")
+        }
+    }
+}


garbados:
This PR fails tests for reasons I can't really understand:

```
error[E0061]: this function takes 1 parameter but 2 parameters were supplied
  --> src/main.rs:51:9
   |
51 |         schema::Context::new("../".into(), src)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 1 parameter
   | 
  ::: src/schema.rs:23:5
   |
23 |     pub fn new<S: Source + Send + Sync + 'static>(source: S) -> Self {
   |     ---------------------------------------------------------------- defined here

error[E0061]: this function takes 1 parameter but 2 parameters were supplied
  --> src/main.rs:56:9
   |
56 |         schema::Context::new("../".into(), src)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 1 parameter
   | 
  ::: src/schema.rs:23:5
   |
23 |     pub fn new<S: Source + Send + Sync + 'static>(source: S) -> Self {
   |     ---------------------------------------------------------------- defined here

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0061`.
error: Could not compile `proxy`.
```

Any thoughts?

garbados:
Nevermind, I understand the issue.

proxy/src/main.rs
```diff
@@ -26,9 +26,39 @@ mod server_warp;
...
+        Err(_e) => {
+            panic!("$DUMMY_VALUE is still unset.")
+        }
+    }
+}
+
 fn main() {
-    std::env::set_var("RUST_BACKTRACE", "full");
-    std::env::set_var("RUST_LOG", "info");
+    // enable backtrace if it isn't already
```
xla: Think we can omit all these comment as the line they are referring to are self-explanatory.

proxy/src/main.rs
```diff
@@ -26,9 +26,39 @@ mod server_warp;
 /// Origin of data required like the on-chain Registry.
 mod source;
 
+/// Convenience function for setting env vars only if they are unset
```
xla:
```suggestion
/// Convenience function for setting env vars only if they are unset.
```

proxy/src/main.rs
```diff
@@ -26,9 +26,39 @@ mod server_warp;
...
+fn set_if_unset (key: &str, val: &str) {
```
xla: Pretty neat and definitely less invasisve than the hardcoded value from before.

proxy/src/main.rs
```diff
@@ -26,9 +26,39 @@ mod server_warp;
...
+    match std::env::var(key) {
```
xla: You can shorten this with the help of `unwrap_or_else` which executs the closure given in the case the result is an `Err`. More details [here](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_else).

garbados: Great!

garbados:
I moved the function into `lib.rs` following the conventions in the Rust Book, and have otherwise responded to any outstanding feedback.

xla:
@garbados I think this [comment](https://github.com/radicle-dev/radicle-upstream/pull/66#discussion_r356114638) got lost in the move to `lib.rs`. :>

garbados:
Great points. I moved the file to `env.rs` so it's accessible with `mod env;`. I tried `use crate::env` but Rust complains it can't disambiguate between that and `std::env`:

```
29 | use crate::env::set_if_unset;
   |            ^^^
   |            |
   |            unresolved import
   |            help: a similar path exists: `std::env`
```

Regarding the comment about unwrap_or_else, you can see it in action [here](https://github.com/radicle-dev/radicle-upstream/pull/66/files#diff-386c416c52e285316f8c53fcdeef0ccdR4). Because `unwrap_or_else` expects the else function to return the same type as the env var (that is, a string) the env var is retrieved and returned after assignment.

PR #67, opened 2019-12-09
Add top-level readme

garbados:
This PR adds a top-level readme with basic information for new committers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,4 +13,3 @@ installation, usage, and testing instructions.
 Notes and guidelines for committers:
 
 - [Use tpope commit norms](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html): Capitalize commit messages, use the imperative present tense, wrap commit messages at 50 characters.
-- Check [Linear](https://linear.app/) for project backlogs, issues, discussion, etc.
--- a/README.md
+++ b/README.md
@@ -1,0 +1,16 @@
+# radicle-upstream
+
+A monorepo for related Radicle projects:
+
+- `app`: Cross-platform desktop app, currently used for product explorations.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,16 @@
...
+
+See the README.md of each respective project for further details, including
+installation, usage, and testing instructions.
+
+## Development
+
+Notes and guidelines for committers:
+
+- [Use tpope commit norms](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html): Capitalize commit messages, use the imperative present tense, wrap commit messages at 50 characters.
+- Check [Linear](https://linear.app/) for project backlogs, issues, discussion, etc.
```
xla: Are Linear backlogs public? If not it's a bit miss-leading to point to it in a public repo.

rudolfs:
Now that the `radicle-upstream` repository represents the actual desktop application that people will use to interact with the radicle protocol, should we maybe make whatever is in the `app` directory top-level and have the proxy as a sub-directory perhaps?

xla:
@rudolfs I still don't agree with that strategy. If we want to correct this, a hierarchy could look like this:

```
radicle-upstream/
|-- proxy/
     |-- src/
     |-- Cargo.toml
     ...
|-- ui
     |-- src/
     |-- package.json
     |-- rollup.config.js
     ...
|-- READM.md
|-- main.js
|-- rollup.config.js
```

Where the top-level is the app container (electron).

rudolfs:
@xla: ok, never mind then. Top-level README.md it is.

README.md
garbados: Oh, this repo is public isn't it. Yes that's true. I'll remove it.

garbados:
Removed the linear link. Merging...

PR #68, opened 2019-12-10
improve error handling MON-19

garbados:
This PR addresses [MON-19](https://linear.app/issue/MON-19), and is a WIP while I implement error handling. As of this writing, the `From` implementations have been converted to `TryFrom`, but compatibility has been retained through use of `.unwrap()` calls, thus defeating the purpose. This will become more resilient over time.


proxy/src/source.rs
```diff
@@ -31,21 +32,23 @@ pub struct ProjectId {
     pub domain: String,
 }
 
-impl From<radicle_registry_client::ProjectId> for ProjectId {
-    fn from(id: radicle_registry_client::ProjectId) -> Self {
-        Self {
+impl TryFrom<radicle_registry_client::ProjectId> for ProjectId {
+    type Error = &'static str;
+    fn try_from(id: radicle_registry_client::ProjectId) -> Result<Self, Self::Error> {
+        Ok(Self {
```
FintanH: I don't understand why this would need to be a `TryFrom` if it's _always_ going to return `Ok`. This seems like you're adding context, `Result`, that isn't actually needed :thinking:

proxy/src/source.rs
```diff
@@ -31,21 +32,23 @@ pub struct ProjectId {
...
 }
 
-impl Into<radicle_registry_client::ProjectId> for ProjectId {
-    fn into(self) -> radicle_registry_client::ProjectId {
-        (
+impl TryInto<radicle_registry_client::ProjectId> for ProjectId {
+    type Error = &'static str;
+    fn try_into(self) -> Result<radicle_registry_client::ProjectId, Self::Error> {
+        Ok((
             ProjectName::from_string(self.name).expect("project name creation failed"),
```
FintanH: This seems like an anti-pattern. You `expect` inside of a thing that returns a `Result`. I'd expect the use of `?` to propagate the error.

proxy/src/source.rs
```diff
@@ -77,13 +81,13 @@ impl From<radicle_registry_client::Project> for Project {
             })
             .collect();
 
-        Self {
-            id: p.id.clone().into(),
+        Ok(Self {
+            id: p.id.clone().try_into().unwrap(),
```
FintanH: Again, finding it weird that we call a partial function, `unwrap`, inside something that returns a `Result`.

FintanH:
Just to follow up with some of those comments. It's a common pattern to "propagate" errors with `Result` and `Option` by using the `?` operator (you can read more about error handling and propagating errors in [the book](https://doc.rust-lang.org/1.29.0/book/2018-edition/ch09-02-recoverable-errors-with-result.html))

The tl;dr is that if you have two functions that return `Result` with the same error type you can do:
```rust
let result1 = fn_that_returns_result()?;
let result2 = fn_that_returns_another_result()?;
do_something_with_results(result1, result2) // this also returns a Result
```

My other point was around the safety of `unwrap` and `expect`. They're what we would call partial functions because they either:
* Do not handle all cases of  a function and return a result
* Stop the execution of the program

We should try and minimise these kinds of functions and leave them to the outer edges of our program. For example, I would want to return `Result` for functions that can error, but if I'm using them in a CLI and there's no way to continue execution then I would `panic` and report the error to the user.

I hope that sheds more light on where I'm coming from :heart:

xla:
> We should try and minimise these kinds of functions and leave them to the outer edges of our program. For example, I would want to return `Result` for functions that can error, but if I'm using them in a CLI and there's no way to continue execution then I would `panic` and report the error to the user.

For context this is exactly what this PR tries to address and is outlined in the issue [MON-19](https://linear.app/issue/MON-19). Good to get reassurance that we try to improve in the right direction.

FintanH:
> For context this is exactly what this PR tries to address and is outlined in the issue MON-19. Good to get reassurance that we try to improve in the right direction.

Ah yes, I should have looked at the issue :sweat_smile: So I guess a follow up for me, after reading that, is that if we always return `Ok`, it only needs to be `From` and if it can fail that's when we would use `TryFrom` :)

garbados:
Thanks for the feedback! Yes, this PR is all about improving error handling. I've inventoried each instance of `.unwrap` or `.expect` [here](https://gist.github.com/garbados/1c78a5188569767d4823af74d1813269) so that each can be discussed and handled. I'll convert that inventory to a checklist shortly.

Regarding `From` and `TryFrom` I wonder how these type conversions can fail, as TryFrom implies such a possibility though it's hard for me to see that possibility in the code. In the inventory I've annotated these type conversions with the concern "What could go wrong?" because I don't yet understand what could go wrong but I assume something could, something which could be handled.

garbados:
Checklist:

source.rs
---------
- Source
  - get_all_projects
  	- [ ] expect: Retrieving project list from registry client failed.
      - This should bubble as the method simply wraps the registry client.
  	- [ ] expect: Retrieving project by ID failed.
      - This should bubble as the method simply wraps the registry client.
    - [ ] unwrap: radicle_registry_client::Project -> Project
      - Simple type conversion. What could go wrong?
  - get_project:
    - [ ] unwrap: ProjectID -> radicle_registry_client::ProjectID
      - Simple type conversion. What could go wrong?
  	- [ ] expect: Retrieving project by ID failed.
      - Failure should bubble, not be obscured by .expect
    - [ ] unwrap: radicle_registry_client::Project -> Project
      - Simple type conversion. What could go wrong?
  - register_project:
  	- [ ] expect: Project name creation failed.
      - This is a simple type conversion. What could go wrong?
    - [ ] expect: Project domain creation failed.
      - This is a simple type conversion. What could go wrong?
    - [ ] expect: Failed to register project with registry.
      - Called by schema::Mutation, which already wraps it in a Result.
      - So this method can be converted to return a Result, eliminating .expect
    - [ ] unwrap: radicle_registry_client::ProjectId -> ProjectId
      - Simple type conversion. What could go wrong?
- Ledger:
  - [x] enrich_members:
    - unwrap_or: Retrieves account or creates an anonymous account.

schema.rs
---------
- only one expect, inside a test (where i suspect it is appropriate?)

main.rs
-------
- [ ] expect: Missing <source> CLI parameter.
  - Should print a helpful message and quit -- not panic.
- [ ] expect: Creating registry client failed.
  - Print a helpful message about the failure and quit.

server_warp.rs
--------------
- build_response:
  - [ ] expect: "response is invalid"?
  - [ ] expect: "status code is invalid"?

FintanH:
> Regarding From and TryFrom I wonder how these type conversions can fail, as TryFrom implies such a possibility though it's hard for me to see that possibility in the code. In the inventory I've annotated these type conversions with the concern "What could go wrong?" because I don't yet understand what could go wrong but I assume something could, something which could be handled.

Do you mean any types at all or just the ones in this case? I think it's safe to say that if your type `T` has matching fields in type `U` where no conversion can fail then it's `From`. If there's the possibility that `T` may not have a field present that is needed for `U`, or if you can't go from `T.field -> U.field` safely, then you use `TryFrom`.

For example, given:
```rust
struct T {
  x: u8,
  y: u8,
}

struct U {
  x: u16,
  y: u16,
}
```

We can safely write `From` for `T -> U`, because it's safe to go from `u8 -> u16`. If you flip it around it's not safe and we'd have to do a check on the boundary of `u8` to convert to `u16`.

garbados:
Consider this type conversion:

```rust
impl From<radicle_registry_client::ProjectId> for ProjectId {
    fn from(id: radicle_registry_client::ProjectId) -> Self {
        Self {
            name: id.0.to_string(),
            domain: id.1.to_string(),
        }
    }
}
```

Is it possible for the `id` tuple to have a length other than 2? What should happen in that case? I see from the [registry source](https://github.com/radicle-dev/radicle-registry/blob/5cd58ad81c6cd1359b5754212fe63f57c4bb0ac4/runtime/src/registry.rs#L30) that it will always be a 2-length tuple with 2 String32s in it. Thus, wouldn't it make sense to use `From` over `TryFrom`?

Or consider the `Into` version of that conversion:

```rust
impl Into<radicle_registry_client::ProjectId> for ProjectId {
    fn into(self) -> radicle_registry_client::ProjectId {
        (
            ProjectName::from_string(self.name).expect("project name creation failed"),
            ProjectDomain::from_string(self.domain).expect("project domain creation faile"),
        )
    }
}
```

These `::from_string` calls effectively convert two strings to two String32s. The use of `.expect` indicates the system expects that failure is possible, but what should we do in that case? What could fail?

cloudhead:
Think of `.0` and `.1` like accessing a struct field, eg. `.id` or `.name`, so no it's not possible for eg. the `(A, B)` type to not have `.0` or `.1`, in fact it would result in a compile error.

garbados:
Great, then I think it makes sense for those conversions to remain `From` and `Into` rather than to become their `Try` variants, since failure is not to be expected.

cloudhead:
As an aside, the way to think about `panic!/expect/unwrap` is: *if this fails, is it a bug in the software?* If yes, then it should `panic`, otherwise it should return an error.

FintanH:
Ya so it seems like `From` works because on application side, we don't care about the 32 length. On the other hand, going `Into`, `from_string` is probably doing a check for the `String` length and failing. So I would say we want `TryInto`, propagating the conversion failure.

proxy/src/source.rs
```diff
@@ -31,21 +32,23 @@ pub struct ProjectId {
...
-        }
+        })
     }
 }
 
-impl Into<radicle_registry_client::ProjectId> for ProjectId {
-    fn into(self) -> radicle_registry_client::ProjectId {
-        (
+impl TryInto<radicle_registry_client::ProjectId> for ProjectId {
+    type Error = &'static str;
```
FintanH: So I think we want `type Error = String` to keep consistent with [`from_string`](https://github.com/radicle-dev/radicle-registry/blob/master/runtime/src/string32.rs#L17)

Then we can do:
```rust
let project_name = ProjectName::from_string(self.name)?;
let project_domain = ProjectDomain::from_string(self.domain)?;
Ok(project_name, project_domain)
```

garbados: Perfect, I was just about to ask this!

proxy/src/source.rs
```diff
@@ -217,7 +214,7 @@ where
         let project_name =
             ProjectName::from_string(name.to_string()).expect("project name creation failed");
         let project_domain =
-            ProjectDomain::from_string("rad".to_string()).expect("project domain creation faile");
+            ProjectDomain::from_string("rad".to_string()).expect("project domain creation failed");
```
FintanH: You could probably clean these two `let`s up using the `try_into` we just defined above, possibly?

garbados: No, that didn't work. Throws these sorts of errors:

```
error[E0382]: use of moved value: `name`
   --> src/source.rs:234:13
    |
210 |     fn register_project(&self, name: String, description: String, img_url: String) -> Result<Project, RegistryError> {
    |                                ---- move occurs because `name` has type `std::string::String`, which does not implement the `Copy` trait
...
215 |             ProjectName::from_string(name.into()).unwrap(),
    |                                      ---- value moved here
...
234 |             name,
    |             ^^^^ value used here after move
```

and

```
error[E0277]: `?` couldn't convert the error to `substrate_subxt::error::Error`
   --> src/source.rs:215:53
    |
215 |             ProjectName::from_string(name.try_into()?).unwrap(),
    |                                                     ^ the trait `std::convert::From<std::convert::Infallible>` is not implemented for `substrate_subxt::error::Error`
```

Not knowing what to do about it, I left it as `to_string` and replaced the `.expect` calls with `.unwrap` for now, until I understand enough to apply a more robust approach.

garbados:
OK! I've updated the checklist with the unwraps and expects that I unwrapped and expected. What remains has to do with type conversions, particularly where failure is not expected such as the String -> String32 conversion that populating registry::{ProjectName,ProjectDomain} requires. For now these conversions still use unwrap.

xla:
> Consider this type conversion:
> 
> ```rust
> impl From<radicle_registry_client::ProjectId> for ProjectId {
>     fn from(id: radicle_registry_client::ProjectId) -> Self {
>         Self {
>             name: id.0.to_string(),
>             domain: id.1.to_string(),
>         }
>     }
> }
> ```
> 
> Is it possible for the `id` tuple to have a length other than 2? What should happen in that case? I see from the [registry source](https://github.com/radicle-dev/radicle-registry/blob/5cd58ad81c6cd1359b5754212fe63f57c4bb0ac4/runtime/src/registry.rs#L30) that it will always be a 2-length tuple with 2 String32s in it. Thus, wouldn't it make sense to use `From` over `TryFrom`?
> 
> Or consider the `Into` version of that conversion:
> 
> ```rust
> impl Into<radicle_registry_client::ProjectId> for ProjectId {
>     fn into(self) -> radicle_registry_client::ProjectId {
>         (
>             ProjectName::from_string(self.name).expect("project name creation failed"),
>             ProjectDomain::from_string(self.domain).expect("project domain creation faile"),
>         )
>     }
> }
> ```
> 
> These `::from_string` calls effectively convert two strings to two String32s. The use of `.expect` indicates the system expects that failure is possible, but what should we do in that case? What could fail?

These two are great examples, in the first case we only do some data massaging. Ensuring it's in the right shape, nothing can fail here so a `From` implementation is perfectly fine.

For the second example we could actually pass in something that is not acceptable as `name` or `domain`. These might be for now just length constraints, but could easily be more elaborate down the line (subset of ASCII characters, reserved words, etc.). So the conversion failing is significant and we want a `TryInto`. The associated `Error` type should carry enough information to meaningfully signal back to the caller. In this case that a provided name or domain is invalid for certain reasons.

I hope with us starting to explore explicit error handling in the context of these conversions we begin to map out a catalogue of possible errors. This could take shape in form of an `enum` collecting all cases.

proxy/src/source.rs
FintanH: Ah, I might have been unclear. But I think you could roughly change this to:
```rust
let (project_name, project_domain) = ProjectId { name: name.to_string(), "rad".to_string }.try_into().expect("Failed to create registry::ProjectId");
```

That's me guessing that application's `ProjectId` constructor looks like this :sweat_smile: But basically the idea was to use the `try_into` so you're not repeating the `from_string` logic in both places :)

proxy/src/source.rs
```diff
@@ -91,11 +92,11 @@ impl From<radicle_registry_client::Project> for Project {
 pub trait Source {
```
FintanH: I would actually do one of two things here:
1. Parameterise the trait with a generic `Error`, i.e. `Source<Error>`
2. Add an associated `Error` type.

The reason being that traits should be general to the implementation, and `RegistryError` is specific to the final `registry` implementation. The benefit is that you may want to a test a "pure" implementation where you have guaranteed data and you can set the `Error` to just be `()`.

proxy/src/source.rs
```diff
@@ -171,50 +172,49 @@ where
...
-                    .registry_client
-                    .get_project(id)
-                    .wait()
-                    .expect("get_project failed");
-
-                match maybe_project {
-                    Some(project) => Some(self.enrich_members(Project::from(project))),
-                    None => None,
+                match self.get_project(id.into()) {
+                    Ok(maybe_project) => match maybe_project {
```
FintanH: We can use [`ok_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or) if you can return an error for `None` case here.

proxy/src/source.rs
```diff
@@ -171,50 +172,49 @@ where
...
-                let maybe_project = self
-                    .registry_client
-                    .get_project(id)
-                    .wait()
-                    .expect("get_project failed");
-
-                match maybe_project {
-                    Some(project) => Some(self.enrich_members(Project::from(project))),
-                    None => None,
+                match self.get_project(id.into()) {
```
FintanH: This can be `let maybe_project = self.get_project(id.into())?`.

Any time you do `match` followed by `Err(err) => Err(err)` you should be able to convert that to using a `?` :)

proxy/src/source.rs
```diff
@@ -171,50 +172,49 @@ where
...
         let maybe_project = self
             .registry_client
-            .get_project(id.into())
-            .wait()
-            .expect("get project failed");
+            .get_project(id.try_into().unwrap())
+            .wait()?;
 
-        match maybe_project {
+        let result = match maybe_project {
```
FintanH: This can be simplified to `maybe_project.map(|p| self.enrich_members(Project::from(p)))`

proxy/src/source.rs
```diff
@@ -227,16 +227,15 @@ where
...
                 },
             )
-            .wait()
-            .expect("osc project registration failed");
+            .wait()?;
 
-        Project {
-            id: registry_id.into(),
+        Ok(Project {
+            id: registry_id.try_into().unwrap(),
```
FintanH: I think we want to move this outside to `let id = registry_id.try_into()?;` and then this would propagate the error :)

proxy/src/schema.rs
```diff
@@ -54,11 +54,11 @@ impl Query {
     }
 
     fn projects(ctx: &Context) -> FieldResult<Vec<Project>> {
-        Ok(ctx.source.get_all_projects())
+        Ok(ctx.source.get_all_projects()?)
```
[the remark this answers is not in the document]
garbados: This fails because the return type `FieldResult` only expects certain kinds of errors, namely ones to do with Juniper:

```
error[E0308]: mismatched types
  --> src/schema.rs:61:9
   |
61 |         ctx.source.get_project(id.into())
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `juniper::FieldError`, found enum `substrate_subxt::error::Error`
```

Somehow wrapping it in an Ok satisfied the compiler, which is either magical or concerning. I'll switch it out to use an enum of errors.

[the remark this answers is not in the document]
FintanH: Hmmm maybe this is something to do with `FieldResult`.

So [`FieldResult`](https://docs.rs/juniper/0.14.1/juniper/type.FieldResult.html) is just a type wrapper around `Result`. So I guess there's some magic going on with type conversions using `?`.

Sorry for the wild goose chase!

proxy/src/source.rs
garbados: That makes sense to me but apparently not the compiler:

```
error[E0277]: `?` couldn't convert the error to `substrate_subxt::error::Error`
   --> src/source.rs:220:40
    |
220 |         let id = registry_id.try_into()?;
    |                                        ^ the trait `std::convert::From<std::convert::Infallible>` is not implemented for `substrate_subxt::error::Error`
```

FintanH: I _think_ that's because you need to match the error type that is expected for this function. So in this case it's `RegistryError`, but `try_into` errors with a `String`.

So, we're getting into error handling territory that's probably worth breaking down some more. The main thing to keep in mind is that we need an error type that can encapsulate any error that can be thrown. The choices we make here depend on ergonomics (and sometimes personal preference).

If we just look at `Result` and it's instance for `Try`, allowing us to use `?`, we can get an idea of what I'm alluding to above. Let's say we have two results `registry_result: Result<Foo, RegistryError>` and `application_result: Result<Bar, ApplicationError>`.

If we were to do manually matching, in place of `?`, this would look like:
```rust
let process_registry = match registry_result {
  Ok(foo) => bar(foo)
  Err(registry_err) => return Err(registry_err)
}

let process_application = match application_result {
  Ok(bar) => baz(bar),
  Err(application_err) => Err(application_err),
}
```

The thing is that upon type-checking `RegistryError` is not the same as `ApplicationError`. So we will get a type error saying so.

**BUT**, there's something else that's sneaky about `?` that I only noticed by accident. Say if you had an impl for `Into` from `RegistryError` to `ApplicationError`, `?` will use that an convert it under the hood. So if your top-level signature is `Result<Baz, ApplicationError>` it will know that `RegistryError` can be converted and do that for you. I'm not sure about the internals here.

So there are two tactics that come to mind here, but maybe people can chime in with more options.

1. Make an `enum` of the types of errors that you can run into, and if there's outside cases then you can add that to the `enum` and write an `Into` instance. You can see `radicle-surf` using this [here](https://github.com/radicle-dev/radicle-surf/blob/master/src/vcs/git.rs#L49-L56).

2. Use `String. It's easy to turn things into `String` so you can always just use that.

You may be asking why would you use one over the other? 

- `enum` captures the error cases that your functions can call (but technically can still do things like `panic`), but are hard to extend and you get into a kind of `enum` hell, matching on layer after layer.

- `String` has little information as to what the error was, you can't match on particular cases, but extends easily because of being able to just represent anything as a `String`.

Welcome to the world of explicit error handling :hugs:

garbados: Coming from JavaScript where error handling is all about opaque strings, I'm inclined to go with the enum approach... or a ProxyError type with From methods for everything we expect?

FintanH: I think the `enum` approach is the `ProxyError` type with `From` method for `RegistryError` :)
So you'd have an `enum ProxyError` that has the specific errors that occur in Proxy, and then one of the branches is `Registry(RegistryError)`

proxy/src/source.rs
```diff
@@ -88,14 +94,14 @@ impl From<radicle_registry_client::Project> for Project {
 }
 
 /// Abstraction used to fetch information from the registry.
-pub trait Source {
+pub trait Source<Error = RegistryError> {
```
FintanH: I think something got lost in translation :sweat_smile: What I was saying that you want to be generic over the `Error` instead of committing to `RegistryError`, which you're doing by use `Error = RegistryError` here.

What I'd probably expect is:
```rust
pub trait Source {
  type Error;
  /* rest of the code */
}
```

Then for here https://github.com/radicle-dev/radicle-upstream/blob/13388f4d3d4f569ab7f981c968abca6d13036d41/proxy/src/source.rs#L157 you would have:
```rust
impl<R> Source for Ledger<R> {
  type Error = RegistryError;
}
```

garbados: oh, fascinating. that makes more sense.

FintanH: Hence why I mentioned that a "pure" implementation might do `Error = ()` because you're guaranteed to get the data since you control it, for example :)

FintanH: And that would be for testing of course

xla:
First part of #71

proxy/src/schema.rs
```diff
@@ -1,4 +1,6 @@
 use juniper::{FieldResult, ParseScalarResult, ParseScalarValue, RootNode, Value};
+use radicle_registry_client::{Error as RegistryError};
+use substrate_subxt::error::{Error as SubstrateError};
```
geigerzaehler: `radicle_registry_client::Error` is a type alias for `substrate_subxt::Error`. So for the compiler these two types are the same. This means that you only need to import the one. We’re trying to move away from the subxt errors with https://github.com/radicle-dev/radicle-registry/issues/109

garbados:
Made a lot of progress today with help from @xla but ultimately ran into this issue:

```
error[E0271]: type mismatch resolving `<S as source::Source>::Error == schema::Error`
  --> proxy/src/schema.rs:79:21
   |
79 |             source: Arc::new(source),
   |                     ^^^^^^^^^^^^^^^^ expected enum `schema::Error`, found associated type
   |
   = note: expected type `schema::Error`
              found type `<S as source::Source>::Error`
```

The context is that the Source used to created a Context requires an associated `Error` type, but I haven't been able to figure out the syntax for passing it correctly. For reference, here's the declaration of Context where the `source` property is given certain type presumptions:

```rust
/// Container for data access from handlers.
#[derive(Clone)]
pub struct Context {
    /// Intermediate repo used to serve dummy data to be presented to the API consumer.
    dummy_repo_path: String,
    /// Origin of data needed to server APIs.
    source: Arc<dyn Source<Error = Error> + Send + Sync>,
}
```

I also implemented `Error::String` as a kind of placeholder for errors we haven't fully handled yet, so you can do things like this:

```
Err(format!("Directory at path {} is missing.", path))
```

proxy/src/schema.rs
```diff
@@ -1,18 +1,59 @@
...
 use radicle_surf::{
     file_system::{Path, SystemType},
     git::{git2, BranchName, GitBrowser, GitRepository, Sha1, TagName},
 };
 
 use crate::source::{AccountId, Project, ProjectId, Source};
 
+enum Error {
+    GitError(radicle_surf::git::GitError),
+    String(String)
```
FintanH: `String` is a bit of a weird error name 🤔 Maybe just use the name `Proxy` if they're specific to the application side of things.

proxy/src/schema.rs
```diff
@@ -366,19 +367,22 @@ impl Query {
...
+            Ok(root_dir)
         } else {
-            root_dir.find_directory(&path).expect(&format!(
-                "directory listing failed: {} -> {} | {:?}",
-                path,
-                path.is_root(),
-                prefix,
-            ))
-        };
+            // TODO log error
```
FintanH: This hopefully should be going away soon once I clean up some logic in surf :)

proxy/src/schema.rs
xla: Agreed with that, we can use something like `CatchAll` for the time being, as this is intended to be temporary.

xla:
> The context is that the Source used to created a Context requires an associated `Error` type, but I haven't been able to figure out the syntax for passing it correctly. For reference, here's the declaration of Context where the `source` property is given certain type presumptions:

As mentioned yesterday we should drop the construct of the associated error type for `Source`. While it is fun as an exercise I doubt that the `Source` abstraction is going to survive for much longer. As a sidenote the associated type is declared in the `impl Source for` of the concrete implementation.

xla:
To help with the matter you can remove the call to the registry from `schema.rs` all together.

proxy/src/schema.rs
garbados: `CatchAll` applied.

garbados:
Great, thanks everyone. I've got tests passing and made some dents in unwrap/expect usage. For now I'm focusing on eliminating them syntactically and leaving TODOs to highlight the need for more detailed handling, ex: the CatchAll error.

garbados:
Ok! I've expanded the `Error` enum to remove more instances of `.expect`, and to disambiguate more `CatchAll` errors into more specific types, but I seem to have caused a test failure in the refactor:

```
Diff < left / right > :
<[
<    ExecutionError {
<        location: SourcePosition {
<            index: 77,
<            line: 1,
<            col: 16,
<        },
<        path: [
<            "tree",
<        ],
<        error: FieldError {
<            message: "Repository has an empty commit history.",
<            extensions: Object(
<                Object {
<                    key_value_list: [
<                        (
<                            "type",
<                            Scalar(
<                                String(
<                                    "EMPTY_COMMIT_HISTORY",
<                                ),
<                            ),
<                        ),
<                    ],
<                },
<            ),
<        },
<    },
<]
>[]

...

failures:
    schema::tests::query_tree_root
```

I'm not sure why my changes have caused this failure. Here's where the error originates, in `Query::tree`:

```
let last_commit = match &browser.last_commit(&path)
    Some(last_commit) => Ok(Commit::from(last_commit)),
    None => Err(radicle_surf::git::GitError::EmptyCommitHistory),
}?;
```

Any guidance appreciated.

garbados:
So upon further reading and analysis, it seems the error I've run into *should be thrown* as the given repository is never given a commit history. Instead it is simply initialized, in `setup_fixtures`. In order to satisfy the test `query_tree_root`, the repository should be given a commit history, but I'm not sure how given the current stack.

proxy/src/schema.rs
```diff
@@ -209,8 +323,15 @@ impl Mutation {
         name: String,
         description: String,
         img_url: String,
-    ) -> FieldResult<Project> {
-        Ok(ctx.source.register_project(name, description, img_url))
+    ) -> Result<Project, Error> {
+        let err_name = &name.clone();
+        match ctx.source.register_project(name, description, img_url) {
```
FintanH: This pattern of keeping the `Ok` intact while changing the `Err` side is captured with `map_err` on a `Result` :)

proxy/src/schema.rs
```diff
@@ -223,36 +344,28 @@ impl Query {
...
         p.append(&mut Path::from_string(&path));
-        let file = root
-            .find_file(&p)
-            .expect(&format!("unable to find file: {} -> {}", path, p));
-        let last_commit = browser
-            .last_commit(&p)
-            .expect(&format!("[blob] unable to get last commit: {}", p));
+        let file = match root.find_file(&p) {
+            Some(result) => Ok(result),
+            None => Err(Error::FileNotFound(format!("unable to find file: {} -> {}", path, p))),
```
FintanH: And this pattern is captured with `ok_or`, where it turns `Some` into `Ok` and `None` into an `Err` with the error you supply

proxy/src/schema.rs
```diff
@@ -223,36 +344,28 @@ impl Query {
...
+        }?;
+        let last_commit = match browser.last_commit(&p) {
```
FintanH: Same here :)

proxy/src/source.rs
```diff
@@ -1,7 +1,31 @@
 use futures::future::Future;
 use std::collections::HashMap;
+use std::convert::{TryInto};
 
-use radicle_registry_client::{CryptoPair as _, ProjectDomain, ProjectName, H256};
+use radicle_registry_client::{
+    CryptoPair as _, ProjectDomain, ProjectName, H256, Error as RegistryError
+};
+
+type BadInput = String;
```
FintanH: I find it confusing having two `BadInput`s, this one and the variant in `SourceError`

FintanH:
> So upon further reading and analysis, it seems the error I've run into should be thrown as the given repository is never given a commit history. Instead it is simply initialized, in setup_fixtures. In order to satisfy the test query_tree_root, the repository should be given a commit history, but I'm not sure how given the current stack.

@garbados There is a history since a browser is initialised with HEAD implicitly. This just means we couldn't find the `last_commit` for a BLOB. The thing is though that the logic in that function isn't correct anyway... It's something I've promised to look into and fix :innocent: I think throwing a `EmptyHistory` in the code is probably the wrong thing to do.

Going forward I would say:
* Add a case for `LastCommitNotFound` since in reality no last commit found would be due to looking up a `Path` in the `History` that doesn't exist (maybe because the history changed in between rendering the directory and doing `last_commit`, for example).
* Possibly disable that test so that it doesn't bog you down for now. It is strange that you're getting an error now though because there was an `expect` which should have caused a failure too.

garbados:
Thanks for the guidance @FintanH :)

I've added a LastCommitNotFound error case and tried to make the code more idiomatic. I'm having some issues with `execute_query` but I'll write more about that this afternoon -- at present I have to rush out the door :dash:

proxy/src/schema.rs
```diff
@@ -223,36 +358,30 @@ impl Query {
...
-            .expect(&format!("[blob] unable to get last commit: {}", p));
+        let file = if let Some(result) = root.find_file(&p) {
+            Ok(result)
+        } else {
+            let error_message = format!("unable to find file: {} -> {}", path, p);
+            let error = Error::FileNotFound(error_message);
+            Err(error)
+        }?;
+        let last_commit = browser.last_commit(&p)
+            .ok_or(radicle_surf::git::GitError::EmptyCommitHistory)?;
```
FintanH: Looks like you missed using `LastCommitNotFound` here

proxy/src/schema.rs
```diff
@@ -343,34 +462,34 @@ impl Query {
...
                     path.push(label.clone());
                     path
                 };
-                let last_commit = Commit::from(&browser.last_commit(&entry_path).expect(&format!(
-                    "[tree] unable to get entry last commit: {}",
-                    entry_path
-                )));
+                let last_commit = match &browser.last_commit(&path) {
+                    Some(last_commit) => Ok(Commit::from(last_commit)),
+                    None => Err(radicle_surf::git::GitError::EmptyCommitHistory),
```
FintanH: Also here?

garbados:
OK! I've eliminated the `CatchAll` errors in place of more specific handlers. Further, the `String` values previously associated with most errors have been replaced with more indicative types such as `radicle_surf::file_system::Path`. The `String` type is still used when the underlying error is an opaque string, such as for `BadProjectName` which is just a wrapper around [a String32 conversion error](https://github.com/radicle-dev/radicle-registry/blob/83003d6c6173a22cd47542a612214e0a9a164d5b/runtime/src/string32.rs#L34-L37).

FieldError handlers for the `schema::Error::Registry` variant have been expanded to match the implementation in the [registry](https://github.com/radicle-dev/radicle-registry/blob/a32733a831fecd43e9129bab66518c962b6798da/client/src/error.rs). The way it's written now angers rustc because this branch doesn't use the latest radicle-registry code, but it should be ready for review.

Thanks again for all your help and guidance, everyone! I feel much more confident in reading and writing Rust now, and very appreciative for the compiler's assistance and insistence.

garbados:
PR has been updated to comply with, and build with, the latest registry client code.

proxy/src/error.rs
```diff
@@ -0,0 +1,155 @@
...
+            SourceError::Registry(error) => Self::Registry(error),
+        }
+    }
+}
+
+impl IntoFieldError for Error {
+    fn into_field_error(self) -> FieldError {
+        match self {
+            Self::Git(git_error) => {
+                match &git_error {
```
xla: These sub-matches are extending the function body quite a bit, maybe we can split them out. Or if possible implement `IntoFieldError` for the contained errors.

proxy/src/error.rs
```diff
@@ -0,0 +1,155 @@
...
+                            format!("Internal Git error: {:?}", error),
+                            graphql_value!({
+                                "type": "INTERNAL"
+                            })
+                        )
+                    },
+                }
+            },
+            Self::Registry(reg_error) => {
+                match reg_error {
```
xla: Same as above.

proxy/src/error.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[derive(Debug)]
+/// Enumerable of expected error types.
+pub enum Error {
+    /// File at a given path was irretrievable.
+    FileNotFound(Path),
+    /// Directory at a given path was irretrievable.
+    DirectoryNotFound(Path),
+    /// Project name exceeded 32 characters.
+    BadProjectName(String),
```
xla:
```suggestion
    ProjectNameValidation(String),
```
This would be my suggestion, disregard if not an improvement.

xla: This could hold all possible project name validation errors. Which to start with would be an enum with a single variant.

``` rust
enum ProjectNameValidation {
    TooLong(u32)
}
```

proxy/src/error.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Project domain exceeded 32 characters.
+    BadProjectDomain(String),
```
xla:
```suggestion
    ProjectDomainValidation(String),
```

xla: Same as for name, could carry another more specific error.

proxy/src/source.rs
```diff
@@ -1,7 +1,27 @@
...
+    CryptoPair as _, ProjectDomain, ProjectName, H256, Error as RegistryError
+};
+
+#[derive(Debug)]
+/// Enumerable of expected error types.
+pub enum Error {
+    /// Registry client errors.
+    Registry(RegistryError),
+    /// Thrown when a project name exceeds 32 characters.
+    BadProjectName(String),
```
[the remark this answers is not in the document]
garbados: Short answer: Yes. I'm reconfiguring to eliminate the repetition.

Long answer: The validation errors originate in source.rs, so I gave it its own error implementation which error.rs then wrapped. As you point out, it makes more sense to place those definitions in error.rs and have source.rs import the necessary enums.

garbados:
This is great feedback, thanks. I'm applying the suggested changes as able. Once I've wrapped that up, I'll reply to individual comments about how I handled raised issues.

proxy/src/error.rs
garbados: I can't implement IntoFieldError directly on the Git error types because they belong to radicle_surf, not the proxy. Instead I broke the matching out into a `convert_git_error_to_field_error` function -- same for RegistryError, which belongs to radicle_registry_client.

proxy/src/schema.rs
```diff
@@ -752,6 +875,20 @@ mod tests {
...
+        let mut vars = Variables::new();
+        let mut id_map: IndexMap<String, InputValue> = IndexMap::new();
+
+        id_map.insert("domain".into(), InputValue::scalar("rad"));
+        id_map.insert("name".into(), InputValue::scalar("upstream"));
+        vars.insert("id".into(), InputValue::object(id_map));
+
+        let (_res, errors) = execute_query("query($id: IdInput!) { tags(id: $id) }", &vars);
+
+        assert_eq!(errors, []);
```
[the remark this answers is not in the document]
garbados: That's true. Looking into this...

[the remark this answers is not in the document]
garbados: I added this error with intent to try and trigger errors but, alas, I'm having trouble getting it to throw an error!

proxy/src/error.rs
garbados: This is good. Implemented.

proxy/src/schema.rs
```diff
@@ -209,8 +357,9 @@ impl Mutation {
         name: String,
         description: String,
         img_url: String,
-    ) -> FieldResult<Project> {
-        Ok(ctx.source.register_project(name, description, img_url))
+    ) -> Result<Project, Error> {
+        ctx.source.register_project(name, description, img_url)
+            .map_err(|error| error.into())
```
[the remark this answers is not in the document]
garbados: I added a new `From` implementation that eliminates the need for this `.into` call entirely :+1:

garbados:
I figure next up is eliminating any merge conflicts with master.

garbados:
Ok, so...

It appears that in master, all references to `source.rs` have been removed. It's still named as a mod, but nothing appears to use it. Meanwhile the use of `librad` has changed which errors are possible, and those errors are encapsulated by `schema/error.rs`. Given this, there are two paths I see:

1. Delete source.rs, schema.rs, and error.rs, as their functionality has been moved or otherwise made redundant.
2. Integrate error.rs into schema/error.rs, somehow...?

garbados:
I hesitated to enumerate all the possible git2 errors, but I certainly could if we decide that's worthwhile.

PR #69, opened 2019-12-11
Upgrade rollup-plugin-terser to avoid vulnerability

rudolfs:
There is a potential XSS vulnerability in `serialize-javascript`: https://github.com/radicle-dev/radicle-upstream/network/alert/app/yarn.lock/serialize-javascript/open


.buildkite/run.sh
```diff
@@ -29,7 +29,7 @@ else
 fi
 
 echo "--- Building proxy"
-(cd app && yarn proxy:build)
+(cd app && ELECTRON_ENABLE_LOGGING=1 yarn proxy:build)
```
xla: This is intended to stay?

rudolfs: Yeah, we have a case of a flaky test, this will allow us to catch it. It's also good to keep it in on CI permanently, so we see all the Electron console messages that would otherwise be hidden on CI.

xla: @rudolfs Can we open an issue for that test with some output?

rudolfs: @xla https://linear.app/issue/MON-22

PR #70, opened 2019-12-11
Project creation

rudolfs:
This will enable users to create new projects backed by either an existing git repository or a newly created one. Since we wired up the `+` button in the sidebar we had to move the design system to a modal. For this we also added hotkeys: `?` for help and `d` for design system.

Better validations and tests are coming in separate PRs.


kim:
I'm not sure I can parse all this correctly, so here are just a few notes on how radicle-dev/radicle-link#18 works currently:

* A project is created from an existing `git` repo, which has to have at least 1 commit
* I settled on requiring to pass a `UserProfile` value to initialise a project, which minimally must contain the `nick`. This is not ideal, because I think we ultimately want the default experience to point to a "profile project" instead -- this requires a bit more thought, because that project obviously needs to exist first locally, and we want only one hop of URLs. For now, inline profiles should do, perhaps you can store a "preset" in local storage like I did for the CLI.

rudolfs:
Thanks for the feedback @kim!

> A project is created from an existing git repo, which has to have at least 1 commit

I had a chat with @cloudhead, this actually makes things easier for us. Before we thought of letting the user choose to either create a new repo or use an existing one. This means we will only have one option. I'll have a chat with Julien to update the UX flow for this.

Btw. here's the original spec that this implementation is based on: 
https://www.notion.so/monadic/Project-Creation-a24698de083a45a18abd75c4993ddb39

> I settled on requiring to pass a UserProfile value to initialise a project, which minimally must contain the nick. This is not ideal, because I think we ultimately want the default experience to point to a "profile project" instead -- this requires a bit more thought, because that project obviously needs to exist first locally, and we want only one hop of URLs. For now, inline profiles should do, perhaps you can store a "preset" in local storage like I did for the CLI.

We haven't thought about identities or profiles yet, so for now we'll just use something hardcoded and once we have the identity we'll pass it down.

app/src/App.svelte
```diff
@@ -40,26 +45,37 @@
...
 
-<style>
-  .container {
-    position: relative;
-    left: var(--slim-sidebar-width);
-    width: calc(100vw - var(--slim-sidebar-width));
-    overflow-x: hidden;
-  }
-</style>
+  hotkeys("d,shift+d", () => {
```
xla: Love the start of hotkeys in the app. Can we capture a follow-up to modularise properly?

app/src/DesignSystem/Button.svelte
```diff
@@ -2,12 +2,15 @@
...
+  export let disabled = null;
   export let icon = null;
+  export let style = null;
 
   let iconClass = icon ? (size === "big" ? "icon" : "icon-small") : null;
-  let disabledClass = disabled ? "disabled" : null;
-  let buttonClass = [variant, size, iconClass, disabledClass].join(" ");
+
+  // we want to dynamically change whether a button is disabled or not
+  $: disabledClass = disabled ? "disabled" : null;
```
xla: Why not have `disabled` be `null|"disabled"`.

app/src/DesignSystem/DirectoryInput.svelte
```diff
@@ -0,0 +1,32 @@
...
+  import Text from "./Text";
+
+  export let placeholder = "";
+  export let style = null;
+  export let path = "";
+
+  let files;
+
+  const openFileDialog = async () => {
+    path = await window.electron.ipcRenderer.invoke(MAIN_IPC_CHANNEL);
```
xla: Can you add a comment to documentation on how this works? As I remember the normal file dialog not being sufficient.

app/src/pages/DesignSystem.svelte
```diff
@@ -55,284 +57,379 @@
...
-  <div>
-    <Select items={['master', 'dev']} value={'dev'} />
+  <div class="section">
+    <div class="first-column">
+      <Title.Big>Components</Title.Big>
+      <Text.Regular style="color: var(--color-gray)">
+        More complex UI elements
+      </Text.Regular>
+    </div>
+    <div class="second-column">
```
xla: Can we add a heading or some similar text to show that this is a notifiction we are looking at?

app/src/components/NotificationFaucet.svelte
```diff
@@ -0,0 +1,59 @@
+<script>
```
xla: Cute file name, like it!

app/src/components/Projects.svelte
```diff
@@ -1,26 +1,31 @@
...
     }
   `;
 
-  const projects = query(getClient(), { query: GET_PROJECTS });
+  const client = getClient();
+  const projects = query(client, { query: GET_PROJECTS });
+  projects.refetch();
+
+  const createProject = () => {
+    push("/projects/new");
```
xla: Why do we use push directly, instead of an href? Also can centralise routing strings?

app/src/pages/CreateProject.svelte
```diff
@@ -0,0 +1,364 @@
...
+    }
+  };
+
+  let localBranches = "";
+
+  const fetchBranches = async path => {
+    // Reset to defaults whenever the path changes so that we show the defaults
+    // in case this query fails or the user clicks cancel in the directory
+    // selection dialog.
+    localBranches = "";
```
xla:
```suggestion
    localBranches = [];
```

app/src/pages/CreateProject.svelte
```diff
@@ -0,0 +1,364 @@
...
+          ((prevHash << 5) - prevHash + currVal.charCodeAt(0)) | 0,
+        0
+      );
+  };
+
+  const createProject = async () => {
+    let response;
+
+    try {
+      response = await mutate(client, {
```
xla: Is this how we do manual gql queries? Pretty cool.

app/src/pages/CreateProject.svelte
```diff
@@ -0,0 +1,364 @@
+<script>
```
xla: This is getting quite large, can we capture a follow-up where we break it apart in components? Or would that be not warranted?

app/src/App.svelte
xla: Maybe let's stick to one modifier for now and not use single characters to avoid surprises.
```suggestion
  hotkeys("shift+d", () => {
```

app/src/pages/Project.svelte
```diff
@@ -79,15 +76,18 @@
...
+    <ProjectSidebar />
 
-  <div class="container">
-    <ProjectBreadcrumbs
-      style="position: fixed; top: 0;"
-      project={result.data.project} />
-    <div class="layout">
-      <Router {routes} />
+    <div class="container">
+      <NotificationFaucet />
```
xla: Shouldn't notifications be shown globally in the app?

app/src/stores.js
```diff
@@ -3,3 +3,9 @@ import { writable } from "svelte/store";
 export const revision = writable(null);
 export const objectPath = writable(null);
 export const objectType = writable(null);
+
+export const notification = writable(null);
+
+export const showNotification = message => {
```
xla: Great use of the store.

app/src/components/Projects.svelte
rudolfs: Because it is an async action via a button push. How would you do it with a href?

rudolfs: +1 for routing strings.

app/src/pages/CreateProject.svelte
rudolfs: Yep: https://github.com/timhall/svelte-apollo#mutate

app/src/App.svelte
rudolfs: 👍https://linear.app/issue/MON-37

app/src/DesignSystem/Button.svelte
rudolfs: I don't think I follow. How would that work?

app/src/pages/CreateProject.svelte
rudolfs: I would love to!
https://linear.app/issue/MON-38

app/src/pages/Project.svelte
rudolfs: Indeed, this was my first stab at notifications. Improvements will go here: https://linear.app/issue/MON-34.

app/src/stores.js
rudolfs: 🙏

PR #76, opened 2020-01-09
Improve project creation validations

rudolfs:
Handle the following edge cases:

- new project + new repo:
  - [x] directory has already a git repo in it -> validation error
- new project + existing repo:
  - [x] directory is empty -> validation error
  - [x] directory is already managed by Radicle -> validation error

Talk with @julien about error feedback improvements (show it ~~on the right-hand side?~~ below).

### DIFF
--- a/app/cypress/integration/navigation.spec.js
+++ b/app/cypress/integration/navigation.spec.js
@@ -35,7 +35,7 @@ context("navigation", () => {
       cy.get('[data-cy="page"]').should("contain", "Profile");
 
       cy.get('[data-cy="sidebar"] [data-cy="new-project"]').click();
-      cy.get('[data-cy="page"]').should("contain", "Create a new project");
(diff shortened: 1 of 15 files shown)


xla:
@rudolfs Are we testing these validatiosn?

rudolfs:
> The CreateProject page is growing in size, one first step to make it more pallatable is to pull out the validation rules into a standard js file - this way they can be shared as well.

It's a bit tricky, as the validations are coupled to some of the logic on the page, but it's on my radar: https://linear.app/issue/MON-38

rudolfs:
Added specs for most of the validations. However, I couldn't properly test the directory validations as Cypress doesn't yet have support for native electron testing merged into their pre-release [MON-40](https://linear.app/issue/MON-40). This also blocks us from properly testing project creation [MON-28](https://linear.app/issue/MON-28).

PR #78, opened 2020-01-10
Integrate new transaction flow

garbados:
This PR implements the Radicle client's new registration syntax. This is a draft implementation for discussion.

Still TODO:

- [ ] Eliminate documentation TODOs.
- [ ] Implement a project registration endpoint.


garbados:
Some questions:

- Does `register.rs` make sense as a filename for this functionality?
- Should the registration implementation file live somewhere in particular?
- How does `POST /register` sound for an endpoint?

rudolfs:
> How does POST /register sound for an endpoint?

I'd rather prefer `registerProject` so it is in-line with the rest of our GraphQL schema:
https://github.com/radicle-dev/radicle-upstream/blob/master/app/schema.graphql#L108.

xla:
> * Does `register.rs` make sense as a filename for this functionality?
> * Should the registration implementation file live somewhere in particular?

As this will be the place where we bundle registry functionality it might be better named `registry.rs`. While I love to see module(s) emerge which truly hold our product feature surface and can ideally be reused, we are not there yet. A testament to this is our sunsetted `source.rs`. That said I'd propose we keep endpoint and handlers under `schema`. To get an idea how that could work have a look at `schema/git.rs`.

> * How does `POST /register` sound for an endpoint?

Seeing the `POST` verb here indicates you were thinking of exposing it directly via HTTP. Did you plan to not expose it as a GraphQL mutation like our other API calls?

garbados:
> Seeing the POST verb here indicates you were thinking of exposing it directly via HTTP. Did you plan to not expose it as a GraphQL mutation like our other API calls?

Oh, I misunderstood the task parameters. I need to do more studying up on GraphQL. Exposing it as a GraphQL mutation sounds fine :+1:

rudolfs:
Had a look at the [project registration flow](https://www.figma.com/file/dnYlebm0ehgyJyT3Po9Kcl/Radicle-app?node-id=2442%3A1), here's what we need from the UI perspective:

  - get project registration costs (+ how much transaction fees cost)
  - get wallet balance given `walletId` (so we can check if we have enough funds to perform registration)
  - check if a project name is already taken
  - perform the project registration given `walletId`: xxx, `transactionType`: project-registration and return `transactionId` and `transactionStatus`
  - query the status of a transaction given `transactionId`
  - get a list of all transactions

Stretch-goal:
  - have a GraphQL subscription for the transaction status, so we can update it in real-time

xla:
@rudolfs Thanks for putting this together. Looks good for the most part, I think it's too early to dive into wallet functionality as its not exposed yet. For the transaction cost we can fake that for now, but it's good to get in the habit of accounting for it. A more flexible query would be:

``` graphql
type Query {
    transactionCosts(type: TransactionType!) TransactionCosts!
}
```

@rudolfs What did you refer to by `duration` in your OG proposal? Think we can hold of on orgs, until they solidified.

rudolfs:
> I think it's too early to dive into wallet functionality as its not exposed yet.
👍 

> @rudolfs What did you refer to by duration in your OG proposal?

@xla in "wallet explorations" -> 2.1 we have a screen that let's the user choose for how long the name should be registered for: "register for: 1 year".

Screenshot 2020-01-13 at 18 31 07


```
type Query {
    transactionCosts(type: TransactionType!) TransactionCosts!
}
```
I was thinking about this initially, but then I couldn't figure out how to parametrise this query (for example the PROJECT_REGISTRATION tx costs depend on the duration).

xla:
> @xla in "wallet explorations" -> 2.1 we have a screen that let's the user choose for how long the name should be registered for: "register for: 1 year".

Good point, I think we agreed to not make it configurable for now in the last flow review session and default to 1yr for now. Which doesn't solve the general issue that cost query needs to know about the shape of the transaction in order to give correct feedback. What could serve us well here are [Unions](http://spec.graphql.org/June2018/#sec-Unions) as they allow for a `one of` semantic.

rudolfs:
Hm, union output types solve this for now as we don't need to pass in the duration of the registration.

How would we do it if we needed to pass input params (like duration) in the future? Seems like union input types are still in [draft stage](https://github.com/graphql/graphql-spec/blob/master/rfcs/InputUnion.md).

I'd also like to point out that we'll have to figure out how to properly encode Rad. On the rust side maybe something like [rust_decimal](https://docs.rs/rust_decimal/1.1.0/rust_decimal/)? And JS one of these: [decimal.js](https://github.com/MikeMcl/decimal.js/) or [bignumber.js](https://github.com/MikeMcl/bignumber.js/).

Here's the updated schema:
  - dropped wallet
  - dropped `duration`
  - adjusted the `transactionCosts` query

```
Scalar Rad

union TransactionCosts = ProjectRegistrationCosts

ProjectRegistrationCosts {
  amount: Rad!
  fees: Rad!
  total: Rad!
}

enum TransactionState {
  PENDING
  SUCCEEDED
  FAILED
}

TransactionStatus {
  state: TransactionState!
  message: String
}

enum TransactionType {
  PROJECT_REGISTRATION
}

type Transaction {
  id: ID!
  status: TransactionStatus!
  costs: TransactionCosts!
  type: TransactionType!
  to: String!
  from: String!
  timestamp: Datetime!
}

type Query {
  transactionCosts(): TransactionCosts!
}

type Mutation {
  registerProject(projectId: ID!, name: String!, organization: String!): Transaction!
}
```

xla:
> Hm, union output types solve this for now as we don't need to pass in the duration of the registration.
> 
> How would we do it if we needed to pass input params (like duration) in the future? Seems like union input types are still in [draft stage](https://github.com/graphql/graphql-spec/blob/master/rfcs/InputUnion.md).

You are right, the limitations of the GraphQL type system strike again. This would mean we have to supprot distinct queries for now.

> I'd also like to point out that we'll have to figure out how to properly encode Rad. On the rust side maybe something like [rust_decimal](https://docs.rs/rust_decimal/1.1.0/rust_decimal/)? And JS one of these: [decimal.js](https://github.com/MikeMcl/decimal.js/) or [bignumber.js](https://github.com/MikeMcl/bignumber.js/).

So far there hasn't been a mention that we are going to have sub-rads/fractions of a rad. Which would indicate that the smallest possible unit is a single rad. Would you still want to have deicmal handling then?

Couple of comments regarding the schema:

* `TransactionCosts` doesn't need to be a union, the costs for transactions will always have the same shape
* there won't be a transaction type, rather
* a transaction will carry message(s) with the intended operation
* we should find a richer model for the status of a transaction to carry additional info

updated schema:

``` graphql
scalar Rad

type TransactionCosts {
  deposit: Rad!
  fees: Rad!
  total: Rad!
}

union TransactionState = Accepted | Applied | Rejected

type Accepted {
  # Address or other identity of the node that accepted the transaction.
  node: String!
}

type Applied {
  # Number of blocks that this transaction has been applied to.
  blocks: Int!
}

type Rejected {
  # The issue that prevented this transaction to be accepted/applied.
  reason: String!
}

union Message = ProjectRegistration | Transfer

type ProjectRegistration {
  name: String!
  domain: String!
}

 # Account id of sorts.
scalar AccountID

type Transfer {
  amount: Rad!
  from: AccountID!
  to: AccountID!
}

type Transaction {
  id: ID!
  messages: [Message!]!
  costs: TransactionCosts!
  state: TransactionState!
  timestamp: Datetime!
}

type Mutation {
  registerProject(domain: String!, name: String!): Transaction!
}

input TranscationCostInput {
  messages: [Message!]!
}

type Query {
  transactionCosts(draft: TranscationCostInput): TransactionCosts!
}
```

Disclaimer, I included `Transfer` and `AccountID` to illustrate the point, they are out of scope for this PR.

rudolfs:
> So far there hasn't been a mention that we are going to have sub-rads/fractions of a rad. Which would indicate that the smallest possible unit is a single rad. Would you still want to have deicmal handling then?

Clarified this with @cloudhead, who promptly [wrote it down](https://www.notion.so/monadic/Currency-57fe5cb1bc6d493aafd75acce4c774e1). So we can use some kind of `int` for the underlaying storage and always store and handle everything in μRAD. We'll convert them to RAD only in the UI.

xla:
> μRAD

Cute! Alternatively: Radrons

xla:
@garbados In order to make good on our intentions to keep PRs smaller I'd propose that we bring this one to conclusion by documentating the leftover TODOs and audit if we can test further. Break out the design discussion for the GraphQL endpoint into an issue and follow up with another PR which implements that. @rudolfs Any objections to that?

rudolfs:
xla: sounds good to me!

xla:
Broken this out into #81 - let's continue follow-ups there.

proxy/src/register.rs
```diff
@@ -0,0 +1,117 @@
...
+where
+    R: ClientT,
+{
+    /// TODO
+    fn new(client: R) -> Self {
+        Self { client }
+    }
+
+    /// TODO
+    async fn register_project(
```
xla: We should not jump into `async` just yet. It seems juniper will not play nicely. So for now we should use the blocking calls on the client with `wait` and later on `block_on` once radicle-dev/radicle-registry#138 is done.

garbados: Yes, I found this out the hard way. Sounds good.

garbados:
Just checked in where I'm at. I've been wrestling with async/await, discovering the hard way that it won't play nicely with Juniper and learning more about GraphQL derivations. I'll wrap this up by addressing the TODOs that @xla laid out.

garbados:
Note this issue:

> expected enum `url::host::Host`, found a different enum `url::host::Host`

This is because the proxy uses url [2.1](https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/Cargo.toml#L28) and the registry uses [1.7](https://github.com/radicle-dev/radicle-registry/blob/master/client/Cargo.toml#L26).

garbados:
The last TODO now lies in trying to figure out synchronous future handling.

proxy/src/schema/registry.rs
```diff
@@ -0,0 +1,88 @@
...
+use crate::schema::error::{Error, ProjectValidation};
+
+/// Registry client wrapper.
+#[derive(Clone)]
+pub struct Registry<R>
+where
+    R: ClientT,
+{
+    /// Registry client, whether an emulator or otherwise.
+    client: R,
```
geigerzaehler: It’s not necessary to make this polymorphic in the client. We changed this recently and now there exits only one implementation of client which is `Client`. So you could remove the type parameter `R` every where and replace it with `Client`.

Of course you might only want to rely on methods defined by `ClientT` in which case you can keep it like it is.

garbados:
TODOs addressed. Ready for review!

proxy/src/schema/registry.rs
garbados: This causes some confusing issues so I stuck with ClientT.

garbados:
Tests were failing due to a bracket that went missing during merge resolution. The bracket has been restored.

PR #82, opened 2020-01-14
Adjust to latest surf changes

xla:
Some changes from our upstream `radicle-surf` dependency. Adjusting to the new APIs and correct tests.

### DIFF
--- a/proxy/src/schema/error.rs
+++ b/proxy/src/schema/error.rs
@@ -79,7 +79,7 @@ fn convert_fs(error: &radicle_surf::file_system::error::Error) -> FieldError {
         radicle_surf::file_system::error::Error::Label(label_error) => match label_error {
             radicle_surf::file_system::error::Label::ContainsSlash => "Label contains slashes",
             radicle_surf::file_system::error::Label::Empty => "Label is empty",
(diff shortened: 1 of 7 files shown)


proxy/src/schema/error.rs
```diff
@@ -64,8 +73,29 @@ impl From<url::ParseError> for Error {
     }
 }
 
+/// Helper to convert `radicle_surf` `FileSystem` error to `FieldError`.
+fn convert_fs(error: &radicle_surf::file_system::error::Error) -> FieldError {
+    let error_str = match &error {
+        radicle_surf::file_system::error::Error::Label(label_error) => match label_error {
+            radicle_surf::file_system::error::Label::ContainsSlash => "Label contains slashes",
+            radicle_surf::file_system::error::Label::Empty => "Label is empty",
+            radicle_surf::file_system::error::Label::InvalidUTF8 => "Lable is not valid utf8",
```
rudolfs:
```suggestion
            radicle_surf::file_system::error::Label::InvalidUTF8 => "Label is not valid utf8",
```

xla:
@rudolfs Only one, related to still missing last commit for directories, which is forthcoming.

PR #94, opened 2020-01-21
First stab at macOS builds

rudolfs:
First attempt of mac builds. It is limited to building only the `master` branch of radicle-upstream, because making the host fool-proof and properly locked down via docker containers and file system isolation is out of scope for this quarter.

Depends on: https://github.com/oscoin/ci/pull/44


rudolfs:
Here's an example of a working mac-n-cheese build: https://buildkite.com/monadic/radicle-upstream/builds/895#2f63d0f0-5b01-4d53-bbe2-0ff8c3094aa6

And tete-de-moine still works as well: https://buildkite.com/monadic/radicle-upstream/builds/897#64d7ba14-36cf-4c09-894b-9a56114b5a76

PR #95, opened 2020-01-21
FIx project creation test assertions

xla:
:woman_facepalming:


xla:
@MeBrei Are the test names not indicative enough?

Issue #81, opened 2020-01-14
Project registration endpoint


xla:
We want to support registration on the Registry in the context of existing projects. For that we need a new mutation.

### Design
The schema addition proposal:

``` graphql
scalar Rad

type TransactionCosts {
  fees: Rad!
  total: Rad!
}

union TransactionState = Accepted | Applied | Rejected

type Accepted {
  # Address or other identity of the node that accepted the transaction.
  node: String!
}

type Applied {
  # Number of blocks that this transaction has been applied to.
  blocks: Int!
}

type Rejected {
  # The issue that prevented this transaction to be accepted/applied.
  reason: String!
}

union Message = ProjectRegistration

type ProjectRegistration {
  name: String!
  domain: String!
}

type Transaction {
  id: ID!
  messages: [Message!]!
  costs: TransactionCosts!
  state: TransactionState!
  timestamp: Datetime!
}

type Mutation {
  registerProject(domain: String!, name: String!): Transaction!
}

input TranscationCostInput {
  messages: [Message!]!
}

type Query {
  transactionCosts(draft: TranscationCostInput): TransactionCosts!
}
```

### Implementation roadmap

* [x] Consume Registry client for registration: #78 
* [x] Implement registration mutation
* [x] Migrate to futures-0.3 which landed in the client
* [x] Pass `metadata` once radicle-dev/radicle-registry#167 landed
* [ ] Store `librad::ProjectId` in `metadata`

geigerzaehler:
> ```graphql
> 
> type ProjectRegistration {
>   name: String!
>   domain: String!
> }
> ```

Project registration also requires an existing checkpoint.



> ```graphql
> type Mutation {
>   registerProject(domain: String!, name: String!): Transaction!
> }
> ```

Does `registerProject` submit the transaction? What is the purpose of the `Transaction.state` field?

xla:
> Project registration also requires an existing checkpoint.

Is that already the case and part of the examples in the client?

> Does `registerProject` submit the transaction? What is the purpose of the `Transaction.state` field?

Yes it submits the transaction. The state is to surface the lifecycle of transactions in the UI (notifications, progress, etc.).

geigerzaehler:
> Yes it submits the transaction. The state is to surface the lifecycle of transactions in the UI (notifications, progress, etc.).

[...]

PR #96, opened 2020-01-21
Project registration endpoint

xla:
Add mutation for project registration to enable integration in the UI. First part of #81.

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -122,7 +122,7 @@ union TransactionState = Applied
 
 type Applied {
   # Hash of the block it was applied to.
-  blocks: ID!
+  block: ID!
 }
 
 type Mutation {
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -104,10 +104,31 @@ ProjectMetadataInput {
   defaultBranch: String!
 }
 
-type Mutation {
-  createProject(project: ProjectMetadataInput!, path: String!, publish: Boolean!): Project!
+Transaction {
+  id: ID!
+  messages: [Message!]!
+  state: TransactionState!
+  timestamp: Datetime!
 }
 
+union Message = ProjectRegistration
+
+type ProjectRegistration {
+  domain: String!
+  name: String!
+}
+
+union TransactionState = Applied
+
+type Applied {
+  # Hash of the block it was applied to.
+  blocks: ID!
+}
+
+type Mutation {
+  createProject(project: ProjectMetadataInput!, path: String!, publish: Boolean!): Project!
+  registerProject(domain: String!, name: String!): Transaction!
+}
 
 ################
 # Exploration. #
--- a/proxy/src/schema/mod.rs
+++ b/proxy/src/schema/mod.rs
@@ -31,6 +31,7 @@ pub struct Context {
     dummy_repo_path: String,
     /// Root on the filesystem for the librad config and storage paths.
     librad_paths: Paths,
+    /// Wrapper to interact with the Registry.
     registry: registry::Registry,
 }
 
--- a/proxy/src/schema/mod.rs
+++ b/proxy/src/schema/mod.rs
@@ -2,6 +2,7 @@ use juniper::{RootNode, ID};
 use std::str::FromStr;
 
 use librad::paths::Paths;
+use radicle_registry_client::ed25519;
 use radicle_surf as surf;
 
 /// Error definitions and type casting logic.
@@ -30,14 +31,20 @@ pub struct Context {
     dummy_repo_path: String,
     /// Root on the filesystem for the librad config and storage paths.
     librad_paths: Paths,
+    registry: registry::Registry,
 }
 
 impl Context {
     /// Returns a new `Context`.
-    pub const fn new(dummy_repo_path: String, librad_paths: Paths) -> Self {
+    pub fn new(
+        dummy_repo_path: String,
+        librad_paths: Paths,
+        registry_client: radicle_registry_client::Client,
+    ) -> Self {
         Self {
             dummy_repo_path,
             librad_paths,
+            registry: registry::Registry::new(registry_client),
         }
     }
 }
@@ -71,6 +78,28 @@ impl Mutation {
         Ok(project::Project {
             id: id.to_string().into(),
             metadata: meta.into(),
+        })
+    }
+
+    fn register_project(
+        ctx: &Context,
+        domain: String,
+        name: String,
+    ) -> Result<registry::Transaction, Error> {
+        // TODO(xla): Get keypair from persistent storage.
+        let fake_pair = ed25519::Pair::from_legacy_string("//Robot", None);
+        ctx.registry
+            .register_project(&fake_pair, domain.clone(), name.clone())?;
+
+        Ok(registry::Transaction {
+            id: juniper::ID::new("123"),
+            messages: vec![registry::Message::ProjectRegistration(
+                registry::ProjectRegistration {
+                    domain: domain,
+                    name: name,
+                },
+            )],
+            timestamp: "now".to_string(),
         })
     }
 }
@@ -342,7 +371,11 @@ mod tests {
     where
         F: FnOnce(Value, Vec<ExecutionError<DefaultScalarValue>>) -> (),
     {
-        let ctx = Context::new(REPO_PATH.into(), librad_paths);
+        let ctx = Context::new(
+            REPO_PATH.into(),
+            librad_paths,
+            radicle_registry_client::Client::new_emulator(),
+        );
         let (res, errors) =
             juniper::execute(query, None, &Schema::new(Query, Mutation), vars, &ctx)
                 .expect("test execute failed");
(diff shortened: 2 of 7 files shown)


proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
+            .submit_transaction(checkpoint_tx)
+            .await?
+            .await?
             .result?;
+
+        // Prepare and submit project registration transaction.
+        let register_message = message::RegisterProject {
+            id: (project_name, project_domain),
+            checkpoint_id,
+            metadata: registry::Bytes128::from_vec(vec![]).expect("unable construct metadata"),
```
garbados: Why not use `?`?

xla: It's not meaningful to implement `From` for a `String` as it looses the context of the origin of the error. For that reason I created radicle-dev/radicle-registry#185.

proxy/src/schema/mod.rs
```diff
@@ -73,6 +82,45 @@ impl Mutation {
             metadata: meta.into(),
         })
     }
+
+    fn register_project(
+        ctx: &Context,
+        domain: String,
```
NunoAlexandre: Considering https://github.com/radicle-dev/radicle-registry/issues/152, shouldn't we here also validate the `domain` to be `.rad`. Or better, to not receive `domain` since only `.rad` will be supported?

proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
 };
 
 use crate::schema::error::{Error, ProjectValidation};
 
+/// A container to dissiminate and apply operations on the [`Registry`].
+#[derive(GraphQLObject)]
+pub struct Transaction {
+    /// Unique identifier, in actuality the Hash of the transaction. This handle should be used by
+    /// the API consumer to query state changes of a transaction.
+    pub id: juniper::ID,
```
NunoAlexandre: It seems to me that if `id` is the `hash`, the types should reflect that. or do you want, intentionally, want to keep that separation?

proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
+    /// Creation time.
+    pub timestamp: String,
+}
+
+/// Required information to issue a new project registration on the [`Regisry`].
+#[derive(juniper::GraphQLObject)]
+pub struct ProjectRegistration {
+    /// The domain/namespace the project should be registered for.
+    pub domain: String,
+    /// The name of the project, which MUST be uniqure for the domain.
```
NunoAlexandre: typo on "uniqure".

proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
-    pub fn register_project(
+    pub async fn register_project(
         &self,
-        author: &Pair,
-        name: String,
+        author: &ed25519::Pair,
         domain: String,
-    ) -> Result<(), Error> {
+        name: String,
+    ) -> Result<H256, Error> {
```
geigerzaehler: You can use `radicle_registry_client::TxHash` to make the type more descriptive. (It’s an alias)

proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
+    /// Current state of the transaction.
+    pub state: TransactionState,
+    /// Creation time.
+    pub timestamp: String,
+}
+
+/// Required information to issue a new project registration on the [`Regisry`].
+#[derive(juniper::GraphQLObject)]
+pub struct ProjectRegistration {
+    /// The domain/namespace the project should be registered for.
```
NunoAlexandre: nit: the "/namespace" is unnecessary and noisy imo.

proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
+    /// List of operations to be applied to the Registry. Currently limited to exactly one. We use
+    /// a Vec here to accommodate future extensibility.
+    pub messages: Vec<Message>,
+    /// Current state of the transaction.
+    pub state: TransactionState,
+    /// Creation time.
+    pub timestamp: String,
+}
+
+/// Required information to issue a new project registration on the [`Regisry`].
```
NunoAlexandre: typo on "Regisry"

proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
-            .wait()?
-            .wait()?
-            .result
-            .map_err(|error| error.into())
+            .submit_transaction(register_tx)
+            .await?
+            .await?
+            .result?;
+
+        Ok(H256::random())
```
geigerzaehler: You can use [`TransactionApplied::tx_hash`](https://github.com/radicle-dev/radicle-registry/blob/8689536f7e30d5e4839141658d2357e2cc550755/client/src/interface.rs#L47) to get the transaction hash`

proxy/src/schema/mod.rs
```diff
@@ -73,6 +82,45 @@ impl Mutation {
...
+        Ok(registry::Transaction {
+            id: juniper::ID::new(transaction_id.to_string()),
+            messages: vec![registry::Message::ProjectRegistration(
+                registry::ProjectRegistration {
+                    domain: domain,
+                    name: name,
+                },
+            )],
+            state: registry::TransactionState::Applied(registry::Applied {
+                block: juniper::ID::new(radicle_registry_client::H256::random().to_string()),
```
geigerzaehler: You can get this from [`TransactionApplied::block`](https://github.com/radicle-dev/radicle-registry/blob/8689536f7e30d5e4839141658d2357e2cc550755/client/src/interface.rs#L49)

proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
+    pub state: TransactionState,
+    /// Creation time.
+    pub timestamp: String,
+}
+
+/// Required information to issue a new project registration on the [`Regisry`].
+#[derive(juniper::GraphQLObject)]
+pub struct ProjectRegistration {
+    /// The domain/namespace the project should be registered for.
+    pub domain: String,
```
geigerzaehler: Alternatively, you could use `String32` here. That would require you to define a custom GraphQL scalar but would move these errors to the parsing stage.

proxy/src/schema/registry.rs
NunoAlexandre: Why returning a random hash? `submit_transaction` gives you a `TransactionApplied` back that contains a hash.

proxy/src/schema/registry.rs
```diff
@@ -1,82 +1,149 @@
...
     let registry = Registry::new(client);
-    let alice = Pair::from_legacy_string("//Alice", None);
-    let result = registry.register_project(&alice, "hello".into(), "world".into());
+    let alice = ed25519::Pair::from_legacy_string("//Alice", None);
+    let result = futures::executor::block_on(registry.register_project(
+        &alice,
+        "hello".into(),
+        "world".into(),
+    ));
     assert_eq!(result.is_err(), false);
```
NunoAlexandre: you can simplify to `assert!(result.is_ok());`

rudolfs:
Please update `app/schema.graphql` with the new endpoints.

proxy/src/schema/registry.rs
NunoAlexandre:
```suggestion
/// Required information to issue a new project registration on the [`Registry`].
```

proxy/src/schema/registry.rs
NunoAlexandre:
```suggestion
    /// The name of the project, which MUST be unique for the domain.
```

proxy/src/schema/registry.rs
xla: We can use none juniper (the crate we use to drive our GraphQL API) supported types, but that needs custom `Scalar` implementations. I captured that in the follow-up issue #99.

proxy/src/schema/registry.rs
xla: Will be handled later, I captured it in #99.

proxy/src/schema/mod.rs
xla: Not yet certain where the best way is to make the cut off, we want to support multiple domains down the road, maybe even sooner with the advent of orgs.

xla:
@NunoAlexandre @rudolfs @geigerzaehler All comments have been addressed, thanks again for the valuable feedback.

app/schema.graphql
```diff
@@ -104,11 +104,32 @@ ProjectMetadataInput {
...
+type ProjectRegistration {
+  domain: String!
+  name: String!
+}
+
+union TransactionState = Applied
+
+type Applied {
+  # Hash of the block it was applied to.
+  blocks: ID!
```
geigerzaehler:
```suggestion
  block: ID!
```

xla:
> Just one type. Otherwise

Typo inception. hahaha

proxy/src/schema/registry.rs
garbados: That makes sense.

PR #102, opened 2020-01-23
Don't allow registering already taken names

rudolfs:
- [x] failing test case
- [ ] fix


rudolfs:
We're rewriting project registration, which deprecates this PR.

Issue #98, opened 2020-01-21
Two-way attestation of project identity between librad and registry


garbados:
The project ID generated by librad should be included in the metadata of project registration and the project ID of the registered project (a tuple of `(domain, name)`) should be included in the metadata of the librad project, so that each can attest to the other's veracity.

To accomplish this:

- The proxy must include the librad project ID (`librad::project::ProjectId`) in the metadata sent during project registration, as part of the `metadata` field of `radicle_registry::client::message::RegisterProject`. This will require casting a `String` to `Bytes128`.
- The proxy must include the registry's project ID (the tuple) in the librad project's metadata (`librad::meta::project::Project#add_rel`) either using the Relation::Label variant (as `Label("radicle", "({domain}, {name})")`) or as a new Relation variant which does not require serializing and deserializing the project ID.

These requirements raise two issues:

- Can librad-ProjectID exceed 128 bytes? If so, how should we handle compressing the ID?
- De/serializing a `(domain, name)` tuple to and from a String raises questions about forbidden characters, to distinguish elements of the domain and the name from special characters like those separating them. If, as an example, `||` is used as a separator, how would we handle the project name `He||o`?

xla:
Thanks for putting this together. The serialisation story is a bit more intricate as we need to address forward and backward compatibility. To ensure that metadata written today can still be made sense of many years down the road, which is esp. important for the `Registry` as the metadata is immutable. For binary we looked at [CBOR](https://cbor.io/) in the past and might be a good first candidate to evaluate. If we stick to the string value in `librad` we should turn to JSON and investigate how we could do versioning.

garbados:
CBOR seems like a good candidate. Particularly the [serde_cbor}(https://github.com/pyfisch/cbor) library seems appropriate.

garbados:
After diving into the issue, I've come up with two questions:

1. It is possible to serialize the `(domain, name)` tuple into CBOR as `vec<u8>` but converting that to a string presents problems. `String::from_utf8` accepts `vec<u8>` as a parameter, but can fail if the vector doesn't contain valid UTF-8, which isn't important to our purposes. I know CBOR has an explicit string form so I suspect I'm missing something that would otherwise make this easy. How would you recommend serializing to a string?
[...]

PR #105, opened 2020-01-24
Implement one-way attestation

garbados:
This PR addresses https://github.com/radicle-dev/radicle-upstream/issues/98

This is a draft implementation of the core functionality to get feedback. As librad project metadata cannot be altered (https://github.com/radicle-dev/radicle-upstream/issues/98#issuecomment-577618492) I'm not sure how to handle that side of attestation.


proxy/src/schema/error.rs
```diff
@@ -297,6 +308,9 @@ fn convert_url_parse_error_to_field_error(error: &url::ParseError) -> FieldError
 impl IntoFieldError for Error {
     fn into_field_error(self) -> FieldError {
         match self {
+            Self::Cbor(cbor_error) => {
```
xla: This is the first time I think we are facing a range of errors which are not meaningful to the upstream layers. We should rethink what our strategy when we encounter a serialisation error. And in turn what we really want to communicate potentially to the end user. A cbor error won't be actionable.

proxy/src/schema/mod.rs
```diff
@@ -87,12 +87,13 @@ impl Mutation {
         ctx: &Context,
         domain: String,
         name: String,
+        pid: Option<String>,
```
xla:
```suggestion
        librad_id: Option<String>,
```

`pid` is otherwise too ambiguous.

proxy/src/schema/registry.rs
```diff
@@ -91,6 +97,7 @@ impl Registry {
         author: &ed25519::Pair,
         domain: String,
         name: String,
+        maybe_pid: Option<String>,
```
xla: Interesting prefix choice, not quite sure how this is usually advocated in Rust.

proxy/src/schema/registry.rs
```diff
@@ -118,11 +125,22 @@ impl Registry {
             .await?
             .result?;
 
+        let register_metadata_vec = if let Some(pid_string) = maybe_pid {
+            let pid_cbor = ProjectIdCbor(pid_string);
+            to_vec(&pid_cbor)?
```
xla: What is the rationale for going for a vec?

Given our requirements it won't help much with backward and forward compatibility as the order of the elements still neds to be known. I think a struct of sorts with fields in it will be safer for that. And we need to a have a way to know which version the metadata we face was serialisaed with.

proxy/src/schema/registry.rs
```diff
@@ -158,18 +176,24 @@ impl Registry {
...
-    let registry = Registry::new(client);
-    let alice = ed25519::Pair::from_legacy_string("//Alice", None);
-    let result = futures::executor::block_on(registry.register_project(
-        &alice,
-        "hello".into(),
-        "world".into(),
-    ));
-    assert!(result.is_ok());
+    #[test]
+    fn test_register_project() {
```
xla: Dunno if this is the right place for it, but we need to extend the tests to be more ellaborate and actually assert that we can get out what we put in and expect to get back from the Registry.

proxy/src/schema/registry.rs
garbados: The vectorization is just how serde_cbor serializes information, casting objects to vectors of bytes so they can be cast back faithfully. Is there another way you think could be more appropriate?

I'll update the `ProjectIdCbor` to a `ProjectCbor` struct that can contain more information, like the metadata version.

proxy/src/schema/registry.rs
garbados: I saw the `maybe_` prefix used in a piece of example Futures code and it stuck in my patterns, but if there's a more appropriate or idiomatic expression I'd be happy to adopt it.

proxy/src/schema/error.rs
garbados: Interesting. I suppose an `.expect` might be appropriate then? Or falling back to an empty vec?

proxy/src/schema/registry.rs
garbados: Yes, fair point. I'll extend the tests. It makes sense to include that extension in the PR, to me, as it's testing features this PR adds.

garbados:
Woo! Tests pass.

garbados:
Ready for a second review @xla

proxy/src/schema/registry.rs
xla: Thanks for the pointer, maybe we should start gathering these little best-practices somewhere.

proxy/src/schema/registry.rs
```diff
@@ -118,11 +130,25 @@ impl Registry {
             .await?
             .result?;
 
+        let register_metadata_vec = if let Some(pid_string) = maybe_pid {
+            let pid_cbor = ProjectCbor {
+                id: pid_string,
+                version: 1,
+            };
+            to_vec(&pid_cbor).expect("unable to serialize project metadata")
```
xla: We should actively work on removing panics in our modules. It shouldn't block any further progress on this PR, let's address it soon, maybe in the context of #137 - so maybe leave a `TODO` here.

proxy/src/schema/registry.rs
```diff
@@ -91,6 +102,7 @@ impl Registry {
         author: &ed25519::Pair,
         domain: String,
         name: String,
+        maybe_pid: Option<String>,
```
xla: Why is this an `Option<String>` and not the actual type librad uses for project ids?

garbados: Because I couldn't figure out how to convert it to a format Serde can deserialize :grimacing:

proxy/src/schema/registry.rs
```diff
@@ -70,7 +81,7 @@ pub struct Applied {
 #[derive(Clone)]
 pub struct Registry {
     /// Registry client, whether an emulator or otherwise.
-    client: Client,
+    pub client: Client,
```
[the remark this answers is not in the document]
garbados: Due to a testing complication it was either this or borrow the client with a lifetime, which I was too sick and tired to figure out. I'm pushing a commit shortly that changes to a static borrow.

The testing complication was that the client is needed in two places in the test suite: to pass to the Registry constructor, and to retrieve the project after registering the project. Since the Registry instance didn't use the client by reference, the client was locked to the instance.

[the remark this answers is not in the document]
garbados: Actually I'm having some issues with the borrow checking while converting it to a borrow :grimacing: Maybe we can talk about this on Monday.

proxy/src/schema/registry.rs
garbados: :+1:

proxy/src/schema/registry.rs
garbados: No wait, that's not true. It's because librad doesn't re-export git2 so I couldn't construct an Oid from a string without adding git2 as a dependency, so I used a string instead and... then forgot to go back and change it. I'm working on a fix now.

garbados: Hmm upon further review it appears I just got very confused and then forgot about the issue when I opted to use a string as an intermediary placeholder. I'm still pretty confused so maybe we can talk about this on Monday?

proxy/src/schema/registry.rs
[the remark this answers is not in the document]
xla: Check 4c6556f for how to reuse the client, it derives `clone`.

proxy/src/schema/registry.rs
xla: Addressed in a608b52 to use proper id types.

Issue #113, opened 2020-02-04
Clean up schema.graphql


rudolfs:
- [ ] ~~make sure it reflects what is currently implemented in the backend~~
- [ ] ~~move it to the top-level directory, because it's a shared API contract between app and proxy~~
- [x] document how to use GraphiQL to query schema from proxy

Refs #97

PR #126, opened 2020-02-06
Clean up schema

rudolfs:
Instead of having a schema file which quickly becomes outdated, document how to use GraphQL introspection to explore the live API and introduce better GraphiQL tooling.

Refs #113 

Screenshot 2020-02-06 at 13 43 32


DEVELOPMENT.md
```diff
@@ -21,6 +21,39 @@ The pull request ref goes in brackets at the end of the subject.
...
+## Work coordination
+
+Work on the app and proxy can happen in parallel. However for this to work the
+teams working on both code bases have to agree on a common API. The definitive
+source-of-truth for this API is the proxy source code.
+
+To change or extend the API, first open an issue on GitHub, discuss what is
+needed and then propose proxy code changes that implement the API.
+
+We don't keep around a copy of a SDL schema file as they tend to get outdated
```
xla:
```suggestion
We don't keep around a copy of an SDL schema file as they tend to get outdated
```

DEVELOPMENT.md
```diff
@@ -21,6 +21,39 @@ The pull request ref goes in brackets at the end of the subject.
...
+
+
+### GraphiQL explorer
+
+The introspective and self-docummenting nature of GraphQL allows us to easily
+explore our API by querying the backend. For better developer ergonomics we
+include a [static version][gs] of [GraphiQL-explorer][ge]. To see it in action:
+
+1. Start the app in development mode: `(cd app && yarn start)`
+2. Open GraphiQL in your browser via the CLI: `(cd app && yarn graphiql)` or
```
xla: This command produces the following error on my Arch machine: `/bin/sh: open: command not found`

rudolfs: Hm, just tried with `bash` and `zsh` both worked, which shell are you using?

xla: Using `fish` on ARch.

rudolfs: Alright, should be fixed now.

PR #134, opened 2020-02-07
Proxy structure improvements

### LINKED ISSUE #125: Improve separation of concerns in proxy
Given that we want to make it easier for contributions to land from dependent teams we need to find a better way of separating our concerns. Currenlty everything somehow ties into the schema module and is expected to live there. Where in reality we want to be true to the fact that we deal with distinct domains and should clearly draw the line for our own. Rather treat the GraphQL APi just as one possible surface we can expose that domain and don't tie it to it.

* [x] propose new module structure
* [x] restructure proxy codebase based on new layout

xla:
Closes #125 

## Objectives

* Enable CoCo to contribute in a directed and safe way
* Enable Registry to contribute in a directed and safe way
* Decouple domain logic from specific API implementation (GraphQL)
* Improve maintainability

## Structure

In essence the new source structure looks like this:

```
proxy/
├── Cargo.lock
├── Cargo.toml
├── README.md
├── rust-toolchain
├── src
│   ├── coco.rs
│   ├── env.rs
│   ├── error.rs
│   ├── graphql
│   │   ├── api.rs
│   │   ├── error.rs
│   │   ├── project.rs
│   │   └── schema.rs
│   ├── graphql.rs
│   ├── lib.rs
│   ├── main.rs
│   └── registry.rs
└── tests
    ├── common
    │   └── mod.rs
    ├── graphql_mutation.rs
    └── graphql_query.rs

```

This will reap the following benefits:
* decoupling API from core domain
* single entry point for protocol teams
* Adhere to 2018 [Path clarity](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html)
* Test high-level API through [integration tests](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests)


proxy/src/coco.rs
```diff
@@ -0,0 +1,467 @@
...
+use librad::peer;
+use librad::project;
+use radicle_surf as surf;
+use radicle_surf::git::git2;
+
+use crate::error;
+
+/// Branch name representation.
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub struct Branch(pub String);
```
cloudhead: Curious why we need `pub` visibility here, since this would allow branches to be constructed incorrectly, eg. with illegal characters from anywhere in the code. Same for `Tag`.

proxy/src/coco.rs
```diff
@@ -0,0 +1,467 @@
...
+    pub email: String,
+    /// Reference (url/uri) to a persons avatar image.
+    pub avatar: String,
+}
+
+/// Representation of a code commit.
+pub struct Commit {
+    /// Identifier of the commit in the form of a sha1 hash. Often referred to as oid or object
+    /// id.
+    pub sha1: String,
```
cloudhead: Any reason why we don't use a binary representation here?

proxy/src/coco.rs
```diff
@@ -0,0 +1,467 @@
...
+    /// The last commmit that touched this object.
+    pub last_commit: Option<Commit>,
+}
+
+/// File data abstraction.
+pub struct Blob {
+    /// Best-effort guess if the content is binary.
+    pub binary: bool,
+    /// Actual content of the file, if the content is ASCII.
+    pub content: Option<String>,
```
cloudhead: Having a type like `enum BlobContent { Binary, Ascii(String) }` would avoid the possibility that this type has `binary: true, content: Some(...)`

proxy/src/coco.rs
```diff
@@ -0,0 +1,467 @@
...
+    /// Extra info for the file.
+    pub info: Info,
+}
+
+/// Result of a directory listing, carries other trees and blobs.
+pub struct Tree {
+    /// Absolute path to the tree object from the repo root.
+    pub path: String,
```
cloudhead: I wonder why we use `String` instead of `PathBuf` here and in a few other places, since PathBuf should more accurately represent the data, as well as represent all paths, which String isn't guaranteed to.

proxy/src/coco.rs
```diff
@@ -0,0 +1,467 @@
...
+pub fn blob(paths: &Paths, id: &str, revision: &str, path: &str) -> Result<Blob, error::Error> {
+    let project_id = project::ProjectId::from_str(&id)?;
+    let project = project::Project::open(&paths, &project_id)?;
+
+    let mut browser = match project {
+        project::Project::Git(git_project) => git_project.browser()?,
+    };
+
+    // Best effort to guess the revision.
+    if let Err(_err) = browser
```
cloudhead: `browser.is_err()` maybe?

proxy/src/coco.rs
```diff
@@ -0,0 +1,467 @@
...
+/// TODO(xla): Add documentation.
+pub fn tags(paths: &Paths, id: &str) -> Result<Vec<Tag>, error::Error> {
+    let project_id = project::ProjectId::from_str(&id)?;
+    let project = project::Project::open(&paths, &project_id)?;
+    let browser = match project {
+        project::Project::Git(git_project) => git_project.browser()?,
+    };
+
+    let mut tag_names = browser.list_tags()?;
+    tag_names.sort();
```
cloudhead: Can we avoid this sort, since we're sorting again right after?

proxy/src/coco.rs
```diff
@@ -0,0 +1,467 @@
...
+/// TODO(xla): Add documentation.
+pub fn tree(paths: &Paths, id: &str, revision: &str, prefix: &str) -> Result<Tree, error::Error> {
+    let project_id = project::ProjectId::from_str(&id)?;
+    let project = project::Project::open(&paths, &project_id)?;
+
+    let mut browser = match project {
+        project::Project::Git(git_project) => git_project.browser()?,
+    };
+
+    if let Err(_err) = browser
```
cloudhead: `is_err()`

proxy/src/coco.rs
```diff
@@ -0,0 +1,467 @@
...
+                    nonempty::NonEmpty::from_slice(&[label.clone()]).ok_or_else(|| {
+                        radicle_surf::file_system::error::Error::Label(
+                            radicle_surf::file_system::error::Label::Empty,
+                        )
+                    })?;
+                surf::file_system::Path(label_path)
+            } else {
+                let mut p = path.clone();
+                p.push(label.clone());
+                p
```
cloudhead: I'd expect you to be able to do `path.join(label.clone())` here, if `path :: Path`

proxy/src/coco.rs
xla: Leftover from the time these where directly used for the GraphQL API.

proxy/src/coco.rs
xla: Addressed by making it private and implementing `std::fmt::Display` for them.

proxy/src/coco.rs
xla: Neat, will rework.

proxy/src/coco.rs
xla: Leftovers from when these structs where used directly for the GraphQL API.

proxy/src/coco.rs
xla: I think you expect this to be a rust std path, but it's from surf.

proxy/src/registry.rs
```diff
@@ -46,35 +33,20 @@ pub struct Metadata {
 
 /// Possible messages a [`Transaction`] can carry.
 pub enum Message {
-    /// Issue a new project registration. See [`ProjectRegistration`] for the shape of the data.
-    ProjectRegistration(ProjectRegistration),
+    /// Issue a new project registration with (domain, name).
+    ProjectRegistration {
+        /// Domain part o the project id.
```
garbados: o -> of

Issue #118, opened 2020-02-05
Refactor "Project" section


rudolfs:
- [x] extract source browsing into a component so it can be re-used by Project Overview and Project Source sub-pages
- [ ] think about how to make "Project" section source structure less confusing

Issue #117, opened 2020-02-04
Define file structure


rudolfs:
- [x] unify setup and use of params/classes (Button vs. Icon, Title, etc)
- [x] propose a consistent way to structure:
  - design system primitives (icons, texts, form elements, colors, etc)
  - reusable components (cards, notifications, teasers, etc)
  - fragments (used for code extraction purposes: SidebarAvatar, SidebarItem, etc)
  - top-level-pages, sub-pages, modals, layouts
- [x] document best practices
- [x] clean up existing code according to previously defined steps
  - [x] move `ProjectRegistration` steps under pages into a folder
  - [x] check if extracted components from other components can be group into folders or simplified
  - [x] investigate if the layout directory can be simplified or moved
  - [x] group `DesignSystem` components by type (eg form components)

Refs #97

Issue #115, opened 2020-02-04
Unify design system, css and markup


MeBrei:
Collects all refactoring around the `DesignSystem` directory and styling in general. 

**Guidelines/README**
- [x] Define what belongs into the `DesignSystem` directory 
- [x] Define what belongs onto the `DesignSystem` page
- [x] Create guidelines for the use of styling

**Consistent design system components**
- [x] move `Numeric`, `Text`, `Title` each into a single component
- [x] Rename Inputs 
- [x] move input validation errors into Input component
- [x] possibly extract any other reusable components 
  - [x]  ~~`Row`~~  _postponed to #117 _
  - [x]  `button-row`
  - [x]  ~~`StepCounter`~~  _postponed to #117 _
  - [x]  `Avatar`/`Title` combination - see eg `PickWalletStep` 

**Reduce inline css**
- [x] ~~move inline css of components into params or classes~~ _There is currently no good way to pass classes and styling to components in a different way. We'll get back to it when an alternative is implemented._
- [x] ~~Can some of the inline styling be moved in to a parent `div`?~~ _We decided that we want to avoid wrapping components into div's for minor styling changes, as it will clutter the code._
- [x] remove all inline css of non-components (`FileList`, `TransactionSummaryStep`, `PickWalletStep`, `PickNameStep`,  `ConfirmTransactionStep`, `CreateProject`, `RegisterProject`)

**Consistent css**
- [x] ~~reduce duplication of css (`Button`, global styling)?~~
- [x] restructure class names consistently according to hierarchy: layout/page > container > content
- [x] ~use explicit class names for `style` if it is specific to a component~

**Consistent markup**
- [x] remove duplicate wrapping of components in div's (eg `Project`) 

Refs #97

rudolfs:
Outstanding points were addressed in #149.

PR #149, opened 2020-02-10
Define file structure

rudolfs:
### Objectives
  - Enable outsiders to make contributions more easily
  - Improve maintainability

### Structure
- design system primitives (icons, typography, form elements, colors, etc)
- reusable components (cards, notifications, teasers, etc), fragments (for code extraction purposes: SidebarAvatar, SidebarItem, etc) and layouts
- top-level-screens, sub-screens
- centralised data stores, grouped by use-case
- `lib` for all remaining `*.js` code

Screenshot 2020-02-12 at 14 19 01

Refs #117
Refs #118
Refs #115


rudolfs:
Got some offline feedback from @cloudhead: the only thing that doesn't make sense is handling all text in one component, as there is a semantic difference between titles, body text, code, footnotes, etc. So we should keep them in separate components as before.

I tend to agree now, and will revert that change.

app/README.md
```diff
@@ -52,9 +52,127 @@ committing changes, otherwise the tests will be skipped on CI.
...
+    components for the sake of code readability.
+
+    Sub-folders in this category should only be created for breaking up larger
+    components into smaller fragments. If a component is broken up in
+    fragments, make sure to only export the component for public use, not the
+    fragments.
+
+    Single-use components that are extracted from screens and used by only one
+    screen should be prefixed with the name of the screen, for example:
+    `DesignSystemGuideSection.svelte`.
```
xla: If it's only relevant for one screen, what's the benefit of moving it into the design system in the first place?

app/README.md
```diff
@@ -52,9 +52,127 @@ committing changes, otherwise the tests will be skipped on CI.
...
+
+
+There is a third group of components, which is not part of the Design System:
```
xla: Given that components have a clear definition up top and in the context of the design system I'd drop the idea of them being also components. Defning them as composition of components to form a single screen/view for the user to see is unambiguous and helps conveying the hierarchy.

app/README.md
```diff
@@ -52,9 +52,127 @@ committing changes, otherwise the tests will be skipped on CI.
...
+"Screens". Screens bring together components from the Design System forming
+what a user sees in the UI as a whole screen. More complex UX flows can be
+broken down into multiple sub-screens, in this case a top-level screen should
```
xla: If I understand this correctly sub-screens are full screens, just switche based on the state of the parent screen. If that assumption is correct, there might be a possibility to distinguish these even further. To avoid the subtleties between screen and sub-screen. For example introducing the concept of page or view which holds 1..N screens.

app/README.md
```diff
@@ -52,9 +52,127 @@ committing changes, otherwise the tests will be skipped on CI.
...
+contain common data fetching and routing for the sub-screens.
+
+When multiple screens share the same layout, it should be extracted into a
+separate component. Layout components are suffixed with "Layout", like so:
+`DesignSystem/Components/ModalLayout.svelt`.
+
+File and directory name casing:
+  - Svelte components and directories containing components - PascalCase
+  - everything else: `*.js` files and folders - camelCase
+
```
xla: Currently there is a blind spot about `.css` files, how they are structured, when and where to make changes and how they play together with the design system. Maybe we can expand on that, too.

xla:
> the only thing that doesn't make sense is handling all text in one component, as there is a semantic difference between titles, body text, code, footnotes, etc. So we should keep them in separate components as before.

Also agree with that, the old way was communicating expectations way better.

> We should extract data fetching logic from components.

Had that intuition initially as well and we went over it a couple of times. It doesn't really align with the GraphQL model that well tho. Where each page/screen/whatever is responsible for its own data needs. It brings encapsulation that way. We should outline the expected benefits of attempting such extraction.

app/README.md
rudolfs: Technically everything in Svelte is a component. But I see your point, it might be confusing to newcomers. Will change the wording around it.

app/README.md
rudolfs: It doesn't have to be a full screen necessarily. It could also be a tab, containing just a part of the top-level screen (like in project registration).

I could change the wording a bit if sub-screens are confusing, something like:

```
`Screens` bring together components from the Design System forming what a user
in the UI sees as a whole screen. More complex screens can be broken down into
multiple partials, in this case a screen should contain common data fetching and
routing for the partials. Partials should be placed in a directory named after the screen.
```

Maybe this way we can also avoid having single-use components in the `DesignSystem/Components` directory.

app/README.md
rudolfs: 👍 forgot to add the CSS guidelines, will work on it.

app/README.md
rudolfs: See below about partials.

app/README.md
rudolfs: Alternatively we could call them fragments, similar to how we do with broken-up components.

xla: That helps a lot already, especially the part about state and routing as I think this is the most important outcome - to determine for which entity to draw the line in regards for example expecting to be parameterised via `location`. I have no strong opinions on the names themselves as long as the concepts/entities are distinct and rules have been established what is allowed for either. i.e.: `A Screen holds a single store which is used to transport down information to its fragments...`

app/README.md
xla: Totally understood. The documentation here focuses on our system, so it should be free of Sveltes abstractions, we can allude to it in the beginning and then move on to outline our own vocabulary.

app/README.md
xla: :heart:

PR #153, opened 2020-02-13
Migrate to async/await

### LINKED ISSUE #136: Migrate proxy to async/await
With juniper now supporting the async/await feature we should properly migrate to it. While that we should evaluate runtime/executor usage esp. in conjunction with the registry client.

xla:
Closes #136

### DIFF
--- a/proxy/Cargo.lock
+++ b/proxy/Cargo.lock
@@ -277,6 +277,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "bson"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


xla:
Updated the branch after our required rewrite due to the release. The adjustment to the api changes in latest juniper have been done, it's currently blocked by graphql-rust/juniper#549.

xla:
I started a first attempt to implement graphql-rust/juniper#549 in graphql-rust/juniper#552. Meanwhile all methods have been migrated to `async` and underlying datatypes use thread-safe versions.

xla:
Obsolete as we move to a REST API.

Issue #141, opened 2020-02-10
Improve linting configuration


rudolfs:
Lets have an explicit config for eslint and prettier to un-break xla's setup.
~While at it, look into husky bug where stashed changes get deleted.~  -> can't reproduce anymore

rudolfs:
Closing, thanks to @znarf! 🎉

PR #156, opened 2020-02-13
Prettier configuration

znarf:
Implement prettier part of https://github.com/radicle-dev/radicle-upstream/issues/141

- add default prettier configuration with svelte plugin (already in package.json)
- apply prettier to svelte files

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -91,7 +91,10 @@
     "cypress:open": "yarn run cypress open",
     "generate:colors": "babel-node --presets \"@babel/env\" scripts/generate-colors.js",
     "release": "babel-node --presets \"@babel/env\" scripts/release.js",
-    "release:finalize": "babel-node --presets \"@babel/env\" scripts/release.js --finalize"
+    "release:finalize": "babel-node --presets \"@babel/env\" scripts/release.js --finalize",
+    "prettier": "prettier \"**/*.@(js|json|svelte|css|html)\" --ignore-path .gitignore",
+    "prettier:check": "yarn prettier --check",
+    "prettier:write": "yarn prettier --write"
   },
   "dependencies": {
     "@babel/runtime": "^7.8.4",
--- a/app/.prettierrc
+++ b/app/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "plugins": ["prettier-plugin-svelte"]
+}


app/package.json
```diff
@@ -91,7 +91,8 @@
     "cypress:open": "yarn run cypress open",
     "generate:colors": "babel-node --presets \"@babel/env\" scripts/generate-colors.js",
     "release": "babel-node --presets \"@babel/env\" scripts/release.js",
-    "release:finalize": "babel-node --presets \"@babel/env\" scripts/release.js --finalize"
+    "release:finalize": "babel-node --presets \"@babel/env\" scripts/release.js --finalize",
+    "prettier": "prettier \"**/*.@(js|json|svelte)\" --ignore-path .gitignore"
```
rudolfs:
```suggestion
    "prettier": "prettier \"**/*.@(js|json|svelte|css|html)\" --ignore-path .gitignore --check"
```
We should also lint the static stuff that we have.

And I think it also makes sense to have the output human readable via `--check`, or did you have a different purpose in mind for this script?

znarf: Definitely! It's usually useful to add commands to check and write. The first command can stay as a base command to keep configuration and parameters in one place.

We could add:
- `"prettier:check": "yarn prettier --check"
- `"prettier:write": "yarn prettier --write"

Sounds good?

rudolfs: Hm, now that I think about it, we actually already have Husky and lint-staged in place which get run just before making a commit and it should bark at us if either the linter or prettier are unhappy.

This is how it looks on my end when I try to commit something funky:
```
$ git commit
husky > pre-commit (node v11.11.0)
  ✔ Preparing...
  ❯ Running tasks...
    ❯ Running tasks for *.{js,css,json,html}
      ✖ prettier --write
    ↓ Running tasks for *.svelte [skipped]
      → No staged files match *.svelte
  ↓ Applying modifications... [skipped]
    → Skipped because of errors from tasks.
  ✔ Reverting to original state...
  ✔ Cleaning up...



✖ prettier found some errors. Please fix them and try committing again.
package.json 37ms

[error] scripts/release.js: SyntaxError: Unexpected token, expected ";" (23:13)
[error]   21 |       console.log("  - https://github.com/github/hub");
[error]   22 |     } else {
[error] > 23 |       raise error;
[error]      |             ^
[error]   24 |     }
[error]   25 |   }
[error]   26 |
husky > pre-commit hook failed (add --no-verify to bypass)
```

I believe prettier is even configured to re-write staged code before the commit. So I wonder how that malformed code even slipped by? Perhaps it was just the missing linter config...

Do you think there is still need for explicit linting and prettifier scripts in this setup?
I'm now leaning towards just mentioning in the README that we have it as a pre-commit hook.

znarf:
There is no guarantee that the pre-commit hook will be executed by developers, especially if you start having contributions from the community. For this reason, based on my experience, I suggest you always double check on CI. The tasks can be useful for that.

What are your plans for CI? Adding that as a GitHub action could be fairly easy. Some recent example: https://github.com/backyourstack/backyourstack/blob/master/.github/workflows/ci.yml#L53

rudolfs:
> For this reason, based on my experience, I suggest you always double check on CI. The tasks can be useful for that.

Good point. Let's keep the tasks then.

> What are your plans for CI?

At the moment we're using Buildkite and our own infra:
https://github.com/radicle-dev/radicle-upstream/blob/master/DEVELOPMENT.md#ci-setup

Will have a think about adding the suggested extra checks there, thanks for pointing out!

xla:
@znarf Thanks for your contribution, can you rewrite your history so that all commits are sigend? Oherwise we can't merge. :>

znarf:
@xla It's rebased and signed!

rudolfs:
@znarf added eslint and prettier checks to our CI pipeline: https://github.com/radicle-dev/radicle-upstream/pull/158. Thanks for the tip!

Issue #71, opened 2019-12-16
Proper error propagation


xla:
Currently we do not address errors throughout the application stack. This means we do not address program states which are expected but outside of the happy path. Including a wide variety of user facing experiences:

* validation of user input
* connectivity (online/offline)
* feedback from user initiated actions
* resource starvation

To improve on this work towards a better end user experience this issue tries to outline and track the work necessary to have proper error propagation through out protoocl libs, proxy and ui. The expected outcome is that there is a clear contract between the ui and the proxy encoded in the GraphQL API. Included in that contract should be a fixed set of errors that come with clear semantics so the client behaviour can be consistent and is predictable. This can be achieved with GraphQL errors field and the `extensions` part. We should strive for a setup outlined in [structured errors](https://graphql-rust.github.io/juniper/master/types/objects/error_handling.html#structured-errors).

More specifically what we want is to unpack protocol errors, translate them into our own error enum, where the variants carry relevant contextual information. These custom errors should then be returned from the GraphQL query in question in the error part and carry a unique `type` for the client to vary on.

To keep single changesets small this is a meta issue to track the overall progress of:

* [x] Explicitly handle errors from protocol libraries
* [x] Remove all calls of `expect` and `unwrap` in the proxy modules
* [x] Implement our own error infrastructure with clear semantics for the UI as outlined by [structured errors](https://graphql-rust.github.io/juniper/master/types/objects/error_handling.html#structured-errors)
* [x] Return errors from GraphQL endpoints with the help of the `extensions` part
* [x] Convert `From` implementations to `TryFrom` where conversion can fail
* [x] Convert string error types to `enum` which will be served as part as the GraphQL API to provide better guarantees to the client
* [ ] Consume errors in the UI and reflect the set of errors.

garbados:
Although the PR for MON-19 has been merged, some runtime `.expect` calls survived in `proxy/src/schema/` in [git.rs](https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/src/schema/git.rs#L138) and [mod.rs](https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/src/schema/mod.rs#L83). These should be eliminated in a future PR.

garbados:
[...]

PR #157, opened 2020-02-14
Convert graphql_value! calls to use graphql enums

garbados:
This branch attempts to follow up on this item of https://github.com/radicle-dev/radicle-upstream/issues/71

> Convert string error types to enum which will be served as part as the GraphQL API to provide better guarantees to the client

I'm having a lot of trouble with it though.


xla:
Not needed anymore with the move to a REST API. :>

PR #159, opened 2020-02-17
Add raw data access for testing

### LINKED ISSUE #133: Integration test strategy
At the moment most of our e2e tests are skipped due to recent changes to proxy. We have to figure out how we want to do e2e tests in the future, so we can fix our test suite.

xla:
Closes #133


rudolfs:
As discussed IRL, we'll need these mutations right away:
- [ ] reset/nuke CoCo state
- [ ] reset/nuke Registry state

Would be nice to isolate the `/control` endpoint so it doesn't show queries/mutation from `/graphql`.

rudolfs:
Almost there:
- [ ] `nukeCocoState` does not seem to be idempotent (I get an error message after running it 2x)
- [ ] tags and the dev branch are not exposed via the api for the git-platinum repo (the remaining 7 test failures)

Screenshot 2020-02-28 at 13 59 22

rudolfs:
One test is still failing: we need to also replicate all branches (like we do for tags).

Screenshot 2020-03-02 at 18 13 21

.buildkite/run.sh
```diff
@@ -58,6 +58,8 @@ fi
 echo "--- Updating submodules"
 time git submodule update --init --recursive
 time git submodule foreach "git fetch --all"
+time git submodule foreach "git checkout -B dev -t origin/dev"
+time git submodule foreach "git checkout master"
```
rudolfs: Will we need to do this for every new branch?
Should this also be added to the testing pre-requisites in DEVELOPMENT.md?

xla: Haven't found a good solution for yet, we want a certain state local branches. Any ideas how to make this more elegant?

rudolfs: No idea at the moment.

xla: Created #207 as follow-up.

PR #154, opened 2020-02-13
Migrate from project domain to org

NunoAlexandre:
It updates `register_project` and introduces `register_org` and `unregister_org`. In terms of api, it also flattens `ProjectId` to two separate params `project_name` and `org_id`, to follow the latest registry.

PR #161, opened 2020-02-18
Fix project registration UI

rudolfs:
There were a couple of regressions introduced in https://github.com/radicle-dev/radicle-upstream/pull/154 due to lacking tests on out end.

This PR fixes these regressions.

### DIFF
--- a/app/src/Screens/RegisterProject.svelte
+++ b/app/src/Screens/RegisterProject.svelte
@@ -19,7 +19,7 @@
   };
 
   let step = 1;
-  let name = $projectNameStore;
+  let projectName = $projectNameStore;
 
   const nextStep = () => {
     step += 1;
@@ -32,13 +32,13 @@
   const client = getClient();
 
   const REGISTER_PROJECT = gql`
-    mutation($domain: String!, $name: String!) {
-      registerProject(domain: $domain, name: $name) {
+    mutation($orgId: String!, $projectName: String!) {
+      registerProject(orgId: $orgId, projectName: $projectName) {
         id
         messages {
           ... on ProjectRegistration {
-            domain
-            name
+            projectName
+            orgId
           }
         }
         state {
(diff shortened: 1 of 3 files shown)


NunoAlexandre:
@rudolfs this PR means that my prior PR broke this, right? If so, that means that the UI tests are not covering enough, as I guided myself after them. Sorry for the noise, anyway!

rudolfs:
@NunoAlexandre you're right, there are no tests for this "feature", as it is just a proof-of-concept, which we wanted to show at team week. The actual feature is being specced right now, so no worries.

NunoAlexandre:
I see. Thanks, @rudolfs.

app/src/Screens/RegisterProject.svelte
```diff
@@ -58,8 +58,8 @@
       response = await mutate(client, {
         mutation: REGISTER_PROJECT,
         variables: {
-          project_name: project_name,
-          org_id: org_id
+          projectName: projectName,
+          orgId: "rad123"
```
MeBrei: This has to be replaced, once the org creation is implemented. It might make sense to indicate this already now and keep the `org_id` variable outside of this function, but hardcode it for now.

NunoAlexandre: To avoid possible confusion, I'd suggest renaming "rad123" to "monadic" to 1) be consistent with the other usages and 2) to not confuse rad(icle) as a project.

app/src/Screens/RegisterProject.svelte
```diff
@@ -20,6 +20,7 @@
 
   let step = 1;
   let projectName = $projectNameStore;
+  let orgId = "monadic"; // TODO(rudolfs): get the proper org id!
```
NunoAlexandre: 👌

PR #170, opened 2020-02-19
Eslint setup for svelte

MeBrei:
So far the `.eslintrc.json` rules were only applied to `.js` files.

### DIFF
--- a/app/.eslintrc.json
+++ b/app/.eslintrc.json
@@ -1,5 +1,12 @@
 {
   "extends": ["eslint:recommended"],
+  "plugins": ["svelte3"],
+  "overrides": [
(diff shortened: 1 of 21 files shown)


MeBrei:
With the current changes, lint errors will be thrown via `.buildkite/run.sh`, but not locally when creating the commit (for both svelte and js files). Only the `prettier` command is run then and automatically fixes the errors. If we want to run `eslint` on buildkite, we should already enforce it locally, but some of the eslint rules that we use can not be automatically fixed via `eslint --fix`, eg `no-unused-vars`. The commit message will be lost when it errors out on creating the commit.

Issue #186, opened 2020-02-25
Implement user profile page screens


rudolfs:
Everything that we need to show the initial user profile screen after a new user identity is created during onboarding.

- [x] make sidebar look like in the new designs
  - [x] search and peer icons
  - [x] user profile avatar
  - [x] active state
  - [x] hover state
  - [x] add new org button
  - [x] query profile identity (avatar)
- [x] user profile top-bar
  - [x] `...` dropdown menu
  - [ ] registered user badge
  - [x] incorporate `New project` button
  - [x] Remove titles from subpages (eg `My projects`)
- [x] project list visuals
- [ ] project list query
  - [x] local projects
  - [ ] ~~registered projects I'm a member of~~ i think this is a proxy concern
  - [x] populate projects list `Stat`
- [ ] ~~number formatting (eg 1100 -> 1.1k) for `Stat`~~ seems to be done in proxy

PR #188, opened 2020-02-25
Restyle sidebar to match current design

rudolfs:
Refs #186


ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -1,113 +1,180 @@
...
+
+    <li
+      class="item"
+      data-cy="profile"
+      class:active={path.active(path.projects(), $location)}>
+      <a href={path.projects()} use:link>
+        <img
+          class="user-avatar"
+          src="https://avatars2.githubusercontent.com/u/2326909?s=400&v=4"
+          alt="user-avatar" />
```
MeBrei: Can you use the `Avatar` component instead?

ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -1,113 +1,180 @@
...
-          dataCy="profile"
-          href={path.profile()}
-          active={path.active(path.profile(), $location)} />
-      </li>
-    </ul>
-  </nav>
+<div class="wrapper" data-cy="sidebar">
+  <ul>
+    <li
+      class="item"
```
MeBrei: Minor thing: like this, the cursor-change and link-action is restricted to only the icon, but the tooltip and grey bar already appear on the `li.item`. So they appear although I can't click on the sidebar item.

rudolfs:
Addressed both comments.

Issue #194, opened 2020-02-26
Abstract ProjectTopbar into Topbar


sarahscott:
`ProjectTopbar` and its associated `Menu` & `MenuItem` will be useful for the Orgs screen as well, so this should be a more generalized `Component`

Issue #160, opened 2020-02-17
Improve typography architecture


rudolfs:
## Rationale

While working on #117 I noticed that we conflate two aspects of typography into our components. On the one hand we have text styles and on the other components which represent a textual entity, like a title or paragraph. Currently we mix these two distinct concepts across our codebase.

At the moment we have these typography related components: `title`, `text`, `caption`, `code` and `numeric`.

In some cases their usage in code is justified, like actual titles have the `<Title>` tag and some body texts are wrapped in `<Text>`. However in other places we misuse these components as a means to have globally defined font styles.

With CSS a font style can only be represented as a combination of `font-family`, `font-size`, `line-height`, `color`, and other typography related CSS attributes.  Yet, neither CSS, nor Svelte provide a native way to share font styles across code in an ergonomic way.

We had a similar problem with colors, which we could solve with CSS variables, as a color is a single value. But in the case of typography CSS variables aren't so useful.

## Possible solutions

Ideally we want to have globally defined typography styles with clearly defined hierarchies, that match our Figma designs and which we can reuse across all our components.

Currently I see two ways to go about this:

### Solution A

Define a separate set of components just for typography styles with a clear naming scheme, so it doesn't collide with components for titles, body texts, etc.

Drawbacks:
  - we can only use these in the markup section of Svelte components
    - we can't, for example, use them in the `<style>` section of components
    - or for inline overrides
  - this also doesn't work in raw css files (like `global.css`)

### Solution B

Another way to go about this is to add CSS pre-processing to our pipeline. This allows us to have globally defined typography styles as mixins which can be reused by many components. As opposed to **Solution A**, this allows us to reuse the typography styles in components' `<style>` section.

The drawbacks here are:
  - still not possible to use in inline styles
  - can't use SASS mixins in raw CSS files (like `global.css`) directly, we'd have to do another round of preprocessing
  - added complexity

Here's a little exploration I did to see how it would look like with a SASS preprocessor: https://github.com/radicle-dev/radicle-upstream/tree/rudolfs/preprocessor-poc.

In either case we should put in a bit more thought on how we name our typography styles in Figma. I'm also curious whether there are alternative solutions to this problem which I haven't thought of.

cloudhead:
Hmm, I'm not sure I understand the issue with solution A. Why can't we do something like:

css:
```
h1, h2, h3 { ... } // Title tags
code { ... } // Code
.small { ... } // Stylistic variations
.x-small { ... } 
```

[...]

PR #195, opened 2020-02-27
Dropdown menu component

### LINKED ISSUE #193: Additional Actions Modal
Ellipses button + corresponding modal next to 'Add project':
Screen Shot 2020-02-26 at 1 12 35 PM

Screen Shot 2020-02-26 at 1 10 26 PM

sarahscott:
Fixes #193 & #194 

- [x] Generalize `ProjectTopbar` (& associated `Menu`) into `Topbar` that can be used in other contexts
- [x] Fix hover highlighting on `Menu` buttons (icons were turning purple but text was not)
- [x] Add an `AdditionalActionsDropdown` to Design System Guide:
addlactionsbtn


rudolfs:
> add an icon-only button (right now, Button accommodates icon + text buttons but not icons on their own)

I'd just make the whole dropdown component with a vanilla html `<button>` and add all the styling and functionality in there, as I don't see that we'd reuse this at the moment. We can always extract later when we see duplication.

Here's some points on the code itself (couldn't inline my comments as Github seems to have some service outage atm):

- So far we've been initialising all props with `null` unless they come with a default value or the prop is a boolean. Is there any reason to break that convention?
```
export let menuItems = [];
export let items = [];
```

- Perhaps we should extract the "loose" pattern matching argument into 
```
const topbarMenuItems = projectId => [
    {
      icon: Icon.Source,
      title: "Source",
      href: path.projectSource(projectId)
      looseActiveStateMatching: true
    },

active={path.active(item.href, $location, item.looseActiveStateMatching)} />

instead of

active={path.active(item.href, $location, i === 0)} />
```

In `ui/Screen/Project.svelte` L101-117: why do we need to wrap everything in <Flex>? It seems to render just fine without, when I remove it.

ui/Screen/Project.svelte
```diff
@@ -60,16 +60,61 @@
...
+    <Topbar
       style="position: fixed; top: 0;"
       avatarUrl={result.data.project.metadata.imgUrl}
-      id={result.data.project.id}
-      name={result.data.project.metadata.name} />
-
+      name={result.data.project.metadata.name}
+      href={path.projectOverview(result.data.project.id)}
+      menuItems={topbarMenuItems(result.data.project.id)}>
+      <Flex>
```
sarahscott: Re: why wrapping in `Flex` here - I was thinking it'd be useful to vertically center the buttons, but if they look fine without it I'll take it out

ui/DesignSystem/Component/Topbar/Menu.svelte
```diff
@@ -0,0 +1,48 @@
+<script>
+  import { location } from "svelte-spa-router";
+  import * as path from "../../../lib/path.js";
+
+  import MenuItem from "./MenuItem.svelte";
+
+  export let items = [];
```
sarahscott: This was something I was curious about - if this is `null`, the `#each` block will throw an error (`null` has no `length`) if the prop isn't received for some reason. If it's an empty array, it'll at least render without error. Do you think that's being too explicit about the type?

ui/Screen/Project.svelte
rudolfs: Vertical centering should already be provided by the `<Topbar>` component.

ui/DesignSystem/Component/Topbar/Menu.svelte
rudolfs: I see what you mean. What I'm wondering is, shouldn't it be an exception in case one forgets a required prop as a reminder?

ui/DesignSystem/Component/NotificationFaucet.svelte
```diff
@@ -9,7 +9,7 @@
   import { blur } from "svelte/transition";
 
   export let style = null;
-  let messages = [];
+  let messages = null;
```
rudolfs: This actually is not a prop and initialising it with `null` throws an exception on app start.

Screenshot 2020-03-02 at 15 31 23

ui/Screen/DesignSystemGuide.svelte
```diff
@@ -52,6 +53,24 @@
...
+    },
+    {
+      title: "Add something else",
+      icon: Icon.Plus,
+      event: () => console.log("event(Add Something Else)")
+    },
+    {
+      title: "Send something",
+      icon: Icon.ArrowUp,
+      event: () => console.log("event(Send Something")
```
rudolfs:
```suggestion
      event: () => console.log("event(Send Something)")
```

ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
```diff
@@ -0,0 +1,106 @@
...
+    box-shadow: 0px 4px 8px rgba(51, 51, 51, 0.08);
+    border-radius: 4px;
+  }
+
+  .header {
+    padding: 12px;
+    border-bottom: solid 1px var(--color-lightgray);
+    display: flex;
+    cursor: pointer;
+
```
rudolfs:
```suggestion
```

ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
```diff
@@ -0,0 +1,106 @@
...
+    color: var(--color-darkgray);
+  }
+
+  .menu {
+    cursor: pointer;
+  }
+
+  .menu-item {
+    display: flex;
+
```
rudolfs:
```suggestion
```

ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
```diff
@@ -0,0 +1,106 @@
...
+    background-color: var(--color-lightgray-tint-10);
+  }
+
+  .modal {
+    position: absolute;
+    top: 100%;
+    right: 0;
+    width: 240px;
+    margin-top: 15px;
+
```
rudolfs:
```suggestion
```

ui/DesignSystem/Primitive/Icon/Copy.svelte
```diff
@@ -0,0 +1,41 @@
+<script>
+  export let style = null;
+</script>
+
+<svg
```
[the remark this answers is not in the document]
rudolfs: This has been fixed in the style guide now: https://www.figma.com/file/D4chIBeYwb3a7zrv1jrCfQdL/App-styleguide?node-id=526%3A75

```
<svg
  {style}
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="var(--color-gray)"
  xmlns="http://www.w3.org/2000/svg">
  <path
    fill-rule="evenodd"
    clip-rule="evenodd"
    d="M10 6H18V14H16H14H10V10V8V6ZM8 8V6C8 4.89543 8.89543 4 10 4H18C19.1046 4
    20 4.89543 20 6V14C20 15.1046 19.1046 16 18 16H16V18C16 19.1046 15.1046 20
    14 20H6C4.89543 20 4 19.1046 4 18V10C4 8.89543 4.89543 8 6 8H8ZM8 10H6L6
    18H14V16H10C8.89543 16 8 15.1046 8 14V10Z" />
</svg>
```

ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  .modal {
+    position: absolute;
+    top: 100%;
+    right: 0;
+    width: 240px;
+    margin-top: 15px;
+    background-color: white;
+    box-shadow: 0px 4px 8px rgba(51, 51, 51, 0.08);
```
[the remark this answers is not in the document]
rudolfs: Actually, I'm not sure we can do this with the way we manage colors currently.

To be able to pass around rgba values we either need mixin support via a pre-processor or pass around global classes. I was playing around with these ideas in these experimental branches:
  - [global classes a la tailwind](https://github.com/radicle-dev/radicle-upstream/tree/rudolfs/utility-css-poc)
  - [sass preprocessor](https://github.com/radicle-dev/radicle-upstream/tree/rudolfs/preprocessor-poc)

There is an issue, which is tracking the progress on this: https://github.com/radicle-dev/radicle-upstream/issues/160.

Both solutions have their benefits and drawbacks, I'm not super happy with either yet.

I think hardcoding properties for shadows would be okay for now.
Besides I already did this for a shadow in the Sidebar:
https://github.com/radicle-dev/radicle-upstream/blob/master/ui/DesignSystem/Component/Sidebar.svelte#L90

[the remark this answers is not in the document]
MeBrei: Talked to @rudolfs in person. For now we agreed to add the rgba color to `global.css`. The `var` can be used directly here then. We should create an issue to support alpha colors in the color.json in the future.

The `background-color` here should still use the standard `var(--color-white)` though.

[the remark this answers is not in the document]
sarahscott: We could also standardize `boxShadow''s à la [Material-UI](https://material-ui.com/system/shadows/)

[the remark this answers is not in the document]
rudolfs: @sarahscott I guess that would be a third option - having generic wrappers which expose styling via props. Somehow I feel like having a wrapper just for styling a bit heavy-handed (then again we do it already for positioning with `<Flex>`).

I think I would rather prefer the simpler CSS-variable approach for now, at least until we figure out whats the right thing for us is in terms of styling abstractions.

public/colors.css
```diff
@@ -27,6 +27,7 @@
   --color-lightgray-tint-10: #e9edf1;
   --color-lightgray: #ced8e1;
   --color-lightgray-shade-10: #b3c3d1;
+  --color-light-shadow-gray: rgba(51, 51, 51, 0.08);
```
sarahscott: @MeBrei @rudolfs does this make sense here or in `global`?

rudolfs: `colors.css` is auto-generated and will be overwritten next time someone runs `yarn generate:colors`.

Let's put this CSS variable in `public/global.css` for now.
I also think `--color-lightgray-transparency-08` would better fit our current naming scheme.

ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
```diff
@@ -48,9 +49,11 @@
     right: 0;
     width: 240px;
     margin-top: 15px;
-    background-color: white;
-    box-shadow: 0px 4px 8px rgba(51, 51, 51, 0.08);
```
rudolfs: Could you also use the variable here: https://github.com/radicle-dev/radicle-upstream/blob/master/ui/DesignSystem/Component/Sidebar.svelte#L90 while at it?

public/colors.css
MeBrei: `color.css` will be overwritten once you run `yarn generate:colors` the next time. I would put it in `general.css` for now, but maybe write an issue in the Housekeeping milestone to include transparent colors in the color generation.

PR #196, opened 2020-02-28
Add ability to escape 'Add Project' and 'Register Project' modals with esc key

sarahscott:


ui/DesignSystem/Component/ModalLayout.svelte
```diff
@@ -3,6 +3,13 @@
   import { Icon } from "../Primitive";
 
   export let dataCy = null;
+
+  // close on ESC
```
xla: This comment should be closer to the `keyCode` making it clear that 27 is the escape key, otherwise a const more appropriatly named would also help communicating the intend.

rudolfs:
Good thinking, I like the idea!

We already use a hotkey library and have a central place for managing hotkeys: `ui/lib/hotkeys.js`.

You could do something like this, which wouldn't be elegant (as you'd have to add each modal's path), but would work for now until we have time to re-think our routing:

```diff
diff --git a/ui/lib/hotkeys.js b/ui/lib/hotkeys.js
index 5075cc6..4c31924 100644
--- a/ui/lib/hotkeys.js
+++ b/ui/lib/hotkeys.js
@@ -21,7 +21,8 @@ export const initializeHotkeys = () => {
   hotkeys("esc", () => {
     if (
       path.active(path.help(), get(location)) ||
-      path.active(path.designSystemGuide(), get(location))
+      path.active(path.designSystemGuide(), get(location)) ||
+      path.active(path.createProject(), get(location))
     ) {
       pop();
     }
```

sarahscott:
Oh sick! Good to know about the hotkey library

PR #199, opened 2020-02-28
Implement Basic User Identity mocks

### LINKED ISSUE #177: Implement mock endpoints for Basic User Identity
Requires: https://github.com/radicle-dev/radicle-upstream/issues/176
Just return some fake data, so we can already work on the UI.

xla:
Closes #177

### DIFF
--- a/proxy/src/graphql/project.rs
+++ b/proxy/src/graphql/project.rs
@@ -65,10 +65,20 @@ impl From<meta::Project> for Metadata {
 
 /// Radicle project for sharing and collaborating.
 #[derive(GraphQLObject)]
-#[graphql(description = "Radicle project")]
+#[graphql(description = "Project")]
 pub struct Project {
     /// Unique identifier of the project in the network.
     pub id: juniper::ID,
     /// Attached metadata, mostly for human pleasure.
     pub metadata: Metadata,
+    /// Coarse set of statistics for the project source code.
+    pub stats: Stats,
+}
+
+#[derive(GraphQLObject)]
+#[graphql(name = "ProjectStats")]
+pub struct Stats {
+    pub branches: i32,
+    pub commits: i32,
+    pub contributors: i32,
 }
--- a/proxy/src/graphql/project.rs
+++ b/proxy/src/graphql/project.rs
@@ -63,6 +63,13 @@ impl From<meta::Project> for Metadata {
     }
 }
 
+#[derive(GraphQLEnum)]
+pub enum Registered {
+    Not,
+    Org,
+    User,
+}
+
 /// Radicle project for sharing and collaborating.
 #[derive(GraphQLObject)]
 #[graphql(description = "Project")]
@@ -71,6 +78,8 @@ pub struct Project {
     pub id: juniper::ID,
     /// Attached metadata, mostly for human pleasure.
     pub metadata: Metadata,
+    /// Signals if a project is on the Registry.
+    pub registered: Registered,
     /// Coarse set of statistics for the project source code.
     pub stats: Stats,
 }
--- a/proxy/src/graphql/project.rs
+++ b/proxy/src/graphql/project.rs
@@ -74,7 +74,7 @@ pub enum Registered {
     User,
 }
 
-/// Radicle project for sharing and collaborating.
+/// Project for sharing and collaborating.
 #[derive(GraphQLObject)]
 #[graphql(description = "Project")]
 pub struct Project {
--- a/proxy/src/graphql/project.rs
+++ b/proxy/src/graphql/project.rs
@@ -63,17 +63,6 @@ impl From<meta::Project> for Metadata {
     }
 }
 
-/// Shows if a project exists on the Registry and distinguishes between Org and User owned.
-#[derive(GraphQLEnum)]
-pub enum Registered {
-    /// Project is not present on the Registry.
-    Not,
-    /// Project is registered under an Org.
-    Org,
-    /// Project is registered under a User.
-    User,
-}
-
 /// Project for sharing and collaborating.
 #[derive(GraphQLObject)]
 #[graphql(description = "Project")]
@@ -83,7 +72,7 @@ pub struct Project {
     /// Attached metadata, mostly for human pleasure.
     pub metadata: Metadata,
     /// Signals if a project is on the Registry.
-    pub registered: Registered,
+    pub registered: bool,
     /// Coarse set of statistics for the project source code.
     pub stats: Stats,
 }
--- a/proxy/src/graphql/project.rs
+++ b/proxy/src/graphql/project.rs
@@ -63,10 +63,14 @@ impl From<meta::Project> for Metadata {
     }
 }
 
+/// Shows if a project exists on the Registry and distinguishes between Org and User owned.
 #[derive(GraphQLEnum)]
 pub enum Registered {
+    /// Project is not present on the Registry.
     Not,
+    /// Project is registered under an Org.
     Org,
+    /// Project is registered under a User.
     User,
 }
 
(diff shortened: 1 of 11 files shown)


proxy/src/graphql/project.rs
```diff
@@ -63,12 +63,39 @@ impl From<meta::Project> for Metadata {
...
+/// Shows if a project exists on the Registry and distinguishes between Org and User owned.
+#[derive(GraphQLEnum)]
+pub enum Registered {
+    /// Project is not present on the Registry.
+    Not,
+    /// Project is registered under an Org.
+    Org,
+    /// Project is registered under a User.
+    User,
+}
```
cloudhead: I'm thinking this might be annoying to use when you just want to know registered or not. Do we need to differentiate whether it's registered under an org or a user somewhere? Alternatively you could implement a `impl From<Registered> for bool`, so at least you can do `if registered.into() { .. }`, but I wonder if the distinction is necessary in the first place :thinking:

proxy/src/graphql/project.rs
```diff
@@ -63,12 +63,39 @@ impl From<meta::Project> for Metadata {
...
+/// Coarse statistics for the Project source code.
+#[derive(GraphQLObject)]
+#[graphql(name = "ProjectStats")]
+pub struct Stats {
+    /// Amount of known branches.
+    pub branches: i32,
+    /// Number of commits on the default branch.
+    pub commits: i32,
+    /// Amount of unique commiters on the default branch.
+    pub contributors: i32,
```
cloudhead: These stats should actually be dependent on which branch you're on, not the default one.

proxy/src/graphql/project.rs
```diff
@@ -63,12 +63,39 @@ impl From<meta::Project> for Metadata {
...
+pub enum Registered {
+    /// Project is not present on the Registry.
+    Not,
+    /// Project is registered under an Org.
+    Org,
+    /// Project is registered under a User.
+    User,
+}
+
 /// Radicle project for sharing and collaborating.
```
MeBrei:
```suggestion
/// Project for sharing and collaborating.
```

proxy/src/graphql/schema.rs
```diff
@@ -185,6 +253,22 @@ impl Query {
             .map(|id| juniper::ID::from(id.0.to_string()))
             .collect::<Vec<juniper::ID>>())
     }
+
+    fn identity(_ctx: &Context, id: juniper::ID) -> Result<Option<Identity>, error::Error> {
+        Ok(Some(Identity {
+            id: id.clone(),
+            shareable_entity_identifier: juniper::ID::new(format!("cloudhead@{}", id.to_string())),
```
cloudhead: :eyes:

proxy/src/registry.rs
```diff
@@ -219,6 +233,29 @@ impl Registry {
...
+        handle: String,
+        id: String,
+    ) -> Result<Transaction, error::Error> {
+        let message_handle = registry::String32::from_string(handle)
+            .map_err(|_| error::UserValidation::HandleTooLong)?;
+        let message_id = registry::String32::from_string(id)
+            .map_err(|_| error::UserValidation::HandleTooLong)?;
+
+        Ok(Transaction {
+            id: registry::H256::random(),
```
cloudhead: Strange to me that the id is an attribute of the transaction, and not just the content hash :eyes:

proxy/src/graphql/schema.rs
```diff
@@ -185,6 +253,22 @@ impl Query {
...
+                handle: "cloudhead".into(),
+                display_name: Some("Alexis Sellier".into()),
+                avatar_url: Some("https://avatars1.githubusercontent.com/u/4077".into()),
+            },
+        }))
+    }
+
+    fn user(_ctx: &Context, handle: juniper::ID) -> Result<Option<juniper::ID>, error::Error> {
+        Ok(None)
+    }
```
MeBrei: At some point we might need a flag here to mock that the user exists so that we can update the user profile (registered badge, remove the register user handle button) and search users and add them as members.

proxy/src/registry.rs
```diff
@@ -219,6 +233,29 @@ impl Registry {
...
+    pub async fn register_user(
+        &self,
+        _author: &ed25519::Pair,
+        handle: String,
+        id: String,
+    ) -> Result<Transaction, error::Error> {
+        let message_handle = registry::String32::from_string(handle)
+            .map_err(|_| error::UserValidation::HandleTooLong)?;
+        let message_id = registry::String32::from_string(id)
+            .map_err(|_| error::UserValidation::HandleTooLong)?;
```
MeBrei:
```suggestion
            .map_err(|_| error::UserValidation:: IdTooLong)?;
```

proxy/src/graphql/project.rs
xla: Are their plans to show stats in the context of a branch switched project?

proxy/src/graphql/schema.rs
xla: @cloudhead Credit where credit is due.

proxy/src/registry.rs
xla: We receive the id from calls to the registry client, we don' really want to know how it's computed, so we return it simple as a key to the caller.

proxy/src/graphql/schema.rs
xla: I hope our mocks have evloved into proper endpoints by that point, otherwise we extend when the need arise.

proxy/src/graphql/project.rs
xla: Was planning on extending this to variants which carry the respective **org id** or **user handle**. An `Into` for `bool` sounds like a fun idea.

proxy/src/graphql/project.rs
cloudhead: Yeah, the project page will show exactly these stats, based on the branch you're on (I thought that's where you got this from)

proxy/src/registry.rs
cloudhead: Ah, gotcha!

proxy/src/graphql/project.rs
xla: Nope, these stats have been based on the project listing, but this info is valuable. WIll adjust documentation.

proxy/src/graphql/project.rs
cloudhead: Ah in that case it makes sense, if you are planning to use a different type for orgs vs user ids, since underneath they have the same format. I'll leave it up to you.

xla: Dropped this enum for now and we revise later.

proxy/src/graphql/project.rs
cloudhead: Ah cool!

xla: Actually this is going to be inconsistent. One of the stats is a amount of branches, whcih won't change in a branch context. Treating these as project global stats and look into branch specific stats independently makes more sense.

rudolfs: Just for reference, when I wrote down the requirements for this endpoint, I had the user profile project listing in mind. For that screen we don't select a branch, so it would make sense to display global values.

Screenshot 2020-03-02 at 10 11 17

proxy/src/graphql/project.rs
rudolfs: I looked at the designs again (and had a chat with @juliendonck), we're using this only for display a "registered" badge in the UI. I also don't see why we would ever want to differentiate whether it's under a project or user, so a bool would be much easier to use in the front-end.

xla: Addressed.

PR #200, opened 2020-02-28
Extract transaction overview

MeBrei:
The transaction detail view will be used in every modal for Registry related actions (register user/org/member, create registered project). The same view will also be used for inspecting (pending) transactions.

For now the "old" register project modal was used to reflect the changes, but the styling was already adjusted to the new figma screens. 

Prepares #202


ui/DesignSystem/Component/index.js
```diff
@@ -12,7 +12,8 @@ import Sidebar from "./Sidebar.svelte";
 import SidebarLayout from "./SidebarLayout.svelte";
 import SourceBrowser from "./SourceBrowser.svelte";
 import StepCounter from "./StepCounter.svelte";
-import TxRow from "./TxRow.svelte";
+import Transaction from "./Transaction.svelte";
+import TxRow from "./Transaction/Row.svelte";
```
rudolfs: Ditto.

MeBrei:
Thanks @rudolfs for the feedback!

> $ amount font is bigger than in mockup

Which one do you mean? I can't see a difference 😬 

> top, bottom and right row padding seems bigger than in the mockup

I talked to @juliendonck about the line padding and he fixed the height to `72px` which should also apply to any other transaction overview. I applied it and the other padding changes.

Issue #146, opened 2020-02-10
Write missing cypress tests for project creation


rudolfs:
Cover the following functionality:

- [x] opening the "create project" modal from sidebar/projects page
- [ ] ~~when the projects page is empty → show create project messaging (blocked by https://github.com/radicle-dev/radicle-upstream/issues/401)~~
- [x] creating a project from scratch
- [x] creating a project from an existing repo
- [x] validations
  - [x] name, image URL, directory selection
  - [x] 'create project' button state
  - [x] directory path
- [ ] branch selection (blocked by https://github.com/radicle-dev/radicle-upstream/issues/434)
- [x] notification about a newly created project
- [ ] ~~notification about failure~~
- [x] project listing should contain newly created project

~~To complete this issue we need to be able to start the proxy back-end without any mock data and be able to delete projects.~~ This is possible now with the `control` endpoints.

~~Blocked by #145.~~ We need to figure out an alternative workaround as Cypress doesn't allow us to test native Electron currently.

Issue #145, opened 2020-02-10
Use cypress electron integration when it's ready


rudolfs:
To properly test project creation we need to be able to test native calls to Electron. This is still in the works at Cypress, when they merge https://github.com/cypress-io/cypress/pull/6143, we can start working on upgrading our cypress setup to the pre-release and start testing native electron.

rudolfs:
Unfortunately the Cypress team won't be working on this feature in the foreseeable future: https://github.com/cypress-io/cypress/pull/6143#issuecomment-616992317.

We'll need to figure out a workaround for testing native electron.

PR #203, opened 2020-02-28
Extract domain entities

### LINKED ISSUE #201: Extract last domain entities from graphql module
During the restructuring of the proxy in #125 - a couple of entities were left in the graphql module and are in need of rescuing. Additionally #199 also introduced some of these and would run out of scope if done in there. We should extract:

* Project
* Identity

xla:
Closes #201


proxy/src/error.rs
```diff
@@ -1,4 +1,4 @@
-//! Domain errors returned by the API.
+//! Domain errors.
```
NunoAlexandre: Just "Domain errors" seems lacking to me.

proxy/src/project.rs
```diff
@@ -1,27 +1,12 @@
-//! Abstractions for projects returned by the API.
+//! Abstractions for projectsI.
```
NunoAlexandre:
```suggestion
//! Abstractions for projects.
```

Could you expand that comment? What abstractions? Why do we need them?

xla:
@rudolfs Thanks for the catch, why is this not surfaced by our tests? Is it because of #159 ?

rudolfs:
@xla yes, #159, #145 and #146.

xla:
> Looks good! `ProjectMetadata` would be enough for me instead of `ProjectMetadataInput` but leave that with you +1

@NunoAlexandre `ProjectMetadata` exists as well. GraphQL distinguishes between input values and returned values. So input to queries and mutations are expressed differently. Therefore the on the surface superfluous looking suffix.

NunoAlexandre:
@xla that's clear! And, In that case, I like the suffix to make things explicit. Thanks for explaining!

Issue #189, opened 2020-02-26
Replace our current font with Inter & Source Code Pro


rudolfs:
We're moving to open source fonts.

https://github.com/rsms/inter/releases/download/v3.12/Inter-3.12.zip
https://github.com/adobe-fonts/source-code-pro/releases/tag/2.030R-ro%2F1.050R-it

rudolfs:
Resolved in #204

PR #204, opened 2020-03-02
Use open source fonts

rudolfs:
Refs #189


ui/Screen/DesignSystemGuide.svelte
```diff
@@ -84,20 +85,57 @@
     title="Typography"
     subTitle="Using GT America and GT America Mono from Grill Type">
```
sarahscott: This should be changed to reflect the new font family names 😉

rudolfs: Good catch ⚡️

Issue #175, opened 2020-02-24
Overview: Identities I


rudolfs:
This issue tracks all of the work (including dependencies) which has to be done to complete the [Basic User Identity](https://www.notion.so/monadic/Basic-User-Identity-0688a802e65040778a459892d6922919) spec.

#### Open questions
  - proxy: when you register your handle, but you pick a different one from the one you picked at identity creation, what should happen with the old/new handle?

### UI
  - [x] copy-to-clipboard entity id component (#180)
  - [x] emojicon generation module (#181)
  - [x] implement onboarding flow screens (#182)
  - [x] implement profile page screens (#186)
  - [x] wire up onboarding and profile screens with app boot logic (#183)
    - [x] fix: navigation from identity creation flow to registering - on cancel/finish, jumps to start of identity flow
  - [x] do not query avatar in user registration flow (#230)
    - [x] proxy: use id to generate the avatar fallback (instead of handle), (the endpoint for querying only     
    - [x] the avatar of a handle will still be needed for org registration)
    - [x] ui: remove avatar query from the user registration modal, always use the avatar fallback of the identity (#276)
  - [x] redo empty profile project page (restyle to match current design) (#263)
  - [x] fix: copy-to-clipboard for identity card (identity creation flow) (#263)
  - [x] fix: use shareable entity identifier for profile `...` dropdown copy-to-clipboard (#263)
  - [x] fix: unregistered project avatar heading (#263)
  - [x] adjust create project flow for local projects to new design & flow (#264)
  - [x] fix: use shareable entity identifier for project and projects listing `...` dropdown copy-to-clipboard (#393)
  - [ ] ~~handle registration: get transaction costs from proxy (#391)~~ will do them when registry will provide them
  - [x] when everything is integrated, write end-to-end tests for the whole identity creation flow (#184)

### Proxy
  - [x] define UI ↔ proxy API (#176)
  - [x] implement mock endpoints (#177)
  - [x] integrate real functionality into API by replacing mock endpoints (#185)
  - [x] switch avatar generation to id(#230)
  - [x] create identity query should return the correct avatarFallback or avatarUrl (it returns the same hardcoded avatar for every handle)
  - [x] integrate identity (the identity should be persisted on refresh or app reboot)
  - [ ] ~~handle registration: endpoint for transaction costs for handle registration (#391)~~
  - [x] add shareable entity identifier to `project` and `projects` endpoints  (#393)

For more info see #171

[...]

Issue #182, opened 2020-02-24
Implement onboarding flow screens


rudolfs:
Implement all the visual parts of the Basic User Identities onboarding flow, including:
- [ ] `cancel` and `x` buttons send back to welcome screen
- [ ] identity card fragment
  - [ ] fall back to emojicon if no avatar url is provided
- [ ] validation for name presence
- [ ] validation for avatar URL
- [ ] enable/disable `create` button depending on validations
- [ ] wire up mock APIs

<img width="1283" alt="Screenshot 2020-02-25 at 09 03 54" src="https://user-images.githubusercontent.com/158411/75227054-c720dd80-57ad-11ea-897b-c2e9a78403ea.png">

**Note**: this issue doesn't include the actual profile page.

Issue #180, opened 2020-02-24
Copy-to-clipboard entity identifier component


rudolfs:
Make a component that shows a shareable entity identifier.

- [ ] depending on the available container space, shorten the identifier with an ellipsis (`...`)
- [ ] when clicked this component should copy the identifier to the clipboard

For the copy-to-clipboard functionality `native/main.js` will have to be extended to provide access to the [Electron clipboard API](https://www.electronjs.org/docs/api/clipboard).

<img width="401" alt="Screenshot 2020-02-24 at 15 04 33" src="https://user-images.githubusercontent.com/158411/75158855-4adbbb00-5717-11ea-973f-484b9788e0ed.png">
<img width="170" alt="Screenshot 2020-02-24 at 15 05 04" src="https://user-images.githubusercontent.com/158411/75158861-4c0ce800-5717-11ea-9ad8-ee08e0889f90.png">
<img width="240" alt="Screenshot 2020-02-24 at 15 05 49" src="https://user-images.githubusercontent.com/158411/75158863-4ca57e80-5717-11ea-959a-c283881c061f.png">

xla:
Very helpful feature and the resulting infrastructure will be helpful for many a context (i.e. navigate to a commit, etc.). I think there is something important which needs to precede this. We need to agree on a consistent uri format for our deep links. We should do this consistently and in tandem with our routing strategy so we can potentially link to all possible parts of the app. Potentially enable even from outside the app itself a la `upstream://proejcts/abc1234/source/tree/master`.

rudolfs:
This issue is only about the visuals of the component and the ability to copy whatever string is passed to the system clipboard. It shouldn't really care what the string is.

I agree that we'll have to re-think our routing strategy.  Will create an issue for that.

xla:
Good point, didn't mean to hijack the issue, let's capture it in a proper way.

nassarhayat:
hi @rudolfs can you add me as member to notion file mentioned?

rudolfs:
Thanks for raising the issue @nassarhayat — we’ve been discussing how to manage this exact issue. The intention is to make all relevant documents as public as possible while also keep them organized and such. We’ll be figuring out if we’ll make Notion pages public or cross-post info to central place like GitHub. We’ve got an internal thread on it now, I’ll keep you updated with any developments!

However to complete this issue, you won't need any of the information from Notion (I also removed the link to it for now).

[...]

PR #211, opened 2020-03-05
New identity flow

sarahscott:
Part of #175 

TODO:
- [x] Respond to feedback
- [x] Test coverage

identity_flow


ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -178,5 +172,11 @@
         <Title>Profile</Title>
       </div>
     </li>
+
+    <li class="item">
```
sarahscott: Pretty sure this is actually where the 'new project' button will go in the sidebar; just using it as a convenience for now

xla:
I think this should be more aptly called `New identity flow`. We might benefit from referring to a user only in the context of a registered handle on chain and otherwise refer to an identity.

xla:
Think this PR relates to #175 - or am I confused @sarahscott ?

ui/Screen/IdentityCreationSuccess.svelte
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  .identity-card {
+    display: flex;
+    align-items: center;
+    background-color: var(--color-almostwhite);
+    padding: 24px;
+    margin-bottom: 34px;
+    width: 100%;
+  }
```
MeBrei: The identity card should have `border-radius: 2px;`.

xla:
Ref #182

ui/DesignSystem/Component/Avatar.svelte
```diff
@@ -7,7 +7,7 @@
   export let title = null;
 
   export let variant = "user"; // user | project
-  export let size = "regular"; // regular | medium | big
+  export let size = "regular"; // regular | medium | big | xl
```
MeBrei: I just saw that in `Title.svelte` the next step after `big` is `huge`. I think it would be nice to make it consistent.

ui/Screen/IdentityCreation.svelte
```diff
@@ -0,0 +1,175 @@
...
+
+  const constraints = {
+    handle: {
+      presence: {
+        message: "You must provide a handle",
+        allowEmpty: false
+      },
+      format: {
+        pattern: SINGLE_WORD_MATCH,
+        message: "Handle should match [a-z0-9][a-z0-9_-]+"
```
rudolfs: As the validation messages are coupled with the matchers (i.e. when changing the matcher, you'd have to change the message as well), does it even make sense to extract the regex matchers into a separate file?

ui/Screen/IdentityCreation.svelte
```diff
@@ -0,0 +1,175 @@
...
+      });
+
+      const responseData = response.data.createIdentity;
+      setCurrentIdentity({
+        handle: responseData.metadata.handle,
+        displayName: responseData.metadata.displayName,
+        avatarUrl: responseData.metadata.avatarUrl,
+        shareableEntityIdentifier: responseData.shareableEntityIdentifier
+      });
+      push("/identity/success");
```
rudolfs: I'd prefer the whole flow just be one screen, like @MeBrei mentioned previously. But if you still want to have it as separate screens, I'd recommend extracting this path in `path.js`.

ui/Screen/IdentityCreation.svelte
```diff
@@ -0,0 +1,175 @@
...
+        message: "Handle should match [a-z0-9][a-z0-9_-]+"
+      }
+    },
+    displayName: {
+      format: {
+        pattern: NAME_MATCH,
+        message: "Display name should match [a-z0-9]"
+      }
+    },
+    avatarUrl: {
```
rudolfs: Could also use the built-in matcher to keep it simple:
```
    imageUrl: {
      optional: {
        url: {
          schemes: ["http", "https"],
          message: "Not a valid avatar URL",
          allowLocal: false
        }
      }
    },
```

We already do it like this in `CreateProject.svelte`.

ui/store/identity.js
```diff
@@ -0,0 +1,18 @@
+import { writable } from "svelte/store";
+
+// TODO: define & structure this store
+
+// e.g. does each value gets its own store...
+//    const handle = "cloudhead" // as writable();
+//    const shareableEntityIdentifier = cloudhead@ViJQHAdeZoiEbaE5vv83dpjEun.rad" // as writable();
```
rudolfs: In my understanding it makes sense to have a separate store for each value in Svelte.

If you store an object in a Svelte store, then each time there is an attribute update, the whole object will have to be re-written, which will trigger all associated components to update. This might not perform/look well.

rudolfs: Then again, it might become unwieldy if we have a gazillion stores and don't have consistent name-spacing. Maybe it doesn't matter if a given object is anyway updated via one back-end query in one swoop.

In other words: I haven't figured out what the best practice is in this regard for Svelte 😬

sarahscott: Interesting - will update to have one value per store. Will also make a research issue for collecting knowledge & questions about best practices 😊

ui/Screen/IdentityCreation.svelte
```diff
@@ -0,0 +1,49 @@
+<script>
+  import { pop, replace } from "svelte-spa-router";
+
+  import { showNotification } from "../store/notification.js";
+  import { ModalLayout } from "../DesignSystem/Component";
+  import { Button } from "../DesignSystem/Primitive";
+  import IdentityCreationForm from "./Identity/IdentityCreationForm.svelte";
+  import IdentityCreationSuccess from "./Identity/IdentityCreationSuccess.svelte";
+
+  const steps = {
```
sarahscott: @xla @rudolfs @MeBrei this is one way to do js "enums"; how do you feel about it? It makes the markup more readable but ofc isn't as pretty as _real_ enums 😜

ui/Screen/IdentityCreationSuccess.svelte
```diff
@@ -0,0 +1,78 @@
...
+          <Avatar
+            size="xl"
+            imageUrl={$currentIdentityStore.avatarUrl || undefined} />
+          <div class="identity-card-text-container">
+            {#if $currentIdentityStore.displayName}
+              <Title>{$currentIdentityStore.displayName}</Title>
+            {/if}
+            <Text style="color: var(--color-darkgray);">
+              {$currentIdentityStore.shareableEntityIdentifier}
+            </Text>
```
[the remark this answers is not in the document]
sarahscott: Will do in a separate PR; issue can be found in #180

ui/Screen/IdentityCreation.svelte
```diff
@@ -0,0 +1,173 @@
...
+      <Input.Text
+        placeholder="Add a display name"
+        bind:value={displayName}
+        valid={!(validations && validations.displayName)}
+        validationMessage={validations && validations.displayName && validations.displayName[0]}
+        style="margin-top: 16px;" />
+      <Input.Text
+        placeholder="Avatar url"
+        bind:value={avatarUrl}
+        style="margin: 16px 0 32px 0;" />
```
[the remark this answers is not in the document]
sarahscott: Also will do in a new PR; made an issue #236

ui/lib/validationHelpers.js
```diff
@@ -0,0 +1,13 @@
...
+export const SINGLE_WORD_MATCH = new RegExp("^[a-z0-9][a-z0-9_-]+$", "i");
+
+// General name match
+export const NAME_MATCH = new RegExp("^[a-z0-9 ]+$", "i");
+
+// TODO: fix this regex
+export const IMAGE_FILENAME = new RegExp(
+  ".*.(gif|jpe?g|tiff|png|webp|bmp)$",
+  "i"
+);
```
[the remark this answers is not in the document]
sarahscott: What would you suggest? Would `IMAGE_FILENAME_REGEXP` work?

ui/Screen/IdentityCreation.svelte
sarahscott: I think it makes sense because some of the regular expressions can be shared (e.g. the one for identity handles and project names). I also think the components look cleaner without them, but if it'd be simpler to leave them in the `<script>` tags instead of extracting them we can go back to that!

ui/Screen/IdentityCreation.svelte
xla: Definitely an improvement over just raw numbers, or ungrouped consts. This is friendlier for the reader to follow what is actually behind the steps.

ui/lib/validationHelpers.js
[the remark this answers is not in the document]
MeBrei: If you keep the regexp for `IMAGE_FILENAME` I think the name is fine, I thought you were adding the url part as well.
As for the others, maybe `SINGLE_WORD_MATCH` could be `HANDLE_MATCH` and `NAME_MATCH` could be `DISPLAY_NAME_MATCH`. If they are used for different purposes than that, maybe we should keep them separate, as they might diverge in the future.

ui/Screen/Identity/IdentityCreationForm.svelte
```diff
@@ -0,0 +1,189 @@
...
+        handle: $handle
+        displayName: $displayName
+        avatarUrl: $avatarUrl
+      ) {
+        id
+        shareableEntityIdentifier
+        metadata {
+          handle
+          displayName
+          avatarUrl
```
MeBrei: Can you also request the `avatarFallback` here, so it can be filled in the store as well?

ui/Screen/Identity/IdentityCreationSuccess.svelte
```diff
@@ -0,0 +1,76 @@
...
+    <Text style="margin: 20px 0; color: var(--color-gray);">
+      This is your peer-to-peer identity. Even though your radicleID is unique,
+      your handle isn't. To get a unique handle, you have to
+      <!-- TODO(sarah): actually link to handle registration flow -->
+      <a class="registration-link" href={path.projects()} use:link>
+        register it.
+      </a>
+    </Text>
+    <div class="identity-card">
+      <Avatar size="huge" imageUrl={$avatarUrlStore || undefined} />
```
MeBrei: Here the fallback Avatar should be shown, if no url is provided (see comment above)

ui/lib/hotkeys.js
```diff
@@ -26,13 +26,21 @@ export const initializeHotkeys = () => {
     push(path.help());
   });
 
+  hotkeys("shift+i", () => {
+    if (path.active(path.createIdentity(), get(location))) {
+      pop();
+    }
+    push(path.createIdentity());
+  });
```
MeBrei: Can you add a `todo` here to remove the temporary hotkey?

ui/Screen/Identity/IdentityCreationForm.svelte
sarahscott: Done! This is a great example of something we can throw in a data access layer at some point & make our lives a lot easier 😇

rudolfs: I'm not sure I follow @sarahscott, did you forget to do the change? I still see that you're doing a separate query in `IdentityCreationSuccess.svelte` to retrieve the `avatarFallback`. It would be better to fetch it right here where you create the identity.

I initially also overlooked the `fallbackAvatar`, here's what the query could look like:
Screenshot 2020-03-23 at 18 09 41

ui/Screen/CreateProject.svelte
```diff
@@ -102,8 +102,8 @@
         allowEmpty: false
       },
       format: {
-        pattern: VALID_NAME_MATCH,
-        message: "Project name should match [a-z0-9][a-z0-9_-]+"
+        pattern: new RegExp(projectNameMatch, "i"),
+        message: `Project name should match ${projectNameMatch}`
```
rudolfs: Neat! 💯

ui/Screen/CreateProject.svelte
```diff
@@ -273,8 +272,8 @@
   }
 
   .create-project {
-    text-align: left;
-    width: 540px;
+    text-align: center;
+    flex: 1;
```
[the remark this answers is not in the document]
rudolfs: Still one to go:
Screenshot 2020-03-23 at 18 16 01

ui/Screen/Identity/IdentityCreationSuccess.svelte
```diff
@@ -0,0 +1,113 @@
...
+
+  import { Avatar, Button, Text, Title } from "../../DesignSystem/Primitive";
+
+  export let onClose,
+    onError = null;
+
+  const client = getClient();
+  let avatarFallback = null;
+
+  const getAvatarUrl = async () => {
```
rudolfs: We can get this data already in the `createIdentity` mutation before.

Issue #202, opened 2020-02-28
User registration


MeBrei:
All of the work which has to be done to complete the [User registration](https://www.notion.so/monadic/User-Registration-6bcb1ebd6f334d0290c95c5a05f54649) spec.

- [x] UI
  - [x] user registration modal (#216)
  - [x] modal follow ups
    - [x] wire up identity store (#252)
    - [x] open modal via buttons & remove the temp. hotkey (after new identity profile) (#232)
    - [x] add tests for availability of handles (after mocks replacement)
  - [x] show pending transactions
    - [x] transaction icons (#235)
    - [x] transaction center (#242)
    - [x] transaction detail view (#242)
    - [x] ~~transaction storage and reactivity of transaction center~~ (moved to #307)
    - [x] ~~transaction center animations (slide up/down of cards, slide in of center on first transaction)~~ (moved to #307)
    - [x] ~~implement status/progress/error & transaction costs after proxy changes~~ (moved to #225)
- [x] proxy
  - [x] define GraphQL schema for User Registration (#198)
  - [x] implement mock endpoints
  - [x] implement transaction endpoints
  - [x] integrate real functionality into API by replacing mock endpoints

Partially depends on #175

Issue #181, opened 2020-02-24
Emojicon generation module


rudolfs:
This module will be used whenever we have to show an avatar for a user or org, but no avatar image url is provided. This could simply be a function that, given a user/project name, returns an object:
```
{
  emoji: "👹"
  color: "#D976E3"
}
```

The mapping should always be the same, meaning a given user/project name will always return the same emoji/color combination.

User emojicons always return "🙂", only the background color differs depending on the id.
<img width="101" alt="Screenshot 2020-02-25 at 08 53 03" src="https://user-images.githubusercontent.com/158411/75226528-c9cf0300-57ac-11ea-91b3-a63880957804.png">

For orgs we pick from a pre-defined set of emojis and colors.
<img width="83" alt="Screenshot 2020-02-25 at 08 53 28" src="https://user-images.githubusercontent.com/158411/75226534-cc315d00-57ac-11ea-8214-6ef25ac934f3.png">

The org emoji set is TBD: @juliendonck.

Requires: #179.

rudolfs:
@cloudhead you mentioned that you might already have some code for this?

cloudhead:
Here's a working prototype: https://gist.github.com/cloudhead/545fd3fd9f89070672375eda75ab1093

* We need to choose an emoji font for Linux (as macOS/windows have one already)
* UTF-8 charset is required everywhere, for this to work.

cloudhead:
Example output:

![2020-02-25-123817_1184x104_scrot](https://user-images.githubusercontent.com/40774/75244174-9e0f4580-57cb-11ea-8cf3-b80ca6f2447d.png)

cloudhead:
Possible font candidate: https://github.com/MorbZ/OpenSansEmoji

xla:
Similar to #179 this information should come through the proxy API as part of the response for a proejct/org.

xla:
@cloudhead Can you port this to a cute little Rust module in the proxy source? PRs are open.

juliendonck:
Do I still need to make an emoji selection for this?

xla:
> Do I still need to make an emoji selection for this?

@cloudhead @rudolfs Maybe you can answer this.

PR #216, opened 2020-03-06
User handle registration flow

MeBrei:
Implements the user handle registration modal of #202 and connects it to the graphql mock endpoints.
To open the handle registration use the temporary hotkey `shift + t`.

Depends partially on #175, especially #181 and #186 
- [x]  query fallback avatar for changed user handle
- [ ] retrieve actual identity info when opening the modal


ui/lib/path.js
```diff
@@ -43,6 +43,7 @@ export const registerProject = id => `/projects/${id}/register`;
 export const wallet = _params => "/wallet";
 export const profile = _params => "/profile";
 export const help = _params => "/help";
+export const registerUser = _params => "/register-user";
```
xla: Is this path temporary? I'd expect building up a namespace similar to `projects`, i.e.: `/users/register`

ui/lib/hotkeys.js
```diff
@@ -11,6 +11,14 @@ export const initializeHotkeys = () => {
     push(path.designSystemGuide());
   });
 
+  // temporary hotkey to open user registration
```
xla:
```suggestion
  // TODO(merle): Remove temporary hotkey to open user registration
```

Make it more actionable. And helps audits for outstanding cleanups.

ui/App.svelte
```diff
@@ -17,6 +17,7 @@
   import Project from "./Screen/Project.svelte";
   import Projects from "./Screen/Projects.svelte";
   import RegisterProject from "./Screen/RegisterProject.svelte";
+  import RegisterUser from "./Screen/RegisterUser.svelte";
```
xla: This naming convention might not be ideal. It sounds like an active action instead of a concept. Concretely one **registers** the **user** on the `UserRegistration` screen. So I'd advocate for `UserRegistration` wich also groups the prefixes more naturally.

ui/Screen/RegisterUser.svelte
```diff
@@ -0,0 +1,103 @@
+<script>
+  import { ModalLayout, StepCounter } from "../DesignSystem/Component";
+
+  import PickHandleStep from "./RegisterUser/PickHandleStep.svelte";
+  import SubmitRegistrationStep from "./RegisterUser/SubmitRegistrationStep.svelte";
+
+  import { Text, Title } from "../DesignSystem/Primitive";
+  import { gql } from "apollo-boost";
+  import { getClient, mutate } from "svelte-apollo";
+  import { pop } from "svelte-spa-router";
```
xla: Would group third-party imports at the top and group owned ones closer to the implementation.

ui/Screen/RegisterUser.svelte
```diff
@@ -0,0 +1,103 @@
...
+          }
+        }
+      }
+    }
+  `;
+
+  let response;
+  let errorMessage;
+  const registerUser = async () => {
+    console.log("Register handle, id ", handle, id);
```
xla:
```suggestion
    // TODO(merle): Remove log statement or rework to some debug form which can be removed for production code.
    console.log("Register handle, id ", handle, id);
```

ui/Screen/RegisterUser.svelte
```diff
@@ -0,0 +1,103 @@
...
+          handle: handle,
+          id: id
+        }
+      });
+    } catch (error) {
+      errorMessage = error;
+      console.log("Register user error: ", errorMessage);
+    } finally {
+      console.log("Register user respone", response);
+      pop();
```
xla: It's strange to see a stray `pop` without any context, what does it do?

ui/Screen/RegisterUser.svelte
```diff
@@ -0,0 +1,103 @@
...
+          selectedStep={step}
+          steps={['Prepare', 'Submit']}
+          style="margin-bottom: 16px" />
+      </div>
+
+      <Title variant="big" style="margin: 48px 0 24px 0; text-align: center">
+        Register your handle
+      </Title>
+
+      {#if step === 1}
```
xla: Can we name alias these steps, maybe with `const`s?

ui/Screen/RegisterUser/PickHandleStep.svelte
```diff
@@ -0,0 +1,108 @@
...
+    } catch (error) {
+      validations = { handle: [error] };
+    }
+  };
+
+  validatejs.options = {
+    fullMessages: false
+  };
+
+  const VALID_NAME_MATCH = new RegExp("^[a-z0-9][a-z0-9_-]+$", "i");
```
xla: This match is not according to the spec in https://github.com/radicle-dev/registry-spec/ - skeptical if we should maintain this in the UI at all as it easily gets out of sync. Can we defer to the proxy to be authoritative on this?

ui/lib/path.js
MeBrei: It's temporary until the general routing regarding users/orgs is figured out.

xla: Makes perfect sense!

ui/Screen/RegisterUser.svelte
MeBrei: Maybe I am misunderstanding you, but I find having the numbers directly here makes the code easier to read. If I would replace it with `if step === pickHandleStep` I would have to find the `const` definition before knowing the order of steps (besides them being in order in the html). Especially since there are only 2 steps I would stick with the current implementation.

ui/Screen/RegisterUser/PickHandleStep.svelte
MeBrei: Issue to move validations into proxy: #219

The specifications in https://www.notion.so/monadic/Basic-User-Identity-0688a802e65040778a459892d6922919 also diverge from the `registry-spec`. I would leave the regexp as is for now.

ui/Screen/RegisterUser.svelte
MeBrei: `pop` brings you to your last location, in this case your user profile and specifically the tab which was opened before the register modal was opened.

xla: Understood, thanks for the explanation.

ui/Screen/RegisterUser/PickHandleStep.svelte
xla: The regexp in the Notion document is not in regards to the user handle input allowed on chain, but the restrictions for handles that are part of the identity metadata in CoCo. The spec is authoritative for all Registry concerns.

It's fine to leave it for now and capture the follow-up in #219.

ui/Screen/RegisterUser.svelte
xla: Very fair points, as the reviewer it was hard to imagine the state of the app when you see `step == 2`.

ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -8,12 +8,15 @@
...
+  //     imageUrl: "avatar url of the target (optional)"
   //   },
   //   payer: {
   //     name: "name of the owner of the paying wallet"
-  //     avatar: "avatar of the owner (optional)"
+  //     avatar: "avatar of the owner (optional)",
+  //     avatarFallback: "avatar fallback of the owner",
+  //     imageUrl: "avatar url of the owner (optional)"
   //   }
   // }
```
xla: Is this comment block a leftover or should it stay here for documentation purposes?

ui/DesignSystem/Component/Transaction/Row.svelte
```diff
@@ -2,6 +2,7 @@
   import { Flex } from "../../Primitive";
 
   export let style = null;
+  export let dataCy = null;
```
xla: This is for cypress, right? If so it would be good flag that explicitely to avoid confusion with other component attributes.

ui/DesignSystem/Primitive/Input/Text.svelte
```diff
@@ -71,12 +72,13 @@
     {disabled}
     on:change
     on:input />
-  {#if variant === 'handle' && value && avatar}
+  {#if variant === 'handle'}
```
xla: Instead of all these checks in here for a variant, wouldn't it be cleaner to refactor it into its own component?

ui/DesignSystem/Component/Transaction.svelte
MeBrei: For documentation purposes. It should be possible to look up the variations of the content of `tx` without having to go through the whole code.

ui/DesignSystem/Component/Transaction/Row.svelte
MeBrei: Yes, it is for cypress purposes. But we have a those in all tested components, so I would be hesitant to repeat a comment in every place.

ui/DesignSystem/Primitive/Input/Text.svelte
MeBrei: I'd prefer to keep them in the same component to have consistent styling.

PR #217, opened 2020-03-06
Implement basic avatar generation

### LINKED ISSUE #181: Emojicon generation module

cloudhead:
For now, we start with a short emoji whitelist. Eventually it will make sense to either add more emojis here, or use emoji ranges.

Closes #181

### DIFF
--- a/proxy/src/avatar.rs
+++ b/proxy/src/avatar.rs
@@ -1,0 +1,201 @@
+//! Org and user avatar generation.
+
+/// Emoji whitelist.
+const EMOJIS: &[&str] = &[
+    "⌚️", "📱", "📲", "💻", "⌨️", "💽", "💾", "💿", "📀", "📼", "📷", "📸", "📹", "🎥", "🎞", "📞",
+    "☎️", "📟", "📠", "📺", "📻", "⏰", "🕰", "⌛️", "⏳", "📡", "🔋", "🔌", "💡", "🔦", "💸", "💵",
+    "💴", "💶", "💷", "💰", "💳", "🧾", "💎", "⚖️", "🔧", "🔨", "", "🔩", "⚙️", "⛓", "🔫", "💣", "🔪",
+    "🗡", "⚔️", "🚬", "⚰️", "⚱️", "🏺", "🧭", "🧱", "🔮", "🧿", "🧸", "📿", "💈", "⚗️", "🔭", "🧰", "🧲", "🧪",
+    "🧫", "🧬", "🧯", "🔬", "🕳", "💊", "💉", "🚽", "🚰", "🚿", "🛁", "🛀", "🛀🏻", "🛀🏼", "🛀🏽",
+    "🛀🏾", "🛀🏿", "🧴", "🧵", "🧶", "🧷", "🧹", "🧺", "🧻", "🧼", "🧽", "🛎", "🔑", "🗝", "🚪", "🛍", "🧳",
+    "🛒", "🎁", "🎈", "🎏", "🎀", "🎊", "🎉", "🧨", "🎎", "🏮", "🎐", "🧧", "✉️", "📩", "📨", "📧",
+    "💌", "📥", "📤", "📦", "📪", "📫", "📬", "📭", "📮", "📯", "📜", "📃", "📄", "📑", "📊", "📈",
+    "📉", "📆", "📅", "📇", "📋", "📁", "📂", "🗂", "🗞", "📰", "📓", "📔", "📒", "📕", "📗", "📘",
+    "📙", "📚", "📖", "🔖", "🔗", "📎", "📐", "📏", "📌", "📍", "✂️", "🖊", "🖋", "✒️", "🖌", "📝", "✏️",
+    "🔍", "🔎", "🔏", "🔐",
+];
+
+/// An emoji.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Emoji(&'static str);
+
+/// An avatar.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Avatar {
+    /// The emoji component.
+    pub emoji: Emoji,
+    /// The background color component.
+    pub background: Color,
+}
+
+impl Avatar {
+    /// Generate an avatar from an input string.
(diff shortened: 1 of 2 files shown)


proxy/src/avatar.rs
```diff
@@ -0,0 +1,201 @@
...
+    "💴", "💶", "💷", "💰", "💳", "🧾", "💎", "⚖️", "🔧", "🔨", "", "🔩", "⚙️", "⛓", "🔫", "💣", "🔪",
+    "🗡", "⚔️", "🚬", "⚰️", "⚱️", "🏺", "🧭", "🧱", "🔮", "🧿", "🧸", "📿", "💈", "⚗️", "🔭", "🧰", "🧲", "🧪",
+    "🧫", "🧬", "🧯", "🔬", "🕳", "💊", "💉", "🚽", "🚰", "🚿", "🛁", "🛀", "🛀🏻", "🛀🏼", "🛀🏽",
+    "🛀🏾", "🛀🏿", "🧴", "🧵", "🧶", "🧷", "🧹", "🧺", "🧻", "🧼", "🧽", "🛎", "🔑", "🗝", "🚪", "🛍", "🧳",
+    "🛒", "🎁", "🎈", "🎏", "🎀", "🎊", "🎉", "🧨", "🎎", "🏮", "🎐", "🧧", "✉️", "📩", "📨", "📧",
+    "💌", "📥", "📤", "📦", "📪", "📫", "📬", "📭", "📮", "📯", "📜", "📃", "📄", "📑", "📊", "📈",
+    "📉", "📆", "📅", "📇", "📋", "📁", "📂", "🗂", "🗞", "📰", "📓", "📔", "📒", "📕", "📗", "📘",
+    "📙", "📚", "📖", "🔖", "🔗", "📎", "📐", "📏", "📌", "📍", "✂️", "🖊", "🖋", "✒️", "🖌", "📝", "✏️",
+    "🔍", "🔎", "🔏", "🔐",
+];
```
cloudhead: :information_source: We can't use `char` here, because some emojis use more than one unicode codepoint.

xla: Maybe we can capture that in a inline comment.

proxy/src/avatar.rs
```diff
@@ -0,0 +1,201 @@
...
+    if l < 0.5 {
+        c.lighten(0.125 * (1. - l))
+    } else {
+        c.lighten(0.125 * -l)
+    }
+}
+
+/// Fowler–Noll–Vo hash function. We use this simple hash function
+/// to make it easy to port the avatar generation algorithm to other
+/// platforms.
```
xla: This commenting formatting doesn't seem to respect our `rustfmt.toml`.

cloudhead:
Going to drop this here also if we want to do more fancy emoji things: https://github.com/open-i18n/rust-unic/tree/master/unic/emoji

proxy/src/avatar.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        // This isn't perceptual lightness, but whatever.
+        (r / n + g / n + b / n) / 3.
+    }
+
+    /// Ligthen a color by an amount between `-1.0` and `1.0`.
+    /// Panics if the amount isn't in that range.
+    fn lighten(self, amount: f32) -> Self {
+        assert!(amount <= 1.0);
+        assert!(amount >= -1.0);
```
xla: What does the compiler do to these macros when built with `--releae`? Generally if this is fallible I'd prefer it to return a `Result` to avoid panics in library code.

proxy/src/avatar.rs
```diff
@@ -0,0 +1,201 @@
...
+        Self {
+            emoji: generate_emoji(input),
+            background: compress_color(generate_color(input)),
+        }
+    }
+}
+
+/// A 32-bit RGBA color.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Color {
```
xla:
```suggestion
pub struct RGBA {
```

Could communicate the type better, but both work.

proxy/src/avatar.rs
cloudhead: Running `cargo fmt` doesn't produce a different output though?

proxy/src/avatar.rs
cloudhead: These stay around in release builds. You can use `assert_debug!` for debug-only asserts. The function isn't failible, but it can mis-used. If you don't like this approach, I can also just cap the amount, but this can hide bugs.

proxy/src/avatar.rs
cloudhead: I considered it, I really did. But I like `Color` because it communicates the purpose a little better.

proxy/src/avatar.rs
cloudhead: e34d612

proxy/src/avatar.rs
cloudhead: If you mean comment width, it's because `cargo fmt` will only wrap comment lines longer than the `comment_width` value. I don't think there's a way to force shorter lines to extend to `comment_width`. I usually wrap my comments manually because I find really long lines harder to read.

proxy/src/avatar.rs
xla: On the fence, what I don't want is panics in library code. Esp. from something like a color manipulation input being out of bound. Ideally the input given can never be incorrect, but that might go out of hand quickly for what we want to do. So either we cap it or return a `Result`, we have a already good amount of error infrastrcture in `error.rs`.

proxy/src/avatar.rs
xla: :smirk_cat:

Issue #213, opened 2020-03-05
Expose avatar fallbacks in GraphQL API


xla:
Pending the implementation in #181 the proxy should expose the generate values for an org and user.

- [x] Extend Org and User entities with avatar fallback information
- [x] Add schema to get avatar info for org id
- [x] Add schema to get avatar info for user handle

MeBrei:
@xla We might need an extra endpoint for this, which returns the avatar fallback for a unregistered ie available org handle. When registering your Org, you choose a org handle and the avatar should be shown in the ui before the org is actually created on the Registry.

xla:
@MeBrei Is this true for identity avatars as well?

MeBrei:
For identity the fallback avatar is only shown after creation (and if no custom avatar was provided). 
For user registration, we also want to show the fallback avatar while changing to a new handle.

xla:
There are currently avatars attached to users (meaning a registered handle on the Registry), so avatars only are important in the context of the identity. Does that change our understanding what endpoints we need? @MeBrei

MeBrei:
@xla My understanding was that there are avatars attached to identity and org. When you first open the app and create your identity, you can provide an avatar or get a fallback one depending on the name you choose. When you register your user, you might or have to change your previously picked name (which will become your user handle) and the fallback avatar has to change accordingly.

xla:
@MeBrei Our understandings match and we should have an endpoint previewing the generated avatar info given a `handle`. Updated the todo list up top.

MeBrei:
Proposal for the schema additions:

```graphql
type UpstreamQuery {
  ...
  avatar(handle: ID!): Avatar!
}

type IdentityMetadata {
  ...
  avatarFallback: Avatar!
}
```
with `Avatar` coming from `proxy/src/avatar.rs`. Currently we don't distinguish between user and org when creating the avatar so that one endpoint is enough.

xla:
@MeBrei Looks guuut, one small addition as we declare our types conversion types explicitely:

``` graphql
type Avatar {
    emoji: String!
    background: String!
}
```

PR #218, opened 2020-03-10
Extend schema with avatar fallback

xla:
First follow up to #213 implementing the GraphQL query for getting an avatar given a handle as input.

### DIFF
--- a/proxy/src/avatar.rs
+++ b/proxy/src/avatar.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::as_conversions, clippy::float_arithmetic)]
+
 //! Org and user avatar generation.
 
 use std::fmt;
@@ -5,6 +7,7 @@ use std::fmt;
 /// Emoji whitelist.
 ///
 /// Note that these are `str` and not `char` because an emoji can span multiple unicode scalars.
+#[allow(clippy::non_ascii_literal)]
 const EMOJIS: &[&str] = &[
     "⌚️", "📱", "📲", "💻", "⌨️", "💽", "💾", "💿", "📀", "📼", "📷", "📸", "📹", "🎥", "🎞", "📞",
     "☎️", "📟", "📠", "📺", "📻", "⏰", "🕰", "⌛️", "⏳", "📡", "🔋", "🔌", "💡", "🔦", "💸", "💵",
@@ -40,6 +43,7 @@ pub struct Avatar {
 
 impl Avatar {
     /// Generate an avatar from an input string.
+    #[must_use]
     pub fn from(input: &str) -> Self {
         Self {
             emoji: generate_emoji(input),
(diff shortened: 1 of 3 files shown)


proxy/src/graphql/schema.rs
```diff
@@ -149,6 +150,10 @@ impl Query {
         "1.0"
     }
 
+    fn avatar(handle: juniper::ID) -> Result<avatar::Avatar, error::Error> {
+        Ok(avatar::Avatar::from(&handle.to_string()))
+    }
```
MeBrei: Shouldn't we return `Err` if there is an error?

proxy/src/avatar.rs
```diff
@@ -19,7 +19,7 @@ const EMOJIS: &[&str] = &[
 
 /// An emoji.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct Emoji(&'static str);
+pub struct Emoji(pub &'static str);
```
cloudhead: This kind of defeats the purpose of a newtype, I would rather we implement `Display` on it. Then you can call `to_string()`.

proxy/src/graphql/schema.rs
```diff
@@ -359,6 +364,32 @@ pub struct ControlQuery;
...
+
+    fn g() -> i32 {
+        self.g as i32
+    }
+
+    fn b() -> i32 {
+        self.b as i32
+    }
+}
+
```
cloudhead: Why do we need these? Why can't we just use the `u8` fields directly? Is this a juniper thing?

xla: Juniper doesn't implement their `Value` trait for any other primitive than `i32`.

proxy/src/graphql/schema.rs
xla: None of the operations here a fallible so far, so there will never be an error at the moment. This might change down the road.

proxy/src/avatar.rs
xla: Addressed in 0fc506a.

proxy/src/graphql/schema.rs
cloudhead: Ugh. Maybe for the frontend it would be easier to have a hex representation eventually, eg. `"#ff0011"`, because we won't need the individual channels. You could have eg.

```rust
impl fmt::Display for Color {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "#{:02x}{:02x}{:02x}", self.r, self.g, self.b)
    }
}
```

xla: Thought about that as well, but the UI should determine how to handle colors. Esp. when there are plans to consistently move to `rgba()`. This would give the most flexibilty imho. @MeBrei Thoughts on this. Happy to hex, if it is the best format.

PR #222, opened 2020-03-12
Add official avatar list and usage

cloudhead:
I don't know what I'm doing wrt. the GraphQL stuff.


cloudhead:
HELP

proxy/src/graphql/schema.rs
```diff
@@ -150,8 +150,11 @@ impl Query {
         "1.0"
     }
 
-    fn avatar(handle: juniper::ID) -> Result<avatar::Avatar, error::Error> {
-        Ok(avatar::Avatar::from(&handle.to_string()))
+    fn avatar(handle: juniper::ID, usage: String) -> Result<avatar::Avatar, error::Error> {
+        let usage =
+            avatar::Usage::from_str(usage.as_str()).map_err(|e| error::Error::InvalidQuery(e))?;
```
MeBrei: Can we move the conversion of string to `avatar::Usage` into `avatar.rs`? It would make the code here clearer.

proxy/src/avatar.rs
```diff
@@ -32,6 +61,27 @@ impl fmt::Display for Emoji {
...
+/// Avatar usage.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Usage {
+    /// A user avatar.
+    User,
+    /// An org avatar.
+    Org,
+}
+
+impl std::str::FromStr for Usage {
```
xla: I'd drop this and make the it the callers responsibility to use the appropriate enum variant.

proxy/src/error.rs
```diff
@@ -55,6 +55,8 @@ pub enum Error {
     Runtime(DispatchError),
     /// Errors from handling time.
     Time(SystemTimeError),
+    /// Invalid query.
+    InvalidQuery(String),
```
xla: What is this error for? The comment doesn't really give context either. :D

cloudhead:
Ok, I'm going to leave the GQL integration for later by using a default usage.

xla:
> Ok, I'm going to leave the GQL integration for later by using a default usage.

You can just pass the `User` variant of `Usage` hardcoded for now as the endpoint is not yet used for orgs anyway.

Issue #220, opened 2020-03-12
Evaluate feasibility of a REST API


garbados:
Pursuant to https://github.com/radicle-dev/radicle-upstream/issues/165, this issue explores the feasibility of a REST API by way of designing one. As an example:

## Endpoints

Consider a set of HTTP endpoints scoped under the prefix `/api/v1` such that the path to the endpoint `identities` is actually `/api/v1/identities`. Object types referenced in this outline, such as "identities", refer to entities that are already described within the proxy as Rust structs.

- GET identities
  - list local identities
- GET identity/{id}
  - retrieve info about a single identity
- GET local_branches/{...path}
  - retrieve a list of branches for a locally-stored repository
- GET org/{id}
  - retrieve information about an organization
- GET org/{id}/projects
  - list this org's projects
- GET org/{id}/users
  - list registered users associated with this org
- GET orgs
  - list orgs, scoped using query parameters
- GET project/{id}
  - retrieve the project as of its latest commit
- GET project/{id}/blob/{...path}
  - retrieve a blob from the main branch (ex: master)'s latest commit
- GET project/{id}/branches
  - list branches for the project
- GET project/{id}/commit/{sha1}
  - retrieve the project as of the given commit
- GET project/{id}/commit/{sha1}/blob/{...path}
  - retrieve a blob as of a given commit
- GET project/{id}/commits
  - list this project's commits, scoped using query parameters
- GET project/{id}/tags
  - list this project's tags, scoped using query parameters
- GET project/{id}/tree/{revision}
  - retrieve the project as of the given revspec
- GET project/{id}/tree/{revision}/blob/{...path}
  - retrieve a blob as of the given revspec
- GET project/{id}/contributors
  - list users (or identities?) who have contributed to this repo
- GET projects
  - list projects, scoped using query parameters
- GET transaction/{id}
  - get info about a specific transaction, ex: did it succeed?
- GET transactions
  - list transactions, scoped using query parameters
- GET user/{id}
  - get info about a user (that is, an identity with a registered handle)
- GET users
  - list users, scoped using query parameters
- POST identity
  - create a new identity / keypair
- POST transaction
  - submit a transaction (ex: register user, project, org, etc)

## Entities

Endpoints deal with common, standardized entities, including:

- Org
- User
- Identity
- Project
- Commit
- Branch
- Tag
- Blob
- Transaction

The nature of these entities is already contained within the proxy. For example, here is the current characterization of an Identity:

```
[...]

PR #226, opened 2020-03-17
feat(proxy): demo rest api

garbados:
This PR follows up on https://github.com/radicle-dev/radicle-upstream/issues/220 by implementing a demo REST API. This implementation is incomplete, and currently only intended to demonstrate the flow of writing routes.

This PR adds this file organization for its REST API:

```
src/
  rest/
    <entity>.rs
  rest.rs
```

Endpoints are divided by entity (ex: Org, User, Commit) and tied together in `rest.rs`, which `main.rs` invokes to start the server.

Although the proxy currently uses the [Warp](https://docs.rs/warp/0.2.2/warp/) HTTP framework, Warp is ill-suited to writing a large API for debugging reasons. It is very difficult to determine what filter has or hasn't caught, and what routes are actually exposed at any given time. Instead I chose to use [Rocket](https://rocket.rs/), which inherits the norms of frameworks like [Rails](https://rubyonrails.org/) and [Flask](https://flask.palletsprojects.com/en/1.1.x/).

This API demonstrates writing simple routes and simple tests, but a production API must also manage Librad and Registry clients. For that, Rocket exposes a [State](https://rocket.rs/v0.4/guide/state/#state) system so that routes can safely access shared objects.

This PR is incomplete but, should we decide to pursue a REST API over GraphQL, it will be expanded to encompass operations and queries with all Radicle entities.

### DIFF
--- a/proxy/Cargo.lock
+++ b/proxy/Cargo.lock
@@ -917,6 +917,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "fsevent"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bitflags 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "fsevent-sys 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "fsevent-sys"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.66 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "fuchsia-cprng"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


garbados:
Added examples for project endpoints, particularly [project registration](https://github.com/radicle-dev/radicle-upstream/pull/226/files#diff-ab4e98af651b375e77e56b1a20fc3748R40-R55), which show accepting JSON requests and generating JSON responses.

garbados:
There will be several steps to integrating the REST API into the application. Here is a breakdown and an estimate:

- [ ] API implementation: Of course, the API must be complete and available at a known URL, such as `localhost:12345`.
- [ ] API documentation: Endpoints, their parameters, and their responses should be fully documented. I plan to deliver something in the approximate shape of the [CouchDB API docs](https://docs.couchdb.org/en/stable/api/index.html).
- [ ] JS Tooling: Currently the app can use GraphQL tooling to issue queries and process responses. Alternative tooling must be written to consume a REST API. I plan to release examples and documentation for this alternative tooling using [window.fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch), which is built into modern browsers and won't require external dependencies.

garbados:
I expect to be able to deliver a working API by EOW (end of Friday, March 20). Documentation will follow:

- Documentation by EOD Monday, March 23
- Tooling by EOD Wednesday, March 25

At that point I will pass off integration work to the frontend team, where I will be available for questions. I'm well versed in JS and am happy to dive into code as necessary. Otherwise I will commit time to expanding the routes we support in order to meet our key objectives.

To request a route, or in general if you want something from the API that it doesn't provide, open an issue and describe the sought behavior. I'll reply with some hypothetical API documentation, we'll work out a consensus about it, and then I'll implement it.

xla:
Closing in favor of #250 - thanks @garbados for the great work and giving us a starting point for our explorations.

PR #227, opened 2020-03-17
Integrate tx fee

NunoAlexandre:
- Bump radicle-registry-client to latest master
- Update OrgId and ProjectName ::from_string to ::try_from
- Add 'fee' field to transactions

**Note**: Submitting transactions with insufficient funds to pay the indicated fee will result in failed transactions. 

@rudolfs I think I have covered the changes as far as I should go. Note that I have also integrated the latest changes by @xla on `OrgId` and `ProjectName`, now built with `try_from` instead of `from_string`.

Edit: Still have a lot of things to fix here that `cargo build` wasn't reporting.

### DIFF
--- a/.buildkite/run.sh
+++ b/.buildkite/run.sh
@@ -56,10 +56,11 @@ else
 fi
 
 echo "--- Updating submodules"
-time git submodule update --init --recursive
+time git submodule update --remote --init --recursive
 time git submodule foreach "git fetch --all"
 time git submodule foreach "git checkout -B dev -t origin/dev"
 time git submodule foreach "git checkout master"
+time git submodule foreach "git pull origin master"
 
 echo "--- Set custom git config"
 cp .buildkite/.gitconfig "$HOME/"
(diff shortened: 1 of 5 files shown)


NunoAlexandre:
@xla I have left a TODO for the fake hardcoded fee. On top of that, I have fixed the registry tests that were failing since the involved org didn't have enough funds to pay for the tx fee. That same issue is still occurring in the `graphql_mutation>register_project` test, where I am not sure how to proceed.

xla:
> Edit: Something else that needs work is the error translation for failing OrgId and ProjectName, where now, here in the Upstream, the errors are outdated and only about the size of those inputs:

Fixed in 9e20351 - having to use a specific account for our tests to succeed is quite brittle as the integration is on a string literal, is there a more effective strategy to setup an account with sufficient funds for our fee dependent transactions?

> Want me to address that here or will you?

That's be dope, the existing errors don't cut it anymore, would be nice to have one validation error per id type and embbed the original description so we can bubbe it up to the UI.

xla:
> That's be dope, the existing errors don't cut it anymore, would be nice to have one validation error per id type and embbed the original description so we can bubbe it up to the UI.

More sensible is to do these changes in a follow-up PR, captured it in #233.

PR #235, opened 2020-03-18
Implement transaction state icon

MeBrei:
This prepares showing the transaction states in the bottom right corner.

The component has the props:
- `state` pending (default), success, error
- `progress` a number between 0 and 6 with 0 being the default. This prop only has an effect in combination with state pending

The blink animation is currently set to 2 seconds and fades equally from 0.35 opacity to 1 and back.
tx-state-animation

### DIFF
--- a/ui/Screen/DesignSystemGuide.svelte
+++ b/ui/Screen/DesignSystemGuide.svelte
@@ -212,6 +212,14 @@
     <Icon.SearchSmall />
     <Icon.Source />
     <Icon.Spinner />
+    <Icon.TxState />
+    <Icon.TxState progress={1} />
+    <Icon.TxState progress={2} />
+    <Icon.TxState progress={3} />
+    <Icon.TxState progress={4} />
+    <Icon.TxState progress={5} />
+    <Icon.TxState state="success" />
+    <Icon.TxState state="error" />
   </Section>
 
   <Section
--- a/ui/Screen/DesignSystemGuide.svelte
+++ b/ui/Screen/DesignSystemGuide.svelte
@@ -218,15 +218,14 @@
     subTitle="Icons at 24px and 32px width and height with animations and
     reactive coloring.">
     <Icon.Spinner />
-    <Icon.TxState />
-    <Icon.TxState progress={1} />
-    <Icon.TxState progress={2} />
-    <Icon.TxState progress={3} />
-    <Icon.TxState progress={4} />
-    <Icon.TxState progress={5} />
-    <Icon.TxState state="success" />
-    <Icon.TxState state="error" />
-    <Icon.TxState state="error" progress={3} />
+    <Icon.TxState state="positive" />
+    <Icon.TxState progress={0} />
+    <Icon.TxState progress={10} />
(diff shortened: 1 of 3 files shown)


juliendonck:
I feel it's a bit too slow, I'd try 1 sec or 1,5 sec interval

MeBrei:
Changed the animation speed to 1.5s

ui/Screen/DesignSystemGuide.svelte
```diff
@@ -212,6 +212,14 @@
...
     <Icon.Source />
     <Icon.Spinner />
+    <Icon.TxState />
+    <Icon.TxState progress={1} />
+    <Icon.TxState progress={2} />
+    <Icon.TxState progress={3} />
+    <Icon.TxState progress={4} />
+    <Icon.TxState progress={5} />
+    <Icon.TxState state="success" />
+    <Icon.TxState state="error" />
```
rudolfs: Maybe it would be nice to distinguish the icons a little by putting them into their own section.
The spinner icon could also be separate.

MeBrei:
There has been a change of design for the icons.
Screenshot 2020-03-19 at 15 44 22

MeBrei:
> * does it even make sense to have a spinner or show progress percentage in a "negative" tx state?

For the tx summary, there will be mixed states. For example if there is one tx of state pending with 60% progress an one tx failed, then the summary should show a negative icon, but with the progress of the pending tx.

> * would it make sense to move this component to `Component`, as it is a complex thing consisting of multiple icon primitives, maybe call it TxProgressSpinner or something like that?

I agree that this icon is more complex than the others, but `Icon` is where I would be looking for it.

> * updates might happen without a page reload, we should make sure the component is reactive to prop changes

Thanks for testing that! I will address it.

juliendonck:
> Will the other variants be addressed in another PR?

This is how I would expect the smaller size to look like. Updated it in the designs.
Screenshot 2020-03-24 at 09 50 08

juliendonck:
> From a UX standpoint I also find the animation to be a bit subtle, but @juliendonck, should have a look at how it looks like in the app.

This was about the previous blinking animation, no? If so, the new one where the thingy turns around is addresses this imo. Let me know what you think

juliendonck:
@MeBrei how can I see the spinners in action? I'm in the branch, but can only see the `shift-t` flow

MeBrei:
> @MeBrei how can I see the spinners in action? I'm in the branch, but can only see the `shift-t` flow

@juliendonck you can see the transaction icons on the DesignSystemGuide `shift-d`

juliendonck:
Oh snap, missed it by 3 minutes. I was wondering if it's possible to align the `!` better. It's a bit too high
Screenshot 2020-03-24 at 10 22 53

juliendonck:
I can probably do a little edit myself if you point me in the right direction, @MeBrei

MeBrei:
@juliendonck You would have to replace the path at https://github.com/radicle-dev/radicle-upstream/pull/235/files#diff-76fb94d46fcded1f8c226f0717dd9563R96 with the adjusted one.

PR #237, opened 2020-03-19
feat(proxy): convert registry validation errors correctly

### LINKED ISSUE #233: Convert Registry validation errors
The reason that user input for top level entities (orgs and users) and projects is invalid is more nuanced that just the length since recent additions in radicle-dev/radicle-registry#246, radicle-dev/radicle-registry#247 and radicle-dev/radicle-registry#249. We get back more context which we should propagate up to the user. Addressed leftovers from https://github.com/radicle-dev/radicle-upstream/pull/227#issuecomment-600378189.

xla:
Catches up with the changes in registry inputs and how they are validated.

Closes #233

### DIFF
--- a/proxy/src/registry.rs
+++ b/proxy/src/registry.rs
@@ -104,8 +104,7 @@ impl Registry {
         fee: Balance,
     ) -> Result<Transaction, error::Error> {
         // Verify that inputs are valid.
-        let org_id =
-            OrgId::try_from(org_id.clone()).map_err(|_| error::ProjectValidation::OrgTooLong)?;
+        let org_id = OrgId::try_from(org_id.clone())?;
 
         // Prepare and submit project registration transaction.
         let register_message = message::RegisterOrg {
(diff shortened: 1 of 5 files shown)


proxy/src/registry.rs
```diff
@@ -177,10 +175,8 @@ impl Registry {
         fee: Balance,
     ) -> Result<Transaction, error::Error> {
         // Verify that inputs are valid.
-        let project_name = ProjectName::try_from(name.clone())
-            .map_err(|_| error::ProjectValidation::NameTooLong)?;
-        let org_id =
-            OrgId::try_from(org_id.clone()).map_err(|_| error::ProjectValidation::OrgTooLong)?;
+        let project_name = ProjectName::try_from(name.clone())?;
+        let org_id = OrgId::try_from(org_id.clone())?;
```
NunoAlexandre: Don't you need to map_err to go from the registry error to the proxy upstream one?

xla: We satisfy `try` semantics in our error module where we have the infrastructure to lift protocol errors into upstream domain:

https://github.com/radicle-dev/radicle-upstream/blob/1755dee156c15f85ea183d996a27841900ee5337/proxy/src/error.rs#L140-L144

NunoAlexandre: Ah, sweet. I figured that had to be done explicitly here.

PR #238, opened 2020-03-19
feat(proxy): wire up Registry user registration

### LINKED ISSUE #185: Integrate Registry user registration
The user registration flow as outlined in #202 requires the integration with the user registration provided by radicle-dev/radicle-registry.

- [x] integrate registry client user registration in `regisry.rs`
- [x] integrate user registration in graphql schema

xla:
We replace the mock used so far for user registration with an actual integration with the Registry. Follow up to radicle-dev/radicle-registry#249.

Closes #185

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -74,3 +74,19 @@ Cypress.Commands.add(
     });
   }
 );
+
+Cypress.Commands.add("registerUser", (handle = "nope", id = "123abcd.git") => {
+  client.mutate({
+    variables: {
+      handle: handle,
+      id: id
+    },
+    mutation: gql`
+      mutation RegisterUser($handle: String, $id: String) {
+        registerUser(handle: $handle, id: $id) {
+          handle
+        }
+      }
+    `
+  });
+});
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -82,9 +82,14 @@ Cypress.Commands.add("registerUser", (handle = "nope", id = "123abcd.git") => {
       id: id
     },
     mutation: gql`
-      mutation RegisterUser($handle: String, $id: String) {
+      mutation RegisterUser($handle: ID!, $id: ID!) {
         registerUser(handle: $handle, id: $id) {
-          handle
+          messages {
+            ... on UserRegistration {
+              handle
+              id
+            }
+          }
         }
       }
     `
--- a/proxy/src/graphql/schema.rs
+++ b/proxy/src/graphql/schema.rs
@@ -294,12 +294,14 @@ impl Query {
         }))
     }
 
-    fn user(_ctx: &Context, handle: juniper::ID) -> Result<Option<juniper::ID>, error::Error> {
-        if handle == juniper::ID::new("cloudhead") {
-            Ok(None)
-        } else {
-            Ok(Some(juniper::ID::new("1234")))
-        }
+    fn user(ctx: &Context, handle: juniper::ID) -> Result<Option<juniper::ID>, error::Error> {
+        Ok(futures::executor::block_on(
+            ctx.registry
+                .read()
+                .expect("unable to acquire read lock")
+                .get_user(handle.to_string()),
+        )?
+        .map(juniper::ID::new))
     }
 }
 
--- a/proxy/src/graphql/schema.rs
+++ b/proxy/src/graphql/schema.rs
@@ -129,13 +129,15 @@ impl Mutation {
         id: juniper::ID,
     ) -> Result<registry::Transaction, error::Error> {
         // TODO(xla): Get keypair from persistent storage.
-        let fake_pair = ed25519::Pair::from_legacy_string("//Robot", None);
+        let fake_pair = ed25519::Pair::from_legacy_string("//Alice", None);
+        // TODO(xla): Use real fee define dby the user.
+        let fee = 100;
 
         futures::executor::block_on(
             ctx.registry
                 .read()
                 .expect("unable to acquire read lock")
-                .register_user(&fake_pair, handle.to_string(), id.to_string()),
+                .register_user(&fake_pair, handle.to_string(), id.to_string(), fee),
         )
     }
 }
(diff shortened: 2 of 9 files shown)


xla:
@MeBrei Good catch, wired it up as well as part of this PR.

proxy/src/graphql/schema.rs
```diff
@@ -129,7 +129,18 @@ impl Mutation {
         id: juniper::ID,
     ) -> Result<registry::Transaction, error::Error> {
         // TODO(xla): Get keypair from persistent storage.
-        let fake_pair = ed25519::Pair::from_legacy_string("//Alice", None);
+        let fake_pair =
+            ed25519::Pair::from_legacy_string(&format!("//{}", handle.to_string()), None);
+
+        use radicle_registry_client::CryptoPair;
+        // Give new account some dough so we can perform transactions.
```
NunoAlexandre:
```suggestion
        // The new account needs funds to perform transactions.
```

"dough" sounds cool but such language is unhelpful here, imo. Also, I suggest explaining _why_ we are doing it, the _what_ the code tells already.

proxy/src/graphql/schema.rs
```diff
@@ -295,13 +306,17 @@ impl Query {
...
 
     fn user(ctx: &Context, handle: juniper::ID) -> Result<Option<juniper::ID>, error::Error> {
-        Ok(futures::executor::block_on(
+        let maybe_handle = futures::executor::block_on(
             ctx.registry
                 .read()
                 .expect("unable to acquire read lock")
                 .get_user(handle.to_string()),
-        )?
-        .map(juniper::ID::new))
```
NunoAlexandre: Why the change from `map(juniper::ID::new))` to `match`? It adds another few lines of code with no added value.

proxy/src/graphql/schema.rs
```diff
@@ -366,6 +381,24 @@ impl ControlMutation {
...
     }
+
+    fn register_user(
+        ctx: &Context,
+        handle: juniper::ID,
+        id: juniper::ID,
+    ) -> Result<registry::Transaction, error::Error> {
+        // TODO(xla): Get keypair from persistent storage.
+        let fake_pair = ed25519::Pair::from_legacy_string("//Alice", None);
+        // TODO(xla): Use real fee define dby the user.
```
NunoAlexandre:
```suggestion
        // TODO(xla): Use real fee defined by the user.
```

proxy/src/graphql/schema.rs
```diff
@@ -366,6 +381,24 @@ impl ControlMutation {
...
         Ok(true)
     }
+
+    fn register_user(
+        ctx: &Context,
+        handle: juniper::ID,
+        id: juniper::ID,
+    ) -> Result<registry::Transaction, error::Error> {
+        // TODO(xla): Get keypair from persistent storage.
+        let fake_pair = ed25519::Pair::from_legacy_string("//Alice", None);
```
NunoAlexandre: You could make this function receive the seed for the ed25519 pair instead of hardcoding it to Alice.

proxy/src/registry.rs
```diff
@@ -247,6 +247,21 @@ impl Registry {
         Ok(self.client.get_user(user_id).await?.map(|_user| handle))
     }
 
+    pub async fn prepay_account(
```
NunoAlexandre: This is essentially a transfer, but now you can't specify the donator.

proxy/src/graphql/schema.rs
```diff
@@ -366,6 +381,24 @@ impl ControlMutation {
...
+        // TODO(xla): Use real fee define dby the user.
+        let fee = 100;
+
+        futures::executor::block_on(
+            ctx.registry
+                .read()
```
NunoAlexandre: I'm not fully sure as to what this is but should it be `write` since it's registering a user and not, say, getting/reading one?

ui/Screen/UserRegistration.svelte
```diff
@@ -11,7 +11,7 @@
 
   let step = 1;
   // TODO(merle): Get actual user profile (id, name, imageUrl, avatarFallback)
-  let handle = "cloudhead";
+  let handle = "";
```
MeBrei: Why did you remove the handle? This is the identity name aka default handle that automatically fills the user handle registration input field. One of the tests is failing due to this.

xla: How is this related? The tests use `nope` as far as I went through the. It is a weird default to have when opening the screen and was in the way when testing it in the running app.

MeBrei: There are two different tests. `nope` is used to test for a handle that is already taken ("second test"). The first test is about navigating through the modal. The modal should be pre-filled with your chosen identity name, which will come from the storage in the future, for now it is hardcoded.

cypress/support/commands.js
```diff
@@ -74,3 +74,19 @@ Cypress.Commands.add(
...
+      handle: handle,
+      id: id
+    },
+    mutation: gql`
+      mutation RegisterUser($handle: String, $id: String) {
+        registerUser(handle: $handle, id: $id) {
+          handle
+        }
+      }
+    `
```
MeBrei: The query is not correct here. handle and id should be `ID` and required and just `handle` can't be returned as such. See the query in `UserRegistration.svelte`, copying it over here makes the test green.

proxy/src/registry.rs
xla: This method is acting like a faucet, you can bootstrap a new account from the main available pool `Alice`. At this point there is no reason to have more flexibility until we actually wanna test transfers in the app. Also see this as temporary aid .

proxy/src/graphql/schema.rs
xla: It's a temporary measure and I allowed myself some fun with it. Will add a comment to remove the method alltogether.

proxy/src/graphql/schema.rs
xla: Then we also need to ensure that that account has funds, will leave it until the needs arises.

proxy/src/graphql/schema.rs
xla: These are the calls to acquire the `RWLock` and as we don't mutate the field itself we  don't need to acquire a `write` lock.

PR #228, opened 2020-03-17
fix: bump all UI dependencies

rudolfs:
To address security alert.

Screenshot 2020-03-17 at 10 27 26

PR #239, opened 2020-03-19
chore(ui): update major depdencies

xla:
This has been raised by the Github autmoation and all dependencies have been bumped with `yarn upgrade`.

### DIFF
--- a/.buildkite/Dockerfile
+++ b/.buildkite/Dockerfile
@@ -1,13 +1,18 @@
-FROM gcr.io/opensourcecoin/radicle-registry/ci-base
+FROM gcr.io/opensourcecoin/radicle-registry/ci-base@sha256:18d8cf1e39f4f42d871a2a9d3bcd8641bfda343a1fc7b9fbca4f8aea422bae6f
 
 RUN apt-get -y update
 
 # install cypress deps
 RUN apt-get -y install autoconf git nettle-dev m4 gnupg xvfb libgtk-3-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2
 
+# install npm
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN apt-get -y update
+RUN apt-get -y install build-essential nodejs
+
 # install yarn
 RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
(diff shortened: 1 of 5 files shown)


rudolfs:
Latest Svelte seems to break prettier, same as in: https://github.com/radicle-dev/radicle-upstream/pull/228#issuecomment-600017621.

xla:
@rudolfs The reason for the prettier breakage is that the plugin wasn't up-to-date with the introduction of destructing support in await statements, which landed in svelte `3.20.0`. Addressed in UnwrittenFun/prettier-plugin-svelte#83.

.buildkite/Dockerfile
```diff
@@ -1,13 +1,18 @@
-FROM gcr.io/opensourcecoin/radicle-registry/ci-base
+FROM gcr.io/opensourcecoin/radicle-registry/ci-base@sha256:18d8cf1e39f4f42d871a2a9d3bcd8641bfda343a1fc7b9fbca4f8aea422bae6f
 
 RUN apt-get -y update
 
 # install cypress deps
 RUN apt-get -y install autoconf git nettle-dev m4 gnupg xvfb libgtk-3-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2
 
+# install npm
```
rudolfs: Why do we need this now?

.buildkite/Dockerfile
```diff
@@ -1,13 +1,18 @@
-FROM gcr.io/opensourcecoin/radicle-registry/ci-base
+FROM gcr.io/opensourcecoin/radicle-registry/ci-base@sha256:18d8cf1e39f4f42d871a2a9d3bcd8641bfda343a1fc7b9fbca4f8aea422bae6f
```
rudolfs: Where does the hash come from?

.buildkite/Dockerfile
xla: I couldn't get to the bottom of it, but without our `yarn install` step fails.

.buildkite/Dockerfile
xla: It's the latest image the registry provides.

rudolfs:
> And the PR is not addressing the root issue yet as cypress has a transient dependency on an old version of minimist.

Should we merge anyway?

Issue #225, opened 2020-03-16
General transaction handling


xla:
Transaction handling is core to the interactions and feedbacks when interacting with the Registry. To enable functionality vital for the deliveyor [Identities I]() and [Orgs I](), we ought to solve transaction handling, presentation and the handling of non-trivial unlocking of dependent actions in the UI. As it is vital for further development we should make it part of the **Identities I** deliverable.

*This issue superseeds #89, #212 and #214*

Summarising a meeting of the Application team to come to a common understanding of how to address the need for general transaction infrastructure, we assume that complex flows in the app depend on states of certain dependent transactions to advance. These state changes allow for new/added **capabilities** or **permissions**, which control what actions a user can perform.

### Requirementes

* get transactions by id
* get list of known transactions
* global view on transactions
* cache relevant transactions
* configurable confirmation unit (amount of blocks for a bar to go full in the transaction progress indicator)
* derive capabilities and permissions from relevant transaction state

### Constraints

* get realtime updates for transaction state changes and their respective capabilities/permissions
* exisiting entity responses should convey their capabilities/permissions

### API

A single endpoint is proposed to full-fill all requriements, including realtime update, assuming we go with the most naive version of polling in the beginning, so that we don't require extra infrastructure/complexity (e.g.: GraphQL subscriptions, websockets, SSE).

``` graphql
query($ids: [ID!]) {
  listTransactions(ids: $ids): [Transaction!]!
}

type Transaction {
  costs: Cost!
  id: ID!
  messages: [Message!]!
  receiver: Receiver!
  sender: Sender!
  state: State!
  submission_time: Timestamp!
}

type Cost {
  deposit: Int!
  fee: Int!
}

union Receiver = Org | User
union Sender = Org | User

union State = Unconfirmed | Confirmed | Failed

type Unconfirmed {}

type Confirmed {
  block: Block
}

type Failed {
  block: Block
  error: # Come up with error shape
}

type Block {
  hash: ID!
  height: Int
}
```

### Open Questions

* [x] Which model (push/poll) is more feasible for our realtime needs/
* [ ] How does the error state of a transaction look?
* [x] Where does te confirmation config for units live: Upstream or Registry?
* [ ] What is the structure of the capabilities/permission responses?

*There are dependent conversations/issues that need to be fleshed out, i.e.: state/storage/data-access and routing/navigation.*

### UI
- [ ] once in place in the proxy, handle status/progress/error & transaction costs in the ui
[...]

Issue #178, opened 2020-02-24
Set up unit tests for lib code in UI


rudolfs:
Some of our business-logic modules might lend themselves to unit testing.
Set up Jest, so we can write unit tests and run them locally as well as automatically on CI.

https://jestjs.io/

xla:
I recently was exposde [ava](https://github.com/avajs/ava) which seemed low touch and approachable.

sarahscott:
Closed by #413

PR #242, opened 2020-03-20
Implement transaction center

MeBrei:
As part of #202 this component will show transactions and their states in an expandable list.

Depends on #235 

ToDos
- [ ] format txs with actual data
  - [x] tx message - needs fixing of `IntrospectionFragmentMatcher` error to retrieve the tx kind
  - [ ] tx status/progress - needs block heights to compare to the threshold (#225)
  - [ ] error message - to display the cause of a failed transaction (#225)
  - [ ] timestamp - we need timestamps that reflect settle and error events (#225)
- [x] add box shadow & error coloring
- [ ] add (or write issue on) animation
    - [ ] slide up/down cards
    - [ ] on first appearance of the tx center, slide it into view
- [x] Transaction detail view
  - [x] add correct icon design
  - [x] format timestamp
  - [x] retrieve user/project info
- [ ] move transaction query into a central storage

The TxAccordion component can be viewed in the `DesignSystemGuide` `shift-d` with mock data. The Tx center can be viewed by submitting a transaction (eg register your user `shift-t` - note that you have to create an identity first `shift-i`) and opening any view that uses the sidebar layout, as currently transactions are only queried on sidebar layout load.


MeBrei:
### `IntrospectionFragmentMatcher` issue

The transaction `messages` use fragments with different fields. This causes an error with our current apollo implementation (even if we query the same fields for each fragment).
Screenshot 2020-03-26 at 13 52 35
Screenshot 2020-03-26 at 13 52 49

The link in the error does not work, but I found [another one](https://www.apollographql.com/docs/react/data/fragments/#fragments-on-unions-and-interfaces).

MeBrei:
The 2 todos dependent on #225 can't be addressed right now.

I will create a followup issue for the animation todos as they also depend on how we poll the transactions continuously in the future.

For now only the tx message is using the actual data. The progress and status are hardcoded.

ui/DesignSystem/Component/TxAccordion.svelte
```diff
@@ -0,0 +1,171 @@
...
+  .hidden {
+    display: none;
+  }
+</style>
+
+<div class="pipeline" class:negative={iconState === 'negative'} {style}>
+  <div class="cards" class:hidden>
+    {#each transactions as tx}
+      <!-- TODO: Link card to tx detail view -->
+      <div class="card">
```
xla: This nesting is quite expansive, can we refactor this part out into a distinct card component?

MeBrei: In which way do you mean expensive?

I like the idea of having all the logic and styling of this component in one file. For now we also only need the `card` in this component so that I would tend to keeping it here. 
@rudolfs What's your opinion on this?

xla: Not expensive, wrote expansive, as in the nesting of the markup here is quite deep. It's not extraction for reusability but for readability. If there is no strong concern shared it's fine to leave it as is, this was the only place that gave me a stumble while reviewing.

rudolfs: > @rudolfs What's your opinion on this?

I'm fine with it either way. I wish there was a better way of handling `<Flex>` without needing to nest it twice, but I don't know of a better solution at the moment either.

rudolfs:
> add (or write issue on) animation

Have a look at [svelte transitions](https://svelte.dev/docs#transition_fn). We already use them in some parts, like:
```
ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
ui/DesignSystem/Component/NotificationFaucet.svelte
ui/DesignSystem/Component/RadioOption.svelte
```

MeBrei:
I will add the transaction detail view in this PR as well. (Will dismiss the approvals)

ui/DesignSystem/Component/TxAccordion.svelte
```diff
@@ -0,0 +1,171 @@
+<script>
+  import { Flex, Icon, Text } from "../Primitive";
+  // TODO: Make component responsive to transactions change
```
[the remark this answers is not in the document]
MeBrei: I will create a follow up issue to move the transactions into a central store. I will then also address this todo

MeBrei:
In order to not let this PR grow too big, I moved some of the outstanding tasks outside of this PR and noted them in #202

ui/DesignSystem/Component/Transaction/Accordion.svelte
```diff
@@ -0,0 +1,177 @@
...
+  <div class="cards" class:hidden>
+    {#each transactions as tx}
+      <div
+        class="card"
+        on:click={() => {
+          push(path.transactions(tx.id));
+        }}>
+        <Flex>
+          <div slot="left">
+            <Flex align="left">
```
rudolfs, line 134: You could avoid wrapping the content within another `<Flex align="left">
 ` by setting the parent `div`'s inline style to: `style="display: flex"`.

ui/DesignSystem/Component/Transaction/Accordion.svelte
```diff
@@ -0,0 +1,177 @@
...
+              style="vertical-align: middle; margin-right: 16px;" />
+          </div>
+        </Flex>
+      </div>
+    {/each}
+  </div>
+  <div class="summary" on:click={hideCards}>
+    <Flex>
+      <div slot="left">
+        <Flex align="left">
```
rudolfs, line 160: You could avoid wrapping the content within another <Flex align="left"> by setting the parent div's inline style to: style="display: flex".

ui/Screen/DesignSystemGuide.svelte
```diff
@@ -636,6 +674,41 @@
     </Swatch>
   </Section>
 
+  <Section
```
rudolfs, line 677: So far we've grouped all similar components under one section (and use `<Section>` tags to divide components into subgroups). Should we perhaps also do it for transactions for consistency's sake?

ui/Screen/TransactionDetails.svelte
```diff
@@ -0,0 +1,133 @@
...
+
+  const formatMessage = {
+    USER_REGISTRATION: "User registration",
+    PROJECT_REGISTRATION: "Project registration"
+  };
+
+  const formatSubject = msg => {
+    return {
+      name:
+        msg.kind === "USER_REGISTRATION"
```
rudolfs, line 71: This should also be a constant that we import.

ui/Screen/TransactionDetails.svelte
```diff
@@ -0,0 +1,133 @@
...
+  const transactions = query(client, {
+    query: GET_TRANSACTIONS,
+    variables: {
+      ids: [params.id]
+    },
+    fetchPolicy: "no-cache"
+  });
+
+  const formatMessage = {
+    USER_REGISTRATION: "User registration",
```
rudolfs, line 64: We should keep types coming from the proxy in one place.

So far we've kept them as constants (have a look how BLOB and TREE are used in components):
https://github.com/radicle-dev/radicle-upstream/blob/master/native/types.js

ui/Screen/DesignSystemGuide.svelte
MeBrei: It is grouped under the headline `Components`. Do you mean to have an additional `Transaction` group outside of `Components`?

MeBrei:
> Additional thoughts:
> 
> * it seems to me that the `<Transaction>` component is trying to do too many things, maybe we can do a little pairing session to check if it could be refactored?

I agree, but leave it as is for now. When we implement Orgs and integrate wallets, we will have to touch the component again and it will get more complex. That will be a good time to refactor it.

> * this would be a good place to start doing unit or snapshot tests for single components, see: #178

I'd prefer to try to wrap up this PR. I could add a comment in that issue that transaction component tests could be part of that issue.

> * the nested ternary operators are a bit hard to parse but the only alternative which preserves reactivity would be using `&&` and `||`, which is not any better: [`IdentityAvatar.svelte`](https://github.com/radicle-dev/radicle-upstream/blob/master/ui/DesignSystem/Component/IdentityAvatar.svelte#L14:L17)

I agree that they are a bit clumsy right now. As a follow up to this PR I want to work on centralizing transaction handling and working on the reactivity of the components. I expect to have to touch many of those places then and will try to rework them more readable.

ui/Screen/DesignSystemGuide.svelte
rudolfs: Yeah, something like this:
```
Components
  Avatars
  ...
  Transactions
    <Swatch><TransactionRow></Swatch>
    <Swatch><Accordion></Swatch>
    <Swatch><TransactionStatusbar></Swatch>
```

PR #243, opened 2020-03-21
chore(proxy): address clippy errors

xla:
Please the clippy.

### DIFF
--- a/proxy/src/avatar.rs
+++ b/proxy/src/avatar.rs
@@ -1,4 +1,11 @@
-#![allow(clippy::as_conversions, clippy::float_arithmetic)]
+#![allow(
+    clippy::as_conversions,
+    clippy::cast_possible_truncation,
+    clippy::cast_precision_loss,
+    clippy::cast_sign_loss,
+    clippy::float_arithmetic,
+    clippy::integer_arithmetic
+)]
 
 //! Org and user avatar generation.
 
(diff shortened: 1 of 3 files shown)


cloudhead:
Yeah I mean, apart from the use of `saturating_add` (TIL), I think the code is overall worse :man_shrugging:

proxy/src/avatar.rs
```diff
@@ -116,34 +123,35 @@ impl Color {
...
     #[must_use]
-    pub fn lightness(&self) -> f32 {
-        let r = self.r as f32;
-        let g = self.g as f32;
-        let b = self.b as f32;
-        let n = 0xff as f32;
+    pub fn lightness(self) -> f32 {
+        let r = f32::from(self.r);
+        let g = f32::from(self.g);
+        let b = f32::from(self.b);
```
cloudhead: `f32::from` is implemented as `x as f32`, so this is just a more verbose way of saying the same thing.

proxy/src/avatar.rs
```diff
@@ -116,34 +123,35 @@ impl Color {
...
+        let n = 255_f32;
```
cloudhead: Also exactly the same as `0xff as f32` but less clear that we're looking at the max value of a byte. Maybe `u8::MAX as f32` would be even better.

proxy/src/avatar.rs
```diff
@@ -116,34 +123,35 @@ impl Color {
...
 
         // This isn't perceptual lightness, but whatever.
         (r / n + g / n + b / n) / 3.
     }
 
     /// Ligthen a color by an amount between `-1.0` and `1.0`.
+    #[allow(clippy::shadow_reuse)]
```
cloudhead: This lint doesn't seem useful -- shadowing is pretty idiomatic in rust as far as I've seen.

proxy/src/avatar.rs
```diff
@@ -167,10 +176,18 @@ fn generate_emoji(input: &str, usage: Usage) -> Emoji {
             if let Some(s) = EMOJIS.get(ix) {
                 Emoji(s)
             } else {
-                Emoji(EMOJIS_USER[ix - EMOJIS.len()])
+                Emoji(
+                    EMOJIS_USER
+                        .get(ix - EMOJIS.len())
+                        .expect("index of out of range"),
+                )
```
cloudhead: Again here, this is a much more verbose way of saying `foo[index]`. It doesn't give us any additional safety.

proxy/src/avatar.rs
```diff
@@ -209,9 +226,10 @@ fn hash(input: &str) -> u64 {
     hash
 }
 
+#[allow(clippy::float_cmp, clippy::non_ascii_literal)]
 #[cfg(test)]
 mod test {
-    use super::*;
+    use super::{generate_color, generate_emoji, hash, Avatar, Color, Emoji, Usage};
```
cloudhead: Since we're only importing from the current file, it seems redundant to restate everything here, and just adds additional maintenance.

proxy/src/avatar.rs
xla: Not quite:

>  Rust’s as keyword will perform many kinds of conversions, including silently lossy conversions. Conversion functions such as i32::from will only perform lossless conversions. Using the conversion functions prevents conversions from turning into silent lossy conversions if the types of the input expressions ever change, and make it easier for people reading the code to know that the conversion is lossless.

It might be a bit pedantic in this case, but generally avoiding lossy conversions will make the code safer and correct.

proxy/src/avatar.rs
xla: Promoted to crate `allow` in 38adaab.

proxy/src/avatar.rs
xla: The reason these are explicitly imported is for a reader to easier understand the origin of used function or type. Again it might be the case now that only the same file is imported but this tends to change over time and established good hygiene from the get-go will avoid harder to maintain code down the road.

proxy/src/avatar.rs
xla: Given that we want to ultimately remove all expects from library code this will force us to think about what we do for the case if this is out of range.

proxy/src/avatar.rs
cloudhead: Yeah I meant for `u8` sorry, not for all cases. In the case of `u8 -> f32` the conversion is lossless.

proxy/src/avatar.rs
NunoAlexandre: Yeah, I also like this approach to explicitly import stuff.

proxy/src/avatar.rs
xla: That's clear, it's what the comment above hints at that if the code evolves and the `u8` will become something else having the explicit `from` will surface that there is a transition from a lossless conversion to a lossy one, so it's about safety over time.

proxy/src/avatar.rs
cloudhead: I don't see how that will be possible...

PR #247, opened 2020-03-24
Implement naive tx cache

### LINKED ISSUE #244: Expose transaction cache from the proxy API
Proxy part of #225. Aims to provide a cache of locally observed transactions for the UI to be consumed.

* [x] implement transactions endpoint as per #225 
* [x] primitive caching of transactions

xla:
Closes #244


proxy/tests/graphql_query.rs
```diff
@@ -527,6 +526,34 @@ fn tree_root() {
...
+        let mut vars = Variables::new();
+        vars.insert("ids".into(), InputValue::list(vec![]));
+        let query = "query($ids: [ID!]!) {
+            listTransactions(ids: $ids) {
+                messages {
+                    ... on ProjectRegistration {
+                        projectName,
+                        orgId
+                    }
+                },
```
MeBrei: It is a bit confusing to request a specific content although we are asserting that there will be no transactions

xla: Still working on meaningful tests, this is wip.

proxy/src/registry.rs
```diff
@@ -71,31 +75,50 @@ pub enum TransactionState {
...
+    /// Returns all cached transactions.
+    pub async fn list_transactions(
+        &self,
+        ids: Vec<registry::TxHash>,
+    ) -> Result<Vec<Transaction>, error::Error> {
+        Ok(self
+            .transactions
+            .values()
+            .cloned()
+            .filter(|tx| ids.iter().any(|id| *id == tx.id))
```
cloudhead: You should be able to do `ids.contains(t&x.id)` or something in that vein.

proxy/src/registry.rs
```diff
@@ -118,19 +141,22 @@ impl Registry {
...
-        Ok(Transaction {
+        let tx = Transaction {
             id: register_applied.tx_hash,
             messages: vec![Message::OrgRegistration(org_id)],
             state: TransactionState::Applied(register_applied.block),
             timestamp: SystemTime::now(),
-        })
+        };
+
+        self.transactions.insert(tx.id, tx.clone());
```
cloudhead: Might be a good idea to implement a `self.submit_tx` proxy function that calls `self.client.submit_transaction` and inserts the tx in the cache. That way we don't ever forget to cache stuff.

xla: Unfortunately that is not doable at the moment as `Message` in the registry is a trait with a static method. So we can't really pull the code out that constructs registry transactions.

PR #249, opened 2020-03-25
feat(ui): copyable component (#180)

### LINKED ISSUE #180: Copy-to-clipboard entity identifier component

sarahscott:
Closes #180 

copy


ui/lib/nativeUtils.js
```diff
@@ -0,0 +1,14 @@
+import { IPC_CLIPBOARD_WRITETEXT, IPC_DIALOG_SHOWOPENDIALOG } from "./types.js";
```
xla: The `Utils` suffix seems unnecessary, this can be all `native` library functionality, which inherently should provide utility. :>

ui/lib/types.js
```diff
@@ -1,4 +1,5 @@
 export const TREE = "TREE";
 export const BLOB = "BLOB";
 
-export const MAIN_IPC_CHANNEL = "MAIN_IPC_CHANNEL";
+export const IPC_DIALOG_SHOWOPENDIALOG = "IPC_DIALOG_SHOWOPENDIALOG";
```
xla: As these dictate behaviour in the native/electron wrapper, it should live there. The current dependency direction seems off, given that the native part has to import part of the UI.

ui/DesignSystem/Component/Copyable.svelte
```diff
@@ -0,0 +1,27 @@
...
+
+  const copy = () => copyToClipboard(slotContent.textContent);
+</script>
+
+<style>
+  .container {
+    display: flex;
+    cursor: pointer;
+    color: var(--color-gray);
+  }
```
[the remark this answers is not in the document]
sarahscott: I made this more general, so anything with text can be passed in 😊

native/utils.js
```diff
@@ -0,0 +1,14 @@
+import { IPC_CLIPBOARD_WRITETEXT, IPC_DIALOG_SHOWOPENDIALOG } from "./types.js";
```
sarahscott: @xla does `utils.js` make sense to you as a name for this? I don't think these kinds of helper functions belong in `main`; was also thinking about calling it `lib` because it's a library.

sarahscott: Alternatively there could be a `native/index.js`

xla: I usually find utils/helpers and friends an invitation to become catch all modules for everything that didn't find a proper home. Having it `index` sounds alright, More aptly in his case all these methods are concerned with ipc so that might be an appropriate name for the file.

sarahscott:
I like `ipc`. Simple & to the point

PR #250, opened 2020-03-25
feat(proxy): demo warp based REST API endpoints

xla:
Alternative follow-up to #220, in response to the issues I see with pulling in Rocket as a dependency, which is at the core of the great prior work of @garbados in #226. Also just for brevity so we can compare our options as this approach uses our existing web stack based on [warp](https://github.com/seanmonstar/warp).

To show an end-to-end integration I focused on the endpoint around projects as it forced me to handle CoCo and Registry equally as well as read and write operations. This includes serialisation of request inputs and response bodies, while trying to adhere to our strict separation of core domain and materialisation in one of our API. This approach is comparable to what we tried to achieve with our GraphQL API and the isolation of it in the `graphql` module.

* [x] Boostrap warp based REST stack
* [x] Introduce custom error infrastructure to translate core errors into JSON responses
* [x] Pass state down to filters and handlers
* [x] Implement project filters (create, get, list, register)
* [x] Implement project handlers (create, get, list, register)
* [x] Test end-to-end request chain


proxy/src/http/project.rs
```diff
@@ -0,0 +1,422 @@
...
+        let mut reg = registry.write().await;
+        let tx = reg
+            .register_project(&fake_pair, input.project_name, input.org_id, None, fake_fee)
+            .await?;
+
+        Ok(reply::with_status(reply::json(&tx), StatusCode::CREATED))
+    }
+}
+
+impl Serialize for project::Project {
```
jxs: why is Serialize impl'ed here instead of `crate::project` ?

proxy/src/http/project.rs
```diff
@@ -0,0 +1,422 @@
...
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("Project", 2)?;
+        state.serialize_field("id", &self.id.to_string())?;
+        state.serialize_field("metadata", &self.metadata)?;
+        state.end()
+    }
+}
+
+impl Serialize for registry::Transaction {
```
jxs: same as above

proxy/src/http/project.rs
```diff
@@ -0,0 +1,422 @@
...
+        let mut state = serializer.serialize_struct("Transaction", 4)?;
+        state.serialize_field("id", &self.id.to_string())?;
+        state.serialize_field("messages", &self.messages)?;
+        state.serialize_field("state", &self.state)?;
+        state.serialize_field("timestamp", &self.timestamp)?;
+        state.end()
+    }
+}
+
+impl Serialize for registry::Message {
```
jxs: same as above

proxy/src/http/project.rs
```diff
@@ -0,0 +1,422 @@
...
+                    serializer.serialize_struct_variant("Message", 3, "UserRegistration", 2)?;
+                sv.serialize_field("handle", &handle.to_string())?;
+                sv.serialize_field("id", &id.to_string())?;
+                sv.end()
+            },
+        }
+    }
+}
+
+impl Serialize for registry::TransactionState {
```
jxs: same as above, should Serialize impl's be next to referent structs?

proxy/src/http/project.rs
```diff
@@ -0,0 +1,422 @@
...
+#[derive(Deserialize, Serialize)]
+pub struct MetadataInput {
+    /// Name of the proejct.
+    name: String,
+    /// Long form outline.
+    description: String,
+    /// Configured default branch.
+    default_branch: String,
+    /// Display image of the project.
+    img_url: String,
```
jxs: don't know the context so forgive-me if it doesn't make sense, but hyper has an [Uri field that impl's From<String> ](https://docs.rs/hyper/0.13.1/hyper/struct.Uri.html#impl-TryFrom%3CString%3E) and has many methods that may be useful while dealing with url's

proxy/src/project.rs
```diff
@@ -66,3 +71,13 @@ pub struct Stats {
     /// Amount of unique commiters on the default branch.
     pub contributors: u32,
 }
+
+/// TODO(xla): Add documentation.
+pub async fn get(paths: &librad::paths::Paths, id: &str) -> Result<Project, error::Error> {
+    let meta = coco::get_project_meta(paths, id)?;
+
+    Ok(Project {
+        id: librad::project::ProjectId::from_str(&id.to_string())?,
```
jxs: you don't need to convert to string to then get a reference, you can use id directly: `librad::project::ProjectId::from_str(id)`

garbados:
This looks like a good demo of how to do a REST API in Warp. Thanks @xla , this addresses the issues I ran into trying to do it myself. I've read through the implementation here and I'm prepared to move forward with it for the REST API over the Rocket implementation, echoing your concerns about dependency footprints.

proxy/src/http/error.rs
```diff
@@ -0,0 +1,102 @@
...
+    fn from(err: error::Error) -> Self {
+        reject::custom(err)
+    }
+}
+
+/// Error type to carry context for failed requests.
+#[derive(serde_derive::Serialize)]
+struct Error {
+    /// Human readable message to convery error case.
+    message: &'static str,
```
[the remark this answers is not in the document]
xla: That's some great context, addressed in 87ab5a7.

proxy/src/http/project.rs
```diff
@@ -0,0 +1,422 @@
...
+        .and(warp::post())
+        .and(super::with_paths(paths))
+        .and(warp::body::json())
+        .and_then(handler::create)
+}
+
+/// GET /projects/<id>
+fn get_filter(paths: Paths) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    path!("projects" / String)
+        .and(path::end())
```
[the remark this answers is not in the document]
xla: Missed that, cleans up the filters nicely. Addressed in 02c8c22.

proxy/src/http/project.rs
xla: Didn't know about the `Uri` type in hyper, might be useful down the road. But your comment triggered me to enforce our already in use `Url` type in the core methods as it is parse to one anyway. Good pointer!

xla: See 76b1737.

proxy/src/project.rs
xla: Ha! Daily wtf. Thanks for the catch.

proxy/src/http/project.rs
xla: I'm on the fence, on one side it's nice to use the derive macros. On the other hand there will be slight diversion between the core types and how they are represented over something like a JSON API. And that is really a concern of the API layer for us `http`. Happy to hear arguments in the other direction.

proxy/src/http/project.rs
jxs: ah nice, btw I found https://github.com/radicle-dev/radicle-link/blob/master/librad/src/meta/common.rs#L164
nitpick: rather than having a parse method, you can impl either the `FromStr` or `From<String>` traits for Url, std does it [here for example](https://doc.rust-lang.org/src/core/str/mod.rs.html#4119-4121)

rudolfs:
Is there any way to get some kind of introspection for the API? Or maybe generate some documentation on what the API supports? I will surely miss graphiql if we switch...

xla:
@garbados and @sarahscott felt the same need and they are currently working out a solution for how to explore/understand the API.

rudolfs:
A couple more thoughts:
- it would be also interesting to see what a mutation would look like in code (PUT, PATCH, POST, DELETE).
- can we also do websockets via warp? This could be helpful for notifications and transaction updates.
- i'm not sure i follow the current implementation, are there tests for the get request?

xla:
> it would be also interesting to see what a mutation would look like in code (PUT, PATCH, POST, DELETE).

https://github.com/radicle-dev/radicle-upstream/blob/c754fbf6bfc5d4425b4ec649af903f4dccaa805b/proxy/src/http/project.rs#L25-L32

https://github.com/radicle-dev/radicle-upstream/blob/c754fbf6bfc5d4425b4ec649af903f4dccaa805b/proxy/src/http/project.rs#L77-L103

> can we also do websockets via warp? This could be helpful for notifications and transaction updates.

Yes it has full [websocket](https://docs.rs/warp/0.2.2/warp/filters/ws/index.html) support. So I wonder if we need a bi-directional connection given the proxy will only push to the UI.

> i'm not sure i follow the current implementation, are there tests for the get request?

https://github.com/radicle-dev/radicle-upstream/blob/c754fbf6bfc5d4425b4ec649af903f4dccaa805b/proxy/src/http/project.rs#L319-L351

xla:
> can we also do websockets via warp? This could be helpful for notifications and transaction updates.

Inspired by this comment I drafted a notifications endpoint in 6e21def based on [Server-sent events](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events).

proxy/src/http.rs
```diff
@@ -0,0 +1,54 @@
...
+
+mod error;
+mod notification;
+mod project;
+
+/// Main entry point for HTTP API.
+pub async fn run(librad_paths: paths::Paths, reg: registry::Registry) {
+    let subscriptions = crate::notification::Subscriptions::default();
+
+    let api = path("v1").and(
```
garbados: This should be `path("api").and(path("v1"))`, that is, `/api/v1`, in keeping with typical API [(1)][1][(2)][2] namespacing.

[1]: https://docs.joinmastodon.org/methods/apps/
[2]: https://developer.twitter.com/en/docs/accounts-and-users/create-manage-lists/api-reference/get-lists-list

xla:
@garbados Some basic logging was added in fbac784.

proxy/src/http.rs
xla: I don't think there is a general convention, best-practice here. Unless I missed an obvious standard body or something similar.

xla: Having `api` distinctly in the path makes sense if the server offers other endpoints as well. For our case the proxy is literally the API itself.

garbados:
I added an `API.md` which documents the basics of the API, and includes examples for each implemented endpoint.

proxy/src/http.rs
cloudhead: Well, how about telemetry or debugging/metrics endpoints? Won't we have those somewhere in there?

MeBrei:
Regarding the notifications: 
In theory I like the idea of notification especially for the transaction center, ie adding new transaction to the list. Will it be possible to send notifications also if there were changes outside of upstream? eg the number of applied blocks for a transaction has changed. Otherwise we would have to additionally query the transaction, so the notifications might be unnecessary.

With notifications/subscriptions I also see the danger of changes being lost, ie don't make it into the ui, because a view might rely on notifications that were not fired by a certain change.

xla:
> In theory I like the idea of notification especially for the transaction center, ie adding new transaction to the list. Will it be possible to send notifications also if there were changes outside of upstream? eg the number of applied blocks for a transaction has changed. Otherwise we would have to additionally query the transaction, so the notifications might be unnecessary.

@MeBrei I assume what you referring to here is when no explicit interaction/action happened in the application, but some a significant change happened. First of all as the notifications are driven by the proxy it must be happening somewhere in there. Yet for the example you brought up, we can imagine that there is a routine in the background which queries the current height of the Registry and either notifies about the new height and then the UI can react to that, or we emit updates per affect transaction, up to negotiations. In short: yes - it is possible - if we implement it.

> With notifications/subscriptions I also see the danger of changes being lost, ie don't make it into the ui, because a view might rely on notifications that were not fired by a certain change.

They are just a mechanism to signal that some change to the data corpus has occurred. If we want to avoid any data inconsistencies we can treat them simply as signals, not carrying any data themselves and therefore just trigger fresh reads/fetches the UI does from the proxy. Totally agree, notifications should never be used as a source of truth and there should always be a way to fetch a consistent view of the data in question.

proxy/src/http.rs
garbados: Hypothetically even those metrics and debugging endpoints would be namespaced under `/v1/`, as `/v1/stats` in anticipation of a potential `/v2/stats` or `/v2/stats_but_different`.

Mastodon and others use a `/api/v1/` path because they expose other non-API HTTP endpoints. Twitter puts `api` in the subdomain rather than the path because the responding server only handles an API, which is more like our use-case as the proxy *is* the API.

ui/src/remote.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export type Data<T> =
+  { status: Status.NotAsked } |
+  { status: Status.Loading } |
+  { status: Status.Success, data: T } |
+  { status: Status.Error, error: Error };
+
+
+// A Store is a typesafe svelte readable store that exposes `updateStatus`
```
sarahscott, line 19: lol this comment is now out of date oops

ui/src/project.ts
```diff
@@ -0,0 +1,54 @@
+import { writable } from "svelte/store";
+
+import * as api from "./api";
+import * as event from "./event";
+import * as remote from "./remote";
```
sarahscott, line 5: I prefer named imports; I feel they communicate everything upfront and put the compiler errors in a single place if something is missing from the module:
`import { createStore } from "./remote"`

xla: I'm in favour of qualification as it usually ends up that the module name is repeated in the names of methods and types, i.e.: `createStore` -> `createRemoteStore`. And if not it helps to contextualize and give indication that it is a foreign type.

xla:
Superseeded by #296

Issue #230, opened 2020-03-17
Avatar generation in handle registration flow


juliendonck:
@MeBrei ran us through the handle registration flow ( #216 ) and the way that the avatar behaves seems strange to me. 

The way it's implemented here is that it would take the handle that you created in the p2p-identity-creation and prefill this in the app, and this is what we expected. The thing is that right now the avatar is also generated from *just* the p2p-handle, this means that if your p2p-handle is taken in the registry and you change it, your avatar would change as well. This doesn't make sense to me, I would expect that the avatar based on your p2p-handle stays even if your handle is taken in the registry. (I would however expect this behaviour in the org registration). 
A possible solution would be that the avatar is generated not just from the p2p-handle, but the whole handle+public key (e.g.: cloudhead@jdfsfhfhjkshf...) and that one is "unique" and stays with you.

@cloudhead how did you expect this to work?
cc @brandonhaslegs

juliendonck:
Basically, the avatar should be based on handle@<public-key> during the identity creation (if you didn't fill in an avatar_url) and then not change anymore during the handle registration.

cloudhead:
The only way I see this working is that we base the avatar on your radicle-link ID. Not on your handle, because that can change.

When you register your ID, we don't change anything, we still use your radicle-link ID for avatar generation everywhere.

juliendonck:
Yes I agree. @MeBrei that would mean that the avatar should come from the "create identity flow" and not from the "register your handle" flow. Does that make sense?

xla:
@MeBrei @radicle-dev/product Do we need to follow-up/address anything given the latest insights?

MeBrei:
@xla I see the following changes which could be done in PRs connected to this issue:

- proxy: use id to generate the avatar fallback (instead of handle), the endpoint for querying only the avatar of a handle is no longer needed
- ui: remove `avatar` query from the user registration modal, always use the avatar fallback of the identity

PR #251, opened 2020-03-25
feat(proxy): extend project with org/user relation

### LINKED ISSUE #245: Extend Project query with User/Org relation
Currently we only return a boolean which indicates if a project is registered or not. However in the UI we'll need to show whether a project is: **registered**, **registered under an org** or **registered under a user**.

Screenshot 2020-03-23 at 11 28 21

Ideally we'd turn `registered` into a GraphQL relation that returns: `Option(User|Org)`, like so:
```graphql
union Registration = User | Org

type Project {
  id: ID!
  metadata: ProjectMetadata!
  registeredUnder: Registration
  stats: ProjectStats!
}
```

For reference, the status quo is:
```graphql
type ProjectMetadata {
  defaultBranch: String!
  description: String!
  imgUrl: String!
  name: String!
}

type ProjectStats {
  branches: Int!
  commits: Int!
  contributors: Int!
}

type Project {
  id: ID!
  metadata: ProjectMetadata!
  registered: Boolean!
  stats: ProjectStats!
}
```

xla:
Closes #245

### DIFF
--- a/proxy/src/graphql/schema.rs
+++ b/proxy/src/graphql/schema.rs
@@ -95,6 +95,7 @@ impl Mutation {
         Ok(project::Project {
             id: librad::project::ProjectId::from(id),
             metadata: meta.into(),
+            registration: None,
         })
     }
 
@@ -256,6 +257,7 @@ impl Query {
         Ok(project::Project {
             id: librad::project::ProjectId::from_str(&id.to_string())?,
             metadata: meta.into(),
+            registration: None,
         })
     }
 
@@ -269,6 +271,7 @@ impl Query {
         .map(|(id, meta)| project::Project {
             id,
             metadata: meta.into(),
+            registration: None,
         })
         .collect::<Vec<project::Project>>();
 
@@ -371,6 +374,7 @@ impl ControlMutation {
         Ok(project::Project {
             id: id.into(),
             metadata: meta.into(),
+            registration: None,
         })
     }
 
@@ -620,8 +624,23 @@ impl project::Project {
     fn metadata(&self) -> &project::Metadata {
         &self.metadata
     }
-    fn registered(&self) -> bool {
-        false
+    fn registered(&self) -> Option<ProjectRegistration> {
+        if let Some(registration) = &self.registration {
+            match registration {
+                project::Registration::Org(org_id) => {
+                    Some(ProjectRegistration::Org(OrgRegistration {
+                        org_id: juniper::ID::new(org_id.to_string()),
+                    }))
+                },
+                project::Registration::User(user_id) => {
+                    Some(ProjectRegistration::User(UserRegistration {
+                        user_id: juniper::ID::new(user_id.to_string()),
+                    }))
+                },
+            }
+        } else {
+            None
+        }
     }
 
     fn stats(&self) -> &project::Stats {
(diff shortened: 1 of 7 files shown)


proxy/tests/graphql_mutation.rs
```diff
@@ -186,7 +193,7 @@ fn register_project() {
         let query = "mutation($projectName: String!, $orgId: String!) {
                         registerProject(projectName: $projectName, orgId: $orgId) {
                             messages {
-                                ... on ProjectRegistration {
+                                ... on ProjectRegistrationMessage {
                                     projectName,
                                     orgId
                                 }
```
MeBrei: Can you add the `registered` info to the response here?

xla: Here we get back a transaction, which is not the actual project. But it's a good point we should query for it when we test for retrieving a project.

proxy/src/graphql/schema.rs
```diff
@@ -620,8 +624,23 @@ impl project::Project {
...
         &self.metadata
     }
-    fn registered(&self) -> bool {
-        false
+    fn registered(&self) -> Option<ProjectRegistration> {
+        if let Some(registration) = &self.registration {
+            match registration {
+                project::Registration::Org(org_id) => {
+                    Some(ProjectRegistration::Org(OrgRegistration {
+                        org_id: juniper::ID::new(org_id.to_string()),
```
rudolfs: Could we return the actual org/user instead of just the id and maybe mock out some sample data?

So far it looks like we'll need:
 - user handle / org name
 - avatarUrl
 - avatarFallback

xla: To mock this meaningfully we need orgs first. For the rest we have to decide what behaviour we want as:

* neither user nor orgs have avatar urls
* only orgs have a fallback avatar
* only users maybe have an coco identity attested to them
  * which only maybe is available at the time of query

What we can do for now is to show the `foo / bar` info with a returned avatar from the `avatar` query. How we control that a project appears as registered is open for discussion, it could be a query parameter that we allow for now.

rudolfs: > neither user nor orgs have avatar urls

Good point, forgot about that.

> only orgs have a fallback avatar

Why wouldn't a user have a fallback avatar?

> What we can do for now is to show the foo / bar info with a returned avatar from the avatar query. 

Deal.

xla: > Why wouldn't a user have a fallback avatar?

AFAIK only identities have fallback avatar, the handle more specifically. @cloudhead and @MeBrei can give more context.

rudolfs: I wonder why would we otherwise have user emojis:

https://github.com/radicle-dev/radicle-upstream/blob/e1b7572a0364e81f43ae2541c53663ca82f6000b/proxy/src/avatar.rs#L44:L51

rudolfs: I wonder why would we otherwise have user emojis:

https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/src/avatar.rs#L52-L59

rudolfs: I guess this will be answered in: https://github.com/radicle-dev/radicle-upstream/issues/230.

xla: @rudolfs We have user emojis for the identity handles.

xla: Expanded on in https://github.com/radicle-dev/radicle-upstream/issues/230#issuecomment-600516204

PR #257, opened 2020-03-26
feat(proxy): add thresholds to transaction list

### LINKED ISSUE #254: Add thresholds to transactions
Part of #225 

For displaying the transaction states `success` and `pending` in the ui, we need the thresholds for those two events, ie the number of blocks required for the transaction to be `settled` or `confirmed`. 

Ideally (in the future) the thresholds should be part of a general config query which get stored globally.

xla:
Embedds the thresholds in the listTransactions query response for now. We serve two fields: `confirmation` and `settlement`.

Closes #254

### DIFF
--- a/proxy/src/registry.rs
+++ b/proxy/src/registry.rs
@@ -72,6 +72,14 @@ pub enum TransactionState {
     Applied(Hash),
 }
 
+/// Configured thresholds for acceptance criteria of transaction progress.
+pub struct Thresholds {
+    /// Number of blocks after which a [`Transaction`] is assumed to be confirmed.
+    pub confirmation: u64,
+    /// Number of blocks after which a [`Transaction`] is assumed to be settled.
+    pub settlement: u64,
+}
+
 /// Registry client wrapper.
 #[derive(Clone)]
 pub struct Registry {
@@ -351,6 +359,15 @@ impl Registry {
         self.cache_transaction(tx.clone()).await;
 
         Ok(tx)
+    }
+
+    /// Returns the configured thresholds for [`Transaction`] acceptance stages.
+    #[must_use]
+    pub const fn thresholds() -> Thresholds {
+        Thresholds {
+            confirmation: 3,
+            settlement: 9,
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


proxy/src/registry.rs
```diff
@@ -72,6 +72,14 @@ pub enum TransactionState {
     Applied(Hash),
 }
 
+/// Configured thresholds for acceptance criteria of transaction progress.
+pub struct Thresholds {
+    /// Number of blocks after which a [`Transaction`] is assumed to be confirmed.
+    pub confirmation: u64,
```
cloudhead: Important to note that this is *including* the block in which it is confirmed.

proxy/src/registry.rs
```diff
@@ -352,6 +360,15 @@ impl Registry {
 
         Ok(tx)
     }
+
+    /// Returns the configured thresholds for [`Transaction`] acceptance stages.
+    #[must_use]
+    pub const fn thresholds() -> Thresholds {
+        Thresholds {
+            confirmation: 3,
+            settlement: 9,
```
cloudhead: Not sure these values make sense :thinking: -- `confirmation` should be `1` if it's supposed to represent how many blocks to wait for confirmation.

cloudhead: If we're talking about thresholds, we actually only need `settlement`, because `confirmation` is always `1`

xla: Isn't confirmation `1` the equivalent of being applied?

cloudhead: Yeah -- all I'm saying is we don't need both numbers: when a transaction lands in a block, it can be considered "mined" or "confirmed with deph=1". The number that's configurable is how many blocks (confirmations) to wait for the tx to be settled. This is `6` for bitcoin.

cloudhead: It's also confusing because some people use the term "confirmed" to mean "settled".

cloudhead: So you could have just `Thresholds { confirmations: 6 }`, and that would do the job. "Wait 6 confirmations before considering the tx settled."

PR #264, opened 2020-04-01
Adjust project creation UX according to new designs

rudolfs:
Following up on the discussions with the #product team, we decided to:
- remove project avatar images entirely
- publish projects by default (remove "publish" checkbox from existing project option)
- ~~rename button `Create project` -> `Publish project`~~
- fix validation message alignment (if the text exceeds the component width, it gets broken into two lines and centered)

Screenshot 2020-04-01 at 18 31 57

### DIFF
--- a/proxy/src/coco.rs
+++ b/proxy/src/coco.rs
@@ -8,7 +8,7 @@ use std::str::FromStr;
 
 use librad::git;
 use librad::keys;
-use librad::meta::{self};
+use librad::meta;
 use librad::paths::Paths;
 use librad::peer;
 use librad::project;
--- a/proxy/src/coco.rs
+++ b/proxy/src/coco.rs
@@ -555,27 +555,23 @@ pub fn replicate_platinum(
 /// [`librad::paths::Paths`].
 pub fn setup_fixtures(librad_paths: &Paths, root: &str) -> Result<(), error::Error> {
     let infos = vec![
-            (
-                "monokel",
-                "A looking glass into the future",
-                "master",
-            ),
-            (
-                "Monadic",
-                "Open source organization of amazing things.",
-                "master",
-            ),
-            (
-                "open source coin",
-                "Research for the sustainability of the open source community.",
-                "master",
-            ),
-            (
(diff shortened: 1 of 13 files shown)


rudolfs:
@cloudhead I wonder if we should also adjust the title?

proxy/src/coco.rs
```diff
@@ -8,7 +8,7 @@ use std::str::FromStr;
 
 use librad::git;
 use librad::keys;
-use librad::meta::{self, common::Url};
+use librad::meta::{self};
```
xla, line 11:
```suggestion
use librad::meta;
```

rudolfs:
@MeBrei we could fix the button behaviour with:

```diff
diff --git a/ui/Screen/CreateProject.svelte b/ui/Screen/CreateProject.svelte
index ab9d5cf..5ceab35 100644
--- a/ui/Screen/CreateProject.svelte
+++ b/ui/Screen/CreateProject.svelte
@@ -34,7 +34,9 @@
   let newRepositoryPath = "";
   let existingRepositoryPath = "";

-  let validations = false;
+  // validatejs returns undefined if all validations pass
+  let validations = undefined;
+
   let beginValidation = false;

   validatejs.options = {
@@ -387,7 +389,7 @@
             </Button>
             <Button
               dataCy="create-project-button"
-              disabled={!(name && currentSelection)}
+              disabled={!(beginValidation && validations === undefined)}
               variant="primary"
               on:click={createProject}>
               Publish project
```

But then we'd have to throw out all of the validation tests, because we can't trigger the validations unless we fill out the form entirely. Since cypress only tests in a browser at the moment, we can't use the directory input as it relies on Electron bindings to open the system dialog.

We could look at it again once https://github.com/radicle-dev/radicle-upstream/issues/145 is closed.

juliendonck:
@cloudhead @rudolfs as mentioned in IRC, I think if we remove the checkmark for "publish by default" we should just remove that concept in it's entirety for the time being and just "create a project". I think otherwise this will be confusing.

PR #272, opened 2020-04-03
feat(proxy): transaction costs endpoint

### LINKED ISSUE #271: Implement transaction costs endpoint
In order to show the transaction preview in the UI it should be empowered to get a cost estimate given the payload planned to be submitted. This is to give accurate feedback to the user and have them submit with confidence.

Ref #175

xla:
Closes #271

### DIFF
--- a/proxy/src/graphql/schema.rs
+++ b/proxy/src/graphql/schema.rs
@@ -925,6 +925,12 @@ impl Message {
     }
 }
 
+impl juniper::FromInputValue for Message {
+    fn from_input_value(_value: &juniper::InputValue) -> Option<Self> {
+        None
+    }
+}
+
 juniper::graphql_union!(Message: () where Scalar = <S> |&self| {
     instance_resolvers: |_| {
         &OrgRegistrationMessage => match *self {
@@ -1014,9 +1020,3 @@ juniper::graphql_union!(TransactionState: () where Scalar = <S> |&self| {
         &Applied => match *self { TransactionState::Applied(ref a) => Some(a) },
     }
 });
-
-impl juniper::FromInputValue for Message {
-    fn from_input_value(_value: &juniper::InputValue) -> Option<Self> {
-        None
-    }
-}
--- a/proxy/src/graphql/schema.rs
+++ b/proxy/src/graphql/schema.rs
@@ -312,6 +312,25 @@ impl Query {
         })
     }
 
+    fn transaction_costs(
+        ctx: &Context,
+        messages: Vec<Message>,
+    ) -> Result<registry::TransactionCosts, error::Error> {
+        let costs = futures::executor::block_on(
+            ctx.registry
+                .read()
+                .expect("unable to acquire read lock")
+                .transaction_costs(
+                    messages
+                        .iter()
+                        .map(|msg| msg.to_registry().expect("unable to transform message"))
+                        .collect::<Vec<registry::Message>>(),
+                ),
+        )?;
+
+        Ok(costs)
+    }
+
     fn identity(
         _ctx: &Context,
         id: juniper::ID,
@@ -676,7 +695,6 @@ impl project::Metadata {
 
 /// Union to represent possible registration states of a project.
 // TODO(xla): Remove attribute once integrated.
-#[allow(dead_code)]
 enum ProjectRegistration {
     /// Project is registered under an Org.
     Org(OrgRegistration),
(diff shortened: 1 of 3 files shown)


NunoAlexandre:
I will have a look today :+1:

proxy/src/graphql/schema.rs
```diff
@@ -766,39 +784,12 @@ impl registry::Transaction {
...
-                registry::Message::UserRegistration { handle, id } => {
-                    Message::UserRegistration(UserRegistrationMessage {
-                        kind: MessageKind::UserRegistration,
-                        handle: juniper::ID::new(handle.to_string()),
-                        id: juniper::ID::new(id.to_string()),
-                    })
-                },
-            })
-            .collect()
+        self.messages.iter().map(Message::from).collect()
```
NunoAlexandre, line 792: :ok_hand:

proxy/src/registry.rs
```diff
@@ -19,6 +19,8 @@ pub struct Transaction {
     /// Unique identifier, in actuality the Hash of the transaction. This handle should be used by
     /// the API consumer to query state changes of a transaction.
     pub id: registry::TxHash,
+    /// Deposit and fees spent for the transaction.
```
NunoAlexandre, line 22: nit:

```suggestion
    /// Deposit and fees charged for the transaction.
```

proxy/src/registry.rs
```diff
@@ -64,6 +66,15 @@ pub enum Message {
     },
 }
 
+/// Breakdown of costs of a [`Transaction`].
+#[derive(Clone)]
+pub struct TransactionCosts {
+    /// Deposit held for registrations.
+    pub deposit: u64,
+    /// Cost for the transaction to be processed and included.
```
NunoAlexandre, line 74: Given the nature of our fees, it's good to explain this more accordingly. Something like:

```suggestion
    /// Transaction fee defined by the user. The higher the fee, the higher the priority
    /// given to the transaction. It needs to be at least the minimum accepted by the registry.
```

proxy/src/registry.rs
```diff
@@ -348,6 +362,7 @@ impl Registry {
 
         let tx = Transaction {
             id: register_applied.tx_hash,
+            costs: TransactionCosts { deposit: 0, fee: 0 },
```
NunoAlexandre, line 365: I'd set these `fee` values to >1, 0 isn't acceptable.

proxy/src/registry.rs
```diff
@@ -369,6 +384,17 @@ impl Registry {
...
     }
+
+    /// Returns the cost estimate of a transaction which includes the given messages.
+    pub async fn transaction_costs(
+        &self,
+        _messages: Vec<Message>,
+    ) -> Result<TransactionCosts, error::Error> {
+        Ok(TransactionCosts {
+            deposit: 1000,
+            fee: 100,
```
NunoAlexandre, line 395: Is this function meant to give an estimate of how much (minimally speaking) the user should specify for each of the tx costs? If so, the doc should better reflect that. Otherwise, it creates confusion since the `fee` is something the user defines, not us. We really just define a minimum (global, transaction-agnostic) value to avoid the network from being DDoSed/spammed.

NunoAlexandre: Note that we could add a query to the registry client giving us here that minimum fee value.

xla: Created radicle-dev/radicle-registry#324 as a follow-up and addressed the doc string.

xla:
As of now GaphQL doesn't permit unions as input types, see the [RFC](https://github.com/graphql/graphql-spec/blob/master/rfcs/InputUnion.md) and this [writeup](https://brunoscheufler.com/). As a consequence the proposed API where we pass the messages upfront for cost estimation can't be provided.

My suggestion is we revisit this issue once we handle user defined fees and made a decision on the future of our API (REST vs. GraphQL). Wdyt @MeBrei ?

proxy/src/registry.rs
NunoAlexandre: @xla Done: https://github.com/radicle-dev/radicle-registry/pull/326

Will leave the integration up to you since you are still on it.

NunoAlexandre: Fixed by https://github.com/radicle-dev/radicle-registry/pull/326. You can use `client::MINIMUM_FEE` to determine whether a given amount is sufficient and/or to just get the minimal allowed.

MeBrei:
> My suggestion is we revisit this issue once we handle user defined fees and made a decision on the future of our API (REST vs. GraphQL). Wdyt @MeBrei ?

That makes sense!

xla:
Closing as we will pick  it up later.

PR #278, opened 2020-04-08
feat: single commit view header

### LINKED ISSUE #274: Single commit view
Implement the single commit view, as part of **Code Browsing 2**.

* [x] Implement commit header
  * [x] UI
  * [x] Backend
* [ ] Show commit parent(s)
* [ ] Show commit branch(es) from `surf`
* [x] Implement file diff
  * [x] UI
  * [x] Backend

cloudhead:
This only includes the commit header. No changeset yet, since this needs to be provided by coco. Note that since we don't have the *user* from a commit, just author/email, I've deviated slightly from the original [design](https://www.figma.com/file/dnYlebm0ehgyJyT3Po9Kcl/Upstream?node-id=5995%3A2425).

2020-04-09-152625_1226x685_scrot

Addresses parts of #274

### DIFF
--- a/proxy/src/graphql/schema.rs
+++ b/proxy/src/graphql/schema.rs
@@ -506,6 +506,10 @@ impl coco::Commit {
         &self.summary
     }
 
+    fn description(&self) -> &str {
+        self.description()
+    }
+
     fn message(&self) -> &str {
         &self.message
     }
--- a/proxy/src/graphql/schema.rs
+++ b/proxy/src/graphql/schema.rs
@@ -510,6 +510,10 @@ impl coco::Commit {
         &self.message
     }
 
+    fn committer(&self) -> &coco::Person {
+        &self.committer
+    }
+
     fn committer_time(&self) -> String {
         self.committer_time.seconds().to_string()
     }
--- a/ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
+++ b/ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
@@ -1,12 +1,10 @@
 <script>
-  import { getContext } from "svelte";
   import { link } from "svelte-spa-router";
   import { Text } from "../../Primitive";
   import UserCard from "../UserCard.svelte";
   import * as path from "../../../lib/path.js";
 
-  const projectId = getContext("projectId");
-
+  export let projectId = null;
   export let commitMessage = null;
   export let timestamp = null;
   export let commitSha = null;
--- a/ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
+++ b/ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
@@ -1,7 +1,11 @@
 <script>
+  import { getContext } from "svelte";
   import { link } from "svelte-spa-router";
   import { Text } from "../../Primitive";
   import UserCard from "../UserCard.svelte";
+  import * as path from "../../../lib/path.js";
+
+  const projectId = getContext("projectId");
 
   export let commitMessage = null;
   export let timestamp = null;
@@ -49,8 +53,11 @@
   <div class="align-right">
     <Text style="color: var(--color-foreground-level-6)">
       Latest commit
-      <a class="commitSha" href="/projects/123/commits/{commitSha}" use:link>
-        {commitSha}
+      <a
+        class="commitSha"
+        href={path.projectCommit(projectId, commitSha)}
+        use:link>
+        {commitSha.substring(0, 7)}
       </a>
       {timestamp}
     </Text>
--- a/ui/DesignSystem/Primitive/Icon/Branch.svelte
+++ b/ui/DesignSystem/Primitive/Icon/Branch.svelte
@@ -1,6 +1,5 @@
 <script>
   export let style = null;
-  export let color = "foreground-level-5";
 </script>
 
 <svg
@@ -8,7 +7,7 @@
   width="24"
   height="24"
   viewBox="0 0 24 24"
-  fill="var(--color-{color})"
+  fill="var(--color-foreground-level-5)"
   xmlns="http://www.w3.org/2000/svg">
   <path
     fill-rule="evenodd"
--- a/ui/DesignSystem/Primitive/Icon/Branch.svelte
+++ b/ui/DesignSystem/Primitive/Icon/Branch.svelte
@@ -1,5 +1,6 @@
 <script>
   export let style = null;
+  export let color = "foreground-level-5";
 </script>
 
 <svg
@@ -7,7 +8,7 @@
   width="24"
   height="24"
   viewBox="0 0 24 24"
-  fill="var(--color-foreground-level-5)"
+  fill="var(--color-{color})"
   xmlns="http://www.w3.org/2000/svg">
   <path
     fill-rule="evenodd"
--- a/ui/DesignSystem/Primitive/Title.svelte
+++ b/ui/DesignSystem/Primitive/Title.svelte
@@ -17,6 +17,12 @@
     line-height: 29px;
   }
 
+  .large {
+    font-family: var(--typeface-bold);
+    font-size: 20px;
+    line-height: 24px;
+  }
+
   .medium {
     font-family: var(--typeface-bold);
     font-size: 16px;
--- a/ui/DesignSystem/Primitive/Title.svelte
+++ b/ui/DesignSystem/Primitive/Title.svelte
@@ -1,7 +1,7 @@
 <script>
   export let style = null;
 
-  export let variant = "regular"; // regular | medium | big | huge
+  export let variant = "regular"; // regular | medium | large | big | huge
 </script>
 
 <style>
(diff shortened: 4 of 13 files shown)


ui/DesignSystem/Primitive/Title.svelte
```diff
@@ -17,6 +17,12 @@
     line-height: 29px;
   }
 
+  .large {
+    font-family: var(--typeface-bold);
+    font-size: 20px;
+    line-height: 24px;
+  }
```
cloudhead, line 24: This size was missing.

ui/Screen/Project/Commit.svelte
```diff
@@ -0,0 +1,131 @@
...
+        variables: {
+          projectId: projectId,
+          commitHash: commitHash
+        }
+      });
+      const result = await response.result();
+      const commit = result.data.commit;
+
+      // TODO(cloudhead): Fetch branch from backend.
+      commit.branch = "master";
```
cloudhead, line 44: We need to provide this somehow from the proxy, when fetching a commit.

cloudhead:
Early review on this would be nice, and also discussion on how/if to include branch name in commit payload.

ui/Screen/Project/Commit.svelte
xla: What are the semantics of branch association for commits? How do we deal with it occurring in multiple? Is this specced already? If so we can then evaluate if we have the building blocks to implement it or need more functionality in surf.

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
```diff
@@ -1,7 +1,11 @@
 <script>
+  import { getContext } from "svelte";
   import { link } from "svelte-spa-router";
   import { Text } from "../../Primitive";
   import UserCard from "../UserCard.svelte";
+  import * as path from "../../../lib/path.js";
+
+  const projectId = getContext("projectId");
```
xla, line 8: Why did you decide to pass this through the context instead of a component prop? Where is the context set?

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
```diff
@@ -1,7 +1,11 @@
 <script>
```
xla, line 1: Is this component used anywhere?

ui/Screen/Project/Commit.svelte
xla: Would also reckon that in this view it's important to not only show the branch but the `repo`, `branch` tuple unless we solve that in another way.

ui/Screen/Project/Commit.svelte
```diff
@@ -0,0 +1,131 @@
...
+  import { showNotification } from "../../store/notification.js";
+  import { getClient, query } from "svelte-apollo";
+  import { format } from "timeago.js";
+  import * as path from "../../lib/path.js";
+
+  export let params = null;
+  const projectId = params.id;
+  const commitHash = params.hash;
+
+  const QUERY = gql`
```
xla, line 15:
```suggestion
  const GET_COMMIT = gql`
```

ui/Screen/Project/Commit.svelte
```diff
@@ -0,0 +1,131 @@
...
+        sha1
+        summary
+      }
+    }
+  `;
+
+  async function fetchCommit() {
+    try {
+      const response = await query(getClient(), {
+        query: QUERY,
```
xla, line 34:
```suggestion
        query: GET_COMMIT,
```

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
rudolfs: I agree, we should pass it in via a prop and the callers should be adjusted accordingly.

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
cloudhead: Yeah, it's what is shown above source files for eg.

ui/Screen/Project/Commit.svelte
```diff
@@ -0,0 +1,131 @@
...
+</style>
+
+{#await fetchCommit() then commit}
+  <header>
+    <Flex align="space-between">
+      <Title variant="large" style="margin-bottom: 1rem">
+        {commit.summary}
+      </Title>
+      <span class="field">
+        Committed to
```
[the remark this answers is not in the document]
cloudhead: That's what I thought, but unfortunately `Text` is a `<p>`, which doesn't work here. I discussed this with rudolfs some time ago, I don't think we should have a `Text` component, but rather unpack it into `<Body>`, `<Label>` etc.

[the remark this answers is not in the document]
cloudhead: Actually it might work for this specific case, but not the other `<span>`..

ui/Screen/Project/Commit.svelte
```diff
@@ -0,0 +1,131 @@
...
+          <span class="author">{commit.author.name}</span>
+          <span class="email">&lt;{commit.author.email}&gt;</span>
+        </p>
+      </div>
+      <Flex align="left" style="align-items: flex-end">
+        <p class="field">
+          Commit
+          <a href={path.projectCommit(projectId, commit.sha1)} use:link>
+            {commit.sha1}
+          </a>
```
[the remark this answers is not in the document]
cloudhead: Yeah it does, but commit hashes are always links as far as I can tell.

ui/Screen/Project/Commit.svelte
```diff
@@ -0,0 +1,131 @@
...
+            style="vertical-align: bottom" />
+          {commit.branch}
+        </span>
+        {format(commit.committerTime)}
+      </span>
+    </Flex>
+    <!-- TODO(cloudhead): Should be <pre> -->
+    <Code variant="code" style="margin-bottom: 1rem">{commit.message}</Code>
+    <hr />
+    <Flex align="space-between">
```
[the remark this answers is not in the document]
cloudhead: Ah yes you might be right, I forgot about the slot thing.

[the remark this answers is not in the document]
cloudhead: Ok I'm not sure this will work, the complication is that, if you look at the design file, we may sometimes get merge commit information, as well as commiter/author distinction, and want to bottom align in case there are two rows on one side and one row on the other side.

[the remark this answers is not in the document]
cloudhead: Nevermind, got it working!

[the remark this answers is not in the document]
cloudhead: 083f443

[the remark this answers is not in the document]
MeBrei: adding `style="align-items: flex-end;"` to the `Flex` in my code snippet, solves that problem.

proxy/src/graphql/schema.rs
```diff
@@ -510,6 +510,10 @@ impl coco::Commit {
         &self.message
     }
 
+    fn committer(&self) -> &coco::Person {
+        &self.committer
+    }
+
     fn committer_time(&self) -> String {
         self.committer_time.seconds().to_string()
     }
```
cloudhead, line 519: Should I remove this now? Only problem is that the time won't be converted to seconds anymore.

xla: I think you can't remove this as juniper won't know how to convert that type.

ui/DesignSystem/Primitive/Icon/Branch.svelte
```diff
@@ -1,13 +1,14 @@
 <script>
   export let style = null;
+  export let color = "foreground-level-5";
```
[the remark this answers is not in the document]
cloudhead: Oh cool. 9a47b9b

ui/Screen/Project/Commit.svelte
cloudhead: Done. d85aabf

ui/Screen/Project/Commit.svelte
```diff
@@ -0,0 +1,131 @@
+<script>
+  import { Title, Flex, Code } from "../../DesignSystem/Primitive";
+  import { Icon } from "../../DesignSystem/Primitive";
+  import { gql } from "apollo-boost";
+  import { link } from "svelte-spa-router";
+  import { showNotification } from "../../store/notification.js";
+  import { getClient, query } from "svelte-apollo";
+  import { format } from "timeago.js";
+  import * as path from "../../lib/path.js";
```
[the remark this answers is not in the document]
cloudhead: f002076

ui/Screen/Project/Commit.svelte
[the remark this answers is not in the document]
cloudhead: Ok checked with Julien, we're going to make this not a link.

cloudhead:
I've addressed review feedback, thanks!

Since most of the missing functionality requires some changes downstream from `surf`, I'd like to get this merged with the current set of functionality, to not keep things lingering too long. I'm tracking the missing pieces here: https://github.com/radicle-dev/radicle-upstream/issues/274

Another big piece that is missing, that @xla brought up is everything around the "repository". Since that is a cross-cutting change, I also want to address it separately.

Issue #277, opened 2020-04-08
Overview: Orgs I


MeBrei:
### UI
- [x] sidebar: list orgs "I am a member of"
- [x] modal: org registration
- [x] org profile
  - [x] needs: handle, avatar
  - [x] list projects registered under the org
  - [x] list members registered under the org
- [x] modal: create/register a project under an org (and adjust the modal for project creation under a user along the way) (#285)
- [x] modal: register a member under an org
- [ ] block certain actions according to transaction state, show tooltip spinner & badge on finish (#269)

### Proxy
- [x] Query
  - [x] org (#300)
  - [x] list orgs I am a member of (#309, #340)
  - [x] list projects of an org (#343 )
  - [x] ~listTransactions: add new messages of mutations~ (transaction and member handling moved to separate issues #349 #347 )
- [x] Mutation
   - [x] register org (#300)
   - [x] create/register a project under an org (they already exist as separate mutations)
- [x] ~transactions: permissions, errors, progress, "receiver"~ (#225) (transaction and member handling moved to separate issues #349 #347 )

For more info see #171

MeBrei:
This is a first draft of the related schema. Some of it is already implemented and/or adjusted.

```gql

type Org {
  id: ID!
  avatarFallback: Avatar!
}

type Member {
  handle: ID!
  orgId: ID!
  timestamp: String!
}

union Message = ... | OrgRegistrationMessage | OrgUnregistrationMessage | ProjectRegistrationMessage | MemberRegistration

"""Kind of the transaction message."""
enum MessageKind {
  ...

  """Registration of a new member."""
  MEMBER_REGISTRATION
}

type UpstreamMutation {
  ...
  registerProject(projectName: String!, orgId: String!, maybeLibradIdInput: ID): Transaction!
  registerMember(handle: ID!, orgId: ID!): Transaction!
}

type UpstreamQuery {
  ...
  avatar(handle: ID!, usage: AvatarUsage!): Avatar!
  project(id: ID!): Project!
  projects: [Project!]!
  """List of projects on the registry by id (org or user)"""
  listRegistryProjects(id: ID!): [ID!]!
  """Lists all orgs the handle is a member of"""
  orgs(handle: ID!): [Org!]!
  """Returns the org if it exists"""
  org(id: ID!): Org
  """Returns all registered members of an org"""
  members(id: ID!): [Member!]!
}

"""Context data for a project registered under an Org."""
type OrgRegistration {
[...]

Issue #166, opened 2020-02-19
Consistent validations


MeBrei:
Follow up issue of #119 

We currently use `validate.js` for validations in ui. The same logic has to be repeated in each component where it is used and custom validation and error handling add to its complexity. At the time of closing #119 there was no svelte library that would be feasible to use.

- [ ] investigate validation libraries
- [ ] extract ui validation logic as much as possible
- [ ] avoid mismatched validations and define authorities of validation between ui and API

sarahscott:
We should think about validating with respect to the needs of the registry, as transactions will fail if certain constraints aren't met. Ideally we would have one source of truth for what registry-legible data looks like (e.g. instead of having regular expressions defined all over the app, we would pull them from proxy as-needed).

We already have [specs for registry data](https://github.com/radicle-dev/registry-spec/), so this might be a good place to start.

Also, we copy & paste a lot of validation code in our components. Here's an overview of what validation looks like app-wide:

![Screen Shot 2020-04-22 at 3 30 23 PM](https://user-images.githubusercontent.com/2712962/80042379-ab873a00-84b3-11ea-85ea-76bc51e04e4d.png)

Perhaps a reactive `validations` store would be a better approach?

MeBrei:
### Idea for moving the validations into the proxy

Currently most validations live in the ui and some of them have diverged from those in the protocol. In the proxy, before transactions are sent to the registry, little validations happen so that the api relies on the ui to provide the correct data. This concepts aim is to consolidate the validations in the proxy and provide an endpoint for the ui to consume the validations. This will prepare for eventually deriving validations from the protocols. 

Idea: A path prefix `validations` is introduced that would be prepended to each post request we need validations for in the ui. Similar like the current `control` prefix, but the `filters` of the original post requests should be used to not add duplication and diverging paths. Validations are added to the original post request handlers. Whenever the request goes over the `validations` path, only the validations are executed (not the actual sending of the transaction), by for example passing a validation flag. Failing validations would return a validation error with a message that could be consumed in the ui to display the error message in forms.  Some validations could still remain in the ui (like input presence)

rudolfs:
Not sure if pushing validations back to the proxy is worth the effort right now. It seems to me that it would only pay off if we did all validations for every field in one place and not validate some things in the proxy and others in the UI.

Would we provide the validation messages via the API then as well?

[...]

PR #280, opened 2020-04-14
feat(ui): org registration

### LINKED ISSUE #281: Register Org Flow (UI)
UI for this modal:

Screen Shot 2020-04-14 at 1 57 10 PM
Screen Shot 2020-04-14 at 1 57 20 PM
Screen Shot 2020-04-14 at 1 57 29 PM
Screen Shot 2020-04-14 at 1 57 38 PM

sarahscott:
UI for org registration 🎉 

- [x] New org path
- [x] Make it a little easier to do hotkey modals
- [x] UI for org registration
- [x] Integrate register org endpoint
- [x] `Avatar` takes `circle`/`square` as variant
- [x] Cool new `NavigationButtons` that work on any modals with next/cancel
- [x] `Validated` input with 4 states: `NotStarted`, `Loading`, `Error`, and `Success`

orgggg

Related to #277 
Closes #281


ui/DesignSystem/Component/Registration/IdentifierSelectionStep.svelte
```diff
@@ -0,0 +1,87 @@
...
+</Text>
+<Input.Text
+  style="--focus-outline-color: var(--color-primary); width: 100%;
+  margin-bottom: 32px;"
+  placeholder={inputPlaceholder}
+  {imageUrl}
+  variant="avatar"
+  bind:value={identifier}
+  valid={!(beginValidation && validations && validations.identifier)}
+  validationMessage={beginValidation && validations && validations.identifier && validations.identifier[0]}
```
xla, line 84: These chains look like they are vital business logic. All of the code around validation and what to show could be well suited to go into pure lib/js code, which in turn could be unit tested. Food for thought.

ui/DesignSystem/Component/Registration/NavigationButtons.svelte
```diff
@@ -0,0 +1,36 @@
+<script>
```
xla, line 1: Can we just name it `Navigation` or do we have another thing that is a navigation?

ui/DesignSystem/Component/Registration/NavigationButtons.svelte
```diff
@@ -0,0 +1,36 @@
...
+  import { Button } from "../../Primitive";
+
+  export let onCancel = null;
+  export let onNextStep = null;
```
xla, line 5:
```suggestion
  export let onNext = null;
```

ui/DesignSystem/Component/Registration/NavigationButtons.svelte
```diff
@@ -0,0 +1,36 @@
...
+  export let disabled = null;
+  export let nextStepTitle = "Next";
```
xla, line 7:
```suggestion
  export let nextTitle = "Next";
```

ui/DesignSystem/Component/Registration/NavigationButtons.svelte
```diff
@@ -0,0 +1,36 @@
...
+      dataCy="cancel-button"
+      variant="transparent"
+      on:click={onCancel}
+      style="margin-right: 24px;">
+      Cancel
+    </Button>
+  </div>
+
+  <div>
+    <Button {disabled} on:click={onNextStep} variant="primary">
```
xla, line 32:
```suggestion
    <Button {disabled} on:click={onNext} variant="primary">
```

ui/DesignSystem/Component/Registration/NavigationButtons.svelte
```diff
@@ -0,0 +1,36 @@
...
+      {nextStepTitle}
```
xla, line 33:
```suggestion
      {nextTitle}
```

ui/Screen/OrgRegistration.svelte
```diff
@@ -0,0 +1,74 @@
...
+  import { Title } from "../DesignSystem/Primitive";
+
+  const steps = {
+    PREPARE: 1,
+    SUBMIT: 2
+  };
+
+  let currentStep = steps.PREPARE;
+
+  const nextStep = () => currentStep++;
```
xla, line 19: It would be nice to move away from the current step pattern and rather push for a proper state machine for these multi-stage flows. Which could be implemented in ts and consumed here.

ui/DesignSystem/Component/Registration/IdentifierSelectionStep.svelte
```diff
@@ -0,0 +1,87 @@
+<script>
```
xla, line 1: Would drop the `Step` suffix.

MeBrei:
I'm a bit on the fence if we actually gain a lot by extracting the registrations steps out. Especially the `IdentifierSelectionStep`. Currently we are planing on having 3 registrations that use a similar code, but they also have several differences. For user registration the `avatarFallback` is static, but for org registration it depends on the id you are typing in, so it requires an additional query. For member registration the avatarFallback & imageUrl depends on the response from the availability query and the validation is oposite of the one for org and user, ie it is an error if the member is not returned.
With including all these cases, do we still gain from extracting the component?

ui/DesignSystem/Component/Registration/TransactionSubmissionStep.svelte
```diff
@@ -0,0 +1,17 @@
+<script>
+  import NavigationButtons from "./NavigationButtons.svelte";
+  import Transaction from "../Transaction.svelte";
+
+  export let transaction = null;
+  export let onCancel = null;
+  export let onNextStep = null;
```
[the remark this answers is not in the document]
sarahscott: This pattern is great 💯 very happy to implement it

ui/App.svelte
```diff
@@ -43,7 +44,9 @@
       if ($session.data.identity === null) {
         push(path.createIdentity());
       } else {
-        push(path.profile());
+        if ($location === "/") {
```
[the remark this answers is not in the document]
sarahscott: lol maybe at some point this will be more sophisticated but for now it helps

ui/DesignSystem/Component/Registration/IdentifierSelectionStep.svelte
sarahscott: Yeah I think so much of this validation code can be shared - can discuss in #166

ui/DesignSystem/Component/Registration/NavigationButtons.svelte
sarahscott: I feel like `Navigation` is too vague. Also there are other ways we navigate through the app and this is specific to the registration flow so maybe it can be `RegistrationNavigation`?

sarahscott:
I think the UI should always be extracted if it's shared by several different components. Otherwise, it will be a nightmare to make design changes later down the line and remain consistent.

sarahscott:
These specs all look essentially the same, so I don't really understand why we would copy & paste them when we could just have one set of shared components:

Screen Shot 2020-04-23 at 12 58 32 PM
Screen Shot 2020-04-23 at 12 58 40 PM
Screen Shot 2020-04-23 at 12 58 57 PM
Screen Shot 2020-04-23 at 12 59 02 PM
Screen Shot 2020-04-23 at 12 59 05 PM
Screen Shot 2020-04-23 at 12 59 46 PM

sarahscott:
In fact, I think part of the reason they're designed this way is to make our job easier and less repetitive 😄

ui/DesignSystem/Component/index.js
```diff
@@ -2,6 +2,7 @@ import AdditionalActionsDropdown from "./AdditionalActionsDropdown.svelte";
 import Copyable from "./Copyable.svelte";
 import HorizontalMenu from "./HorizontalMenu.svelte";
 import IdentityAvatar from "./IdentityAvatar.svelte";
+import NavigationButtons from "./Registration/NavigationButtons.svelte";
```
xla, line 5: If this becomes a generally exported component, it should have a more descriptive name. Navigation is a broad term. As it lives in our design system I'm curious to hear what @radicle-dev/product thinks this re-occurring element should be named.

ui/lib/path.js
```diff
@@ -18,6 +18,7 @@ export const registerUser = () => "/user-registration";
 export const createIdentity = _params => "/identity/new";
 
 export const orgs = id => `/orgs/${id}`;
+export const orgRegistration = id => `/orgs/${id}/register`;
```
xla, line 21: I don't think we know the `id` upfront before going through the registration flow, shouldn't it be `/orgs/register`?

ui/DesignSystem/Primitive/Input/Validated.svelte
```diff
@@ -0,0 +1,108 @@
...
+
+  import { ValidationStatus } from "../../../src/validation.ts";
+
+  export let style = null;
+  export let placeholder = null;
+  export let value = null;
+  export let dataCy = null;
+
+  export let disabled = null;
+  export let variant = "vanilla"; // vanilla | avatar
```
xla, line 14: That this component has variants, where is avatar and imports the `Avatar` component is an indicator that it tries to too much and is too couple with a certain application of itself. Almost 50% of it's prop surface is to manipulate that part of the presentation.

A more reusable and less invasive way would be to allow [slots](https://svelte.dev/docs#slot) to be passed which are rendered in pre-defined spots. Similar to how the Remote component works: https://github.com/radicle-dev/radicle-upstream/blob/105a43081a27b3ad1f110df67f80a7befd0a7a89/ui/DesignSystem/Component/Remote.svelte#L7-L15

We can then build a higher-level component on top of the two primitives (Avatar and Validated), or combine them where needed, or use with ease in other contexts, without an avatar.

ui/DesignSystem/Primitive/Input/Validated.svelte
```diff
@@ -0,0 +1,108 @@
...
+      absolute; top: 0px; left: 10px" />
+  {/if}
+
+  {#if variant === 'avatar'}
+    {#if validation.status === ValidationStatus.Pending}
+      <Icon.Spinner
+        style="justify-content: flex-start; position: absolute; top: 12px;
+        right: 10px;" />
+    {:else if validation.status === ValidationStatus.Success}
+      <Icon.CheckCircle
```
xla, line 92: Shouldn't we show these indicators no matter the variant?

ui/DesignSystem/Primitive/Input/Validated.svelte
xla: We likely would get away with a single default slot for this use-case.

ui/DesignSystem/Primitive/Input/Validated.svelte
```diff
@@ -0,0 +1,108 @@
...
+  import Text from "../Text.svelte";
+  import Icon from "../Icon";
+  import Avatar from "../Avatar.svelte";
+
+  import { ValidationStatus } from "../../../src/validation.ts";
+
+  export let style = null;
+  export let placeholder = null;
+  export let value = null;
+  export let dataCy = null;
```
xla, line 11: We had a longish disucssion about this on IRC, afaik we want to move away from cypress specific leakage in non-test code. /cc @cloudhead @rudolfs

ui/DesignSystem/Component/index.js
xla: Also convinced that they don't need to live under `Registration`, it's a simple component which can be used in any form context with *cancle* and *submit* button.

ui/DesignSystem/Component/Registration/NavigationButtons.svelte
```diff
@@ -0,0 +1,36 @@
+<script>
+  import { Button } from "../../Primitive";
+
+  export let onCancel = null;
```
xla, line 4: An interface for this component which would be general and meaningful could look like this:

``` svelte
    <NavigationButtons
      style="margin-top: 32px;"
      cancelLabel={$cancelLabel}
      submitLabel={$nextButtonLabel}
      on:cancel={cancel}
      on:submit={next} />
```

Where both labels have sensible defaults.

ui/DesignSystem/Primitive/Input/Validated.svelte
xla: Large part of my rationale is that the caller has the information on how to construct an Avatar already, as they are required by this component. Therefore the caller can construct the Avatar and pass it as a whole if desired.

ui/src/validation.ts
```diff
@@ -0,0 +1,62 @@
+import validatejs from "validate.js"
+import { derived, writable, Readable, get } from "svelte/store"
+
+export enum ValidationStatus {
+  NotStarted = "NOT_STARTED",
+  FormatValidation = "FORMAT_VALIDATION",
+  Pending = "PENDING",
```
xla, line 7: Should we be consistent with `Remote` and call it `Loading`, or rename the state in `Remote`?

ui/src/validation.ts
```diff
@@ -0,0 +1,62 @@
...
+  Error = "ERROR",
+  Success = "SUCCESS"
+}
+
+type ValidationState =
+  { status: ValidationStatus.NotStarted } |
+  { status: ValidationStatus.FormatValidation; input: string } |
```
xla, line 14: Is this ever constructed?

ui/src/validation.ts
```diff
@@ -0,0 +1,62 @@
...
+  { status: ValidationStatus.Pending; input: string } |
+  { status: ValidationStatus.Error; input: string; message: string } |
+  { status: ValidationStatus.Success; input: string }
+
+interface ValidationStore extends Readable<ValidationState> {
+  validate: (input: string) => void;
+}
+
+const VALID_NAME_MATCH = new RegExp("^[a-z0-9][a-z0-9_-]+$", "i");
+const constraints = {
```
xla, line 24: These seem to be specific to orgs, should they live here?

ui/Screen/OrgRegistration.svelte
```diff
@@ -0,0 +1,108 @@
...
+    switch (state) {
+      case RegistrationFlowState.NameSelection:
+        pop();
+        break;
+      case RegistrationFlowState.TransactionConfirmation:
+        state = RegistrationFlowState.NameSelection;
+    }
+  };
+
+  const nextButtonTitle = () => {
```
xla, line 54: Can't this be a reactive statement instead of a function?

ui/src/validation.ts
sarahscott: I'm planning on this sort of thing being a context-specific variable that consumers can add

ui/src/validation.ts
sarahscott: Nope, it's something I was playing with as a state in the state machine but I don't think it's actually necessary

package.json
```diff
@@ -77,7 +77,7 @@
     "wait-on": "^4.0.1"
   },
   "scripts": {
-    "start": "run-p --race svelte:watch proxy:start graphiql:serve electron:start",
```
sarahscott, line 80: ignore this; is part of #318

ui/DesignSystem/Component/index.js
sarahscott: Waiting on @radicle-dev/product  for this one

ui/src/validation.ts
sarahscott: `Loading` works

ui/DesignSystem/Primitive/Input/Validated.svelte
sarahscott: What's the status on whether or not we should use cypress-specific syntax in production code @rudolfs @cloudhead ?

ui/Screen/OrgRegistration.svelte
```diff
@@ -0,0 +1,128 @@
...
+        Registering an org allows you to give others in your org the right to
+        sign transactions, like adding other members or adding projects.
+      </Text>
+      <Input.Validated
+        placeholder="Org name (e.g. Flowerpot)"
+        bind:value={orgName}
+        style="--focus-outline-color: var(--color-primary); width: 100%;"
+        validation={$validation}>
+        <div slot="avatar">
+          <Avatar imageUrl={org.avatar.imageUrl} variant="square" />
```
MeBrei, line 113: Org's don't have an imageUrl for now

ui/lib/hotkeys.js
```diff
@@ -31,6 +31,8 @@ export const initializeHotkeys = () => {
     console.log(`Switch to ${nextTheme}`);
   });
 
+  // TODO(sos): rethink how we do esc hotkey; any time we show a modal, it should
+  // be dismissable with esc. this can be incorporated into Modal component
```
MeBrei, line 35: I would move this into an issue instead.

ui/src/org.ts
```diff
@@ -0,0 +1,76 @@
...
+  avatarFallback: null,
+  imageUrl: imageUrl
+};
+
+export const subject = {
+  name: "",
+  kind: "org",
+  avatarFallback: null,
+  imageUrl: imageUrl
+};
```
MeBrei, line 76: Why are `payer` and `subject` part of `org.ts`? They should also be filled correctly, ie the payer is always the currently registered user with their own avatar/image.

ui/src/org.ts
```diff
@@ -0,0 +1,76 @@
...
+    message: `Org name should match [a-z0-9][a-z0-9_-]+`
+  }
+};
+
+// DUMMY DATA
+
+export const org: Org = {
+  name: "",
+  avatar: {
+    imageUrl: imageUrl
```
MeBrei, line 39: Same as above, org doesn't have a image url for now. It only has the avatarFallback.

ui/DesignSystem/Primitive/Input/Validated.svelte
```diff
@@ -0,0 +1,112 @@
+<script>
```
MeBrei, line 1: `Input/Text.svelte` already implemented this variant. Maybe making a new component out of it is clearer, but the naming `Validated` is confusing, as all inputs should be validated and the variant should then be removed from `Input/Text.svelte`

ui/Screen/OrgRegistration.svelte
rudolfs: That's correct, the Registry stores only the org handle, no images or any other metadata is planned at the moment.

ui/DesignSystem/Primitive/Input/Validated.svelte
sarahscott: 👍

sarahscott:
This is meant to be UI-only (as in, the api is not ready yet so I'm just implementing the visuals). Once we have an availability check endpoint, I can integrate it in a new PR. Same goes for the avatar (I don't think we have an endpoint for generating it on-the-fly unless I'm missing something in the swagger docs).

ui/lib/hotkeys.js
```diff
@@ -4,20 +4,20 @@ import { push, pop, location } from "svelte-spa-router";
...
-      pop();
-    }
-    push(path.help());
-  });
+export const initializeHotkeys = () => {
+  hotModal("shift+d", path.designSystemGuide());
+  hotModal("shift+/", path.help());
+
+  // TODO(sarah): Remove temporary hotkey for org registration
+  hotModal("shift+o", path.orgRegistration());
```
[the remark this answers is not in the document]
sarahscott: TBD in subsequent api integration pr

ui/src/org.ts
sarahscott: TBD in subsequent api integration pr

ui/Screen/OrgRegistration.svelte
sarahscott: TBD in subsequent api integration pr

ui/src/org.ts
sarahscott: Dummy data! TBD in subsequent api integration pr

MeBrei:
For the availability check, the endpoint is already in place: `GET /orgs/<id>` 

The avatar endpoint seems to have been removed while transitioning to the rest api. I will implement the endpoint, but might be better to add it then in another PR here

ui/DesignSystem/Primitive/Input/Validated.svelte
sarahscott: Makes sense! I updated `Input/Text`

rudolfs:
eslint seems to be complaining: https://buildkite.com/monadic/radicle-upstream/builds/1972#1b504093-f297-4369-87ff-ca06c74e87e1/155-157

ui/DesignSystem/Primitive/Input/Text.svelte
```diff
@@ -35,8 +37,13 @@
     background-color: var(--color-background);
   }
 
+<<<<<<< HEAD
```
rudolfs, line 40: Merge snafu.

rudolfs:
I tried to finish a couple of leftovers to get this thing merged, as I could use some of the changes to finish the PR I'm working on #292.

Once I got the small stuff out the way the specs showed some breakage in `CreateProject` due to validations not working properly. Leaving this to @sarahscott to fix up.

ui/DesignSystem/Component/index.js
brandonhaslegs: What does **NavigationButtons** do? Or what does it look like exactly? Is it directly related to the stepper? If so, maybe it could be called **StepperNavigationButtons**. If it's just the pink button and the "transparent" button, they should just be called **PrimaryButton**  and **TransparentButton**.

xla: @brandonhaslegs Thanks for the follow-up. It's the latter, so our navigation elements that we use in any form context. Would something like `FormNavigation` make sense?

ui/DesignSystem/Primitive/Input/Text.svelte
```diff
@@ -68,48 +70,54 @@
...
-      {imageUrl}
-      variant="user"
-      size="small"
-      style="width: 24px; height: 40px; justify-content: flex-start; position:
-      absolute; top: 0px; left: 8px" />
-  {/if}
 
-  {#if variant === 'handle'}
-    {#if validationPending}
+  <div class="avatar-wrapper">
```
xla, line 97: A more general way would be to refer to an icon here.

ui/DesignSystem/Component/index.js
brandonhaslegs: In most cases buttons are navigating forms, but they're definitely used in more contexts that just forms. For example are these `FormNavigation`?

image
image
image


Maybe I'm misunderstanding the component we're talking about though. If it's truly just for forms, `FormNavigation` is fine with me.

PR #293, opened 2020-04-17
feat: switch to REST API for most endpoints

rudolfs:
Forking off of: https://github.com/radicle-dev/radicle-upstream/pull/250 at a place where the tests were green.


ui/Screen/UserRegistration.svelte
```diff
@@ -57,6 +69,8 @@
       pop();
     }
   };
+
+  $: console.log(handle);
```
rudolfs, line 73: Stray console.log.

xla:
> It seems like the identity creation wizard is broken now. There should be a step showing the new identity after the form. Also the newly created identity isn't shown in the app, just the mock data. It was working previously in this PR.

Good catch on the first part, it's addressed in 127282a. For the identity, we now fetch the actual session data, so what actually is known to the proxy, as we get it after the identity creation flow is complete. So we use what identity is stored in the session, this is the correct behaviour and is gonna be resolved once wired up to the new coco identities.

rudolfs:
LGTM 🚀

PR #294, opened 2020-04-17
feat(ui): new project page with repository selector

cloudhead:
Early version of project page with repository selector. Things aren't hooked up yet.

2020-04-17-181544_1088x839_scrot


ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -1,31 +1,168 @@
...
-  `;
-
-  const allRevisions = query(getClient(), {
-    query: ALL_REVISIONS,
-    variables: { projectId: getContext("projectId") }
-  });
+  /* import { gql } from "apollo-boost"; */
+  /* import { getContext } from "svelte"; */
+  /* import { getClient, query } from "svelte-apollo"; */
+  import ClickOutside from "svelte-click-outside";
```
sarahscott, line 5: I think we just removed this dependency - the same functionality can be achieved with `<svelte:window onClick={hideDropdown} />`

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -1,19 +1,132 @@
...
+  const showDropdown = () => {
+    expanded = true;
+  };
+
+  const hideDropdown = () => {
+    expanded = false;
+  };
+
+  const handleClick = ev => {
+    if (dropdown !== ev.target && !dropdown.contains(ev.target)) {
```
rudolfs, line 23: Why do we need these checks?
It seems to work correctly even if I remove them.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -1,19 +1,132 @@
...
+      hideDropdown();
+    }
+  };
 
   const dispatch = createEventDispatcher();
-  const select = () => dispatch("select", revision);
+  const selectRevision = (ev, rev) => {
```
rudolfs, line 29: We can omit passing the event if it's not used.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -1,19 +1,132 @@
...
+          title={repo.identity.metadata.handle}
+          avatarFallback={repo.identity.avatarFallback}
+          size="small"
+          variant="user" />
+      </div>
+      <ul>
+        {#each repo.branches as branch}
+          <li
+            class="branch"
+            on:click|stopPropagation={ev => selectRevision(ev, branch)}>
```
rudolfs, line 122: Should we not also pass the user here?

cloudhead: Yeah, although we won't be using it until we actually implement the routing etc. based on the user/repo.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
cloudhead: If you click inside the dropdown when expanded, eg. on the user handle, it will hide the dropdown.

cloudhead:
> However, the different border thickness might be tricky to get right, so I wonder if we want to do that.

We can do this with a `box-shadow` without pushing the element I think.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
rudolfs: I see, I was just clicking on the branch and `on:click|stopPropagation` was making sure that worked. Didn't think about clicking around elsewhere in the menu :)

rudolfs:
> We can do this with a box-shadow without pushing the element I think.

I'm actually facing the same problem and `box-shadow` was my first thought, but you'll need box shadow for the menu itself elevation, so it might not work.

cloudhead:
> > We can do this with a box-shadow without pushing the element I think.
> 
> I'm actually facing the same problem and `box-shadow` was my first thought, but at least in my case I need box shadow for the menu elevation, so it might not work. Give it a try, I'll steal your solution then ; P

You can have multiple ones :wink:

xla:
> You can have multiple ones wink

mind blown

rudolfs:
> You can have multiple ones 😉
🤯 🤯 🤯

rudolfs:
I still wonder if we shouldn't set `user-select: none` on all button-type components, like the whole menu in this case. I feel what often happens is people (probably just me) double-click by accident which selects the contents. To me seeing stuff getting selected in a "system-widget" when clicking around wildly seems wrong:

double-click

So far we've disabled this on all clickable components:
```
ui/DesignSystem/Primitive/Avatar.svelte
88:    user-select: none;

ui/DesignSystem/Component/RadioOption.svelte
35:    user-select: none;

ui/DesignSystem/Component/Sidebar.svelte
90:    user-select: none;

ui/DesignSystem/Component/Notification.svelte
53:    user-select: none;

ui/DesignSystem/Component/Transaction/Accordion.svelte
91:    user-select: none;

ui/DesignSystem/Component/SourceBrowser/FileSource.svelte
46:    user-select: none;

ui/DesignSystem/Primitive/Input/Checkbox.svelte
9:    user-select: none;

ui/DesignSystem/Component/SourceBrowser/Folder.svelte
34:    user-select: none;
```

cloudhead:
Yeah I don't mind too much either way, as long as there's *some* place where I can select branch names to copy/paste them. As a counter example, file names are selectable in the file tree.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -1,19 +1,142 @@
...
+  const showDropdown = () => {
+    expanded = true;
+  };
+
+  const hideDropdown = () => {
+    expanded = false;
+  };
+
+  const handleClick = ev => {
+    // Any click *outside* the dropdown should hide hte dropdown.
```
rudolfs, line 23: I feel like my review starts hitting diminishing returns ; P

```suggestion
    // Any click *outside* the dropdown should hide the dropdown.
```

rudolfs:
Oh, yeah, you'll have to adjust the tests too:
https://buildkite.com/monadic/radicle-upstream/builds/1809#b70e440c-8b40-4545-a28d-465871739064

rudolfs:
@cloudhead had another look at the specs, and you're right, more effort will be required to test the new functionality. Especially because we can't select tags anymore from the revision-selector.

Could you mark the tests with `it.skip("...` and somewhere in your issues track that we'll have to adjust the tests once we have all functionality wired up?

PR #296, opened 2020-04-20
feat: source browsing in the new API

xla:
Follow-up to #293


proxy/src/coco.rs
```diff
@@ -178,21 +178,30 @@ pub struct TreeEntry {
...
+) -> Result<Blob, error::Error> {
     let project_id = project::ProjectId::from_str(id)?;
     let project = project::Project::open(paths, &project_id)?;
+    let meta = project::Project::show(paths, &project_id)?;
 
     let mut browser = match project {
         project::Project::Git(git_project) => git_project.browser()?,
     };
 
+    let path = maybe_path.unwrap_or("".to_string());
```
cloudhead, line 196: I think you can `unwrap_or_default()` here :tipping_hand_man:

ui/src/source.ts
```diff
@@ -0,0 +1,161 @@
...
+      currentRevisionStore.update(() => msg.revision);
+      break;
+
+    case Kind.UpdatePath:
+      currentPathStore.update(() => msg.path)
+      objectStore.loading();
+
+      switch (msg.type) {
+        case ObjectType.Blob:
+          console.log("FETCH BLOB");
```
cloudhead, line 109: Can we kill these log statements?

proxy/src/http/source.rs
```diff
@@ -0,0 +1,1054 @@
...
+use std::sync::Arc;
+use tokio::sync::RwLock;
+use warp::document::{self, ToDocumentedType};
+use warp::{path, Filter, Rejection, Reply};
+
+use crate::coco;
+
+/// Prefixed filters.
+pub fn routes(
+    paths: Arc<RwLock<Paths>>,
```
cloudhead, line 15: Why is this in a lock? Are we writing to it?

ui/DesignSystem/Component/SourceBrowser.svelte
```diff
@@ -1,86 +0,0 @@
-<script>
```
rudolfs, line 1: Any specific reason why we can't keep it contained in a component? We'll have the same source browser functionality in orgs as well.

xla: I wanted to contain any data access under the screen hierarchy. In what context will source browsing done for orgs? Only projects have repos, right?

proxy/src/http/source.rs
xla: Esp. during testing we reset/replace the path with a new temp location, to nuke the states.

ui/DesignSystem/Component/SourceBrowser.svelte
rudolfs: Right, yeah it'll be only for projects. I remember there was a point where we had a distinct project landing page (which shows the readme and the source browser) and then a separate "Source" tab with only the source browsing. Right now we don't have this in the designs anymore, so maybe we won't need it.

xla:
> why is everything so slow all of a sudden?

Have not found the source yet, I suspect the issue stems from mounting both REST and GraphQL, wanna sit this out until graphql is removed and see where we at, before diving into profiling.

PR #298, opened 2020-04-20
feat(ui): commit changeset UI

cloudhead:
Progress on #274 

This is the UI rendered from mock data. It was suggested that I get this merged before hooking up the backend, to have shorter lived PRs.

2020-04-20-170621_1278x1414_scrot


ui/Screen/Project/Commit.svelte
```diff
@@ -152,4 +466,45 @@
...
+      <span class="additions">
+        {commit.changeset.summary.additions} addition(s)
+      </span>
+      and
+      <span class="deletions">
+        {commit.changeset.summary.deletions} deletion(s)
+      </span>
+    </div>
+    {#each commit.changeset.files as file}
+      <article class="changeset-file">
```
xla, line 481: It would be helpful to make this into a component for readability and nesting.

ui/Screen/Project/Commit.svelte
```diff
@@ -44,6 +44,232 @@
 
       // TODO(cloudhead): Fetch branch from backend.
       commit.branch = "master";
+      commit.changeset = {
```
xla, line 47: Is this data shaped on something? Maybe we can link it here if so.

ui/Screen/Project/Commit.svelte
```diff
@@ -87,6 +313,94 @@
     font-family: var(--typeface-mono-regular);
   }
 
+  .changeset-summary {
+    margin-top: 2rem;
```
rudolfs, line 317: So far we've been used `px` everywhere, I think switching to relative spacing could be beneficial to make the whole styling a bit more flexible, but we should decide on what's the best way and then do one big change across the whole code base.

If it's too much of a hassle to change it for this PR, would be at least nice to have an issue documenting on how we want to move forward with spacings.

ui/Screen/Project/Commit.svelte
rudolfs, line 47: You could move this mock out into a separate file (maybe `commitMocks.js`?), we already have one like that: https://github.com/radicle-dev/radicle-upstream/blob/master/ui/lib/orgMocks.js

ui/Screen/Project/Commit.svelte
cloudhead: I would be in favor of moving to relative spacing, since that is how the app is designed, and it makes it easy to catch mistakes eg. spacing that doesn't fit the system. So I'd rather open an issue for us to move to `rem`, than change this PR and the Commit one too, only to change it back later.

cloudhead:
> Interesting use of html tags like main, article, header, etc. Do you have a good heuristic on what to use when? So far we've mostly stuck to classics like <div>, <span> and <table>. I wonder if there's a good pattern that we could document, so we have more consistency across the whole codebase.

Yeah I tend to avoid `divs` when I'm structuring content that has meaning. Usually:
* `<article>`: a document, page, object, file, etc.
* `<header>`: the header of an article
* `<main>`: the body of an article

ui/Screen/Project/Commit.svelte
rudolfs: Yeah, makes sense, looking forward to the issue.

ui/Screen/Project/Commit.svelte
cloudhead: It's what I could come up with which makes the frontend easy to build with not too much logic. Unless there's a good reason not to, I would say that we transform whatever `surf` gives us into this format. It's also probably a good idea to discuss this with `surf` to nudge them in this direction, or have them convince us otherwise.

rudolfs:
> Yeah I tend to avoid divs when I'm structuring content that has meaning.

Sounds like a good practice! Wanna add some guidelines under a "Markup" section in: https://github.com/radicle-dev/radicle-upstream/blob/master/DEVELOPMENT.md#design-system?

ui/Screen/Project/Commit.svelte
rudolfs: I see @xla moved it into the store, that works for me as well.

ui/Screen/Project/Commit.svelte
cloudhead: https://github.com/radicle-dev/radicle-upstream/issues/303

Issue #246, opened 2020-03-23
Improve dropdown component


rudolfs:
Now that we're actually using the `AdditionalActionsDropdown` component, I noticed that it could be improved:

1) when there are multiple dropdown components on one page, clicking any `...` menu button should close any other open dropdown modals;
2) there seems to be an issue with stacking: if a dropdown menu button comes after another dropdown it will be higher in the Z stack and will appear to be on top of the second dropdown menu
3) it should be possible to specify the menu button background color for the hover state via props (css vars?), otherwise the button blends into the underlying background if the underlying background color matches the hover state background color; alternatively the hover state could be an alpha, so it works on any underlaying background.

<img width="491" alt="Screenshot 2020-03-23 at 13 24 38" src="https://user-images.githubusercontent.com/158411/77316793-54eadc80-6d0a-11ea-8d5e-745bf9a92d97.png">

<img width="317" alt="Screenshot 2020-03-23 at 13 37 18" src="https://user-images.githubusercontent.com/158411/77317742-0dfde680-6d0c-11ea-8d31-3a7a5b5d2f06.png">

When this is fixed, remove any related `force: true` statements from the specs:
```
cypress/integration/user_registration.spec.js
8:        force: true

cypress/integration/project_creation.spec.js
12:        force: true
22:        force: true
```

rudolfs:
#299 seems to have fixed only the stacking problem for the top-bar, we still need a general solution that works even for components that are siblings.

<img width="416" alt="Screenshot 2020-04-23 at 11 22 55" src="https://user-images.githubusercontent.com/158411/80082765-189edc00-8555-11ea-9e1c-d6c050c29c37.png">

rudolfs:
> 3. It should be possible to specify the menu button background color for the hover state via props (css vars?), otherwise the button blends into the underlying background if the underlying background color matches the hover state background color; alternatively the hover state could be an alpha, so it works on any underlaying background.

Will be addressed by: #292

PR #299, opened 2020-04-21
Fix stacking of dropdowns

sarahscott:
Because the dropdowns have `position:relative`, they create a new stacking context. By adding a `z-index` to the `Topbar`, we ensure its child dropdown stays on top:

Also, I removed the `svelte-click-outside` dependency so we'll have an easier time debugging and modifying the behavior of the `AdditionalActionsDropdown`.

fix-zindex

Fixes part of #246

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,6 @@
     "sirv-cli": "^0.4.5",
     "standard-version": "^7.1.0",
     "svelte": "^3.20.1",
-    "svelte-click-outside": "https://github.com/radicle-dev/svelte-click-outside#10e47e6359f943d45b72c2f3fc7ccc8036b60fac",
     "svelte-spa-router": "^2.1.0",
     "typescript": "3.8.3",
     "wait-on": "^4.0.1"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -73,7 +73,6 @@
     "sirv-cli": "^0.4.5",
     "standard-version": "^7.1.0",
     "svelte": "^3.20.1",
-    "svelte-click-outside": "https://github.com/radicle-dev/svelte-click-outside#10e47e6359f943d45b72c2f3fc7ccc8036b60fac",
```
xla, line 76: Should we nuke our fork?

ui/DesignSystem/Component/Topbar.svelte
```diff
@@ -13,6 +13,7 @@
     height: var(--topbar-height);
     left: var(--sidebar-width);
     border-bottom: 1px solid var(--color-foreground-level-3);
+    z-index: 1;
```
xla, line 16: Hrhr, always fun.

package.json
sarahscott: Yes

PR #300, opened 2020-04-21
feat(proxy): org API endpoints

MeBrei:
Implements the get and register org endpoints in the rest api and connects them to the registry as part of #277

### DIFF
--- a/proxy/src/http/org.rs
+++ b/proxy/src/http/org.rs
@@ -142,7 +142,7 @@ impl ToDocumentedType for registry::Org {
                 .example("monadic"),
         );
 
-        document::DocumentedType::from(properties).description("Unique org")
+        document::DocumentedType::from(properties).description("Org")
     }
 }
 
--- a/proxy/src/http/org.rs
+++ b/proxy/src/http/org.rs
@@ -1,0 +1,256 @@
+//! Endpoints for Org.
+
+use serde::ser::SerializeStruct as _;
+use serde::{Deserialize, Serialize, Serializer};
+use std::sync::Arc;
+use tokio::sync::RwLock;
+use warp::document::{self, ToDocumentedType};
+use warp::{path, Filter, Rejection, Reply};
+
+use crate::avatar;
+use crate::notification;
+use crate::registry;
+
+/// The org with identifier and avatar
+pub struct Org {
+    /// The unique identifier of the org
+    id: String,
+    /// Generated fallback avatar 
+    avatar_fallback: avatar::Avatar
+}
+
+/// Combination of all org routes.
+pub fn filters(
+    registry: Arc<RwLock<registry::Registry>>,
+    subscriptions: notification::Subscriptions,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    get_filter().or(register_filter(registry, subscriptions))
+}
+
+/// POST /orgs/register
+fn register_filter(
+    registry: Arc<RwLock<registry::Registry>>,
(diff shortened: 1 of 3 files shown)


proxy/src/http/org.rs
```diff
@@ -0,0 +1,256 @@
...
+/// POST /orgs/register
+fn register_filter(
+    registry: Arc<RwLock<registry::Registry>>,
+    subscriptions: notification::Subscriptions,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    path!("orgs" / "register")
+        .and(warp::post())
+        .and(super::with_registry(registry))
+        .and(warp::body::json())
+        .and(super::with_subscriptions(subscriptions))
```
xla, line 39: I tend to order the params for the handler with datastores first and inputs, params last. So would swap line 38 and 39.

proxy/src/http/org.rs
```diff
@@ -0,0 +1,256 @@
...
+use tokio::sync::RwLock;
+use warp::document::{self, ToDocumentedType};
+use warp::{path, Filter, Rejection, Reply};
+
+use crate::avatar;
+use crate::notification;
+use crate::registry;
+
+/// The org with identifier and avatar
+pub struct Org {
```
xla, line 15: This should live in `registry.rs`, or alternatively in a seperate `org.rs`. Similar to how we have our own identity, this seems to be upstreams understanding of an org.

proxy/src/http/org.rs
```diff
@@ -0,0 +1,256 @@
...
+
+/// Combination of all org routes.
+pub fn filters(
+    registry: Arc<RwLock<registry::Registry>>,
+    subscriptions: notification::Subscriptions,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    get_filter().or(register_filter(registry, subscriptions))
+}
+
+/// POST /orgs/register
```
xla, line 30: The only real write operation we have on top-level orgs is the registration, so synonymous with creation. Would say we can just do `POST /orgs` for that endpoint.

xla:
@MeBrei Org retrieval should be as trivial as the `get_user` method. This is what we likely wanna use in the registry client;

https://github.com/radicle-dev/radicle-registry/blob/1ecf004bb1e27f152e73716b1b316e0db878b0fd/client/src/interface.rs#L129

MeBrei:
> @MeBrei Org retrieval should be as trivial as the `get_user` method. This is what we likely wanna use in the registry client; https://github.com/radicle-dev/radicle-registry/blob/1ecf004bb1e27f152e73716b1b316e0db878b0fd/client/src/interface.rs#L129

Yes, that part I already have I think. Not sure yet, how I can test it (ie test the querying of an existing user) and I need to add the avatar to the response.

proxy/src/http/org.rs
MeBrei: I moved it to `registry.rs`

proxy/src/http/org.rs
```diff
@@ -0,0 +1,258 @@
...
+        let mut properties = std::collections::HashMap::with_capacity(2);
+        properties.insert("avatarFallback".into(), avatar::Avatar::document());
+        properties.insert(
+            "id".into(),
+            document::string()
+                .description("The id of the org")
+                .example("monadic"),
+        );
+
+        document::DocumentedType::from(properties).description("Unique org")
```
xla, line 145:
```suggestion
        document::DocumentedType::from(properties).description("Org")
```

Used the singular type everywhere else for now.

PR #301, opened 2020-04-21
fix(ui): restyle notification banners

brandonhaslegs:
Aligned notifications with new style.

image


ui/DesignSystem/Component/Notification.svelte
```diff
@@ -10,58 +10,55 @@
...
+    border-color: var(--color-negative-level-2);
+  }
+
+  .close.info {
+    border-color: var(--color-foreground-level-6);
   }
 </style>
 
 <div class={`notification ${level}`} {style}>
   <svelte:component this={icon} style="margin: 8px; height: 24px" />
```
brandonhaslegs, line 55: @MeBrei How can I change the icon depending on if it's an error or a notification?

MeBrei: Currently the icon is passed down by `export let icon = Icon.Info;` in `Notification.svelte`, but I think it is actually not used.
You could remove it and set the icon depending on the `level` input. Something like:
```js
const icon = level === "info" ? Icon.Info : Icon.Important;
```

juliendonck: Fixed this myself

juliendonck:
@brandonhaslegs added the icon picker myself :/

ui/DesignSystem/Component/Notification.svelte
```diff
@@ -10,58 +9,58 @@
...
+    border-color: var(--color-foreground-level-6);
   }
 </style>
 
 <div class={`notification ${level}`} {style}>
-  <svelte:component this={icon} style="margin: 8px; height: 24px" />
-
-  <Title>
+  {#if level == 'info'}
+    <Icon.Info style="margin: 8px; height: 24px" />
```
juliendonck, line 55: @MeBrei Ended up using Icons like this here. Saw it somewhere else. I don't know what the difference is, but is this fine?

MeBrei: I think this is a good solution as you have to pass the variant to one of the icons.

juliendonck:
@brandonhaslegs also added the "no icon" variants. Check it out and let me know. I feel like this could be ready for review.

ui/Screen/DesignSystemGuide.svelte
```diff
@@ -200,7 +200,7 @@
     </TypographySwatch>
 
     <TypographySwatch title="regular Numeric">
-      <Numeric>0123456789</Numeric>
+      <Numeric>01234567890</Numeric>
```
rudolfs, line 203: I remember we changed this just as an exercise...
```suggestion
      <Numeric>0123456789</Numeric>
```

ui/DesignSystem/Primitive/Icon/Important.svelte
```diff
@@ -1,5 +1,6 @@
 <script>
   export let style = null;
+  export let small = false;
```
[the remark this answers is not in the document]
rudolfs: Hm, having variants for icons is an interesting idea! I wonder why I didn't do it from the beginning on?

I think I prefer having variants over different individual files of icon variants, so we should probably open another issue to make this consistent across all icons. Here's a list of which ones should be adapted:
```
ArrowDown.svelte
ArrowUp.svelte
Carret.svelte
CarretBig.svelte
Check.svelte
CheckCircle.svelte
Cross.svelte
CrossBig.svelte
Ellipse.svelte
EllipseBig.svelte
Search.svelte
SearchSmall.svelte
```

ui/DesignSystem/Primitive/Icon/Important.svelte
```diff
@@ -9,23 +10,41 @@
...
-    15 12.0001 15C11.6023 15 11.2208 15.158 10.9395 15.4393C10.6582 15.7206
-    10.5001 16.1022 10.5001 16.5C10.5001 16.8978 10.6582 17.2794 10.9395
-    17.5607ZM11.0001 12.1C11.1201 13.3 12.8701 13.3 13.0001 12.1L13.5001
-    7.10001C13.5142 6.96053 13.4987 6.81967 13.4548 6.68655C13.4108 6.55344
-    13.3394 6.43105 13.2451 6.32732C13.1508 6.2236 13.0357 6.14086 12.9074
-    6.08447C12.7791 6.02808 12.6403 5.9993 12.5001 6.00001H11.5001C11.36 5.9993
-    11.2212 6.02808 11.0929 6.08447C10.9645 6.14086 10.8495 6.2236 10.7552
-    6.32732C10.6609 6.43105 10.5894 6.55344 10.5455 6.68655C10.5016 6.81967
-    10.4861 6.96053 10.5001 7.10001L11.0001 12.1Z" />
+  {#if small}
```
rudolfs, line 13:
```suggestion
  {#if variant === "small"}
```

ui/DesignSystem/Component/Notification.svelte
```diff
@@ -1,67 +1,72 @@
 <script>
-  import { Icon, Title } from "../Primitive";
+  import { Icon, Text, Title } from "../Primitive";
 
-  export let icon = Icon.Info;
   export let style = null;
-
   export let level = "info";
+  export let noIcon = false;
```
rudolfs, line 6: I think avoiding a double negative makes it more readable and also make the "action" configurable (as per styleguide):
```suggestion
  export let showIcon = true;
  export let action = "Close";
```

You'd have to also adjust the `if` branch below for this to work and also use the `action` variable in the template like so:
```
<Title>{action}</Title>
```

ui/DesignSystem/Component/Notification.svelte
```diff
@@ -1,67 +1,72 @@
...
+    color: var(--color-background);
+    background-color: var(--color-negative);
   }
 
   .notification.error :global(svg) {
-    fill: var(--color-negative-level-6);
+    fill: var(--color-background);
   }
 
   .close {
```
rudolfs, line 35: If you chose to make the "action" configurable, I'd rename all the `close` classes to `action`.

ui/DesignSystem/Component/Notification.svelte
```diff
@@ -1,67 +1,72 @@
...
 </style>
 
 <div class={`notification ${level}`} {style}>
-  <svelte:component this={icon} style="margin: 8px; height: 24px" />
-
-  <Title>
-    <slot />
-  </Title>
+  {#if noIcon == false}
+    {#if level == 'info'}
```
rudolfs, line 55:
```suggestion
    {#if level === 'info'}
```

ui/DesignSystem/Component/Notification.svelte
```diff
@@ -1,67 +1,72 @@
...
+      <Icon.Info style="margin: 8px; height: 24px" />
+    {:else}
+      <Icon.Important small style="margin: 8px; height: 24px" />
```
rudolfs, line 58:
```suggestion
      <Icon.Important variant="small" style="margin: 8px; height: 24px" />
```

ui/DesignSystem/Component/Notification.svelte
MeBrei: @rudolfs We wanted to keep this PR simple for now and have the same functionality as before. Adding custom actions here will need some more consideration as we will have to find a way of passing the actual actions to the notification faucet, eg repeat a search query.

ui/DesignSystem/Primitive/Icon/Important.svelte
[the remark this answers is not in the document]
juliendonck: Cool. I'll do that.

ui/DesignSystem/Component/Notification.svelte
juliendonck: Leaving the "action" stuff out for now. But address the double negative.

ui/DesignSystem/Primitive/Icon/Important.svelte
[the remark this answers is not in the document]
rudolfs: @juliendonck keep in mind that you'll also have to update any instance in the rest of the code, if you change the icons.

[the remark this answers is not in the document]
juliendonck: Created issue #302

ui/DesignSystem/Component/Notification.svelte
juliendonck: For a different time ;)

rudolfs:
The width of the action button hasn't been fixed yet. Also, the default notification probably should have an icon, no?
Screenshot 2020-04-22 at 14 41 04

ui/DesignSystem/Component/Notification.svelte
```diff
@@ -1,67 +1,69 @@
 <script>
-  import { Icon, Title } from "../Primitive";
+  import { Icon, Text, Title } from "../Primitive";
 
-  export let icon = Icon.Info;
   export let style = null;
-
-  export let level = "info";
+  export let variant = "info"; // info | error
```
MeBrei, line 5: You will have to change the prop for `Notification` in `NotificationFaucet` as well.

juliendonck: Ok, took a stab at it. I also changed  ` level: message.level` to `variant: message.variant` I don't know if that makes sense or if the message is referenced somewhere else as well.

ui/DesignSystem/Component/NotificationFaucet.svelte
```diff
@@ -19,7 +19,7 @@
     const id = Math.random();
     messages = [
       ...messages,
-      { id: id, message: message.text, level: message.level }
+      { id: id, message: message.text, variant: message.variant }
```
MeBrei, line 22: @juliendonck I would keep it as `level` here and only change it in the prop for `Notification`, otherwise you will have to change it where ever a notification is created. I also think `level` in this part of the code makes more sense

PR #305, opened 2020-04-23
feat: transactions through REST API

xla:
Follow-up to #293 
Related to #225


ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -1,60 +1,46 @@
...
 </script>
 
 <Caption style="color: var(--color-foreground-level-6); margin-bottom: 16px">
   Your transaction
 </Caption>
-<Row dataCy="summary" variant={transaction.id ? 'top' : 'single'}>
+<Row dataCy="summary" variant={tx.id ? 'top' : 'single'}>
   <div slot="left">
-    <Title>{transaction.message}</Title>
+    <Title>{transaction.formatMessage(tx.messages[0])}</Title>
```
MeBrei, line 24: I find it confusing to import the helper functions as `transaction`. Below, you imported the functions directly, which I would prefer for readability.

xla: It's not importing them as `transaction`, it is to have them qualified. `formatMessage` by itself doesn't communicate as nicely as `transaction.formatMessage`, which gives  a lot of extract content just by module prefix.

ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -66,21 +52,17 @@
   Transaction cost
 </Caption>
 
-{#if transaction.stake}
```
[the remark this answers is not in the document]
xla: We can address it when introduce txs without it.

ui/DesignSystem/Component/Transaction/Center.svelte
```diff
@@ -1,30 +1,17 @@
...
-      case PROJECT_REGISTRATION:
-        return "Project registration";
-    }
-  };
-
-  // TODO(merle): Use actual data
-  const formatTransactions = transactions => {
-    return transactions.map(transaction => {
+  // TODO(merle): Use actual data.
+  const formatTransactions = txs => {
```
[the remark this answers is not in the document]
xla: This is an inline variable, I don't think the same rules as for public interfaces apply. Furthermore `tx` is a very ubiquitous abbreviation for it. We have a bit of the problem that we overload it.

ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -1,60 +1,46 @@
...
-  //     name: "name of the owner of the paying wallet"
-  //     avatar: "avatar of the owner (optional)",
-  //     avatarFallback: "avatar fallback of the owner",
-  //     imageUrl: "avatar url of the owner (optional)"
-  //   }
-  // }
-  export let transaction = null;
-
-  const feePosition = transaction.stake ? "middle" : "top";
+  export let tx = null;
```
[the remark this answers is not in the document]
xla: As this determines the public interface of the component I think that point is very valid.

ui/DesignSystem/Component/Transaction.svelte
MeBrei: I agree with the context argument, but if we use `transaction` as a prop as well, it seems confusing to have the same name for 2 different things.

ui/DesignSystem/Component/Transaction/Center.svelte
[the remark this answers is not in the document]
MeBrei: well, I guess for inline it is ok, but we should keep it consistent and have a clear and simple guideline for when to use what.
@rudolfs What do you think? I remember you commenting about this in the past.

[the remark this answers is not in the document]
rudolfs: For consistency's sake we have tried to always spell things out and don't use abbreviations.
```
radicle-upstream/ui » g transaction | wc -l
     145
radicle-upstream/ui » g tx | wc -l
       0
```

yarn.lock
```diff
@@ -1,4 +1,4 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+a THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
rudolfs, line 1: Um, what?

xla: lol, unintended.

ui/DesignSystem/Component/Transaction.svelte
xla: Already addressed.

ui/DesignSystem/Component/Transaction/Center.svelte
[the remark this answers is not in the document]
xla: I addressed the prop of the component. There is no gain spelling it out in this function.

ui/DesignSystem/Component/Transaction.svelte
sarahscott: FWIW even though this is already merged, I agree with spelling things out 😄

PR #309, opened 2020-04-23
feat(proxy): list all orgs by member

MeBrei:
- [x] list all orgs
- [x] clean up code, please clippy & check error handling
- [x] filter list by provided user (for now it is filtered by the public key, while waiting for a [registry change](https://github.com/radicle-dev/radicle-registry/issues/396))

Part of #277


proxy/src/http/org.rs
```diff
@@ -255,4 +293,43 @@ mod test {
...
+            radicle_registry_client::Client::new_emulator(),
+        )));
+        let subscriptions = notification::Subscriptions::default();
+        let api = super::filters(
+            Arc::<RwLock<registry::Registry>>::clone(&registry),
+            subscriptions,
+        );
+
+        // Register the org
+        request()
```
xla, line 309: Instead of calling the API and therefore depend on the correctness of other filters it would be more straightforward to call methods on `registry` directly to set up the state we need.

xla: Esp. as we don't inspect if the request was successful. Calling the methods directly will give use results to panic on.

proxy/src/registry.rs
```diff
@@ -551,6 +584,29 @@ mod tests {
...
+        // Register the org
+        let org_id = OrgId::try_from("monadic").unwrap();
+        let registration = registry
+            .register_org(&alice, org_id.clone().into(), 10)
+            .await;
+        assert!(registration.is_ok());
+
+        // List the orgs
+        let orgs = registry
+            .list_orgs(&ed25519::Public::from(alice.clone()))
```
xla, line 603: Can't we just call `public` on `alice` to get the pub key?

proxy/src/http/org.rs
```diff
@@ -74,6 +77,27 @@ fn get_filter(
         .and_then(handler::get)
 }
 
+/// GET /orgs
+fn list_filter(
+    registry: Arc<RwLock<registry::Registry>>,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    path("orgs")
+        .and(warp::get())
+        .and(super::with_registry(registry))
```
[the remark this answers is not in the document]
MeBrei: @xla There is currently no view planned that requires a list of orgs for a different user than myself. Not sure if it makes sense to "plan ahead" as we don't know what endpoints we would need for that.

proxy/src/registry.rs
MeBrei: b0d295c33b8b3f4a147febaa3eae77c977765302

proxy/src/http/org.rs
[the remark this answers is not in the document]
xla: How do you get the info of who the current user is if not through a query param?

proxy/src/http/org.rs
MeBrei: 0e5a59c

MeBrei:
The remaining local clippy errors are all due to the use of `unwrap()` 

```
error: could not compile `proxy`.
warning: build failed, waiting for other jobs to finish...
error: used `unwrap()` on `an Option` value
   --> src/http/org.rs:247:18
    |
247 |         let tx = txs.first().unwrap();
    |                  ^^^^^^^^^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> src/lib.rs:10:5
    |
10  |     clippy::option_unwrap_used,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: if you don't want to handle the `None` case gracefully, consider using `expect()` to provide a better panic message
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_unwrap_used

error: used `unwrap()` on `an Option` value
   --> src/registry.rs:364:23
    |
364 |             orgs.push(self.client.get_org(org_id).await.unwrap().unwrap())
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: if you don't want to handle the `None` case gracefully, consider using `expect()` to provide a better panic message
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_unwrap_used

error: used `unwrap()` on `a Result` value
   --> src/registry.rs:364:23
    |
364 |             orgs.push(self.client.get_org(org_id).await.unwrap().unwrap())
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> src/lib.rs:11:5
    |
11  |     clippy::result_unwrap_used,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: if you don't want to handle the `Err` case gracefully, consider using `expect()` to provide a better panic message
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_unwrap_used

error: aborting due to 3 previous errors

error: could not compile `proxy`.
```

proxy/src/http/org.rs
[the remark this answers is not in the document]
MeBrei: True, if we switch to filtering the members by users then. Ok, I will add it.

proxy/src/http/org.rs
```diff
@@ -17,10 +17,13 @@ pub fn filters(
     subscriptions: notification::Subscriptions,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     get_filter(Arc::<RwLock<registry::Registry>>::clone(&registry))
+        .or(list_filter(Arc::<RwLock<registry::Registry>>::clone(
```
cloudhead, line 20: FYI you don't need to specify the full type here (and in other places), you can do `Arc::clone(&registry)`

proxy/src/registry.rs
```diff
@@ -348,6 +348,37 @@ impl Registry {
...
+    /// # Errors
+    ///
+    /// Will return `Err` if a protocol error occurs.
+    pub async fn list_orgs(&self, _user_id: String) -> Result<Vec<Org>, error::Error> {
+        // TODO(merle): Remove temp_public_key once members are returned as user ids
+        let temp_public_key = ed25519::Pair::from_legacy_string("//Alice", None).public();
+        let org_ids = self.client.list_orgs().await?.into_iter();
+        let mut orgs = Vec::new();
+        for org_id in org_ids {
+            orgs.push(self.client.get_org(org_id).await?.expect("Get org"));
```
cloudhead, line 362: Strange.. does the client return a nested `Result`?

xla: @cloudhead Yes it does, they are for different stages of the transaction lifecycle.

proxy/src/http/org.rs
```diff
@@ -74,6 +77,31 @@ fn get_filter(
         .and_then(handler::get)
 }
 
+/// `GET /orgs`
```
xla, line 80: Thought about this a bit more. Besides that it is per spec not expected for a server to parse any body for GET request, I think this endpoint should be modeled a bit differently. First I thought along listing with a `by` constraints, i.e. member id. Which is usually expressed with query parameters. Which then begs the question what if the parameter is not passed, multiple permutations are passed for different constraints and so on.

What we actually want is the list of a orgs for a specific user. That user must exist, therefore I think a better semantic would be to have this endpoint in `user.rs`, under `/users/<id>/orgs` and make it unambiguous and easy to use. This will clarify semantics for it, won't require any extensive or complicated documentation on how to satisify the use-case.

proxy/src/http/org.rs
```diff
@@ -168,7 +207,33 @@ impl ToDocumentedType for RegisterInput {
...
+                .description("ID of the user")
+                .example("alice"),
+        );
+
+        document::DocumentedType::from(properties).description("Input for org list")
+    }
+}
+
+#[allow(
+    clippy::non_ascii_literal,
```
xla, line 233: This is only necessary if we have literal unicode like emojis in our test assertions.

proxy/src/http/org.rs
MeBrei: Thanks! Replaced it.

proxy/src/http/org.rs
MeBrei: 👍 Moved!

proxy/src/http/org.rs
MeBrei: 👍 Removed

proxy/src/http/user.rs
```diff
@@ -74,6 +79,31 @@ fn register_filter(
...
+fn list_orgs_filter(
+    registry: Arc<RwLock<registry::Registry>>,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    warp::get()
+        .and(super::with_registry(registry))
+        .and(document::param::<String>(
+            "handle",
+            "ID of the user to query for",
+        ))
+        .and(path("orgs"))
```
xla, line 92: Can we stay consistent with the other filters where path is the first filter?

proxy/src/http/user.rs
```diff
@@ -74,6 +79,31 @@ fn register_filter(
...
+        .and(document::document(document::description(
+            "List all orgs the user is a member of",
+        )))
+        .and(document::document(document::tag("User")))
+        .and(document::document(
+            document::response(
+                200,
+                document::body(registry::Org::document()).mime("application/json"),
```
xla, line 100:
```suggestion
                document::body(document::array(registry::Org::document())).mime("application/json"),
```

PR #310, opened 2020-04-24
refactor(ui): remote component

xla:
In order to reduce the boilerplate and leakage of internals this change introduces a new component which given a remote store exposes a svelte native way to render content for different states. The slot based approach reduces a lot of the deep object referencing and inline if statements.


ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -162,11 +161,11 @@
       class="item indicator"
       data-cy="profile"
       class:active={path.active(path.profile(), $location, true)}>
-      {#if $session.status === remote.Status.Success}
+      <Remote store={session} let:data>
```
sarahscott, line 164: Interesting! Reading up on the `let` directive now

xla:
> Creative and svelty use of svelte! My one concern is that this will lead to `<Remote />` wrappers all over the place, and that it might feel cluttered in the markup. But then the `{#if status === success} {:else if status === error}` pattern also adds quite a bit to the markup, so maybe this is indeed cleaner.

I think so too, the `<Remote />` pattern will reduce the markup by allowing the default slot to be the success, it leans it better to be embedded in the html part and allows for more flexibility around naming the data part.

ui/Screen/Profile/Projects.svelte
```diff
@@ -1,5 +1,26 @@
...
+    projectNameStore.set(project.metadata.name);
+    push(path.projectSource(project.id));
+  };
 </script>
 
-<ProjectList />
+<Remote {store} let:data={projects}>
+  {#if projects.length > 0}
+    <ProjectList {projects} on:select={select} />
+  {:else}{push(path.profileOnboard())}{/if}
```
rudolfs, line 21: I wonder if prettier did this?

xla: Think so, buuuh prettier. I think we should put some love in the svelte plugin.

PR #312, opened 2020-04-27
refactor: Migrate outstanding endpoints to REST

xla:
Follow-up to #250


xla:
> Performance seems to be wack for some reason, maybe it's the underlying protocol, but would be good to investigate.

Yeah performance took a huge hit over the last weeks, will be investigated independently. A hunch is that the long link times are not just coincidence.

rudolfs:
Missed a couple of spots:
```
work/radicle-upstream-2 » g graphql                                                                                                                                                 ‹xla/migrate-endpoints›
graphiql/index.html
47:      var fetchURL = "http://localhost:8080/graphql";
48:      function graphQLFetcher(graphQLParams) {
57:          body: JSON.stringify(graphQLParams)
72:          fetcher: graphQLFetcher

proxy/README.md
4:radicle code collaboration and registry protocols. It provides a GraphQL API to
11:After that the API is served on `http://127.0.0.1:8080/graphql`.

proxy/src/lib.rs
1://! Proxy to serve a specialised `GraphQL` API to radicle-upstream.
```

proxy/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! Proxy to serve a specialised `GraphQL` API to radicle-upstream.
+//! Proxy serving a specialized API to the Upstream UI..
```
rudolfs, line 1:
```suggestion
//! Proxy serving a specialized API to the Upstream UI.
```

PR #314, opened 2020-04-27
fix(ui): timely session fetch

### LINKED ISSUE #313: Regression "register it" link not working
We should start writing tests for the new UI features.
Looks like this got introduced in https://github.com/radicle-dev/radicle-upstream/pull/293.

register_it_broken
register_it_exception

xla:
Fixes #313

### DIFF
--- a/ui/Screen/IdentityCreation.svelte
+++ b/ui/Screen/IdentityCreation.svelte
@@ -43,11 +43,15 @@
         returnToWelcomeStep();
         return;
       case steps.SUCCESS:
-        session.fetch();
         replace(path.profileProjects());
         return;
     }
   };
+
+  // Refetch the session, once we created the identity.
+  $: if (currentStep === steps.SUCCESS) {
+    session.fetch();
+  }
 </script>
 
 <style>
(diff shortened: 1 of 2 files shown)


ui/Screen/IdentityCreation.svelte
```diff
@@ -43,11 +43,15 @@
...
       case steps.SUCCESS:
-        session.fetch();
         replace(path.profileProjects());
         return;
     }
   };
+
+  // Refetch the session, once we created the identity.
+  $: if (currentStep === steps.SUCCESS) {
+    session.fetch();
```
rudolfs, line 53: I guess it's unlikely that the user clicks the register link before this finishes, but is there a way to guarantee that a fetch completes before performing an action? Any other way than using route guards?

xla: It doesn't really matter if the user clicks before this finsihes, as the component that threw the error is guarded by the `Remote` and will only render when the session store is in success state and therefore a avoid the null reference we observed: https://github.com/radicle-dev/radicle-upstream/blob/105a43081a27b3ad1f110df67f80a7befd0a7a89/ui/DesignSystem/Component/Sidebar.svelte#L164-L168

PR #292, opened 2020-04-16
feat(ui): project registration flow

rudolfs:
Partially solves: #285

- [x] wire up all the different entry points for the flow
  - [x] open from profile screen project list context menu
  - [x] open from project topbar context menu
  - [x] open from org topbar context menu
- [x] custom dropdown with text and avatar options
- [x] validations for project selection and project name
- [x] summary has wallet and transaction subject formatting depending on what was chosen
- [x] visual tweaks and fixes for additional actions dropdown menu (closes: #246)
- [x] extracted [elevations](https://www.figma.com/file/z0IYHvWzBKfaAVnGonKaxS/%F0%9F%93%91-Elevation?node-id=0%3A1) into a separate css file
- [x] prevent registering a project before handle is registered
All remaining tasks for this feature are tracked in #285.

reg

PR #315, opened 2020-04-27
fix: style forms to match new design system

juliendonck:
Updated the forms & drop downs to match the design system.

### DIFF
--- a/ui/DesignSystem/Primitive/Input/Dropdown.svelte
+++ b/ui/DesignSystem/Primitive/Input/Dropdown.svelte
@@ -40,8 +40,6 @@
   }
   select:focus {
     outline: none;
-    border: 1px solid
-      var(--focus-outline-color, var(--color-foreground-level-3));
   }
 
   select:disabled {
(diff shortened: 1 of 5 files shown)


juliendonck:
@rudolfs I didn't name my commits properly, but they will get squashed, right? 😇

rudolfs:
@juliendonck yeah, no worries!

btw. I think you shouldn't spend too much time on styling the dropdown, as it will be replaced with a fully custom one in #292.

juliendonck:
@rudolfs I see. But I've noticed that it's still not entirely based on the new design. SHould I just wait for that one to land? Or should I pull that one and fix it?

rudolfs:
@juliendonck yeah, you're right, it is not exactly as per new design. I made it fit the existing design so it would not stick out.

We should aim to make design changes in one go: adjust the component and make sure that it works in all places where it is already used.

I would wait for #292 to be merged before working on the dropdown.

juliendonck:
@rudolfs I guess we can wait for it to be merged, pull master into this one and then fix it here.

rudolfs:
@juliendonck just realised that there will be more work required to get the new dropdown working which is beyond the scope of #292. So I guess we can just style the existing one for now.

juliendonck:
@rudolfs good catch. Removed those.

PR #316, opened 2020-04-27
fix: update buttons to match new design system

juliendonck:
- remove small button size and makes all of them 40px height
- fix all hover & active states
- restyles small overflow button in top bar to match height

### DIFF
--- a/ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
+++ b/ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
@@ -39,8 +39,8 @@
 <style>
   .container {
     position: relative;
-    height: 36px;
-    width: 36px;
+    height: 40px;
+    width: 40px;
   }
 
   button {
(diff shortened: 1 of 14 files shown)


ui/DesignSystem/Primitive/Button.svelte
```diff
@@ -3,54 +3,36 @@
...
-  // big | small
-  export let size = "big";
 
   export let disabled = null;
   export let icon = null;
   export let style = null;
   export let dataCy = null;
 
-  const iconClass = icon ? (size === "big" ? "icon" : "icon-small") : null;
+  const iconClass = icon ? "icon" : null;
```
juliendonck, line 12: @rudolfs wondering if we can simplify this now

ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
```diff
@@ -53,8 +53,15 @@
     outline-style: none;
   }
 
+  button :global(svg) {
```
rudolfs, line 56: We should use a unique classname for the button in this component, something like `.additional-actions-dropdown-button`, otherwise this global style might do something unexpected in case we have a `<button>` tag with an `<svg>` as child somewhere else in the code.

ui/DesignSystem/Primitive/Button.svelte
rudolfs: I don't see a way at the moment.

PR #322, opened 2020-04-29
feat: fetch registered project

xla:
Exposed the ability to fetch registered projects to check for existence. It currently is exposed for orgs, as the user part is waiting for radicle-dev/radicle-registry#389.

### DIFF
--- a/proxy/src/http/org.rs
+++ b/proxy/src/http/org.rs
@@ -238,7 +238,7 @@ impl ToDocumentedType for registry::Project {
             "maybeProjectId".into(),
             document::string()
                 .description("The id project attested in coco")
-                .example("monadic")
+                .example("123abdcd.git")
                 .nullable(true),
         );
 
--- a/proxy/src/http/org.rs
+++ b/proxy/src/http/org.rs
@@ -1,5 +1,4 @@
 //! Endpoints for Org.
-
 use serde::ser::SerializeStruct as _;
 use serde::{Deserialize, Serialize, Serializer};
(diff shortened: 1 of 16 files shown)


proxy/src/http/org.rs
```diff
@@ -145,6 +192,47 @@ impl ToDocumentedType for registry::Org {
...
+            document::string()
+                .description("The id of the org")
+                .example("monadic"),
+        );
+        properties.insert(
+            "maybeProjectId".into(),
+            document::string()
+                .description("The id project attested in coco")
+                .example("monadic")
+                .nullable(true),
```
MeBrei, line 229: The examples for the project name and project id should not be the same as the org id for clarification.

proxy/src/registry.rs
```diff
@@ -43,25 +44,25 @@ pub struct Metadata {
...
     },
 
     /// Issue a user registration for a given handle storing the corresponding identity id.
     UserRegistration {
         /// Globally unique user handle.
-        handle: UserId,
+        handle: Id,
         /// Identity id originated from librad.
-        id: registry::String32,
+        id: Option<String>,
```
MeBrei, line 65: Is it allowed to register a user without this id?

xla: It's not required semantically by the spec

rudolfs:
And another bug, when registering a project in an org the org name and project name are swapped in the response.

Screenshot 2020-04-30 at 10 17 13

PR #325, opened 2020-04-29
feat(ui): implement tracking button

### LINKED ISSUE #260: Implement "Track" toggle switch
We already have a dummy "Track" toggle switch: [TrackToggle.svelte](https://github.com/radicle-dev/radicle-upstream/blob/master/ui/DesignSystem/Component/TrackToggle.svelte).

Implement it according to the UX given in [our Figma designs](https://www.figma.com/proto/3TLScIxjiM93xppRFAIrcF/Track-toggle?node-id=1%3A73&viewport=1145%2C761%2C0.5&scaling=min-zoom).

Screenshot 2020-03-30 at 10 03 26

juliendonck:
Updated the tracking button to have hover/active/tracking states.
Tracking-btn

Closes #260

### DIFF
--- a/ui/DesignSystem/Component/TrackToggle.svelte
+++ b/ui/DesignSystem/Component/TrackToggle.svelte
@@ -1,14 +1,11 @@
 <script>
-  /* TODO(rudolfs): this is a placeholder for now, implement this component
-                    according to UX */
-
   import { Code, Icon, Title } from "../Primitive";
 
   export let style = null;
   export let peerCount = null;
   let hover = false;
   let active = false;
-  let tracking = false;
+  export let tracking = false;
   const enter = () => {
     hover = true;
   };
@@ -73,7 +70,7 @@
     border-bottom-left-radius: 3px;
   }
 
-  .left :global(svg) {
+  .toggle .left :global(svg) {
     fill: var(--color-background);
   }
 
@@ -83,8 +80,8 @@
   .left.active {
     background-color: var(--color-secondary-level-1);
   }
-  .left.tracking :global(svg) {
-    fill: var(--color-forground-level-5);
+  .toggle .left.tracking :global(svg) {
+    fill: var(--color-foreground-level-6);
   }
   .left.tracking {
     background-color: var(--color-foreground-level-3);
--- a/ui/DesignSystem/Component/TrackToggle.svelte
+++ b/ui/DesignSystem/Component/TrackToggle.svelte
@@ -6,41 +6,147 @@
 
   export let style = null;
   export let peerCount = null;
+  let hover = false;
+  let active = false;
+  let tracking = false;
+  const enter = () => {
+    hover = true;
+  };
+  const leave = () => {
+    hover = false;
+  };
+  const down = () => {
+    active = true;
+  };
+  const up = () => {
+    active = false;
+    tracking = !tracking;
+  };
 </script>
 
 <style>
   .toggle {
     display: flex;
-    height: 36px;
+    height: 40px;
     align-items: center;
-    border-radius: 2px;
-    border: 1px solid var(--color-tertiary-level-6);
-    background-color: var(--color-background);
+    border-radius: 4px;
+    border: 1px solid var(--color-secondary);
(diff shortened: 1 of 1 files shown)


rudolfs:
Slightly different than the prototype:
prototype

I like your UX even better than the one with the tooltip and since you're the designer you get to decide what it looks & feels like anyway :P

ui/DesignSystem/Component/TrackToggle.svelte
```diff
@@ -6,41 +6,147 @@
 
   export let style = null;
   export let peerCount = null;
+  let hover = false;
+  let active = false;
+  let tracking = false;
```
rudolfs, line 11: This should be a prop that we can read and write from outside of the component.
```suggestion
  export let tracking = false;
```

ui/DesignSystem/Component/TrackToggle.svelte
```diff
@@ -6,41 +6,147 @@
...
+  const enter = () => {
```
rudolfs, line 12: Since these are simple one-liners and used only in one place you could as well inline them where they're used.
```
  on:mouseenter={() => { hover = true; }}
  on:mousedown={() => { active = true; }}
  on:mouseup={() => { active = false; tracking = !tracking;}}
  on:mouseleave={() => { hover = false; }}>
```

ui/DesignSystem/Component/TrackToggle.svelte
```diff
@@ -6,41 +6,147 @@
...
+    background-color: var(--color-secondary);
     display: flex;
-    height: 36px;
+    height: 40px;
     align-items: center;
+    border-top-left-radius: 3px;
+    border-bottom-left-radius: 3px;
+  }
+
+  .left :global(svg) {
```
rudolfs, line 76: We should be as specific as possible when writing global styles. Otherwise another component with a `.left` class that has an `<svg>` could be affected by this.
```suggestion
  .toggle.left :global(svg) {
```

ui/DesignSystem/Component/TrackToggle.svelte
```diff
@@ -6,41 +6,147 @@
...
+    fill: var(--color-background);
+  }
+
+  .left.hover {
+    background-color: var(--color-secondary-level-2);
+  }
+  .left.active {
+    background-color: var(--color-secondary-level-1);
+  }
+  .left.tracking :global(svg) {
```
rudolfs, line 86:
```suggestion
  .toggle.left.tracking :global(svg) {
```

ui/DesignSystem/Component/TrackToggle.svelte
```diff
@@ -6,41 +6,147 @@
```
rudolfs, line 6: The comment on L2:L3 Can be removed.

rudolfs:
Also updated the PR subject as this is a new feature that you're implementing as par #260 (FYI: anything marked as "chore" won't show up in `CHANGELOG.md`).

cloudhead:
Nice! The icon looks very dark when the project is tracked, compared to the text?

ui/DesignSystem/Component/TrackToggle.svelte
juliendonck: my editor styles this automagically...

ui/DesignSystem/Component/TrackToggle.svelte
juliendonck: updated

ui/DesignSystem/Component/TrackToggle.svelte
juliendonck: updated

ui/DesignSystem/Component/TrackToggle.svelte
juliendonck: updated

ui/DesignSystem/Component/TrackToggle.svelte
rudolfs: lol, never mind then

PR #326, opened 2020-04-29
chore(ui): design cleanup

juliendonck:
Address some messiness I saw around the app:
- fix alignment of carrot arrow in transaction accordion
- change status bar height from 48px to 40px to match design system
- update disabled states of all button types
- fix margin between buttons from 24px to 16px
- visual tweak to color swatches in design system

### DIFF
--- a/ui/DesignSystem/Primitive/Button.svelte
+++ b/ui/DesignSystem/Primitive/Button.svelte
@@ -3,54 +3,36 @@
 
   // vanilla | primary | secondary | transparent | outline
   export let variant = "primary";
-  // big | small
-  export let size = "big";
 
   export let disabled = null;
   export let icon = null;
   export let style = null;
   export let dataCy = null;
 
-  const iconClass = icon ? (size === "big" ? "icon" : "icon-small") : null;
+  const iconClass = icon ? "icon" : null;
 
   // we want to dynamically change whether a button is disabled or not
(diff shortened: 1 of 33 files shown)


ui/DesignSystem/Primitive/Button.svelte
```diff
@@ -155,22 +155,123 @@
...
   .disabled:active {
-    background-color: var(--color-foreground-level-3);
-    border-color: var(--color-foreground-level-3);
-    color: var(--color-foreground-level-5);
+    background-color: var(--color-foreground-level-1);
+    border-color: var(--color-foreground-level-1);
+    color: var(--color-foreground-level-3);
+  }
+
+  .disabled :global(svg) {
```
rudolfs, line 177: We probably should give the button a class and nest the global CSS under that class, just to make sure it doesn't bleed into other components.

juliendonck: good point. Added a `.button` class in general. And then specified most "global styles" under the `.vanilla` style since that's kinda the default. I'm wondering if we should change the variant to `vanilla` instead of `primary`.

rudolfs: >  I'm wondering if we should change the variant to vanilla instead of primary.

👍 

Would also be nice to have the `.vanilla` classes ordered in one place.

PR #330, opened 2020-04-30
feat(proxy): add avatar endpoint

MeBrei:
Implements the endpoint to retrieve the fallback avatar. 
Currently this will only be used for the org registration modal, but might come in handy for identity later on as well.

Related to #277


proxy/src/http/avatar.rs
```diff
@@ -0,0 +1,99 @@
+//! Endpoints for Avatar.
+
+use warp::document::{self, ToDocumentedType};
+use warp::{path, Filter, Rejection, Reply};
+
+use crate::avatar;
+
+/// `GET /avatars/<usage>/<id>`
```
xla, line 8: It seems usage might be better suited as a query param, with the default being `any`. In turn we would return a `400`, if the supplied value is not supported.

MeBrei: I moved `usage` into the query.
For the `400`, do I have to create a new error in `src/error.rs`?

proxy/src/http/avatar.rs
```diff
@@ -0,0 +1,99 @@
...
+    use warp::{reject, reply, Rejection, Reply};
+
+    use crate::avatar;
+
+    /// Get the avatar for the given `id`.
+    pub async fn get(usage: String, id: String) -> Result<impl Reply, Rejection> {
+        let avatar = avatar::Avatar::from(
+            &id,
+            match usage.as_str() {
+                "identity" => avatar::Usage::Identity,
```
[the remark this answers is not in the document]
xla: A different set of emojis is used based on the usage.

proxy/src/http/avatar.rs
```diff
@@ -0,0 +1,111 @@
...
+        id: String,
+        super::GetAvatarQuery { usage }: super::GetAvatarQuery,
+    ) -> Result<impl Reply, Rejection> {
+        let avatar = avatar::Avatar::from(
+            &id,
+            match usage.as_deref() {
+                Some("identity") => avatar::Usage::Identity,
+                Some("org") => avatar::Usage::Org,
+                Some("any") | None => avatar::Usage::Any,
+                _ => return Err(reject::not_found()),
```
xla, line 60: Will this emit a `400`?

PR #333, opened 2020-05-01
feat(ui): org registration flow api integration

sarahscott:
- [x] Org uses `avatarFallback`
- [x] Validation store created in `org.ts`


ui/Screen/OrgRegistration.svelte
```diff
@@ -69,16 +70,19 @@
...
       ? "Submit transaction"
       : "Next";
-
   $: disableSubmit = $validation.status !== ValidationStatus.Success;
+
+  $: payer =
+    $session.status === Status.Success && getPayer($session.data.identity);
+  $: transaction = getTransaction(orgName);
+  $: subject = getSubject(orgName);
+  $: avatarFallback = generateAvatar(orgName);
```
xla, line 85: None of these need to be reactive, is that correct?

ui/Screen/OrgRegistration.svelte
```diff
@@ -69,16 +70,19 @@
...
   $: submitLabel =
     state === RegistrationFlowState.TransactionConfirmation
       ? "Submit transaction"
       : "Next";
-
   $: disableSubmit = $validation.status !== ValidationStatus.Success;
+
+  $: payer =
+    $session.status === Status.Success && getPayer($session.data.identity);
+  $: transaction = getTransaction(orgName);
```
xla, line 83: This is an oddly general method, esp. living in `org.ts`. What kind of transaction are we fetching here and from where?

ui/src/org.ts
```diff
@@ -30,16 +32,32 @@ export const nameConstraints = {
...
-export const getOrg = (id: string): Promise<Org> => api.get<Org>(`orgs/${id}`)
+export const getSubject = (id: string) => {
+  return {
+    name: id,
+    kind: "org",
+    avatarFallback: generateAvatar("")
+  }
+};
 
 interface RegisterOrg {
```
xla, line 62: We are suffixing these with `Input` to make clear that they are used as expected shapes for API calls.

ui/src/validation.ts
```diff
@@ -51,16 +64,34 @@ export const createValidationStore = (constraints: any): ValidationStore => {
...
-        update((store) => {
-          // If the input has changed since this request was fired off, don't update
-          if (get(inputStore) !== input) return store
-          return { status: ValidationStatus.Success }
+    }
+
+    // Check remote validation
+    if (remoteValidation) {
+      try {
+        await remoteValidation.promise(input)
```
xla, line 72: Is this a doubly nested promise? The mix of await and then style promise resolution makes this quite hard to reason about. Is there room to improve and untangle this?

ui/DesignSystem/Primitive/Avatar.svelte
```diff
@@ -105,7 +105,7 @@
     <div
       class={`avatar ${avatarClass}`}
       style="background: {fmt(avatarFallback.background)}">
-      <Title variant={size} style="min-width: 27px; text-align: end;">
+      <Title variant={size} style="min-width: 27px; text-align: center;">
```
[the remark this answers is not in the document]
sarahscott: Oh weird - I'll remove this change in this pr so it can be addressed in a later one

ui/Screen/OrgRegistration.svelte
sarahscott: Not necessarily; it seemed like a quick way to make them dependent on the `orgName` input value and also save some LOC later but I'm open to refactoring

ui/Screen/OrgRegistration.svelte
sarahscott: Maybe something like `generateTransaction` would make more sense - it's fetching the details of the potential transaction to display to the user so it's not quite at the stage of firing it off yet.

sarahscott: But maybe it would fit better in `transaction.ts` with a named event (e.g. `getTransacton("Transactions.AddMemberToOrg")`

ui/src/validation.ts
sarahscott: Are you thinking along the lines of the event loop architecture? That might look better

ui/DesignSystem/Primitive/Avatar.svelte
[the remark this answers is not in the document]
rudolfs: You can check out: https://github.com/radicle-dev/radicle-upstream/pull/339 and let me know if it fixes the alignment on your machine.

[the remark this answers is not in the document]
sarahscott: Fixed 🎉 !

ui/Screen/OrgRegistration.svelte
xla: What we want here is to look for a transaction specifically to a criteria as in type of message that it carries and what's inside it (org name). Seems likely utility `transaction.ts` should expose.

ui/src/transaction.ts
```diff
@@ -136,36 +136,47 @@ export const formatStake = (msg: Message): string => `${formatMessage(msg)} depo
...
   return {
     name: identity.metadata.displayName || identity.metadata.handle,
-    variant: "user",
+    variant: "USER",
     avatarFallback: identity.avatarFallback,
     imageUrl: identity.metadata.avatarUrl
   };
 };
 
+export enum SubjectVariant {
```
sarahscott, line 145: Currently working on a better data structure because `variant` doesn't make a whole lot of sense; for now, this ensures it works.

ui/src/transaction.ts
```diff
@@ -136,36 +136,47 @@ export const formatStake = (msg: Message): string => `${formatMessage(msg)} depo
 export const formatPayer = (identity: identity.Identity): object => {
   return {
     name: identity.metadata.displayName || identity.metadata.handle,
-    variant: "user",
+    variant: "USER",
```
xla, line 139:
```suggestion
    variant: SubjectVariant.User,
```

This should be the actual variant we define down below.

sarahscott: Sweet, made a `Variant` type that can be used by both for now

PR #311, opened 2020-04-24
feat(ui): finalize design on project source view

cloudhead:
Closes #308 

* [x] Design refresh in file browser
* [x] Clickable project root in file header
* [x] Show README when project source is open, and latest commit on branch
* [x] Remove FileList component / Folder view
* [x] Fix tests

2020-04-24-165530_1367x1016_scrot
2020-04-24-165549_1367x1016_scrot

PR #335, opened 2020-05-04
refactor(ui): readme fetch and rendering

xla:
This improves the way the Readme is fetched and rendered and gets rid of a lot of nested conditionals. Achieved through the extension of Remote with an optional context it can create, which is sub-optimal but the best approach at the moment. We also introduce store chaining, where one store is dependent on the successful remote fetch of another.

Follow-up to #311


proxy/src/coco.rs
```diff
@@ -137,6 +137,8 @@ pub struct Blob {
     pub content: BlobContent,
     /// Extra info for the file.
     pub info: Info,
+    /// Absolute path to the object from the root of the repo.
+    pub path: String,
```
cloudhead, line 141: Lol I wanted to add this, but stopped myself.

ui/DesignSystem/Component/Remote.svelte
```diff
@@ -1,15 +1,24 @@
 <script>
   import * as remote from "../../src/remote.ts";
 
+  import WithContext from "./WithContext.svelte";
```
rudolfs, line 4: Wrapper inception :)

cloudhead:
Nice. I can't see where binary readme content is handled though?

xla:
> Nice. I can't see where binary readme content is handled though?

@cloudhead What could we sensibly do with a binary readme? Treated same as a non-existing one, is there a plan for that case?

cloudhead:
> > Nice. I can't see where binary readme content is handled though?
> 
> @cloudhead What could we sensibly do with a binary readme? Treated same as a non-existing one, is there a plan for that case?

Where is that code though? I can't find it.

ui/DesignSystem/Component/Remote.svelte
```diff
@@ -1,15 +1,24 @@
...
+  export let context = null;
 </script>
 
 {#if $store.status === remote.Status.NotAsked}
   <slot name="not-asked" />
 {:else if $store.status === remote.Status.Loading}
   <slot name="loading" />
 {:else if $store.status === remote.Status.Success}
-  <slot data={$store.data} />
+  {#if context && typeof context === 'string'}
```
sarahscott, line 15: Curious about this `typeof context === 'string'` guard - what are the other possibilities for the `context`, and why would we not use a `WithContext` wrapper for them?

ui/Screen/Project/Source.svelte
```diff
@@ -48,33 +47,21 @@
...
-  let readmePath = null;
+  const getRevision = current => {
+    return current !== "" ? current : metadata.defaultBranch;
+  };
 
-  $: readme = null;
-  $: if ($objectStore.status === remote.Status.Success) {
-    if ($objectStore.data.info.objectType === TREE) {
-      readmePath = findReadme($objectStore.data);
+  $: readmeStore = readme(id, getRevision($currentRevision));
```
sarahscott, line 56: Nice, this all looks really slick!

ui/src/source.ts
```diff
@@ -224,3 +221,28 @@ export const findReadme = (tree: Tree): string | null => {
...
   return null;
 }
+
+export const readme = (
+  projectId: string,
+  revision: string,
+): Readable<remote.Data<Readme | null>> => {
+  const readme = remote.createStore<Readme | null>();
+
+  remote
```
sarahscott, line 231: This bit is a little 🤯 for me; can you provide a quick tldr of what the chain is doing here?

ui/DesignSystem/Component/Remote.svelte
xla: We use that prop to name the context variable provided, so we want to make sure it's only used as such, happy to see alternatives.

xla:
> Where is that code though? I can't find it.

Good catch, thought I accounted for it. Addressed in 5d03559.

ui/src/source.ts
xla: `chain` ties one store to another, advancing their loading state in lockstep, and setting the error status if there was one in the dependent store. Once the dependent store has transitioned to `Success` the promise is resolved, that we use here to see  if the loaded object is a tree, search for a readme and load that readme.

PR #336, opened 2020-05-04
fix: correct attestion field name in user

xla:
Corrects a copy-pasta oversight in the user.

### DIFF
--- a/proxy/src/http/user.rs
+++ b/proxy/src/http/user.rs
@@ -137,7 +137,7 @@ mod handler {
 
         let mut reg = registry.write().await;
         let tx = reg
-            .register_user(&fake_pair, input.handle, input.maybe_project_id, fake_fee)
+            .register_user(&fake_pair, input.handle, input.maybe_entity_id, fake_fee)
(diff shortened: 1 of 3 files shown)


proxy/src/http/user.rs
```diff
@@ -202,7 +202,7 @@ pub struct RegisterInput {
     /// Handle the User registered under.
     handle: String,
     /// Optionally passed proejct id to store for attestion.
```
rudolfs, line 204: Comment needs to be adjusted.

proxy/src/registry.rs
```diff
@@ -105,7 +105,7 @@ pub struct User {
     /// Unique handle regsistered on the Regisry.
     pub handle: Id,
     /// Associated coco id for attestion.
```
rudolfs, line 107: Is this comment still correct?

xla:
> What exactly is an `entityId` in this case?

`Entity` is the primitive called in librad, this will be subject to change. For now it's the most accurate descriptor.

PR #337, opened 2020-05-04
feat(ui): commit history view

### LINKED ISSUE #275: Commit history view
Implement commit history view as part of **Code Browsing 2**.

cloudhead:
Closes #275 

UI + Backend. I've ommitted a design detail because I'm not sure yet how we want to do it, and would prefer to merge this as-is first.

Currently there's all commits are loaded. In the future, we may want to paginate.

- [x] Make commit message not blue

2020-05-04-172120_1043x791_scrot


proxy/src/coco.rs
```diff
@@ -290,6 +290,24 @@ pub fn commit(paths: &Paths, id: &str, sha1: &str) -> Result<Commit, error::Erro
     Ok(Commit::from(commit))
 }
 
+/// Retrieves the [`Commit`] history for the given `branch`.
+///
+/// # Errors
+///
+/// Will return [`error::Error`] if the project doesn't exist or the surf interaction fails.
+pub fn commits(paths: &Paths, id: &str, branch: &str) -> Result<Vec<Commit>, error::Error> {
```
xla, line 298: Here we rather want a `revision`, similar to how the `tree` function works as it is equally valid to look at a history for a tag or even a commit. That we only uses with branches for now shouldn't affect this API.

ui/src/source.ts
```diff
@@ -30,6 +30,22 @@ interface Commit {
...
+  summary: string;
+  description: string;
+}
+
+interface CommitGroup {
+  time: number;
+  commits: CommitSummary[];
+}
+
+type CommitHistory = CommitGroup[];
```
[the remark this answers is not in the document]
cloudhead: I'd rather be specific, since there are other types of histories out there..

ui/Screen/Project/Commits.svelte
```diff
@@ -0,0 +1,103 @@
...
+  }
+  .commit-group:after,
+  .commit-group header {
+    border-left: 2px solid var(--color-foreground-level-3);
+  }
+  .commit-group:first-child header {
+    border-left: none;
+  }
+  .commit {
+    border-bottom: 1px solid var(--color-foreground-level-2);
```
juliendonck, line 58: I'd just use the same color as the outer border. Seems off otherwise to me.

```suggestion
    border-bottom: 1px solid var(--color-foreground-level-3);
```

ui/Screen/Project/Commits.svelte
```diff
@@ -0,0 +1,103 @@
...
+    left: 1.25rem;
+    position: relative;
+    display: block;
+  }
+  .commit-group:after,
+  .commit-group header {
+    border-left: 2px solid var(--color-foreground-level-3);
+  }
+  .commit-group:first-child header {
+    border-left: none;
```
juliendonck, line 55: I understand the idea. But I feel or I would remove the border in general or leave it at the top for now. This seems broken.

proxy/src/coco.rs
cloudhead: The coco function takes a branch name though, so not sure that makes sense.

ui/Screen/Project/Commits.svelte
juliendonck: Like this without the line.
Screenshot 2020-05-05 at 11 52 13

proxy/src/coco.rs
xla: This should change in coco then.

ui/src/source.ts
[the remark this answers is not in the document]
xla: Which ones in our domain?

cloudhead:
All feedback addressed, PTAL

proxy/src/coco.rs
cloudhead: Yeah, that's probably right.

ui/src/source.ts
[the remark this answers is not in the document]
cloudhead: Browsing history, edit histories (of issues for eg.), project/org histories and probably more

rudolfs:
```
(!) Plugin svelte: CommitTeaser has unused export property 'commitMessageColor'. If it is for external reference only, please consider using `export const commitMessageColor`
ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
 7:   export let projectId = null;
 8:   export let commitMessage = null;
 9:   export let commitMessageColor = null;
                 ^
10:   export let timestamp = null;
11:   export let commitSha = null;
```

rudolfs:
What's also interesting is that the commit order is different from what you see via `git log` because of grouping by date. But that's by design, right?

cloudhead:
Oh that's strange, looking into it.

cloudhead:
Browsing through the rx commit history, it looks ordered and correct :thinking:

cloudhead:
Might have something to do with merge commit time :thinking:

juliendonck:
> It looks a bit odd when there's a single commit.

@cloudhead should we just remove the vertical line as a whole? I'm more and more thinking it'll be more elegant if we don't have the little circles etc.

cloudhead:
@rudolfs fixed the date grouping -- my algorithm wasn't correct, as I was checking time difference instead of just checking when a new day lapses.

However, there is still one issue and it has to do with `committerTime`, because that isn't what the array is ordered as. We need to decide what to do about that, but I'm going to open an issue regarding that, as I'm not sure where it should be fixed.

Issue #290, opened 2020-04-16
Misaligned avatars


cloudhead:
It seems like it's just me so far. If I remove the `min-width: 27px`, the problem goes away though..

![2020-04-15-180802_1037x627_scrot](https://user-images.githubusercontent.com/40774/79443301-f1d32980-7fd9-11ea-9b12-9f36271e244c.png)

rudolfs:
Weird! There shouldn't be any differences in the electron renderer between platforms, but so far we haven't really tested much on Linux, might be that.

Maybe @MeBrei can comment more as she implemented the icons.

MeBrei:
If I remove the `min-width` the emojis are too far left for me. It was a hack to get them to be centered as most emojis themselves are not. 
So we should look into that again anyway, but I also don't understand why it is different for you.

<img width="78" alt="Screenshot 2020-04-16 at 14 22 15" src="https://user-images.githubusercontent.com/16262137/79455774-af1b4c80-7fed-11ea-8137-eef4aa8dd758.png">

rudolfs:
Hmm a quick google turned up this: https://stackoverflow.com/questions/44142464/how-do-i-horizontally-center-an-emoji

Could also be a retina vs non-retina issue.

Maybe we can look into this deeper when we choose a font for emojis, so they're all consistent across platforms: https://github.com/radicle-dev/radicle-upstream/issues/231.

rudolfs:
Closed by #339

PR #339, opened 2020-05-04
feat(ui): use Twitter svg emojis in Avatar component

### LINKED ISSUE #231: Choice of emoji set for avatar generation
To guarantee that every user on any OS would get a consistent avatar generated and displayed it would make sense to use a custom emoji set then to rely on the OS's system fonts.

Here are some possible options for that:
- https://twemoji.twitter.com/
- https://www.joypixels.com/

cc @brandonhaslegs @cloudhead

rudolfs:
I got tired of the emoji alignment issues.

The implementation is based on this discussion: https://github.com/radicle-dev/radicle-upstream/issues/231#issuecomment-604914935.

Screenshot 2020-05-04 at 20 58 55
Screenshot 2020-05-04 at 20 59 09
Screenshot 2020-05-04 at 20 59 17

Closes #231, #290.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ dist
 .DS_Store
 native/main.comp.js
 public/bundle.*
+public/twemoji
 cypress/videos
 cypress/screenshots
 cypress/fixtures/example.json
--- a/README.md
+++ b/README.md
@@ -19,11 +19,22 @@ If you have questions or would like to get in touch, check out
 [radicle.community][rc].
 
 
+### Attribution
+
+Upstream uses:
+  - [Twemoji by Twitter][tw]
+  - [The Inter typeface family by Rasmus Andersson][ra]
+  - [Source Code Pro font family by Adobe][so]
+
+
 [ba]: https://badge.buildkite.com/4fb43c6b471ab7cc26509eae235b0e4bbbaace11cc1848eae6.svg?branch=master
 [de]: DEVELOPMENT.md
 [pr]: proxy
+[ra]: https://rsms.me/inter
 [rc]: https://radicle.community
 [ru]: https://www.rust-lang.org
 [st]: https://buildkite.com/monadic/radicle-upstream
+[so]: https://adobe-fonts.github.io/source-code-pro
 [sv]: https://svelte.dev
+[tw]: https://twemoji.twitter.com
 [ui]: ui
--- a/cypress/integration/routing.spec.js
+++ b/cypress/integration/routing.spec.js
@@ -43,8 +43,6 @@ context("routing", () => {
             cy.location().should(loc => {
               expect(loc.hash).to.eq("#/search");
             });
-
-            cy.screenshot();
           });
         }
       );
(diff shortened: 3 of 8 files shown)


sarahscott:
Screen Shot 2020-05-04 at 5 47 39 PM


epic LOC count there; looks great for me!

rudolfs:
> Is it really necessary that we pull in all these files? Esp. these thousands of binary files and what looks like the repositories documentation and tooling files.

I think we won't get around of pulling in the SVGs, as the npm package doesn't have them. If we don't include the SVGs then Upstream have emojis while offline.

I could delete the rest of the files, but we might have to keep the LICENCE around and add a line of attribution to our documentation.

I kept everything initially for the reason that it is a dependency that usually goes into a `vendor` dir and I find it easier to spot mistakes when doing an upgrade of the dependency when all of the files are present.

xla:
> Alternatively we can clone their repo in our org and add it as a submodule to this repo.

I think something along those lines might be better, otherwise having all these files tracked in our repo is a bit of a burden.

rudolfs:
Just had a discussion with @juliendonck: there is another option: just cherry-picking SVGs to cover our emoji whitelist. This would add 538 SVGs to the repo.

The downside would be that we wouldn't be able to use this subset for user input like issue/revision titles, bodies and show them when viewing file content, because it wouldn't cover all the possible emojis.

If we only use the SVG emojis for avatars, in all other places the regular font-based emojis would be displayed.

rudolfs:
So our options are:

1) adding all of the SVGs to this repo
    - adds 3248 files / 14M
    - would be in the repo history forever
    - every update would bloat the repo size
    - we can use all of the emojis in both avatars as well as texts

2) picking only the whitelisted emojis:
    - can't use it for other emoji use-cases (texts, user input)
    - adds 538 files / 1,4M
    - takes effort to hand-pick and update them
    - would be in the repo history forever
    - every update would bloat the repo size

3) add as git submodule
    - would have to find a way to do a shallow clone otherwise CI and the packaged app size could be affected (the twemoji repo weighs 861M)
    - added complexity for dev/CI setup
    - the Upstream repo stays lean
    - we can use all of the emojis in both avatars as well as texts

@cloudhead any thoughts?

rudolfs:
I also wonder if this is enough for proper attribution: https://github.com/radicle-dev/radicle-upstream/blob/c89febc94a90c49b651075875ea6028f9ddb9004/README.md#attribution and whether we need to add the licenses of the emojis and fonts somewhere in the repo?

cloudhead:
I agree that we shouldn't check all these files in. A sub-module could work, or we could create an npm package and depend on it.

rudolfs:
Added Twemoji as npm package, the source lives in this repo: https://github.com/radicle-dev/twemoji-svg-assets

The assets are copied via a `postinstall` script into `public/twemoji` whenever `yarn install` is run: https://github.com/radicle-dev/twemoji-svg-assets/blob/master/package.json#L10.

We should test if emojis:
  - [ ] work on Linux
  - [ ] work on macOS
  - [ ] work on CI

PR #341, opened 2020-05-05
feat: expose current user orgs list in session

### LINKED ISSUE #340: Extend session endpoint with list of orgs
For my registered user handle I want to get a list of orgs that I'm a member of.

Similar to https://github.com/radicle-dev/radicle-upstream/pull/300, just as part of http://localhost:8080/docs#get-/v1/session.

This is for the org list in the sidebar. We don't want to do multiple queries to the backend for this.

Screenshot 2020-05-05 at 09 30 25

xla:
Closes #340

### DIFF
--- a/ui/Screen/ProjectRegistration.svelte
+++ b/ui/Screen/ProjectRegistration.svelte
@@ -6,6 +6,7 @@
   import {
     NavigationButtons,
     ModalLayout,
+    Remote,
     StepCounter,
     Transaction
   } from "../DesignSystem/Component";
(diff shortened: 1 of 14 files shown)


ui/Screen/ProjectRegistration.svelte
```diff
@@ -1,25 +1,26 @@
 <script>
+  import { getContext } from "svelte";
   import { pop } from "svelte-spa-router";
 
   import { Flex, Title } from "../DesignSystem/Primitive";
   import {
     NavigationButtons,
     ModalLayout,
-    Remote,
```
rudolfs, line 8: `<Remote>` is still referenced in the file down below.

PR #343, opened 2020-05-05
feat(proxy): org project list endpoint

MeBrei:
Endpoint to return all projects registered under an org.
- [x] filter project list by org id
- [x] return `Project` instead of only the project name (*)
- [x] expose list in the api `GET orgs/<id>/projects`

(*) Open issue/dependency:
- when registering a project, currently `None` is always passed as the `maybe_project_id` in the proxy. In order to show any other info beside the name of the project, we need to pass the project id.

Solves part of #277

### DIFF
--- a/proxy/src/http/project.rs
+++ b/proxy/src/http/project.rs
@@ -224,10 +224,9 @@ mod handler {
         let fake_fee: Balance = 100;
 
         let mut reg = registry.write().await;
-        let maybe_coco_id = match input.maybe_coco_id {
-            Some(id) => Some(librad::project::ProjectId::from_str(&id).expect("Project id")),
-            None => None,
-        };
+        let maybe_coco_id = input
+            .maybe_coco_id
+            .map(|id| librad::project::ProjectId::from_str(&id).expect("Project id"));
         let tx = reg
             .register_project(
                 &fake_pair,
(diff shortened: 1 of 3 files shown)


MeBrei:
Added an [issue in the registry](https://github.com/radicle-dev/radicle-upstream/pull/343) to retrieve the projects of an org/user more direct.

proxy/src/http/project.rs
```diff
@@ -223,8 +224,18 @@ mod handler {
         let fake_fee: Balance = 100;
 
         let mut reg = registry.write().await;
+        let maybe_coco_id = match input.maybe_coco_id {
```
xla, line 227: You should be able to use `Option::map` to transform the `Some` value and return `None` if no value is present.

proxy/src/registry.rs
```diff
@@ -339,11 +361,19 @@ impl Registry {
...
-                name: project.name,
-                org_id: project.org_id,
-                // TODO(xla): Proper conversion of ProjectIds.
-                maybe_project_id: None,
+            .map(|project| {
+                let metadata_vec: Vec<u8> = project.metadata.into();
+                Project {
+                    name: project.name,
+                    org_id: project.org_id,
+                    maybe_project_id: if let [] = metadata_vec[..] {
```
xla, line 369: The [`is_empty`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.is_empty) would make this more readable.

PR #345, opened 2020-05-05
feat(ui): wire up orgs

rudolfs:
- [x] in sidebar
- [x] org screen
- [x] navigation

References: #277.


ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -1,15 +1,11 @@
...
-  import { orgMocks } from "../../lib/orgMocks.js";
   import * as path from "../../lib/path.js";
-  import { session } from "../../src/session.ts";
-
   import { Avatar, Icon, Title } from "../Primitive";
-  import Remote from "./Remote.svelte";
-
   import AddOrgButton from "./Sidebar/AddOrgButton.svelte";
+
+  const session = getContext("session");
```
xla, line 8: As this is a component in the design system it should not depend on global state. The idea was that the context is set up for screens consistently have access to dependent data when routed to. This component should depend on the things in the form of props and the screen which render them should get them out of the context. Here `orgs` and `identity` , which would also make it clear what data this component needs to be rendered.

ui/Screen/OrgRegistration.svelte
```diff
@@ -62,6 +62,7 @@
   const registerOrg = async () => {
     try {
       await register(orgName);
+      await fetchSession();
```
xla, line 65: Good catch on getting the fresh session. This seems to be a common pattern throughout the app, wonder how to address this consistently. Maybe with notifications about new session state. Open question I guess.

ui/src/org.ts
```diff
@@ -1,22 +1,54 @@
...
 
-export interface Org {
-  name: string;
-  avatarFallback: EmojiAvatar;
+// STATE
+const orgStore = remote.createStore<Org>();
+export const org = orgStore.readable;
+
+// EVENTS
+enum Kind {
```
xla, line 36: This should be defined before usage, thought our liniting catches that.

ui/src/org.ts
```diff
@@ -1,22 +1,54 @@
...
 
 export enum RegistrationFlowState {
   NameSelection,
   TransactionConfirmation
 }
 
-export interface Org {
-  name: string;
-  avatarFallback: EmojiAvatar;
+// STATE
```
xla, line 31: Usually we order: types, state, event, helpers in our other files.

ui/Screen/OrgRegistration.svelte
rudolfs: Yea, was wondering about how the Svelte context API will scale for us, as it is not reactive.

> Context is not inherently reactive. If you need reactive values in context then you can pass a store into context, which will be reactive.
https://svelte.dev/docs#setContext

In the future we'll want new data to be pushed to the front-end, then it would be good that the store changes actually propagate to the components automatically. Maybe we could just use global stores instead of the context API then?

rudolfs: Here's more from Svelte docs:
> Contexts vs. stores
>
>Contexts and stores seem similar. They differ in that stores are available to any part of an app, while a context is only available to a component and its descendants. This can be helpful if you want to use several instances of a component without the state of one interfering with the state of the others.
>
>In fact, you might use the two together. Since context is not reactive, values that change over time should be represented as stores
>
>https://svelte.dev/tutorial/context-api

ui/src/org.ts
rudolfs: It catches some of the ordering issues, but not all. Weird.

ui/src/org.ts
rudolfs: What is out of order here?

ui/Screen/OrgRegistration.svelte
xla: That is actually really good insight and definitely something we need to change going forward. My point was more about taming the dependence on fetching the session fresh when an action has been submitted. We can rework the session passing into a store.

ui/src/org.ts
xla: The Msg, Kind and Fetch are all part of the event system, so is update. The state should just move directly after the types and then all the event stuff.

ui/DesignSystem/Component/Sidebar.svelte
rudolfs: Makes sense. Moved the data fetching up to `<SidebarLayout>`. Eventually we'll have to move these "smarter" components out of the design system. But that is out of scope for this issue.

ui/src/org.ts
rudolfs: Like so: https://github.com/radicle-dev/radicle-upstream/pull/345/commits/75034f7cf4fd4a19a2b26f96f8acaf6c19c24982?

I guess the `// TYPES` section threw me off, I just stuck everything that looked like a type in there.

PR #351, opened 2020-05-06
fix: bundle missing library libhogweed

### LINKED ISSUE #112: AppImage: libhogweed.so.4 missing on Arch Linux
When using the AppImage on arch the proxy fails to start with the following message
> error while loading shared libraries: libhogweed.so.4: cannot open shared object file: No such file or directory

`libhogweed` is part of the `nettle` package and arch installs `libhogweed.so.5`. The Rust `nettle` package links dynamically against `libhogweed`.

To fix this the proxy would need to link statically against `libhogweed`.

rudolfs:
Trying out the "official" solution: https://github.com/electron-userland/electron-builder/issues/2732.

Closes #112

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
       ],
       "extraFiles": [
         {
-          "from": "/usr/lib/libhogweed.so.4",
+          "from": "./usr/lib/x86_64-linux-gnu/libhogweed.so.4",
           "to": "usr/lib"
         }
       ]
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
       "extraFiles": [
         {
           "from": "/usr/lib/x86_64-linux-gnu/libhogweed.so.4",
-          "to": "usr/lib"
+          "to": "usr/lib/libhogweed.so.4"
         }
       ]
     },
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
       ],
       "extraFiles": [
         {
-          "from": "./usr/lib/x86_64-linux-gnu/libhogweed.so.4",
+          "from": "/usr/lib/x86_64-linux-gnu/libhogweed.so.4",
           "to": "usr/lib"
         }
       ]
--- a/package.json
+++ b/package.json
@@ -31,6 +31,12 @@
       "target": [
         "Appimage",
         "snap"
+      ],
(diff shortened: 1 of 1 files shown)


rudolfs:
> I’m not sure what adding the shared library to extraFiles does. Does it install the library to /usr/lib? This might mess up the users system.

@geigerzaehler: as we're using electron bundler for distribution, I assume it would bundle the files with the package somehow. Not sure what it uses under the hood. I don't expect it to actually write stuff into your filesystem, and assume that the `usr/lib` path is within the package that it spits out.

I thought the comment on the builder thread was worth giving a shot before spending more time to figure this out.

@cloudhead also suggested looking into [AppImage documentation](https://docs.appimage.org/packaging-guide/from-source/linuxdeploy-user-guide.html#packaging-binaries-and-other-resources-manually). There's an option to bundle libs:
> --library/-l
Bundle a shared library (.so file) into the AppDir. Set up everything so that other libraries, executables etc. use this one instead of a system one.

But would have to investigate how exactly bundler works under the hood for this to be of any help.

rudolfs:
@geigerzaehler could you give this build a try on your machine (assuming that you remove the nettle libs from your system temporarily): https://buildkite.com/monadic/radicle-upstream/builds/2103#76b36227-4c83-4adb-ba59-6e8783da7cac?

geigerzaehler:
The app image is still not working. 

> libhogweed.so.4: cannot open shared object file: No such file or directory

I have `libhogweed.so.5.` installed. I guess you need to add `.so.4` to the extra files.

rudolfs:
Ok, I changed it to `.so.4` and pushed another commit, buildkite didn't pick it up, so I force pushed to kick it. But it failed to pick that build up as well : |

rudolfs:
Will try a manual build.

rudolfs:
So now we got `libhogweed.so.4` in the AppImage, somebody should test the latest package.

```
» ls squashfs-root/usr/lib
libappindicator.so.1  libgconf-2.so.4  libhogweed.so.4	libindicator.so.7  libnotify.so.4  libXss.so.1	libXtst.so.6
```

geigerzaehler:
@rudolfs this fixed it for me!

PR #354, opened 2020-05-07
fix(ui): identity creation regression

### LINKED ISSUE #353: Identity creation flow skips Success state
If you create a new identity, upon confirmation the Welcome screen flashes and this view is skipped:

Screen Shot 2020-05-06 at 4 51 46 PM

sarahscott:
Fixes #353 

Currently, `App.svelte` subscribes to the `$session` store, so when we send `session.fetch` from the `IdentityCreation` flow, `App.svelte` is notified and re-renders everything. 

This approach moves the state management for the identity creation flow into a store, but it still doesn't seem quite right. @xla , curious what you foresee for the `App.svelte` launch process because that could change how we do this. Perhaps `App.svelte` will have more control of the identity creation, or maybe there are more states to consider as we add the ability to login/logout.

Other ideas:
- `App.svelte` has a separate Login router, and `IdentityCreation` is broken up into smaller UI components that a `Launch`/`Login` flow would manipulate (this is actually what I think would be better, but it requires more work & depends on our larger vision for launching and authenticating)
- `IdentityCreation` remains as complex as it is and its `launchFlowStore` is made into something more specific than a `writable`


xla:
> `App.svelte` has a separate Login router, and `IdentityCreation` is broken up into smaller UI components that a `Launch`/`Login` flow would manipulate (this is actually what I think would be better, but it requires more work & depends on our larger vision for launching and authenticating)

@sarahscott This seems like a viable path forward. The top-level should control which routes to render given the main session state. Ideally the flow on app start is bundled in an Onboard screen, already took the first step by introducing the `onboard.ts` which carries the state and variants. It's not desirable that identity creation is tight to login/onboarding/launch.

PR #355, opened 2020-05-07
style(ui): replace html dropdown with custom one

rudolfs:
Now that we have a custom dropdown that matches our style guide we should use it everywhere.

### DIFF
--- a/ui/Screen/CreateProject.svelte
+++ b/ui/Screen/CreateProject.svelte
@@ -9,7 +9,6 @@
   import { getLocalBranches } from "../src/source.ts";
   import { getValidationState } from "../src/validation.ts";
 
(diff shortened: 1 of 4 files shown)


ui/Screen/CreateProject.svelte
```diff
@@ -303,27 +306,31 @@
             <Text
               style="margin-bottom: 12px; color:
               var(--color-foreground-level-6); text-align:left">
-              Choose an existing repository
+              Choose the existing repository
```
sarahscott, line 309: Nitpicky grammar point but I think 'Choose an existing repository' sounds better 🤓

sarahscott:
Big aesthetic improvement 🎨 ! LGTM

PR #356, opened 2020-05-07
fix(ui): new dropdown update

juliendonck:
- Dropdown width is always only the size of the content. Not the longest option in the list.
- Dropdown disabled state doesn't have a hover effect anymore.


juliendonck:
Merged into #355 will, resolve both there.

juliendonck:
Merged into #355 will, resolve both there.

PR #357, opened 2020-05-07
refactor(ui): consolidate to ts

xla:
A set of consolidations to make the codebase more uniform and have almost all non Svelte ui code in typescript. Will leave hotkeys as a follow-up as it currently relies on the router which can't be imported as a normal module as it is implemented as a Svelte component for whatever reason. I do have an idea of how to address that, but it would bloat the scope here.

See individual commits they should be self-contained.


native/ipc.js
```diff
@@ -1,4 +1,5 @@
-import { IPC_CLIPBOARD_WRITETEXT, IPC_DIALOG_SHOWOPENDIALOG } from "./types.js";
+export const DIALOG_SHOWOPENDIALOG = "IPC_DIALOG_SHOWOPENDIALOG";
+export const CLIPBOARD_WRITETEXT = "IPC_CLIPBOARD_WRITETEXT";
```
rudolfs, line 2:
```suggestion
export const DIALOG_SHOWOPENDIALOG = "DIALOG_SHOWOPENDIALOG";
export const CLIPBOARD_WRITETEXT = "CLIPBOARD_WRITETEXT";
```

xla:
> Once we don't have any more `.js` files it would make sense to rename `/ui/src` to `/ui/lib`.

Why?

> `orgMocks.js` can be deleted as it is not used anymore, but the rest should be converted to `.ts` files, it's okay to convert `hotkeys.js` in another PR.

Alluded to the hotkeys follow-up in the description up top. For the `index.js` files it would mean we are forced to import i.e. `Component/index.ts`.

rudolfs:
> Why?

Seems more logical to me.

> For the index.js files it would mean we are forced to import i.e. Component/index.ts.

That's no good, can we then remove the .gitignore for `ui/**/*.js` files, as adding new index.js files would then not work:
https://github.com/radicle-dev/radicle-upstream/pull/357/files#diff-a084b794bc0759e7a6b77810e01874f2

xla:
> Seems more logical to me.

It's the source of the ts based compilation to js. Don't care either way, neither has a strong meaning.

rudolfs:
Notifications seem to be broken, I don't see any when creating a new project.
Here's an easy way to test notifications:
```
diff --git a/ui/App.svelte b/ui/App.svelte
index 804f917..70152f8 100644
--- a/ui/App.svelte
+++ b/ui/App.svelte
@@ -68,6 +68,8 @@
     "/transactions/:id": TransactionDetails,
     "*": NotFound,
   };
+
+  notification.error({ mesage: "Session could not be fetched" });
 </script>

 <Remote {store} context="session">
```

Screenshot 2020-05-07 at 15 30 10

xla:
> Notifications seem to be broken, I don't see any when creating a new project.

Properly eradicated in 07c5990.

sarahscott:
I'm in favor of `src/` because it's the typescript convention and would make the repo less confusing to people coming in with that context.

ui/DesignSystem/Component/Notification.svelte
```diff
@@ -1,9 +1,12 @@
 <script>
+  import { Level } from "../../src/notification.ts";
+
   import { Icon, Text, Title } from "../Primitive";
 
   export let style = null;
-  export let variant = "info"; // info | error
+  export let level = Level.Info;
```
sarahscott, line 7: very nice

ui/Screen/IdentityCreation.svelte
```diff
@@ -72,11 +69,11 @@
       </Button>
     </div>
   {:else if $store === State.Form}
-    <IdentityCreationForm
+    <Form
```
sarahscott, line 72: very happy about the events instead of prop callbacks

PR #359, opened 2020-05-07
test(ui): improve identity creation test coverage

rudolfs:
Discovered a regression: when hitting `Cancel` or the modal close `x` button on the identity form screen there's an exception.

```
Uncaught TypeError: $store.set is not a function
    at IdentityCreationForm.returnToWelcomeStep (IdentityCreation.svelte:16)
    at index.mjs:620
    at Array.forEach (<anonymous>)
    at index.mjs:619
    at click_handler (IdentityCreationForm.svelte:138)
    at index.mjs:637
    at Array.forEach (<anonymous>)
    at bubble (index.mjs:637)
    at HTMLButtonElement.click_handler (Button.svelte:12)
```

Fixed in: https://github.com/radicle-dev/radicle-upstream/pull/359/commits/ec651d0c91b3d454cc76e03b5f0e58e248449c67.

### DIFF
--- a/cypress/integration/identity_creation.spec.js
+++ b/cypress/integration/identity_creation.spec.js
@@ -68,6 +68,9 @@ context("identity creation", () => {
 
           // Now try to close the modal via the "x" button
           cy.get('[data-cy="modal-close-button"]').click();
+
+          // We should land back on the intro screen
+          cy.get('[data-cy="get-started-button"]').should("exist");
         });
       }
     );
--- a/cypress/integration/identity_creation.spec.js
+++ b/cypress/integration/identity_creation.spec.js
@@ -12,6 +12,89 @@ context("identity creation", () => {
     });
   });
 
+  context("navigation", () => {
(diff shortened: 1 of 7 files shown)


cypress/integration/identity_creation.spec.js
```diff
@@ -12,6 +12,89 @@ context("identity creation", () => {
...
+      () => {
+        it("sends the user back to the intro screen", () => {
+          cy.get('[data-cy="get-started-button"]').click();
+          cy.get('[data-cy="cancel-button"]').click();
+
+          // We should land back on the intro screen
+          cy.get('[data-cy="get-started-button"]').click();
+
+          // Now try to close the modal via the "x" button
+          cy.get('[data-cy="modal-close-button"]').click();
```
MeBrei, line 70: Should there be a check that the modal is still open?

rudolfs: Good catch.

PR #361, opened 2020-05-07
feat(ui): wire up org project list

rudolfs:
Screenshot 2020-05-12 at 12 06 26

Screenshot 2020-05-12 at 12 07 09

Implements part of #277.


ui/Screen/Org/Projects.svelte
```diff
@@ -2,54 +2,81 @@
...
-  import Placeholder from "../../DesignSystem/Component/Placeholder.svelte";
+  import { Flex, Icon, Text } from "../../DesignSystem/Primitive";
+  import {
+    AdditionalActionsDropdown,
+    List,
+    ProjectCard,
+    Remote,
+    Stats,
+  } from "../../DesignSystem/Component";
+  import { orgs as orgsStore, fetchProjectList } from "../../src/org.ts";
```
xla, line 13: Would organise these in clusters, starting with external libs, then internal libs, then primitives and componets.

ui/Screen/Org/Projects.svelte
```diff
@@ -2,54 +2,81 @@
...
-  const registerProject = () => push(path.registerProject(params.id));
-  const registerMember = () => push(path.memberRegistration(params.id));
+  const select = (event) => {
+    const orgProject = event.detail;
+    if (orgProject.maybeProject) {
+      push(path.projectSource(orgProject.maybeProject.id));
+    }
+  };
+
+  $: fetchProjectList({ id: params.id });
```
xla, line 24: Would put this at the end of the script tag after all props and consts have been set up.

ui/Screen/Org/Projects.svelte
```diff
@@ -2,54 +2,81 @@
...
+          <div slot="left">
+            <ProjectCard {...projectCardProps(orgProject)} />
+          </div>
+          <div slot="right" style="display: flex; align-items: center;">
+            <AdditionalActionsDropdown headerTitle={orgProject.name} />
+          </div>
+        </Flex>
+      {/if}
+    </List>
+  {:else}{push(path.orgOnboard(params.id))}{/if}
```
xla, line 77: Having routing invocations which are not triggered by user interaction in the markup is burying crucial behaviour in an unexpected place. Can we solve this differently?

ui/DesignSystem/Component/List.svelte
```diff
@@ -1,11 +1,9 @@
 <script>
```
xla, line 1: Is it really a general list now? It still uses references like `project-card`?

ui/Screen/Org/Projects.svelte
```diff
@@ -2,54 +2,81 @@
...
-        variant="vanilla"
-        on:click={registerMember}
-        dataCy="add-member-button">
-        Add a member
-      </Button>
-    </div>
+<Remote store={orgsStore} let:data={orgProjects}>
+  {#if orgProjects.length > 0}
+    <List items={orgProjects} on:select={select} let:item={orgProject}>
+      {#if orgProject.maybeProject}
```
xla, line 53: Shouldn't we move this out into its own component e.g. `OrgProjectCard`, which can wrap `ProjectCard` and knows how to render itself depending if a coco project is present.

ui/src/org.ts
```diff
@@ -1,17 +1,26 @@
...
 // Types
 export interface Org {
   name: string;
   avatarFallback: avatar.EmojiAvatar;
 }
 
+export interface OrgProject {
+  name: string;
+  orgId: string;
+  maybeProjectId: string;
```
xla, line 18: This is a `maybeProject` which should be of type `project::Project` I guess.

ui/src/org.ts
```diff
@@ -1,17 +1,26 @@
...
-import * as user from "./user"
+import * as user from "./user";
 
 // Types
 export interface Org {
   name: string;
   avatarFallback: avatar.EmojiAvatar;
 }
 
+export interface OrgProject {
```
xla, line 15:
```suggestion
export interface Project {
```

This is already in the `org` module, so can just be the unprefixed version.

ui/src/org.ts
```diff
@@ -1,17 +1,26 @@
...
+  name: string;
+  orgId: string;
+  maybeProjectId: string;
+}
+
+type OrgProjects = OrgProject[];
```
xla, line 21:
```suggestion
type Projects = Project[];
```

ui/src/org.ts
```diff
@@ -1,17 +1,26 @@
...
+
+
```
xla, line 23:
```suggestion
```

ui/src/org.ts
```diff
@@ -21,24 +30,33 @@ export enum RegistrationFlowState {
 const orgStore = remote.createStore<Org>();
 export const org = orgStore.readable;
 
-// Api 
-export const getOrg = (id: string): Promise<Org> => api.get<Org>(`orgs/${id}`)
+const orgProjectsStore = remote.createStore<OrgProjects>();
+export const orgs = orgProjectsStore.readable;
```
xla, line 34:
```suggestion
export const projects = projectsStore.readable;
```

Otherwise this is miss-leading.

ui/src/org.ts
```diff
@@ -21,24 +30,33 @@ export enum RegistrationFlowState {
 const orgStore = remote.createStore<Org>();
 export const org = orgStore.readable;
 
-// Api 
-export const getOrg = (id: string): Promise<Org> => api.get<Org>(`orgs/${id}`)
+const orgProjectsStore = remote.createStore<OrgProjects>();
```
xla, line 33:
```suggestion
const projectsStore = remote.createStore<OrgProjects>();
```

ui/Screen/Org/Projects.svelte
rudolfs: Can do, got the inspiration from you: https://github.com/radicle-dev/radicle-upstream/pull/310/files#diff-1f4e722feb2be79d26b81358601f6f3cR21 ; P

ui/DesignSystem/Component/List.svelte
rudolfs: Forgot to remove that unused class... But yeah, apart from that it's pretty generic.

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/361/commits/1918f137453a68839171a7ac8dcd1ea6f40afa37

ui/Screen/Org/Projects.svelte
rudolfs: We can move it to a wrapper if there is more than one place where we need it.

proxy/src/http/org.rs
```diff
@@ -201,7 +201,7 @@ mod handler {
         for p in &projects {
             let maybe_project = if let Some(id) = &p.maybe_project_id {
                 let paths = paths.read().await;
-                Some(project::get(&paths, &id).await.unwrap())
+                Some(project::get(&paths, id).await.expect("Project"))
```
MeBrei, line 204:
```suggestion
                Some(project::get(&paths, id).await.expect("Project not found"))
```

proxy/src/http/org.rs
```diff
@@ -313,12 +313,15 @@ impl ToDocumentedType for registry::Project {
     }
 }
 
-/// Bundled input data for org registration.
+/// Object the API returns for an project that is registered under an org.
```
MeBrei, line 316:
```suggestion
/// Object the API returns for a project that is registered under an org.
```

PR #368, opened 2020-05-10
refactor(ui): improve scalability of hotkeys

xla:
This change reduces the points we need to touch when introducing modals, as it uses a general way for a modal to be escapable whenever created. It also makes the `ModalLayout` more generally usable and wraps the last remaining two (DesignSystemGuide and Help) that used custom layouts.

This is the follow-up to #357


rudolfs:
The modal close hot-key (similar to the close button) should lead you back to the intro screen before the identity is created. After the identity is created the hotkey should land you on the profile page.

Adding specs for this would be nice too.

Screenshot 2020-05-10 at 11 10 28

xla:
@rudolfs All addressed.

cypress/integration/identity_creation.spec.js
```diff
@@ -71,6 +71,12 @@ context("identity creation", () => {
 
           // We should land back on the intro screen
           cy.get('[data-cy="get-started-button"]').should("exist");
```
rudolfs, line 73: You wanna be on the second screen when testing wether the `esc` key works.

```suggestion
          cy.get('[data-cy="get-started-button"]').click();
```

cypress/integration/identity_creation.spec.js
```diff
@@ -57,7 +57,7 @@ context("identity creation", () => {
     });
 
     context(
-      "when clicking cancel or modal close button before the identity is created",
+      "when cancel, close or escape before the identity is created",
```
rudolfs, line 60:
```suggestion
"when clicking cancel, close or hitting esc before the identity is created",
```

ui/Screen/ProjectCreation.svelte
```diff
@@ -226,7 +226,6 @@
   .wrapper {
     display: flex;
     justify-content: center;
-    margin: 92px 0 72px 0;
```
rudolfs, line 229: That margin might have been useful:

margin

Issue #347, opened 2020-05-06
Feature: Transactions


xla:
Outstanding tasks to complete transactions.

**proxy**

- [x] poll for latest block height and update transactions in cache accordingly (#407)
- [x] extend transaction states with error variants (#407)
- [x] clarify transaction state machine (#407)
- [x] consistent timestamps in status variants (#407)
- [ ] pass additional contextual data (i.e. full user/identity)
- [x] delete receiver
- [ ] control actions depending on tx state
- [x] persist txs cache to disk

**ui**

- [x] polling transaction states and update the transaction center and detail view accordingly (#407)
- [x] once in place in the proxy, handle status/progress/error & transaction costs in the ui (#407)
- [x] animate the transaction center: slide up/down of cards, slide in of the whole center on first transaction
- [x] extract receiver from message (407)
- [ ] control actions depending on tx state (#269)
- [x] more nuanced subject formatting (#346)

Part of #277

xla:
Done.

PR #370, opened 2020-05-11
feat(proxy): persist transactions

xla:
Implements persistance of transactions in preparation to progress them properly with the blockchain advancing. Major changes:

* break apart registry interactions from transaction caching
* use existing store to persist txs in to `transactions` bucket
* streamline and correct serialisations
* introduce new trait for better composibility

Part #347


proxy/src/http/transaction.rs
```diff
@@ -51,45 +48,31 @@ fn list_filter(
...
-        reg: Arc<RwLock<registry::Registry>>,
+    pub async fn list<C: registry::Cache>(
+        cache: http::Container<C>,
         input: super::ListInput,
     ) -> Result<impl Reply, Rejection> {
         let tx_ids = input
             .ids
             .iter()
             .map(|id| {
+                println!("INPUT ID: {}", id);
```
rudolfs, line 65: Stray debug log.

xla:
>     * where is the kv store persisted on disk? does the location align with where coco config is stored?

It's stored in the XDG directories which is a common way to store application data on different platforms and is the same mechanism coco uses.

> how do we invalidate the transaction cache?
>  how do we clear the cache manually?

For testing we could introduce another control endpoint, for normal operation we need to incorporate that in the upcoming **Settings**.

rudolfs:
And what's the plan for cache invalidation in general? Do we just keep all transactions around forever until the user clears them?

xla:
> And what's the plan for cache invalidation in general? Do we just keep all transactions around forever until the user clears them?

Yeah for now this shouldn't impact anything much, given that it's the transactions a single user issued from a single machine. I don't expect this to exceed 10s  or 100s of items.

proxy/src/registry/transaction.rs
```diff
@@ -0,0 +1,338 @@
...
+    async fn cache_transaction(&mut self, tx: Transaction) -> Result<(), error::Error>;
+
+    /// Returns all cached transactions.
+    ///
+    /// # Errors
+    ///
+    /// Will return `Err` if a protocol error occurs.
+    async fn list_transactions(
+        &self,
+        ids: Vec<protocol::TxHash>,
```
cloudhead, line 88: Does the function need to own the TxHashes? I'd expect this to take a slice normally.

proxy/src/registry/transaction.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /// Returns all cached transactions.
+    ///
+    /// # Errors
+    ///
+    /// Will return `Err` if a protocol error occurs.
+    async fn list_transactions(
+        &self,
+        ids: Vec<protocol::TxHash>,
```
cloudhead, line 142: Yeah so this could be a slice, since you're just doing a `contains` on it.

proxy/src/registry/transaction.rs
```diff
@@ -0,0 +1,338 @@
...
+        id: registry::Id,
+    },
+
+    /// Issue a new project registration with a given name under a given org.
+    ProjectRegistration {
+        /// Actual project name, unique for org.
+        project_name: registry::ProjectName,
+        /// The Org in which to register the project.
+        org_id: registry::Id,
+    },
```
cloudhead, line 52: I would have thought that the `URN` is also needed to register a project :thinking: maybe the registry hasn't implemented this yet?

proxy/src/registry/transaction.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    /// Issue a user registration for a given handle storing the corresponding identity id.
+    UserRegistration {
+        /// Globally unique user handle.
+        handle: registry::Id,
+        /// Identity id originated from librad.
+        id: Option<String>,
```
cloudhead, line 59: Kind of expect this to be binary data, ie. `Vec<u8>`. Definitely not a utf8 string :thinking:

Issue #285, opened 2020-04-15
Feature: Register project


rudolfs:
Implement the [Register project](https://www.figma.com/file/dnYlebm0ehgyJyT3Po9Kcl/Upstream?node-id=3917%3A26083) flow.

- access the flow via
  - [x] my profile page project list -> choose existing local unregistered project -> `...` dropdown
  - [x] project view topbar `...` context menu
  - [x] org page context `...` dropdown menu in the topbar
  - [x] permissions should prevent invalid user actions (eg. registering a project when you haven't yet registered your own handle) -> grey out the menu item and show tooltip
- components
  - [ ] ~~extract "start new / continue with existing repo" into a reusable component and use it in `CreateProject.svelte` and this register project flow~~ (the UX flow got simplified considerably on 24th of Apr)
  - [x] "register under org/identity" dropdown component
- ~~transactions~~ tracked in #347
  - [ ] ~~transaction detail view should show the org under which the project is going to be registered~~
  - [ ] ~~transaction detail view should show the correct wallet~~
- integration
  - [x] wire up to proxy endpoint that checks if a name is already taken
  - [x] replace orgs mock endpoint with the real thing
  - [x] use `identity.registered` as id 
- tests
  - [x] e2e for basic navigation through the happy paths of the flow #372
- refactor
  - [x] replace html dropdown in "create project" flow with the new dropdown #355
  - [ ] refactor dropdown/option props https://github.com/radicle-dev/radicle-upstream/pull/292#discussion_r417132729
  - [ ] replace manual validations with the ones from `validation.ts`
  - [ ] remove permissions duplication for register menu item
  - [ ] move details step into `ProjectRegistration` screen once it's cleaned up

<img width="1228" alt="Screenshot 2020-04-28 at 16 06 50" src="https://user-images.githubusercontent.com/158411/80497094-4d060400-896a-11ea-97e0-7291ccc06433.png">

PR #372, opened 2020-05-11
test(ui): add project registration specs

rudolfs:
Part of #285


cypress/integration/project_registration.spec.js
```diff
@@ -0,0 +1,216 @@
...
+};
+const project2 = () => {
+  return {
+    name: "upstream",
+    description: "code collab",
+    defaultBranch: "master",
+  };
+};
+const org1 = () => "monadic";
+const org2 = () => "github";
```
xla, line 17: None of these need to functions.

rudolfs: Good point!

PR #373, opened 2020-05-12
feat(ui): streamline transaction formatting

sarahscott:
Because we can pull up transaction details at any time, the `Transaction` screen should be able to figure out what it needs to display regardless of its parent.

- [x] update `formatSubject` to work for all currently-supported messages
- [x] change `registrar` to `domain`
- [x] tests

Part of #347


cypress/integration/org_member_addition.spec.js
```diff
@@ -29,45 +29,62 @@ context("add member to org", () => {
...
+      cy.pick("submit-button").click();
 
-      cy.pick("add-member-modal", "name-input").clear();
-      cy.pick("add-member-modal", "avatar").should("not.be.visible");
+      cy.pick("message").contains("Org member registration");
+      cy.pick("subject").contains("coolname");
     });
+
+    // TODO(sos): add actual transaction details check once we can make this tx
+    // it("submits correct transaction details to proxy", () => {
```
xla, line 73: Let's use `skip` here instead of commenting out the whole block

ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -17,25 +19,44 @@
...
+    switch (subject.type) {
+      case SubjectType.User:
+      case SubjectType.Member:
+      case SubjectType.UserProject:
+        return "circle";
+      default:
+        return "square";
+    }
+  };
+  console.log(transaction);
```
xla:
```suggestion
```

ui/DesignSystem/Primitive/Avatar.svelte
```diff
@@ -20,6 +20,7 @@
     return `rgb(${background.r}, ${background.g}, ${background.b})`;
   };
 
+  $: console.log("av", title, avatarFallback);
```
xla:
```suggestion
```

ui/Screen/ProjectRegistration.svelte
```diff
@@ -22,48 +22,36 @@
...
-      variant: registrarVariant,
-    };
-  };
-
-  const tx = {
-    messages: [{ type: transaction.MessageType.ProjectRegistration }],
-  };
+  const wallet = () => transaction.formatPayer(session.identity);
+
+  // TODO(sos): coordinate message format for project registration with proxy
```
xla, line 44: This should live in `transaction.ts`.

ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -17,25 +19,43 @@
...
-  <div slot="right">
-    <Avatar
-      dataCy="subject-avatar"
-      title={subject.name}
-      imageUrl={subject.imageUrl}
-      avatarFallback={subject.avatarFallback}
-      variant={subject.variant === Variant.User ? 'circle' : 'square'}
-      style="color: var(--color-foreground)" />
+  <div slot="right" data-cy="subject">
+    {#await subject.avatarSource then avatar}
```
xla, line 46: Don't think it's ideal that components in the design system are dependent on outcomes of side-effects. It'd be better if the result of this is passed as a prop.

ui/src/avatar.ts
```diff
@@ -23,5 +23,5 @@ export enum Usage {
 }
 
 // EVENTS
-export const get = (usage: Usage, id: string): Promise<EmojiAvatar> =>
+export const getAvatar = (usage: Usage, id: string): Promise<EmojiAvatar> =>
```
xla, line 26: Still not a fan of prefix/suffix of the module inside the module.

ui/src/transaction.ts
```diff
@@ -1,9 +1,11 @@
 import { Readable } from "svelte/store";
 
 import * as api from "./api";
+import { Avatar, getAvatar, Usage, EmojiAvatar } from "./avatar"
 import * as event from "./event";
 import * as remote from "./remote";
 import { Identity } from "./identity";
+import { Domain } from "./project"
```
xla, line 8: Can we order these import alphanumerically. #OCD

ui/src/avatar.ts
sarahscott: I get that but I also think wildcard imports lead to ugliness like `avatar.Avatar` or `avatar.EmojiAvatar`.

ui/DesignSystem/Component/Transaction.svelte
sarahscott: das macht sinn

PR #375, opened 2020-05-12
style(ui): settings

juliendonck:
## Done
- adds settings page ui
Settings screen
- adds new `destructive` button variant
Screenshot 2020-05-12 at 15 04 16

## todo
- [x] replace `tabs` with reusable component

Issue #128, opened 2020-02-06
Protocol network configurations


xla:
A point that will become more important the more the protocols will be deployed into the wild in different configurations, is how we will enable users of the application to confgure which network to connect to or to join. For the Registry we'll likely want a set of predefined ones (devnet-00, testnet-12, game-of-memes) on top of manual declaration of seed nodes. For CoCo there will always be the need for some set of nodes to be configured, this could be Monadic operated seed nodes, or down the road drag-and-droppable file with the network configuration inside.

xla:
Superseeded by other issues.

PR #383, opened 2020-05-13
feat(proxy): extend session with settings

xla:
In preparation of a range of user controlled parameters this change
extends the session with settings and the ability to update said
settings. Currently we only accept the entire settings struct to be
updated. In the future we'll likely allow more granular updates.

* add `Settings` to session
* add endpoint to update settings

Ref #128 
Ref #375

### DIFF
--- a/proxy/src/http/session.rs
+++ b/proxy/src/http/session.rs
@@ -1,5 +1,6 @@
 //! Endpoints and serialisation for [`session::Session`] related types.
 
+use std::collections::HashMap;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 use warp::document::{self, ToDocumentedType};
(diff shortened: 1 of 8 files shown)


proxy/src/http/session.rs
```diff
@@ -80,18 +175,19 @@ mod test {
...
         let registry = registry::Registry::new(radicle_registry_client::Client::new_emulator());
         let store = kv::Store::new(kv::Config::new(tmp_dir.path().join("store"))).unwrap();
-        let api = super::get_filter(
+        let api = super::filters(
             Arc::new(RwLock::new(registry)),
             Arc::new(RwLock::new(store)),
         );
 
-        let res = request().method("GET").path("/session").reply(&api).await;
+        let res = request().method("GET").path("/").reply(&api).await;
```
rudolfs, line 190: Is session now on `/`?

xla: Nope, it's mounted under "/session" in the `routes` function.

xla:
Part of #385

PR #395, opened 2020-05-15
refactor(proxy): check applied transaction result

### LINKED ISSUE #387: Check applied transaction response in proxy
See https://github.com/radicle-dev/radicle-upstream/blob/02576f02377556d893d3ad69c1a25214927b91ca/proxy/src/registry.rs#L403

xla:
Fixes #387

### DIFF
--- a/cypress/integration/org_registration.spec.js
+++ b/cypress/integration/org_registration.spec.js
@@ -1,5 +1,6 @@
 context("org registration", () => {
   beforeEach(() => {
+    cy.nukeRegistryState();
     cy.nukeSessionState();
     cy.createIdentity();
     cy.registerUser();
@@ -87,6 +88,8 @@ context("org registration", () => {
     });
 
     it("prevents the user from registering an unavailable org name", () => {
+      cy.registerOrg("coolname");
+
       cy.pick("org-reg-modal", "name-input").type("coolname");
       cy.pick("org-reg-modal").contains("Sorry, this name is already taken");
       cy.pick("org-reg-modal", "submit-button").should("be.disabled");
--- a/proxy/src/registry.rs
+++ b/proxy/src/registry.rs
@@ -172,7 +172,7 @@ pub struct Thresholds {
 #[serde(rename_all = "camelCase")]
 pub struct Org {
     /// The unique identifier of the org
-    pub id: String,
+    pub id: Id,
     /// Unambiguous identifier pointing at this identity.
     pub shareable_entity_identifier: String,
     /// Generated fallback avatar
@@ -209,7 +209,7 @@ pub trait Client: Clone + Send + Sync {
     /// # Errors
     ///
     /// Will return `Err` if a protocol error occurs.
-    async fn get_org(&self, id: String) -> Result<Option<Org>, error::Error>;
+    async fn get_org(&self, id: Id) -> Result<Option<Org>, error::Error>;
 
     /// List orgs of the Registry.
     ///
@@ -226,7 +226,7 @@ pub trait Client: Clone + Send + Sync {
     async fn register_org(
         &self,
         author: &protocol::ed25519::Pair,
-        org_id: String,
+        org_id: Id,
         fee: protocol::Balance,
     ) -> Result<Transaction, error::Error>;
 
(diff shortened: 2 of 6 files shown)


proxy/src/registry.rs
```diff
@@ -403,21 +399,24 @@ impl Client for Registry {
...
+                block: Hash(applied.block),
             },
             timestamp: SystemTime::now(),
         };
 
         // TODO(xla): Remove autmoatic prepayment once we have proper balances.
-        let org = self.client.get_org(org_id).await?.expect("org not present");
+        let org = self
+            .client
+            .get_org(org_id.0)
```
cloudhead, line 417: This is a bit ugly -- it's weird that some functions don't expect the `Id`, and others do.

proxy/src/registry.rs
```diff
@@ -445,30 +441,28 @@ impl Client for Registry {
...
+        org_id: Id,
+        project_name: ProjectName,
     ) -> Result<Option<Project>, error::Error> {
-        let org_id = protocol::Id::try_from(id.clone())?;
-        let project_name = protocol::ProjectName::try_from(project_name)?;
-
         Ok(self
             .client
-            .get_project(project_name, org_id)
+            .get_project(project_name.0, org_id.0)
```
cloudhead, line 465: Same for project_name

proxy/src/registry.rs
xla: This is a method on the underyling client not on our own implementation.

proxy/src/registry.rs
xla: Same as the other comment.

proxy/src/registry.rs
cloudhead: Since you don't control the API, perhaps having an `into()` that does what you want is cleaner, since it's a one-way function, vs. giving full access to the inner value.

cloudhead: Or `as_str()` if that's what you need.

cypress/integration/org_registration.spec.js
```diff
@@ -87,6 +88,8 @@ context("org registration", () => {
     });
 
     it("prevents the user from registering an unavailable org name", () => {
+      cy.registerOrg("coolname");
```
rudolfs, line 91: It would be nice if this would be decoupled from the other test.

proxy/src/registry.rs
```diff
@@ -486,16 +480,15 @@ impl Client for Registry {
...
-    async fn list_org_projects(&self, id: String) -> Result<Vec<Project>, error::Error> {
-        let org_id = protocol::Id::try_from(id.clone())?;
-        let project_ids = self.client.list_projects().await?.into_iter();
+    async fn list_org_projects(&self, org_id: Id) -> Result<Vec<Project>, error::Error> {
+        let ids = self.client.list_projects().await?;
         let mut projects = Vec::new();
-        for project_id in project_ids {
-            if project_id.1 == org_id {
+        for id in &ids {
+            if id.1 == org_id.clone().0 {
```
cloudhead, line 487: Would be good to have a getter here instead of the `.1`, eg. `id.domain() == org_id.as_str()` would be more readable

proxy/src/registry.rs
cloudhead: You can then make the inner value non-`pub`

proxy/src/registry.rs
xla: This would be feedback for the radice-registry client implementation.

cypress/integration/org_registration.spec.js
xla: @rudolfs Could you expand what you mean exactly?

rudolfs: Ah, never mind, I missed the `nukeRegistryState` in the before. All good!

proxy/src/registry.rs
cloudhead: Oh I see :(

proxy/src/registry.rs
xla: As these newtypes are just stop-gap solutions until we have reflected these concerns upstream in radicle-registry I will hold off to improve the ergonomics for now. Valid input tho and thanks for pointing it out.

PR #407, opened 2020-05-19
feat: tx polling

xla:
Implements constant polling of transaction progress throughout the
stack. In that context we update our understanding of transction states
and what info those carry.

Part of #347


ui/Screen/DesignSystemGuide.svelte
```diff
@@ -111,37 +112,110 @@
 
   const transactions1 = [
     {
-      message: "Project registration",
-      state: "pending",
-      progress: 50,
```
MeBrei, line 116: There should be one case to visualize the `confirmed` state.

ui/DesignSystem/Primitive/Icon/TransactionState.svelte
```diff
@@ -20,20 +21,19 @@
...
 
-  const defaultDashLength = {
-    caution: progress || 100 / 6,
-    positive: 100,
-    negative: 0,
-  };
-
-  $: dashLength =
-    progress === 0 ? 100 / 6 : progress || defaultDashLength[state];
+  $: dashLength = progress === 0 ? defaultDashLength : progress;
```
MeBrei, line 36: The mixed states are not shown properly anymore, I believe because of this change. It should be possible to be in the `negative` variant (red) but with a rotating dash.

proxy/src/http/transaction.rs
```diff
@@ -173,17 +226,20 @@ mod test {
...
             id: registry::Hash(radicle_registry_client::TxHash::random()),
             messages: vec![registry::Message::ProjectRegistration {
                 project_name: registry::ProjectName::try_from("upstream").unwrap(),
                 org_id: registry::Id::try_from("radicle").unwrap(),
             }],
-            state: registry::State::Applied {
-                block: registry::Hash(radicle_registry_client::Hash::random()),
+            state: registry::State::Confirmed {
+                block: 1,
+                progress: 1,
```
cloudhead, line 239: According to the API docs, this would start at `0`. However if this is the number of confirmations, this is correct, and the docs are wrong.

proxy/src/registry.rs
```diff
@@ -204,6 +203,12 @@ pub struct User {
 /// Methods to interact with the Registry in a uniform way.
 #[async_trait]
 pub trait Client: Clone + Send + Sync {
+    /// Fetch latest height by virtue of checking the block header of the best chain.
+    ///
+    /// # Errors
+    ///
+    /// Will return `Err` if a protocol error occurs.
+    async fn latest_height(&self) -> Result<u32, error::Error>;
```
cloudhead, line 211: This should maybe be called `chain_height` or `best_height`, I think using "latest" could be confusing, because it's not a matter of time.

proxy/src/registry/transaction.rs
```diff
@@ -4,13 +4,43 @@ use async_trait::async_trait;
...
+use std::time::{self, Duration, SystemTime};
 
 use radicle_registry_client as protocol;
 
 use crate::error;
 use crate::registry;
 
+/// Amount of blocks we assume to have been mined before a transaction is
+/// considered to have settled.
+const SETTLEMENT_OFFSET: u32 = 6;
```
cloudhead, line 16: More commonly known as "minimum confirmations". Also note that you're going to be off by one if you treat this as an offset, because when the tx is applied, that counts as the 1st confirmation, so you only need 5 more after that.

proxy/src/registry/transaction.rs
```diff
@@ -65,10 +114,36 @@ pub enum Message {
...
-    Applied {
-        /// The hash of the block the transaction has been applied to.
-        block: registry::Hash,
+    /// [`Transaction`] has been applied to a block, carries the height of the block.
+    Confirmed {
+        /// The height of the block the transaction has been applied to.
+        block: protocol::BlockNumber,
+        /// Amount of progress made towards settlement. We assume height+6 to
+        /// be mathematically impossible to be reverted.
+        progress: u32,
```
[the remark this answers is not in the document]
cloudhead: Ok, seems like it's the docs that are incorrect, since I never see `progress: 0`

proxy/src/registry/transaction.rs
```diff
@@ -65,10 +114,36 @@ pub enum Message {
...
+        timestamp: Timestamp,
+    },
+
+    /// [`Transaction`] failed to be applied or processed.
+    // TODO(xla): Embbed original [`protocol::TransactionError`] and serialize it.
+    Failed {
+        /// Description of the error that occurred.
+        error: String,
+        /// Time when it failed.
+        timestamp: Timestamp,
```
cloudhead, line 134: BlockNumber could be useful here too.

proxy/src/registry/transaction.rs
```diff
@@ -127,6 +202,60 @@ where
...
+                        tx.state = State::Confirmed {
+                            block,
+                            progress,
+                            timestamp,
+                        };
+                    }
+
+                    self.cache_transaction(tx.clone())?;
+                },
+                State::Failed { .. } | State::Pending { .. } | State::Settled { .. } => continue,
```
cloudhead, line 253: Don't we want to do something if it failed?

proxy/src/registry/transaction.rs
```diff
@@ -127,6 +202,60 @@ where
...
+        let mut txs = self.list_transactions(vec![])?;
+
+        for tx in &mut txs {
+            match tx.state {
+                State::Confirmed {
+                    block, timestamp, ..
+                } => {
+                    let target = block
+                        .checked_add(SETTLEMENT_OFFSET)
+                        .unwrap_or(SETTLEMENT_OFFSET);
```
cloudhead, line 232: If `SETTLEMENT_OFFSET == MIN_CONFIRMATIONS - 1`, then this is correct. Otherwise it's off by one.

proxy/src/registry/transaction.rs
```diff
@@ -127,6 +202,60 @@ where
...
+
+                    if latest_height >= target {
+                        tx.state = State::Settled {
+                            timestamp: Timestamp::now(),
+                        };
+                    } else {
+                        let offset = latest_height
+                            .checked_add(SETTLEMENT_OFFSET)
+                            .unwrap_or(SETTLEMENT_OFFSET);
+                        let progress = offset.saturating_sub(target);
```
cloudhead, line 242: Shouldn't this be `latest_height - block + 1`?

ui/src/transaction.ts
```diff
@@ -60,44 +70,120 @@ interface UserRegistration {
 type Message
   = OrgRegistration
   | OrgUnregistration
-  | OrgMemberRegistration
-  | OrgMemberUnregistration
+  | MemberRegistration
+  | MemberUnregistration
```
cloudhead, line 74: :partying_face:

ui/src/transaction.ts
```diff
@@ -60,44 +70,120 @@ interface UserRegistration {
...
   | ProjectRegistration
   | UserRegistration;
 
 export enum StateType {
-  Applied = "applied",
```
cloudhead, line 69: :+1:

ui/src/transaction.ts
```diff
@@ -60,44 +70,120 @@ interface UserRegistration {
...
 
+export const summarizeTransactions = (txs: Transactions): Summary =>
+  txs.reduce((acc, tx): Summary => {
+      acc.counts.sum += 1;
+      acc.counts[tx.state.type] += 1;
+      
+      if (tx.state.type === StateType.Confirmed) {
+        acc.progress += tx.state.progress;
+      } else if (tx.state.type === StateType.Settled) {
+        acc.progress += 6;
```
cloudhead, line 140: A magic number `6` appears

proxy/src/registry/transaction.rs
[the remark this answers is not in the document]
cloudhead: If you want this to be frontend-friendly, then I'd suggest having two values: `confirmations: u32, min_confirmations: u32`. That way it's easy to calculate progress (`min_confirmations/confirmations`), and also you avoid having to define the magic number `6` in the frontend.

ui/DesignSystem/Component/Transaction/List.svelte
```diff
@@ -0,0 +1,23 @@
+<script>
+  import { createEventDispatcher } from "svelte";
+
+  import Item from "./Item.svelte";
+
+  export let transactions = null;
+
+  const disaptch = createEventDispatcher();
```
sarahscott, line 8:
```suggestion
  const dispatch = createEventDispatcher();
```

ui/DesignSystem/Component/Transaction/List.svelte
```diff
@@ -0,0 +1,23 @@
...
+  const select = (id) => disaptch("select", id);
```
sarahscott, line 9:
```suggestion
  const select = (id) => dispatch("select", id);
```

ui/DesignSystem/Component/Transaction/Summary.svelte
```diff
@@ -0,0 +1,46 @@
...
+    display: flex;
+  }
+</style>
+
+<div class="summary" on:click>
+  <Flex>
+    <div slot="left" class="text">
+      <Icon.TransactionState
+        {progress}
+        required={transaction.REQUIRED}
```
sarahscott, line 32: What is a 'required' transaction?

ui/Screen/DesignSystemGuide.svelte
```diff
@@ -25,7 +26,7 @@
     Stats,
     StepCounter,
     TrackToggle,
-    TransactionAccordion,
+    TransactionCenter,
```
sarahscott, line 29: Nice name change

ui/Screen/DesignSystemGuide.svelte
```diff
@@ -835,17 +909,20 @@
...
               transactions={transactions2}
               style="position: absolute; bottom: 0; right: 0;" />
           </div>
           <div style="position: relative; height: 200px; width: 280px;">
-            <TransactionAccordion
+            <TransactionCenter
+              summary={transaction.summarizeTransactions(transactions3)}
               transactions={transactions3}
               style="position: absolute; bottom: 0; right: 0;" />
           </div>
```
sarahscott, line 928: The `TransactionStatusBar`'s are looking a little funny:

Screen Shot 2020-05-19 at 5 32 50 PM

ui/src/remote.ts
```diff
@@ -44,7 +44,7 @@ declare type Starter = () => void;
 export const createStore = <T>(): Store<T> => {
   let starter: Starter | null;
   const initialState = { status: Status.NotAsked } as Data<T>
-  const internalStore = writable(initialState, () => { starter && starter() })
+  const internalStore = writable(initialState, () => { if (starter) { return starter() } })
```
sarahscott, line 47: Can't you just return `starter()` here with the same effect?

ui/src/transaction.ts
```diff
@@ -1,4 +1,5 @@
-import { Readable } from "svelte/store";
+import { Readable, Writable, get, derived, writable } from "svelte/store";
+import * as timeago from "timeago.js";
```
sarahscott, line 2: 🎉 nice extraction

ui/src/transaction.ts
```diff
@@ -60,44 +70,120 @@ interface UserRegistration {
...
+  sum: number;
+}
+
+interface Summary {
+  counts: SummaryCounts;
+  progress: number;
+}
 
+export const summarizeTransactions = (txs: Transactions): Summary =>
+  txs.reduce((acc, tx): Summary => {
```
sarahscott, line 133: fancy fancy

ui/Screen/TransactionDetails.svelte
```diff
@@ -22,7 +22,9 @@
   if (session.identity !== null) {
     identity = session.identity;
   }
-  const store = fetch(params.id);
+
+  $: payer = transaction.formatPayer(identity);
```
sarahscott, line 26: Can one's `identity` pay for transactions? Does this need to be a `user`?

ui/DesignSystem/Primitive/Icon/TransactionState.svelte
xla: Addressed in 69b77fc

ui/Screen/TransactionDetails.svelte
xla: The formatting needs avatar information and therefore relies on identity. This makes sense I think.

ui/DesignSystem/Component/Transaction/Summary.svelte
xla: It's the minimum amount of blocks that have to be mined on top of the one the tx is in for it to be settled.

ui/src/remote.ts
xla: That would crash if starter is undefined.

proxy/src/registry/transaction.rs
```diff
@@ -65,10 +114,36 @@ pub enum Message {
...
+    /// [`Transaction`] has been send but not yet applied to a block.
+    Pending {
+        /// Time when it was applied.
+        timestamp: Timestamp,
+    },
+
+    /// [`Transaction`] has been settled on the network and is unlikely to be reverted.
+    Settled {
+        /// Time when it settled.
+        timestamp: Timestamp,
```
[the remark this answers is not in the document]
xla: The reason we need this timestamp at the moment is that the visual specs show an `at` time for when the tx settled. Same for failure.

proxy/src/registry/transaction.rs
xla: The failure could be before the tx was applied, in this case we wouldn't have a block. We could make it a maybe.

proxy/src/registry/transaction.rs
xla: This routine is about updating the progress (soon renamed to number of confirmations), if tx is already in a failed state there is nothing to update.

MeBrei:
1 example:
I cleared the cache via settings and register my user handle
-> transaction center shows:
 - summary: pending tx but the icon is not rotating
 - 1 tx card with status "confirmed" and non rotating icon
I open the detail view of the transaction -> the icon in the status bar is rotating
I close the detail view:
-> transaction center shows:
 - summary: pending tx but the icon is not rotating
 - 1 tx card with status "confirmed" and this time the icon is rotating

xla:
> summary: pending tx but the icon is not rotating

This is according to the design specs, unless there is at least one tx in the list which is not settled or failed it will always show pending. We could push for changes in the design specs to reflect the confirmation state.

The rest should be addressed in d193b0f.

MeBrei:
> > summary: pending tx but the icon is not rotating
> 
> This is according to the design specs, unless there is at least one tx in the list which is not settled or failed it will always show pending. We could push for changes in the design specs to reflect the confirmation state.
> 

but the icon of the summary should still rotate if there is only one unsettled tx, right?

xla:
> but the icon of the summary should still rotate if there is only one unsettled tx, right?

@MeBrei Can't tell from the design specs, doesn't look like it to me, maybe I'm interpreting them wrong.

proxy/src/registry/transaction.rs
cloudhead: Yeah, could be a maybe. Though I'm not entirely sure what kind of failures can happen before the tx is applied. Would be good to figure that out because I think it's a very different type of failure.

proxy/src/registry/transaction.rs
cloudhead: Ah ok, so you get the information about failure from somewhere else I suppose?

MeBrei:
- [ ] In the Design System Guide, the tx summary in the middle has a static icon, although both transactions are pending

Screenshot 2020-05-20 at 15 57 04

- [ ] for me, the progress of confirmed transactions never advance. I have 3 different transactions, the api response shows advancing `block` entries but `confirmations` is 1 for all of them

proxy/src/registry/transaction.rs
xla: Yeah it's in the initial transaction submission. We haven't figured out a way yet to determine if a tx is not part of the best chain anymore.

xla:
@cloudhead @MeBrei @sarahscott Thanks for the thorough reviews and extremely valuable input. I addressed all outstanding feedback. With the latest changes of the registry we pulled in, transactions should advance automatically on a roughly 10s timer, all affected parts of the UI should be reactive to these updates. Please take another look and let me know if we are good to go.

proxy/src/registry/transaction.rs
xla: Failures before application to a block include codec, rpc and tx invalid errors. For reference `radicle_registry_client::Error`. (docs.rs doesn't have the latest for the crate so I can't link to it)

ui/Screen/TransactionDetails.svelte
sarahscott: Ok, since we haven't implemented wallets yet I guess that makes sense. I've been operating under the assumption that anything registry-related depends on a `user`, and the avatar might come from an `identity` attested to by that `user`.

sarahscott:
Looks absolutely sick, amazing work, 🚀 it

proxy/src/registry/transaction.rs
cloudhead: Cool yeah, that's what I expected. This is most likely client errors then, eg. a bug in the client, or the client needs to be upgraded.

Issue #302, opened 2020-04-22
Create variants for different size icons


juliendonck:
Merge different size icons into variants in stead of seperate files, e.g.:
```
ArrowDown.svelte
ArrowUp.svelte
Carret.svelte
CarretBig.svelte
Check.svelte
CheckCircle.svelte
Cross.svelte
CrossBig.svelte
Ellipse.svelte
EllipseBig.svelte
Search.svelte
SearchSmall.svelte
```

rudolfs:
Closed by #411.

PR #411, opened 2020-05-20
style(ui): icon refresh

juliendonck:
Updated the icons that were cleaned up by @brandonhaslegs and reworked how they are presented in the design system

Screenshot 2020-05-20 at 17 59 27


brandonhaslegs:
All icons from the icons file aren't represented here, is that intentional?

brandonhaslegs:
The box/projects icon wasn't updated.

xla:
@brandonhaslegs You can use the review feature, to request changes to the PR, so the assignee knows there is outstanding feedback to be addressed.

ui/DesignSystem/Primitive/Icon/Cross.svelte
```diff
@@ -1,50 +1,63 @@
...
       clip-rule="evenodd"
-      d="M7.29292 16.7071C6.9024 16.3166 6.9024 15.6834 7.29292 15.2929L15.2929
-      7.29289C15.6834 6.90237 16.3166 6.90237 16.7071 7.29289C17.0977 7.68342
-      17.0977 8.31658 16.7071 8.70711L8.70714 16.7071C8.31661 17.0976 7.68345
-      17.0976 7.29292 16.7071Z" />
-  </svg>
-{:else if (size = 'big')}
-  <svg
-    {style}
-    on:click
```
xla, line 32: These directives are important for events to bubble up to components who use these icons. It's the reason the tests in your last commit fail, check [this docs section](https://svelte.dev/docs#on_element_event) and especially the part on fowarding. I added this to all icons in 49dff34.

ui/DesignSystem/Primitive/Icon/Cross.svelte
```diff
@@ -1,50 +1,63 @@
 <script>
   export let style = null;
-  export let size = "small"; // normal | "big"
+  export let variant = "small"; // small | medium | big | filled | outline
```
sarahscott, line 3: I'm actually in favor of not conflating `variant` and `size`; they're different things. A `variant` should refer to the UI element's visual state depending on its intended context (e.g. `filled`, `square`, `error`).

juliendonck:
Closes #302

ui/DesignSystem/Primitive/Icon/Cross.svelte
rudolfs: I second that.

juliendonck:
> All icons from the icons file aren't represented here, is that intentional?

@brandonhaslegs I tried to see which ones we were actually using and which ones were a leftover. Are there any glaring ones you think are missing?

ui/DesignSystem/Primitive/Icon/Cross.svelte
juliendonck: Cool

ui/DesignSystem/Primitive/Icon/Cross.svelte
juliendonck: I can see that. But then that would mean that you'd need to have different sizes for each variants. Which is not always the case.

juliendonck: The way I would resolve this is by splitting out the "Filled in Cross" from the "Cross" component. Otherwise it seems messy.

brandonhaslegs:
> @brandonhaslegs I tried to see which ones we were actually using and which ones were a leftover. Are there any glaring ones you think are missing?

The only one glaring is the box/projects icon.

rudolfs:
Part of the section subtitle: "with 2px stroke weight, multiple color variations" should probably be removed for both icon sections. The icons have various strokes and colors variants are not shown in the design system.

ui/DesignSystem/Primitive/Icon/Cross.svelte
juliendonck: @sarahscott @rudolfs what do you think?

juliendonck:
> Part of the section subtitle: "with 2px stroke weight, multiple color variations" should probably be removed for both icon sections. The icons have various strokes and colors variants are not shown in the design system.

@rudolfs updated.

ui/DesignSystem/Component/Transaction/Accordion.svelte
```diff
@@ -0,0 +1,181 @@
+<script>
```
rudolfs, line 1: Where does this file come from?

ui/DesignSystem/Primitive/Icon/Cross.svelte
rudolfs: I had another look at our icons and I think we'll have to put in a bit more thought to clean them up properly, but that is beyond the scope of this PR.

To me these things stand out:
  - we have two ways of specifying variants: via component name (`ArrowDown.svelte`, `ArrowUp.svelte`) and via the `variant` prop; ideally we settle on one way of doing things
  - we're mixing components of different complexities and use cases together: plain icons, and various animated spinners and transaction components.

PR #413, opened 2020-05-20
feat (ui): unit testing

sarahscott:
This PR contains examples for testing the `org` module in jest and AVA. Here are my takeaways:

### **jest**

pros
-  watch mode is fast af; writing & refining  tests is a breeze
- well-documented
-  mocking system works perfectly out of the box
-  easy to setup/configure
-  robust typescript support
- svelte snapshots are already a thing if we want to go down that route

cons	
-  facebook repo

### **ava**
pros
- looks cuter in terminal
- assertions & chaining approach presents a compelling alternative to mocha
- highly diy/configurable 
- maintainers seem friendly & active
- chance to play around w interesting concurrency experiments in the future 

cons
- really, really slow compared to jest for a project of our size, as in i stared at its cute spinner for like 5 seconds every time i ran the most basic of tests (see: https://gist.github.com/jamestalmage/a7444257d844fb627c71)
- no mocking, would have to integrate & learn another dependency (https://github.com/avajs/ava/issues/1829#issuecomment-578265927)
- typescript support is still somewhat experimental
- watch mode isn't ready to support TDD or quick iteration; you have to wait for the thing to set up its entire dependency graph upon every save (https://github.com/avajs/ava/blob/master/docs/recipes/watch-mode.md#help-us-make-watch-mode-better)

tldr
I may be biased because I'm much more familiar with jest, but it seems like the right move for us at this stage due to its speed, reliability, & ease-of-use. If we want a lot of test coverage, we should reduce the friction to writing tests as much as possible, and jest seems like the friendliest option at this time. I wanted to like AVA, but it was painfully slow and we don't yet have the use cases that would take advantage of its flexibility.


package.json
```diff
@@ -66,13 +68,14 @@
...
     "husky": ">=4.2.3",
+    "jest": "^26.0.1",
     "lint-staged": "^10.2.2",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.0.5",
     "prettier-plugin-svelte": "^1.1.0",
     "rollup": "^2.8.0",
     "rollup-plugin-babel": "^4.3.3",
-    "rollup-plugin-commonjs": "^10.0.0",
+    "rollup-plugin-commonjs": "^10  .0.0",
```
xla, line 78:
```suggestion
    "rollup-plugin-commonjs": "^10 .0.0",
```

jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  roots: ["<rootDir>/ui/src"],
+  transform: {
+    "^.+\\.ts$": "ts-jest",
+  },
+  testRegex: "(/__tests__/.*|(\\.|/)(test))\\.ts$",
+  moduleFileExtensions: ["ts", "js", "json", "node"],
```
xla, line 7: Do we need the `node` extension?

jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  roots: ["<rootDir>/ui/src"],
+  transform: {
+    "^.+\\.ts$": "ts-jest",
+  },
+  testRegex: "(/__tests__/.*|(\\.|/)(test))\\.ts$",
```
xla, line 6: If we control the test regex, can we avoid the `__` convention?

package.json
```diff
@@ -83,13 +86,16 @@
...
     "svelte-spa-router": "^2.1.0",
+    "ts-jest": "^26.0.0",
     "typescript": "3.8.3",
     "wait-on": "^5.0.0"
   },
   "scripts": {
     "start": "run-p --race svelte:watch proxy:start electron:start",
     "test": "TZ='UTC' run-p --race proxy:start:test wait:test",
     "test:debug": "TZ='UTC' run-p --race svelte:watch proxy:start:test wait:debug",
+    "jest": "jest",
```
xla, line 97:
```suggestion
    "test:unit": "jest",
```

I'd recommend that we move to intention over technology names, this way we have stable commands no matter what runners are used under the hood.

.babelrc
```diff
@@ -5,9 +5,15 @@
...
-    ["@babel/plugin-transform-runtime", { "regenerator": true }]
+    [
+      "@babel/plugin-transform-runtime",
+      {
+        "regenerator": true
+      }
+    ]
   ]
-}
+}
```
xla, line 19: Your editor might need a config touch up to insert proper line breaks at the end of files.

ui/src/org.test.ts
```diff
@@ -0,0 +1,22 @@
+import { getOrg } from "./org"
+
+jest.mock("./api")
```
xla, line 3: magic

ui/src/__mocks__/api.ts
```diff
@@ -0,0 +1,26 @@
+import { Org } from "../org"
```
xla, line 1: Do we have to set up this boilerplate every time we want to mock an endpoint call?

package.json
sarahscott: Agree completely - I'm using `jest` so the cypress tests aren't affected by this experiment but I think `test:unit` should be the command we end up using

ui/src/__mocks__/api.ts
sarahscott: We only have to make one mock of the `api` module, but to make sure we use it in a particular test context we have to include `jest.mock("./api")` in the file.

jest.config.js
sarahscott: Genau. It's not the prettiest convention but it is immediately recognizable, so maybe we leave it? Open to thinking about a better directory name

ui/src/tests/org.test.ts
```diff
@@ -0,0 +1,43 @@
+import test from "ava"
+import { Org, orgIdValidationStore } from "../org"
+import { rewiremock } from '../rewiremock'
+import { get } from 'svelte/store'
+import { ValidationStatus } from '../validation'
+
+// doesn't actually work this simply out of the box
+// rewiremock("../api").with("../__mocks__/api.ts")
```
sarahscott, line 8: Didn't want to go too far down the rabbithole of third-party mocking libraries before we decide that we need one.

rudolfs:
Nice work! 💯 

For some reason, I'm having trouble getting it to work locally (maybe would make sense to have separate branches for different exploration options in the future):
```
work/radicle-upstream-2 » yarn
yarn install v1.22.4
[1/4] 🔍  Resolving packages...
warning rollup-plugin-commonjs@10.1.0: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-commonjs.
[2/4] 🚚  Fetching packages...
error ava@3.8.2: The engine "node" is incompatible with this module. Expected version ">=10.18.0 <11 || >=12.14.0 <12.16.0 || >=12.16.0 <13 || >=13.5.0 <14 || >=14.0.0". Got "11.11.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

I think I'd also prefer `jest` as I've worked with it before.

sarahscott:
@rudolfs good call - next time I'm trying out multiple new dependencies I'll make separate branches. I had to change my node version to be compatible with AVA as well. Thanks for the feedback! Will rework this into a jest-only mergeable PR.

package.json
```diff
@@ -83,13 +86,17 @@
...
     "svelte-spa-router": "^2.1.0",
+    "ts-jest": "^26.0.0",
+    "ts-node": "^8.10.1",
     "typescript": "3.8.3",
     "wait-on": "^5.0.0"
   },
   "scripts": {
     "start": "run-p --race svelte:watch proxy:start electron:start",
-    "test": "TZ='UTC' run-p --race proxy:start:test wait:test",
+    "test": "yarn test:unit && TZ='UTC' run-p --race proxy:start:test wait:test",
```
rudolfs, line 96:
```suggestion
    "test": "TZ='UTC' yarn test:unit && yarn test:integration",
    "test:integration": "run-p --race proxy:start:test wait:test",
```

ui/src/__mocks__/api.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  switch (prefix) {
+    case "orgs":
+      response = param === "radicle" ? radicleMock : null
+      break;
+    case "user":
+      response = userMock
+      break;
+  }
+
```
rudolfs, line 35:
```suggestion
```

ui/src/org.test.ts
```diff
@@ -0,0 +1,91 @@
...
+        background: {
+          r: 255, g: 67, b: 34
+        },
+        emoji: "🔥"
+      }
+    })
+  })
+})
+
+describe("validation", () => {
```
rudolfs, line 24: This is nice, we'll be able to remove the bulk of the integration validation tests and make the integration test suite faster.

yarn.lock
```diff
@@ -5101,7 +7061,7 @@ rollup-plugin-babel@^4.3.3:
     "@babel/helper-module-imports" "^7.0.0"
     rollup-pluginutils "^2.8.1"
 
-rollup-plugin-commonjs@^10.0.0:
+"rollup-plugin-commonjs@^10  .0.0":
```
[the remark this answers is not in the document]
sarahscott: This is so annoying lol my prettier formatter keeps adding this space here. Will remove

jest.config.js
sarahscott: nope

package.json
```diff
@@ -84,13 +87,18 @@
...
     "svelte-spa-router": "^2.1.0",
+    "ts-jest": "^26.0.0",
+    "ts-node": "^8.10.1",
     "typescript": "3.8.3",
     "wait-on": "^5.0.0"
   },
   "scripts": {
     "start": "run-p --race svelte:watch proxy:start electron:start",
-    "test": "TZ='UTC' run-p --race proxy:start:test wait:test",
-    "test:debug": "TZ='UTC' run-p --race svelte:watch proxy:start:test wait:debug",
```
rudolfs, line 93: "test:debug" is actually useful, I'll bring it back.

PR #414, opened 2020-05-22
feat(proxy): coco ascension

### LINKED ISSUE #434: Create new project branch selector shows remote branches
This dropdown should only contain local branches, no remote branches should be shown.

Screenshot 2020-05-29 at 13 18 52

xla:
Fixes #434


proxy/src/coco.rs
```diff
@@ -1,649 +1,457 @@
...
+    fn list_projects(&self) -> Vec<(RadUrn, project::Project)>;
+
+    /// Create a copy of the git-platinum repo, init with coco and push tags and the additional dev
+    /// branch.
+    ///
+    /// # Errors
+    ///
+    /// Will return [`error::Error`] if any of the git interaction fail, or the initialisation of
+    /// the coco project.
+    fn replicate_platinum(
```
FintanH, line 72: This seems like it shouldn't be in the client. It can be defined in terms of `init_project`, I think.

proxy/src/coco.rs
```diff
@@ -1,649 +1,457 @@
...
-pub struct Tag(String);
-
-impl fmt::Display for Tag {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.0)
-    }
+/// The set of data and capabilities that are needed for interacting with `radicle-link`.
+/// It implements [`Client`], which is a collection of these trait capabilities. `Client` should be
+/// used by functions up the stack, while `Coco` should be passed down from the top.
+pub struct Coco<
```
FintanH, line 83: Curious as to why the `User` isn't in here anymore? I thought it would make sense because someone would be logged in on a device -- that user -- but maybe it doesn't make sense?

proxy/src/coco.rs
```diff
@@ -1,649 +1,457 @@
...
+        _revision: u64,
+    ) -> Result<user::User, entity::Error> {
+        todo!()
     }
+}
 
-    Ok(())
+/// Constructs a fake user to be used as an owner of projects until we have more permanent key and
+/// user management.
+pub fn fake_owner() -> user::User {
```
FintanH, line 454: Another organisational thing: it'd probably make sense to have `mock` module for putting this and the `Noop` stuff into.

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,452 @@
...
+    #[must_use]
+    pub fn description(&self) -> &str {
+        self.message
+            .strip_prefix(&self.summary)
+            .unwrap_or(&self.message)
+            .trim()
+    }
+}
+
+impl From<&surf::git::Commit> for Commit {
```
FintanH, line 74: I think I mentioned it in the commit, but for posterity: I think it'd make more sense to own the `surf::Commit` here rather than cloning the fields separately. Worst comes to worse, we just have to clone the reference that uses the `From` call.

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,452 @@
...
+/// Will return [`error::Error`] if the project doesn't exist or a surf interaction fails.
+pub async fn blob<C: coco::Client>(
+    client: &C,
+    project_urn: String,
+    revision: Option<String>,
+    maybe_path: Option<String>,
+) -> Result<Blob, error::Error> {
+    let (mut repo, project) = client.get_repo(project_urn).await?;
+    let locked = repo.locked();
+    let mut browser = locked.browser()?;
```
FintanH, line 184: It annoyed me that I couldn't go a `get_browser` because of lifetimes. Maybe we could do something like `with_browser` in `Client` that allows us to scope the `Browser` and apply a call back?

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,452 @@
...
+        info,
+    })
+}
+
+/// Initialize a [`radicle_surf::git::git2::Repository`] at the given path.
+///
+/// # Errors
+///
+/// Will return [`error::Error`] if any of the git interactions break.
+pub fn init_repo(path: String) -> Result<(), error::Error> {
```
FintanH, line 433: It'd be good track down where this is used. Using the link backend to this would make more sense possibly?

proxy/src/http/control.rs
```diff
@@ -131,13 +142,15 @@ mod handler {
...
-        let new = Paths::from_root(dir.path()).expect("unable to get paths");
-
-        let mut paths = librad_paths.write().await;
-
-        *paths = new;
+    pub async fn nuke_coco<C>(_coco: http::Shared<C>) -> Result<impl Reply, Rejection>
+    where
+        C: coco::Client,
+    {
+        // let tmp = coco::Coco::tmp()?;
```
FintanH, line 149: What happened here?

proxy/src/coco.rs
```diff
@@ -1,649 +1,457 @@
...
+        description: &str,
+        default_branch: &str,
+    ) -> Result<(RadUrn, project::Project), error::Error> {
+        // Set up storage
+        let key = self.get_key()?.secret_key;
+        let storage = Storage::init(&self.paths, key)?;
+
+        // Fetch the owner and build the repo path
+        let path = uri::Path::from_str(path)?;
+        let urn = RadUrn::new(owner.root_hash().clone(), uri::Protocol::Git, path);
```
FintanH, line 247: @kim: For creating a project, should the `Path` component just be empty?

FintanH:
We're currently not verifying any of the entities. We should :)

proxy/src/coco.rs
```diff
@@ -1,649 +1,472 @@
...
```
FintanH, line 262:
```suggestion
        let urn = owner.urn();
```

proxy/src/http/control.rs
```diff
@@ -131,13 +132,12 @@ mod handler {
     }
 
     /// Reset the coco state by creating a new temporary directory for the librad paths.
-    pub async fn nuke_coco(librad_paths: Arc<RwLock<Paths>>) -> Result<impl Reply, Rejection> {
-        let dir = tempfile::tempdir().expect("tmp dir creation failed");
-        let new = Paths::from_root(dir.path()).expect("unable to get paths");
-
-        let mut paths = librad_paths.write().await;
+    pub async fn nuke_coco(_peer: http::Shared<coco::UserPeer>) -> Result<impl Reply, Rejection> {
```
xla, line 135: What we want here is to reset that the coco state, meaning having a new librad monorepo tree which is fresh and backed by a temp directory. We used to swap the `Paths`, dunno how to do it now, replace the `UserPeer` entirely maybe?

proxy/src/http/source.rs
```diff
@@ -259,59 +258,59 @@ fn tree_filter(
...
     pub async fn branches(
-        librad_paths: Arc<RwLock<Paths>>,
-        project_id: String,
-    ) -> Result<impl Reply, Rejection> {
-        let paths = librad_paths.read().await;
-        let branches = coco::branches(&paths, &project_id)?;
+        peer: http::Shared<coco::UserPeer>,
+        project_urn: String,
+    ) -> Result<impl Reply, Rejection>
+where {
```
xla, line 287:
```suggestion
    ) -> Result<impl Reply, Rejection> {
```

proxy/src/http/source.rs
```diff
@@ -713,42 +714,48 @@ mod test {
...
-    async fn blob() {
-        let tmp_dir = tempfile::tempdir().unwrap();
-        let librad_paths = Paths::from_root(tmp_dir.path()).unwrap();
-        let (platinum_id, _platinum_project) = coco::replicate_platinum(
-            &tmp_dir,
-            &librad_paths,
-            "git-platinum",
-            "fixture data",
-            "master",
+    fn assert_status(res: &Response<Bytes>) {
```
xla, line 720: It would be nice to have a `http::test::assert_response` which takes the expected `StatusCode` and deserialised response value, errors early if the statuscode doesn't match. Pretty much extracting this pattern:

``` rust
let have: Value = serde_json::from_slice(res.body()).unwrap();

assert_eq!(res.status(), StatusCode::OK);
assert_eq!(
    have,
    json!(registry::Org {
        id: org_id.clone(),
        shareable_entity_identifier: format!("%{}", org_id.to_string()),
        avatar_fallback: avatar::Avatar::from(&org_id.to_string(), avatar::Usage::Org),
        members: vec![user]
    })
);
```

proxy/src/main.rs
```diff
@@ -45,20 +45,21 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            .await
+            .expect("failed to create /tmp user peer");
+        user_peer
+            .setup_fixtures(tmp_path)
+            .await
+            .expect("fixture creation failed");
+        user_peer
     } else {
-        librad::paths::Paths::new()?
+        todo!()
```
xla, line 60: Production peer.

FintanH:
This currently depends on https://github.com/radicle-dev/radicle-link/pull/164

proxy/src/coco.rs
```diff
@@ -1,649 +1,358 @@
...
-            message: commit.message.clone(),
-            committer: Person {
-                name: commit.committer.name.clone(),
-                email: commit.committer.email.clone(),
-                avatar: avatar(&commit.committer.email),
-            },
-            committer_time: commit.author.time,
-        }
+impl Peer {
+    /// We create a default `Peer` using the `tmp_dir_path` we provide.
```
xla, line 66:
```suggestion
    /// We create a default `Peer` using the `config` we provide.
```

proxy/src/coco.rs
```diff
@@ -1,649 +1,358 @@
...
+    pub async fn new<I>(
+        config: net::peer::PeerConfig<discovery::Static<I, SocketAddr>>,
+    ) -> Result<Self, error::Error>
+    where
+        I: Iterator<Item = (peer::PeerId, SocketAddr)> + Send + 'static,
+    {
+        // Initialise the storage
+        let _ = storage::Storage::init(&config.paths, config.key.clone())?;
```
xla, line 74: What's does it return?

proxy/src/coco.rs
```diff
@@ -1,649 +1,358 @@
...
-        .list_branches(None)?
-        .into_iter()
-        .map(|b| Branch(b.name.name().to_string()))
-        .collect::<Vec<Branch>>();
-
-    branches.sort();
-
-    Ok(branches)
-}
+    /// Fetch a browser for the `project_urn` we supplied to this function.
```
xla, line 100:
```suggestion
    /// Fetch a repository for the `project_urn` we supplied to this function.
```

Says `browser` here but actually return the `repository`.

proxy/src/coco.rs
```diff
@@ -1,649 +1,291 @@
...
-    let tag_names = browser.list_tags()?;
-
-    let mut tags: Vec<Tag> = tag_names
-        .into_iter()
-        .map(|tag_name| Tag(tag_name.name().to_string()))
-        .collect();
-
-    tags.sort();
+    /// Get the project found at `project_urn`.
+    pub async fn get_project(
```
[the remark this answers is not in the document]
xla: Bump.

proxy/src/coco.rs
```diff
@@ -1,649 +1,358 @@
...
+    ) -> Result<project::Project<entity::Draft>, error::Error> {
+        // TODO(finto): we need the storage to be a resolver
+        let urn = project_urn.parse()?;
+        let project = self.resolve(&urn).await?;
+        Ok(project)
+    }
 
-    let history = browser.get();
-    let commit = history.first();
+    /// Initialize a [`librad::project::Project`] in the location of the given `path`.
```
xla, line 141:
```suggestion
    /// Initialize a [`librad::project::Project`] in the location of the given `path`.
```

It doesn't take a path anymore.

proxy/src/coco.rs
```diff
@@ -1,649 +1,358 @@
...
+                        .set_default_branch(default_branch.to_string())
+                        .add_key(key.public())
+                        .build()?;
+                meta.sign_owned(key)?;
+
+                let storage = api.storage().reopen()?;
+                let _repo = storage.create_repo(&meta)?;
+                Ok(meta)
+            })
+            .and_then(|res| res); // unfortunately, flatten is a nightly feature.
```
xla, line 172: Aren't we using a recent nightly anyway?

proxy/src/coco.rs
```diff
@@ -1,649 +1,358 @@
...
-
-    Ok(())
+/// Results in an error if the [`paths::Paths`] could not be created.
+pub fn default_config(
+    key: keys::SecretKey,
+    path: impl AsRef<std::path::Path>,
+) -> Result<PeerConfig, error::Error> {
+    let gossip_params = net::gossip::MembershipParams::default();
+    let listen_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 0);
+    // TODO(finto): could we initialise with known seeds from a cache?
```
xla, line 347: What would testify a known seed in here?

proxy/src/coco.rs
FintanH: A `Storage`, but we're doing it for the side-effect of initialising the mono-repo. I think it should really be done via something like https://github.com/radicle-dev/radicle-link/issues/160

proxy/src/coco.rs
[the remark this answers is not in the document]
FintanH: So, I think this function will later be in `PeerApi` and we'll just call `api.get_project(project_urn.parse()?)` wherever it's needed. It's living in here now because of faking the resolver.

proxy/Cargo.toml
```diff
@@ -25,18 +25,30 @@ serde = { version = "1.0", features = [ "derive" ] }
...
+thiserror = "1.0"
 tokio = { version = "0.2", features = [ "macros", "time" ] }
 url17 = { package = "url", version = "1.7" }
 url = "2.1"
 warp = { git = "https://github.com/radicle-dev/warp", branch = "openapi", features = [ "openapi" ] }
 
 [dependencies.librad]
 git = "https://github.com/radicle-dev/radicle-link.git"
-rev = "4cedafaeb46deee4ec11d0e4b779a1a0d124d599"
+rev = "416d6a411fc77b5f6dc408ad983ea73d0c76ee02"
```
FintanH, line 36:
```suggestion
rev = "1a60d565e0fa53eec85ff10dbde7aa97420b91f0"
```

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,427 @@
...
+use librad::surf;
+use librad::surf::git::git2;
+
+use crate::error;
+
+use super::Peer;
+
+/// Branch name representation.
+#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd, Deserialize, Serialize)]
+pub struct Branch(pub(super) String);
```
FintanH, line 16: Some food for thought: I wonder what the advantage of `proxy` defining these types is. Would it be better to reuse the `surf` types and add de/serialization to them?

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// Used to store file data.
+    Blob,
+}
+
+/// Set of extra information we carry for blob and tree objects returned from the API.
+pub struct Info {
+    /// Name part of an object.
+    pub name: String,
+    /// The type of the object.
+    pub object_type: ObjectType,
```
FintanH, line 123: I don't understand why we have `ObjectType` here when `Info` gets embedded in `Blob`, `Tree`, and `TreeEntry`.

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,427 @@
...
+
+/// Returns the [`Blob`] for a file at `revision` under `path`.
+///
+/// # Errors
+///
+/// Will return [`error::Error`] if the project doesn't exist or a surf interaction fails.
+pub fn blob(
+    peer: &Peer,
+    project_urn: &str,
+    default_branch: String, // TODO(finto): This should be handled by the broweser surf#115
```
FintanH, line 181: We can ditch this now I think. You were right the first time where we should fetch the project to get the default branch.

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,427 @@
...
+    revision: Option<String>,
+    maybe_path: Option<String>,
+) -> Result<Blob, error::Error> {
+    let api = peer.api.lock().map_err(|_| error::Error::LibradLock)?;
+    let repo = api.storage().open_repo(project_urn.parse()?)?;
+    let mut browser = repo.browser()?;
```
FintanH, line 187: Once https://github.com/radicle-dev/radicle-link/pull/174 lands, I think we can do:
```rust
let mut browser = repo.browser(revision.unwrap_or_else(|| default_branch))?;
```

And get rid of `browser.revspec(&revision)`

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,427 @@
...
+///
+/// # Errors
+///
+/// Will return [`error::Error`] if the project doesn't exist or a surf interaction fails.
+pub fn blob(
+    peer: &Peer,
+    project_urn: &str,
+    default_branch: String, // TODO(finto): This should be handled by the broweser surf#115
+    revision: Option<String>,
+    maybe_path: Option<String>,
```
FintanH, line 183: Why's the path optional actually? :thinking:

proxy/src/coco/source.rs
```diff
@@ -0,0 +1,427 @@
...
+    branches: Vec<Branch>,
+    managed: bool,
+}
+
+/// Given a path to a repo returns the list of branches and if it is managed by coco.
+///
+/// # Errors
+///
+/// Will return [`error::Error`] if the repository doesn't exist.
+pub fn local_state(repo_path: &str) -> Result<LocalState, error::Error> {
```
FintanH, line 254: Is this needed anymore?

proxy/src/http/control.rs
```diff
@@ -115,23 +126,26 @@ mod handler {
...
-            &paths,
-            &input.name,
-            &input.description,
-            &input.default_branch,
-        )?;
+        let owner = &*owner.read().await;
+        let mut peer = peer.lock().await;
+
+        let meta = peer
+            .replicate_platinum(
```
FintanH, line 137: Ah `replicate_platinum` is here in pub code. I think we should create a ticket to remove this since we're able to browse existing projects now and whatnot. So creating a project should really be creating a project from scratch.

proxy/src/http/control.rs
```diff
@@ -163,13 +177,25 @@ mod handler {
...
+            let temp_dir = tempfile::tempdir().expect("test dir creation failed");
+            temp_dir.path().to_path_buf()
+        };
 
-        let mut paths = librad_paths.write().await;
+        let config = coco::default_config(
+            SecretKey::new(),
+            tmp_path.to_str().expect("path extraction failed"),
+        )?;
+        let new_peer = coco::Peer::new(config).await?;
```
FintanH, line 195: Something I'm worried about is when we start handling the run loop in peer. How will this work with nuking o_O

proxy/src/project.rs
```diff
@@ -67,13 +72,15 @@ pub struct Stats {
...
+    let meta = peer.get_project(project_urn).await?;
+    let id = meta.urn();
+    let metadata = meta.into();
 
     Ok(Project {
-        id: librad::project::ProjectId::from_str(id)?,
-        shareable_entity_identifier: format!("%{}", id),
-        metadata: meta.into(),
+        id,
+        shareable_entity_identifier: project_urn.to_string(),
```
FintanH, line 82:
```suggestion
        shareable_entity_identifier: format!("%{}", project_urn)),
```
This might be why the project page broke?

xla: Nah it was incorrect UI behaviour see a17d58c.

FintanH: ach so, but do we still expect the `%` before the URN?

proxy/rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2020-02-05
+nightly-2020-06-09
```
rudolfs, line 1: This will need a docker image update for CI, right?

yarn.lock
```diff
@@ -8600,6 +8600,11 @@ yargs@^15.3.1:
     y18n "^4.0.0"
     yargs-parser "^18.1.1"
 
+yarn@^1.22.4:
```
rudolfs, line 8603: Wonder where this is coming from?

proxy/rust-toolchain
xla: See 3aa9fcc

proxy/src/coco.rs
```diff
@@ -1,649 +1,499 @@
...
-                last_commit: None,
-            };
+    /// Create a copy of the git-platinum repo, init with coco and push tags and the additional dev
+    /// branch.
+    ///
+    /// # Errors
+    ///
+    /// Will return [`error::Error`] if any of the git interaction fail, or the initialisation of
+    /// the coco project.
+    pub async fn replicate_platinum(
```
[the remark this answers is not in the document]
xla: It is used in the control endpoints, which is non-test code but used for UI integration tests.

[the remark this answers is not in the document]
FintanH: Ya fair enough. I'd like to eventually separate it so it doesn't become canonical.

proxy/src/coco/source.rs
xla: This is to distinguish tree entries in the returned list.

proxy/src/http/control.rs
xla: CHANNELS!

yarn.lock
xla: Me too, just ran `yarn` in the top level.

proxy/src/http/control.rs
```diff
@@ -146,13 +160,25 @@ mod handler {
...
     /// Reset the coco state by creating a new temporary directory for the librad paths.
-    pub async fn nuke_coco(librad_paths: Arc<RwLock<Paths>>) -> Result<impl Reply, Rejection> {
-        let dir = tempfile::tempdir().expect("tmp dir creation failed");
-        let new = Paths::from_root(dir.path()).expect("unable to get paths");
+    pub async fn nuke_coco(peer: Arc<Mutex<coco::Peer>>) -> Result<impl Reply, Rejection> {
+        // TmpDir deletes the temporary directory once it DROPS.
+        // This means our new directory goes missing, and future calls will fail.
+        // The Peer creates the directory again.
+        //
+        // N.B. this may gather lot's of tmp files on your system. We're sorry.
```
rudolfs, line 168: LOL. So the temp files will never be cleaned up after the tests finish?

ui/Screen/ProjectCreation.svelte
```diff
@@ -198,9 +195,9 @@
     beginValidation = true;
 
     try {
-      localBranches = await getLocalBranches(path);
+      localState = await getLocalState(path);
```
rudolfs, line 198: The branch selector should not contain any remote branches. See: https://github.com/radicle-dev/radicle-upstream/issues/434.

yarn.lock
rudolfs: Looks like it's coming from adding `yarn` to `package.json`.

rudolfs:
I also noticed that the URNs are not consistent anymore:

Screenshot 2020-06-15 at 10 26 44
Screenshot 2020-06-15 at 10 28 02

proxy/src/http/control.rs
FintanH: Nope... Only when you reset your computer :innocent:

FintanH:
> Biggest question is: why do my project disappear after I restart the app?

@rudolfs: is this happening running the application in "production" mode? It might be something I need you to walk myself or xla through :eyes:

proxy/src/http/control.rs
rudolfs: I see. And where do these files get created when we run it on CI?

rudolfs:
@FintanH what is the "production mode"? I start it as I've always done with `yarn start`, create my projects, quit the app and start it again with `yarn start`.

FintanH:
> @FintanH what is the "production mode"? I start it as I've always done with yarn start, create my projects, quit the app and start it again with yarn start.

That sounds like "production mode" to me :sweat_smile: There's a flag that checks are we testing or not and depending on that the setup for coco is different, i.e. uses tmp dirs.

I would expect the projects to be persisted so we should look  into that.

proxy/src/http/control.rs
FintanH: Probably under `/tmp` or `/run/user/1000/.tmp*`. These get cleared out by the OS upon restart, afaik :thinking:

.buildkite/Dockerfile
```diff
@@ -15,4 +15,3 @@ RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
 RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
 RUN apt-get -y update
 RUN apt-get -y install yarn
-RUN rustup component add clippy --toolchain nightly-2020-02-05-x86_64-unknown-linux-gnu
```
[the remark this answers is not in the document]
xla: Nope, it's just part of the base image.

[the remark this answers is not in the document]
rudolfs: And does this registry base image have the same rust-toolchain version the coco dependency? Will both dependencies always be on the same version?

proxy/src/http/control.rs
rudolfs: I guess on CI that would be in the docker image which gets thrown away after the tests run. So should be okay.

rudolfs:
> I would expect the projects to be persisted so we should look into that.

> xla: The reason the state is wiped on restart is that we fake the resolvers with an in-memory hashmap for now.

Will be addressed in: https://github.com/radicle-dev/radicle-upstream/issues/456

.buildkite/Dockerfile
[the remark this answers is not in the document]
xla: We currently don't actively coordinate the toolchain versions across projects. Historically registry was the first to adopt nightly, so we have been trailing theirs. They also off the base image conveniently. This could be raised in radicle-decisions as a cross-engineering concern.

Issue #349, opened 2020-05-06
Feature: Members


xla:
Outstanding work for member management.

**proxy**

- [x] list members for org
- [x] add member endpoint
- [x] ~remove member endpoint~ (later feature)
- [x] refactor user org list to filter by handle instead of the author key

**proxy <> ui**
- [x] use registration endpoint when submitting the transaction in the modal
- [x] validation: do not allow registering an existing member

**ui**

- [ ] add member flow
  - [x] first member flow iteration ( #334 )
  - [ ] show avatar/imageURL of user in text input
- [ ] member list
  - [x] first member list iteration ( #402 )
  - [ ] navigate to user page
  - [x] use actual member list
- [x] ensure correct handling of member txs in transaction center
- [x] thorough testing: additional users currently can only be created in the testing environment

Part of #277

MeBrei:
### Open tasks

1. **Navigate to the user page from the org member list** it is currently not possible in the ui to show the user page of a different user than stored in the session. As far as I can see, there is also no figma screen for a "different" user. So should this be postponed and remain a console log for now?
2. **Copying the id of the member** Currently only the `User` is returned for the org member list. This is waiting on the decisions/implementation of partial user data and how we will connect User and Identity data.
3. **Show avatar/imageURL of the user** It should be shown in the member registration modal and the members list. This is waiting on the decisions/implementation of partial user data and how we will connect User and Identity data as well.
4. **Validating if the user is already a member of the org** This could be added as a validation in the ui as a first version, but validations will move into the proxy in the long run #166 

The action dropdown for the individual members in the org member list currently differs from the one shown in figma, I would argue changing it once the identity and navigation is in place.

cloudhead:
> Navigate to the user page from the org member list

As you noted, this isn't designed, and so is out of scope of this feature.
We *are* working on user profiles, which is part of `Identities 1.` and we're hoping to get in the B1 release. But I would consider that a different feature.

MeBrei:
Closing this issue, as the remaining **open tasks** are actually part of other features and problems.

PR #416, opened 2020-05-22
Register member endpoint

MeBrei:
This implements the api endpoint to register a user as a member under an org. In a followup PR, it will be integrated in the ui and the transaction views will be adjusted to support the new transaction type.

part of #349 

**Note** This can't fully be tested via the api playground, as we use the same `author`/key_pair for all transactions. This means we currently can't register a second user handle to then register it as a member.

### DIFF
--- a/proxy/src/http/org.rs
+++ b/proxy/src/http/org.rs
@@ -137,8 +137,8 @@ fn register_filter<R: registry::Client>(
     http::with_shared(registry)
         .and(http::with_subscriptions(subscriptions))
         .and(warp::post())
-        .and(warp::body::json())
         .and(path::end())
+        .and(warp::body::json())
         .and(document::document(document::description(
             "Register a new unique Org",
         )))
@@ -166,8 +166,8 @@ fn register_member_filter<R: registry::Client>(
     .and(warp::post())
     .and(document::param::<String>("id", "Unique ID of the Org"))
     .and(path("members"))
-    .and(warp::body::json())
     .and(path::end())
+    .and(warp::body::json())
     .and(document::document(document::description(
         "Register a member",
     )))
--- a/proxy/src/http/org.rs
+++ b/proxy/src/http/org.rs
@@ -24,7 +24,10 @@ pub fn routes<R: registry::Client>(
         get_filter(Arc::clone(&registry))
             .or(get_project_filter(Arc::clone(&registry)))
             .or(get_projects_filter(paths, Arc::clone(&registry)))
-            .or(register_filter(Arc::clone(&registry), subscriptions.clone()))
+            .or(register_filter(
+                Arc::clone(&registry),
+                subscriptions.clone(),
+            ))
             .or(register_member_filter(registry, subscriptions)),
     )
 }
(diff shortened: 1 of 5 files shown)


proxy/src/http/org.rs
```diff
@@ -250,6 +281,30 @@ mod handler {
...
+
+    /// Register a member under an org on the Registry.
+    pub async fn register_member<R: registry::Client>(
+        registry: http::Shared<R>,
+        subscriptions: notification::Subscriptions,
+        id: String,
+        input: super::RegisterMemberInput,
+    ) -> Result<impl Reply, Rejection> {
+        // TODO(xla): Get keypair from persistent storage.
+        let fake_pair = radicle_registry_client::ed25519::Pair::from_legacy_string("//Alice", None);
```
sarahscott, line 293: Just reading through this to get more familiar with proxy - is this keypair specific to the user and necessary for authenticating every transaction on the registry?

proxy/src/http/org.rs
```diff
@@ -250,6 +281,30 @@ mod handler {
...
+        // TODO(xla): Use real fee defined by the user.
+        let fake_fee: Balance = 100;
+
+        let reg = registry.read().await;
+        let org_id = registry::Id::try_from(id)?;
+        let handle = registry::Id::try_from(input.handle)?;
+        let tx = reg.register_member(&fake_pair, org_id, handle, fake_fee).await?;
+
+        subscriptions
+            .broadcast(notification::Notification::Transaction(tx.clone()))
```
sarahscott, line 303: sick

proxy/src/http/org.rs
MeBrei: Once a user handle is registered with a specific author key, this key will be associated with the user on the registry. Not only does every transaction has to be authored with a key, some transaction require the association with the user handle to check for permissions to do the transaction (eg registering a member under an org can only be done by a key associated with a user who is a member of the org)

proxy/src/registry.rs
```diff
@@ -597,6 +643,8 @@ impl Client for Registry {
         id: Option<String>,
         fee: protocol::Balance,
     ) -> Result<Transaction, error::Error> {
+        // TODO(xla): Remove automatic prepayment once we have proper balances.
+        self.prepay_account(author.public(), 1000).await?;
```
MeBrei, line 647: This is to make sure an author always has funds before registering their handle.

xla: Are these funds used to pay the fee? I was under the impression that the org pays the registration of its members.

MeBrei:
> One conceptual question around payment origin.

This prepayment happens when registering a user (not the org). Since we need several registered users for registering a new member, we need to give users other than `alice` funds automatically. As explained in the `Note` of the description of the PR, this only helps for running `cargo test` as we can bypass the api directly. The api and thus the playground always uses alice as the author for every transaction

proxy/src/registry.rs
MeBrei: This prepayment happens when registering a user (not the org), so these funds are used for paying the transaction of registering the same authors user handle. Member registration is payed by the org.
Since we need several registered users for registering a new member, we need to give users other than alice funds automatically. As explained in the Note of the description of the PR, this only helps for running cargo test as we can bypass the api directly. The api and thus the playground always uses alice as the author for every transaction

PR #431, opened 2020-05-27
feat(ui): include orgId in member registration tx & consolidate wallet logic

sarahscott:
- [x] Pass `params.orgId` to new transaction in `MemberRegistration`
- [x] Consolidate payer logic to `Transaction.svelte` & `transaction.ts`

In the spec, the payer can be either an `org` or the current `user`, but we don't actually send any information to proxy about which wallet to use. I suggest we assume the `payer` is the current session `identity` until we support org wallets. Alternatively, we can fake it in the UI and allow for it  to look like orgs are paying for something, but we're going to have to revisit this stuff in the Wallets I milestone anyway.

horsegirls420


ui/src/transaction.ts
```diff
@@ -278,15 +278,25 @@ export enum PayerType {
...
+}
+
+export const formatPayerFromIdentity = (identity: Identity): Payer => (identity && {
+  name: identity.metadata.displayName || identity.metadata.handle,
+  type: PayerType.User,
+  avatarFallback: identity.avatarFallback,
+  imageUrl: identity.metadata.avatarUrl
+});
+
+export const formatPayerFromOrg = (orgId: string): Payer => ({
```
sarahscott, line 296: This will be removed if we go with the user-only wallet approach

ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -36,6 +37,7 @@
   const updateAvatar = async () => (avatar = await subject.avatarSource);
 
   $: updateAvatar();
+  $: payer = formatPayerFromIdentity($session.data.identity);
```
xla, line 40: The payer should be passed in, having a component in the DesignSystem depend on a store is backwards of what we have been working towards. Esp. as the data can be undefined depending on the state of the remote store.

ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -118,13 +120,15 @@
...
 <Row style="background-color: var(--color-foreground-level-1)">
   <div slot="left">
-    <Avatar
-      dataCy="payer-avatar"
-      title={payer.name}
-      imageUrl={payer.imageUrl}
-      avatarFallback={payer.avatarFallback}
-      variant={payer.type === PayerType.User ? 'circle' : 'square'}
-      style="color: var(--color-foreground)" />
+    {#if payer}
```
xla, line 123: A transaction should always expect a `payer` passed in.

ui/src/transaction.ts
```diff
@@ -278,15 +278,20 @@ export enum PayerType {
...
-    avatarFallback: identity.avatarFallback,
-    imageUrl: identity.metadata.avatarUrl
-  };
-};
+// TODO(sos): coordinate payer shape with proxy/registry
+interface Payer {
+  type: PayerType;
+  name: string;
+  avatarFallback?: EmojiAvatar;
+  imageUrl?: string;
```
xla, line 286: Instead of maybe fields, we could make the Payers variants with guaranteed fields.

sarahscott: The `avatarFallback` can be guaranteed but isn't the imageUrl always a maybe field?

sarahscott:
@rudolfs I mistakenly put the `orgId` in the transaction's `id` field; we don't actually have the transaction `id` until proxy generates it, so there isn't anything displayed there now. Also fixed up the project registration flow - didn't realize it was using the `context` session instead of the session store.

ui/Screen/ProjectRegistration/RegistrationDetailsStep.svelte
```diff
@@ -5,6 +5,7 @@
 
   import * as notification from "../../src/notification.ts";
   import { getOrgProject, Domain } from "../../src/project.ts";
+  import { session } from "../../src/session.ts";
```
rudolfs, line 8: Why are all these changes necessary in this file?

ui/Screen/Org/MemberRegistration.svelte
```diff
@@ -17,19 +17,22 @@
...
     validating = false;
   const validation = memberHandleValidationStore();
 
   const next = () => {
     switch (state) {
       case RegistrationFlowState.Preparation:
         if ($validation.status === ValidationStatus.Success) {
-          transaction = registerMemberTransaction("monadic", userHandle);
+          transaction = registerMemberTransaction(orgId, userHandle);
           payer = formatPayer($session.data.identity);
```
MeBrei, line 36: Member registration should be payed by the org. 
If it is outside of the scope of this PR, maybe you could add a Todo.

cloudhead:
These screens are going to be reworked soon, so I wouldn't change too much in the UI. It does make sense to have the org id in there though, we will take that into account in the redesign.

ui/Screen/Org/MemberRegistration.svelte
sarahscott: See @cloudhead's comment; the avatar is intentionally not there because we haven't decided on a way to reconcile CoCo identity (which contains the user's avatar) and registry users (which does not contain any avatar information). Also the wallet work will be encompassed by the Wallets I milestone, TBD.

PR #446, opened 2020-06-03
Register member endpoint integration

MeBrei:
Implements actual member registration and replaces the org member list mock. It also adds a new control endpoint to add additional users so that adding members can be tested in cypress.

Related to #349

### DIFF
--- a/proxy/src/registry/transaction.rs
+++ b/proxy/src/registry/transaction.rs
@@ -109,6 +109,14 @@ pub enum Message {
         /// Identity id originated from librad.
         id: Option<String>,
     },
+
+    /// Issue a member registration for a given handle under a given org.
+    MemberRegistration {
+        /// Globally unique user handle.
+        handle: registry::Id,
+        /// The Org in which to register the member.
+        org_id: registry::Id,
+    },
 }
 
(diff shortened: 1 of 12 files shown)


proxy/src/registry/transaction.rs
```diff
@@ -109,6 +109,14 @@ pub enum Message {
         /// Identity id originated from librad.
         id: Option<String>,
     },
+
+    /// Issue a member registration for a given handle under a given org.
+    MemberRegistration {
```
xla, line 114: We should force consistent camelcasing by using `#[serde(rename_all = "camelCase")]` on the variants as well.

MeBrei: Does [L83](https://github.com/radicle-dev/radicle-upstream/blob/0f5dd37f97d869dc5f5a7cff2d75df8197d2b953/proxy/src/registry/transaction.rs#L83) not affect the variants as well?

xla: No unfortunately it doesn't as serde treats every struct variant as its own container.

MeBrei: addressed in #449

PR #447, opened 2020-06-03
docs: add Linux dev dependencies to docs

rudolfs:
Noticed that we don't have all the dependencies listed when I was trying to build Upstream from scratch on Linux.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -61,8 +61,7 @@ On Linux:
   - [GNU M4](https://www.gnu.org/software/m4)
   - [Nettle](http://www.lysator.liu.se/~nisse/nettle)
   - [OpenSSL](https://www.openssl.org)
-
-On both platforms make sure to have [`yarn`][ya] installed.
+  - [Yarn](https://yarnpkg.com)
 
 1. Get Upstream: `git clone git@github.com:radicle-dev/radicle-upstream.git`.
 2. Install dependencies: `cd radicle-upstream && yarn install`.
@@ -464,4 +463,3 @@ Release v0.0.11 successfully completed! 👏 🎉 🚀
 [sv]: https://github.com/conventional-changelog/standard-version
 [tp]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
 [wa]: https://github.com/seanmonstar/warp
-[ya]: https://yarnpkg.com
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -43,14 +43,24 @@ make sure only properly formatted and lint-free code lands into master.
...
 sudo xcodebuild -license
 brew install pkgconfig nettle
 ```
 
+On Debian/Ubuntu:
+```
+sudo apt-get install libssl-dev
+sudo apt-get install libgmp3-dev
+sudo apt-get install m4
+sudo apt-get install clang
```
NunoAlexandre, line 61: I'd suggest to inline these in one command

rudolfs:
@NunoAlexandre addressed your comments, is it more clear now?

NunoAlexandre:
@rudolfs yes it's great! One last suggestion is that we can have a script `install-dependencies-macos` and `install-dependencies-linux` that users just have to run, instead of providing them with commands they need to copy-paste in sequence.

DEVELOPMENT.md
```diff
@@ -43,12 +43,25 @@ make sure only properly formatted and lint-free code lands into master.
...
 ```
 xcode-select --install
 sudo xcodebuild -license
-brew install pkgconfig nettle
+brew install yarn pkgconfig nettle
+```
+
+On Debian/Ubuntu:
+```
+# Install yarn (UI dependency):
```
xla, line 58: I'd advocate against commands which are specific to certain distros. As it doesn't even help a majority of our core devs. We should just list the dependencies we expect to be present and link to the respective projects which will have guidance for installation. This information will be stale very quickly.

rudolfs:
@NunoAlexandre hm, not sure I like that idea, I usually prefer doing it by hand, because sometimes I already have half of the dependencies a project needs.

NunoAlexandre:
> @NunoAlexandre hm, not sure I like that idea, I usually prefer doing it by hand, because sometimes I already have half of the dependencies a project needs.

If you already have them the tool should know and do nothing, saving you that time. It's just a suggestion, feel free to ignore it.

DEVELOPMENT.md
rudolfs: Fair point. How about now?

PR #448, opened 2020-06-04
chore(dependencies): Bump dependencies

NunoAlexandre:
Getting started on my new machine, I had to install electron (`npm install electron --save-dev`), which resulted in a bump to the version we use.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12221 @@
+{
+  "name": "radicle-upstream",
+  "version": "0.0.11",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "7zip-bin": {
+      "version": "5.0.3",
+      "resolved": "https://registry.npmjs.org/7zip-bin/-/7zip-bin-5.0.3.tgz",
+      "integrity": "sha512-GLyWIFBbGvpKPGo55JyRZAo4lVbnBiD52cKlw/0Vt+wnmKvWJkpZvsjVoaIolyBXDeAQKSicRtqFNPem9w0WYA==",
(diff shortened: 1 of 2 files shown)


NunoAlexandre:
@xla @rudolfs it seems like I added `package-lock.json`. Is there a reason as to why that wasn't included so far?

rudolfs:
@NunoAlexandre we use `yarn` as the default package manager, not `npm`.

rudolfs:
We also normally do upgrades for the whole UI stack via `yarn upgrade-interactive --latest`. As we'll have to manually check some thing for which we don't have tests (visuals, packaging, etc).

NunoAlexandre:
Running `yarn start` failed because `electron` wasn't installed. The only way I could find to install the electron was via `npm`. How are/have you doing/done it yourselves?

rudolfs:
@NunoAlexandre I think you missed a point from: https://github.com/radicle-dev/radicle-upstream/blob/master/DEVELOPMENT.md#running-upstream

> Install dependencies: cd radicle-upstream && yarn install.

rudolfs:
Btw. happy to adjust the doc so it's more obvious to newcomers, let me know if you have suggestions on how to restructure it.

NunoAlexandre:
> @NunoAlexandre I think you missed a point from: https://github.com/radicle-dev/radicle-upstream/blob/master/DEVELOPMENT.md#running-upstream
> 
> > Install dependencies: cd radicle-upstream && yarn install.
> 
> So just run `yarn install` and you should be set.

I think I tried `yarn install` but from the `ui` dir :facepalm:

Issue #441, opened 2020-06-02
Add ability to register project under a user


sarahscott:
In [proxy](https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/src/registry/transaction.rs#L97), we can only register projects under orgs at the moment, but the UI implies that it's possible to register one under the current user too.

sarahscott:
Currently proxy rejects user projects with `ERROR proxy::http::error > Rejection([Transaction(RegistryError(InexistentOrg)), MethodNotAllowed])`

NunoAlexandre:
> Currently proxy rejects user projects with `ERROR proxy::http::error > Rejection([Transaction(RegistryError(InexistentOrg)), MethodNotAllowed])` (as it should)

@sarahscott how do you get to this response? From the UI I appear to be able to register a project under a user and I don't see any errors in the logs.

sarahscott:
My steps to reproduce are:
- Register my handle
- Go though project registration flow
- Submit transaction

The error in the console is `Incorrect input` and in terminal it's `Rejection([Transaction(RegistryError(InexistentOrg)), MethodNotAllowed])`

sarahscott:
I'm able to register projects under orgs without any issues

NunoAlexandre:
Thank you, @sarahscott. I was missing the last step of actually registering the project on-chain, now I get the error :ok_hand:

Issue #391, opened 2020-05-14
Integrate transaction costs


rudolfs:
Show correct transaction costs in transaction summary.

<img width="50%" alt="Screenshot 2020-05-14 at 14 47 35" src="https://user-images.githubusercontent.com/158411/81936055-e29fc580-95f1-11ea-9b5d-9a39140e53e8.png">

**proxy**
- [x] provide transaction costs via an endpoint

**ui**
- [x] wire-up transaction costs in the following screens
  - [x] user registration
  - [x] org registration
    - [x] add member to org
  - [x] project registration

Depends on:
- [x] https://github.com/radicle-dev/radicle-registry/issues/485

rudolfs:
@cloudhead: can I assume we won't show the Rad <-> $ conversion in the first iteration of this feature?

cloudhead:
I think we can have it in there if we keep it in the design (we're in a redesign of the wallet atm), and keep the conversion at 1:1 for now.

PR #450, opened 2020-06-04
chore(proxy): bump registry crate

rudolfs:
Required by: #391, #441

### DIFF
--- a/proxy/src/registry.rs
+++ b/proxy/src/registry.rs
@@ -514,7 +514,10 @@ impl Client for Registry {
     ) -> Result<Option<Project>, error::Error> {
         Ok(self
             .client
-            .get_project(project_name.clone().0, protocol::ProjectDomain::Org(org_id.clone().0))
+            .get_project(
+                project_name.clone().0,
+                protocol::ProjectDomain::Org(org_id.clone().0),
+            )
             .await?
             .map(|project| {
                 let metadata_vec: Vec<u8> = project.metadata.into();
@@ -940,7 +943,10 @@ mod test {
         assert!(result.is_ok());
 
         let maybe_project = client
-            .get_project(project_name.clone().0, protocol::ProjectDomain::Org(org_id.clone().0))
+            .get_project(
+                project_name.clone().0,
+                protocol::ProjectDomain::Org(org_id.clone().0),
+            )
             .await?;
 
         assert!(maybe_project.is_some());
--- a/proxy/src/registry.rs
+++ b/proxy/src/registry.rs
@@ -10,7 +10,7 @@ use std::convert::TryFrom;
 use std::fmt;
 use std::str::FromStr;
 
-use radicle_registry_client::{self as protocol, ClientT, CryptoPair, ProjectDomain};
+use radicle_registry_client::{self as protocol, ClientT, CryptoPair};
 
 use crate::avatar;
 use crate::error;
@@ -514,7 +514,7 @@ impl Client for Registry {
     ) -> Result<Option<Project>, error::Error> {
         Ok(self
             .client
-            .get_project(project_name.clone().0, ProjectDomain::Org(org_id.clone().0))
+            .get_project(project_name.clone().0, protocol::ProjectDomain::Org(org_id.clone().0))
             .await?
             .map(|project| {
                 let metadata_vec: Vec<u8> = project.metadata.into();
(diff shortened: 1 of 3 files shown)


proxy/src/registry.rs
```diff
@@ -514,13 +514,13 @@ impl Client for Registry {
...
-            .get_project(project_name.0, org_id.0)
+            .get_project(project_name.clone().0, ProjectDomain::Org(org_id.clone().0))
             .await?
             .map(|project| {
                 let metadata_vec: Vec<u8> = project.metadata.into();
                 Project {
-                    name: ProjectName(project.name),
-                    org_id: Id(project.org_id),
+                    name: project_name.clone(),
+                    org_id: org_id.clone(),
```
NunoAlexandre, line 523: To be transparent about this bit: we request a project under the given `org_id` domain, and when we get it and go on building this `Project` value, instead of unwrapping the `org_id` domain from the registry Project we just use the input param, as they can _only_ be the same. The same goes for the `project_name`. I would usually avoid this to have one source of truth but in this case, it appears sensible, as it's impossible to request a project under an org X and get back a project use domain is not that same org.

proxy/src/registry.rs
```diff
@@ -10,7 +10,7 @@ use std::convert::TryFrom;
 use std::fmt;
 use std::str::FromStr;
 
-use radicle_registry_client::{self as protocol, ClientT, CryptoPair};
+use radicle_registry_client::{self as protocol, ClientT, CryptoPair, ProjectDomain};
```
xla, line 13: We only import traits explicitel for the rest we use the `protocol::` qualification.

```suggestion
use radicle_registry_client::{self as protocol, ClientT, CryptoPair};
```

proxy/src/registry.rs
```diff
@@ -514,13 +514,13 @@ impl Client for Registry {
     ) -> Result<Option<Project>, error::Error> {
         Ok(self
             .client
-            .get_project(project_name.0, org_id.0)
+            .get_project(project_name.clone().0, ProjectDomain::Org(org_id.clone().0))
```
xla, line 517:
```suggestion
            .get_project(project_name.clone().0, protocol::ProjectDomain::Org(org_id.clone().0))
```

proxy/src/registry.rs
```diff
@@ -536,7 +536,7 @@ impl Client for Registry {
         let ids = self.client.list_projects().await?;
         let mut projects = Vec::new();
         for id in &ids {
-            if id.1 == org_id.clone().0 {
+            if id.1 == ProjectDomain::Org(org_id.clone().0) {
```
xla, line 539:
```suggestion
            if id.1 == protocol::ProjectDomain::Org(org_id.clone().0) {
```

proxy/src/registry.rs
```diff
@@ -598,7 +598,7 @@ impl Client for Registry {
         // Prepare and submit project registration transaction.
         let register_message = protocol::message::RegisterProject {
             project_name: project_name.0.clone(),
-            org_id: org_id.0.clone(),
+            project_domain: ProjectDomain::Org(org_id.0.clone()),
```
xla, line 601:
```suggestion
            project_domain: protocol::ProjectDomain::Org(org_id.0.clone()),
```

proxy/src/registry.rs
```diff
@@ -702,7 +702,7 @@ impl Client for Registry {
 )]
 #[cfg(test)]
 mod test {
-    use radicle_registry_client::{self as protocol, ClientT};
+    use radicle_registry_client::{self as protocol, ClientT, ProjectDomain};
```
xla, line 705:
```suggestion
    use radicle_registry_client::{self as protocol, ClientT};
```

proxy/src/registry.rs
```diff
@@ -940,14 +940,14 @@ mod test {
         assert!(result.is_ok());
 
         let maybe_project = client
-            .get_project(project_name.clone().0, org_id.clone().0)
+            .get_project(project_name.clone().0, ProjectDomain::Org(org_id.clone().0))
```
xla, line 943:
```suggestion
            .get_project(project_name.clone().0, protocol::ProjectDomain::Org(org_id.clone().0))
```

proxy/src/registry.rs
```diff
@@ -940,14 +940,14 @@ mod test {
...
             .await?;
 
         assert!(maybe_project.is_some());
 
         let project = maybe_project.unwrap();
         assert_eq!(project.name, project_name.0);
-        assert_eq!(project.org_id, org_id.0);
+        assert_eq!(project.domain, ProjectDomain::Org(org_id.0));
```
xla, line 950:
```suggestion
        assert_eq!(project.domain, protocol::ProjectDomain::Org(org_id.0));
```

xla:
@rudolfs Needs some `fmt`ing and then we good to go.

PR #468, opened 2020-06-09
feat(ui): do not allow registering a member twice

MeBrei:
Implements a validation for the register member modal to check if the user is already a member. In this it also extends the implementation of `createValidationStore` to accommodate multiple remote validations. 

Related to #349


ui/src/org.ts
```diff
@@ -42,6 +42,8 @@ export const getIdAvailability = (id: string): Promise<boolean> =>
   getOrg(id).then(org => !org);
 const validateUserExistence = (handle: string): Promise<boolean> =>
   user.get(handle).then(user => !!user);
+const validateNewMember = (orgId: string) => (handle: string): Promise<boolean> =>
+  getOrg(orgId).then(org => org.members.every((member) => member.handle != handle));
```
NunoAlexandre, line 46: this validates a user by checking that all the members of the involved org have a different handle. I'd expect it to, instead, validate that it doesn't find any member with that handle. So instead of `every`, I'd expect a `! any` (pseudo-code here).

MeBrei: @NunoAlexandre I agree that it would read better, but as far as I could see, the javascript [Array class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) does not provide `any`.

NunoAlexandre: it seems like [`find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) would cut it!

MeBrei: changed in 55a05fbd

Issue #371, opened 2020-05-11
Return permissions via the session endpoint


rudolfs:
As most of the business logic lives in proxy, it's probably the best place to answer which user actions are allowed at a given moment. Here's a list of permissions that we could use in the UI already:

- am I allowed to create a new org?
  - user handle has to be registered
- am I allowed to register a project?
  - I have to be a member of at least one org or a registered user handle
  - I have to have at least one local project
- am I allowed to register this specific project?
  - it is not registered yet
- am I allowed to register a member?
  - I have to be a member of the org

MeBrei:
### V1 permissions
In a first iteration, the permissions would not be connected to the specific transactions, but would be `True` as soon as the user/org information can be retrieved from the registry (ie as soon as the transactions are applied to at least one block).

proposed addition to `session`:
```
permissions: {
	createOrg: True | False,
	registerProject: True | False,
}
```

> * am I allowed to create a new org?
>   
>   * user handle has to be registered

covered by `permissions.createOrg`
> * am I allowed to register a project?
>   
>   * I have to be a member of at least one org or a registered user handle
>   * I have to have at least one local project

covered by `Session.permissions.registerProject`
> * am I allowed to register this specific project?
>   
>   * it is not registered yet

covered by `Session.permissions.registerProject` + checking in the specific project response if it is registered `Project.registered` (ui)
> * am I allowed to register a member?
>   
>   * I have to be a member of the org

check if the org is one of `Session.orgs` (ui)

cloudhead:
This looks good merle. A couple of things to consider:
* If I've registered my handle, I can't register it again (`registerHandle`)
* If I visit an org I am not a member of (this will eventually be possible), I shouldn't be able to register a project there.
* I shouldn't be able to register a member twice.
* I can register a project even if I'm not a member of any org, as long as I have a registered handle

MeBrei:
@cloudhead Thank you for your feedback!

> * If I've registered my handle, I can't register it again (`registerHandle`)

Good point, I am not sure though if this should be part of `permissions` or rather the session should include if you are registered or not (I think  also a badge should be shown if you are registered). I will check what the plan is here.
> * If I visit an org I am not a member of (this will eventually be possible), I shouldn't be able to register a project there.

👍 This should also be covered by `Session.orgs`
> * I shouldn't be able to register a member twice.

This should not be a part of the session as this is org and not user specific. In #349 a validation for this will be added.
[...]

PR #471, opened 2020-06-11
feat: introduce session permissions

MeBrei:
Implements permissions for registering handle/org/project added to the session API endpoint. The use of `registerHandle` in the ui is implemented to block the user from registering a handle a second time. More details can be found in #371. 
The use cases of the other permissions will be implemented in follow up PRs.

### DIFF
--- a/proxy/src/http/session.rs
+++ b/proxy/src/http/session.rs
@@ -306,12 +306,17 @@ mod test {
                         "network": "emulator",
                     },
                 },
+                "permissions": {
+                    "registerHandle": false,
+                    "registerOrg": false,
+                    "registerProject": false,
+                },
             }),
         );
     }
 
     #[tokio::test]
-    async fn udpate_settings() {
+    async fn update_settings() {
         let tmp_dir = tempfile::tempdir().unwrap();
         let store = kv::Store::new(kv::Config::new(tmp_dir.path().join("store"))).unwrap();
         let registry = {
@@ -347,6 +352,11 @@ mod test {
                     "registry": {
                         "network": "emulator",
                     },
+                },
+                "permissions": {
+                    "registerHandle": false,
+                    "registerOrg": false,
+                    "registerProject": false,
                 },
             }),
         );
--- a/proxy/src/session.rs
+++ b/proxy/src/session.rs
@@ -62,9 +62,7 @@ pub async fn current<R: registry::Client>(
     let mut session = get(store, KEY_CURRENT)?;
 
     // Reset the permissions
-    session.permissions.register_handle = false;
-    session.permissions.register_org = false;
-    session.permissions.register_project = false;
+    session.permissions = Permissions::default();
 
     if let Some(mut id) = session.identity.clone() {
         if let Some(handle) = id.registered.clone() {
--- a/proxy/src/session.rs
+++ b/proxy/src/session.rs
@@ -20,8 +20,22 @@ pub struct Session {
     pub identity: Option<identity::Identity>,
     /// List of the orgs of the user associated with the current identity.
     pub orgs: Vec<registry::Org>,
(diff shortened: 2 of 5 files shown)


proxy/src/http/session.rs
```diff
@@ -306,12 +306,17 @@ mod test {
...
+                    "registerHandle": false,
+                    "registerOrg": false,
+                    "registerProject": false,
+                },
             }),
         );
     }
 
     #[tokio::test]
-    async fn udpate_settings() {
```
rudolfs, line 314: heh

proxy/src/session.rs
```diff
@@ -47,14 +61,24 @@ pub async fn current<R: registry::Client>(
 ) -> Result<Session, error::Error> {
     let mut session = get(store, KEY_CURRENT)?;
 
+    // Reset the permissions
+    session.permissions.register_handle = false;
```
xla, line 65:
```suggestion
    session.permissions.register_handle = true;
```

If set to true and in combination with the other suggestion you can drop the last else branch.

proxy/src/session.rs
```diff
@@ -47,14 +61,24 @@ pub async fn current<R: registry::Client>(
...
     if let Some(mut id) = session.identity.clone() {
         if let Some(handle) = id.registered.clone() {
             if registry.get_user(handle.clone()).await?.is_some() {
                 session.orgs = registry.list_orgs(handle).await?;
+                session.permissions.register_org = true;
+                session.permissions.register_project = true;
             } else {
                 id.registered = None;
                 session.identity = Some(id);
+                session.permissions.register_handle = true;
```
xla, line 78:
```suggestion
```

proxy/src/session.rs
```diff
@@ -47,14 +61,24 @@ pub async fn current<R: registry::Client>(
...
             }
+        } else {
+            session.permissions.register_handle = true;
```
xla, line 81:
```suggestion
```

proxy/src/session.rs
```diff
@@ -47,14 +61,24 @@ pub async fn current<R: registry::Client>(
...
+    session.permissions.register_handle = false;
+    session.permissions.register_org = false;
+    session.permissions.register_project = false;
+
     if let Some(mut id) = session.identity.clone() {
         if let Some(handle) = id.registered.clone() {
             if registry.get_user(handle.clone()).await?.is_some() {
                 session.orgs = registry.list_orgs(handle).await?;
+                session.permissions.register_org = true;
+                session.permissions.register_project = true;
```
xla, line 74:
```suggestion
                session.permissions.register_org = true;
                session.permissions.register_project = true;
                session.permissions.register_handle = false;
```

proxy/src/session.rs
MeBrei: I find it easier to read if all permissions are reset to the same value and that permissions are explicitly set to `true`. If we set `register_handle` to true here, the last else branch could be dropped as you suggest, but for consistency we would have to add on for the outermost if checking if an identity exists. If no identity is set in the session, registering a handle should not be permitted.

xla: That's a good point, it would move the complexity to another place, I think it's not worth changing then. What you could try tho is to use the default of the struct, see comment below.

proxy/src/session.rs
```diff
@@ -47,14 +61,24 @@ pub async fn current<R: registry::Client>(
 ) -> Result<Session, error::Error> {
     let mut session = get(store, KEY_CURRENT)?;
 
+    // Reset the permissions
+    session.permissions.register_handle = false;
+    session.permissions.register_org = false;
+    session.permissions.register_project = false;
```
xla, line 67:
```suggestion
    session.permissions = Permissions::default();
```

PR #472, opened 2020-06-11
feat(proxy): option to connect to registry nodes

### LINKED ISSUE #440: Connect app and proxy to different networks and nodes.
As a developer I want to be able to work with the app on the ffnet as well as the devnet, a local node, and the registry emulator so that I can work with persistent data and test the app under realistic conditions. The mechanism for this should be properly documented.

In the future it should be possible to extend this mechanism to allow an app user to set the network/node host through a CLI flag when starting the app.

We will not implement the full network selector described in https://github.com/radicle-dev/radicle-upstream/issues/128.

geigerzaehler:
Fixes #440. The proxy now accepts a node host for the `--registry`
option and can connect to an arbitrary registry node. We also provide
shortcuts for the ffnet and devnet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,8 @@
     "proxy:build:release": "cd proxy && cargo build --release --all-features --all-targets",
     "proxy:start": "cd proxy && cargo run -- --registry=emulator",
     "proxy:start:test": "cd proxy && cargo run -- --registry=emulator --test",
+    "proxy:start:ffnet": "cd proxy && cargo run -- --registry=rpc.ff.radicle.network",
+    "proxy:start:devnet": "cd proxy && cargo run -- --registry=35.241.138.91",
     "cypress:run": "yarn run cypress run",
     "cypress:open": "yarn run cypress open",
     "release": "babel-node --presets \"@babel/env\" scripts/release.js",
--- a/proxy/src/main.rs
+++ b/proxy/src/main.rs
@@ -5,7 +5,8 @@ use proxy::registry;
 
 /// Flags accepted by the proxy binary.
 struct Args {
-    /// Signaling which backend type to use.
+    /// Host name or IP for the registry node to connect to. If the special value "emulator" is
+    /// provided the proxy will not connect to a node but emulate the chain in memory.
     registry: String,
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -111,6 +111,8 @@
     "proxy:build:release": "cd proxy && cargo build --release --all-features --all-targets",
     "proxy:start": "cd proxy && cargo run -- --registry=emulator",
     "proxy:start:test": "cd proxy && cargo run -- --registry=emulator --test",
+    "proxy:start:ffnet": "cd proxy && cargo run -- --registry=rpc.ff.radicle.network",
+    "proxy:start:devnet": "cd proxy && cargo run -- --registry=35.241.138.91",
```
xla, line 115: Can we put this behind a DNS entry, like `rpc.devnet.radicle.network`?

proxy/src/main.rs
```diff
@@ -2,33 +2,53 @@ use proxy::coco;
 use proxy::env;
 use proxy::http;
 use proxy::registry;
+use structopt::{clap::AppSettings, StructOpt};
```
xla, line 5: The dependency graph of the project is alreayd sizable, do we really need to pull in a crate that depends on clap?

geigerzaehler: I opted for `structopt` to have the proxy provide a nice help message. If we can do without then I’m happy to revert to `pico-args`.

package.json
geigerzaehler: https://github.com/radicle-dev/infra/issues/122

geigerzaehler:
>     * as we going to allow the configuration and switching of networks from the ui, it would be preferable to have the info what ip/dns name is associated with the different networks somewhere central, like `registry.rs`

Is it ok if we punt on this for now and handle this in a separate issue? It’s unclear to me how this would be used and what API to expose.

xla:
> Is it ok if we punt on this for now and handle this in a separate issue? It’s unclear to me how this would be used and what API to expose.

Sure thing.

geigerzaehler:
@xla: Could you have another look?

PR #483, opened 2020-06-12
feat(ui): restrict register org sidebar button

MeBrei:
Uses `session.permissions.registerOrg` to disable the register-org sidebar button.
This should cover all current use cases of `session.permissions.registerOrg`. 

Part of #371


ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -9,6 +9,7 @@
   import AddOrgButton from "./Sidebar/AddOrgButton.svelte";
 
   const dispatch = createEventDispatcher();
+  const session = getContext("session");
```
xla, line 12: SidebarLayout already depends on session and should extract the information needed and pass it down to the Sidebar component, this will keep this component KISS and data access centralised.

ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -194,10 +195,17 @@
     {/each}
 
     <li class="item" data-cy="add-org-button">
-      <AddOrgButton on:click={() => dispatch('createorg')} />
-      <div class="tooltip">
-        <Title>Add org</Title>
-      </div>
+      {#if registerOrgPermission}
```
sarahscott, line 198:
```suggestion
    <AddOrgButton 
      on:click={() => registerOrgPermission && dispatch('createorg')} 
      disabled={!registerOrgPermission}
     />
    <div class="tooltip">
      <Title>
        {registerOrgPermission ? 'Register your handle to create an org' : 'Add org}
      </Title>
    </div>
```

juliendonck:
@MeBrei #489 touches some of the things of your PR here, wondering which one makes most sense to merge first. Let me know what you think :)

MeBrei:
@rudolfs @sarahscott I agree with you that there should be tests for all behavior around permissions. I don't think though that the ui should assume knowledge about when the permissions change, ie a cypress test could be checking the sidebar add-org button before and after user registration. I think therefor it is necessary to find a way to mock the session and just testing if the ui changes according to changed permissions. I will need some more time to figure out a way to do that so that I would merge this PR without a test for now.

rudolfs:
@MeBrei I think most of our cypress tests are actually integration tests, so I wouldn't mind if it would test permissions in the same way a user would test it manually. But let's merge this for the sake of velocity.

xla:
> @rudolfs @sarahscott I agree with you that there should be tests for all behavior around permissions. I don't think though that the ui should assume knowledge about when the permissions change, ie a cypress test could be checking the sidebar add-org button before and after user registration. I think therefor it is necessary to find a way to mock the session and just testing if the ui changes according to changed permissions. I will need some more time to figure out a way to do that so that I would merge this PR without a test for now.

We do have ways to mock in the unit tests, if we want to assert that behaviour it should go in that layer. Otherwise I think @rudolfs is on point by saying it's okay to capture these assertions in actual flows. While the ui shouldn't make any assumptions it will also help catch unexpected or unaccounted changes from the API.

cypress/integration/org_registration.spec.js
```diff
@@ -1,3 +1,23 @@
+context("org registration permission", () => {
+  before(() => {
+    cy.nukeAllState();
+    cy.nukeCache();
```
rudolfs, line 4: `nukeCache` is already included in `nukeAllState`.

ui/DesignSystem/Component/Sidebar.svelte
MeBrei: I find it easier to read to keep the `if` statements outside of the component. In general I think avoiding repetition is good, but here it doesn't add too much duplication.

PR #485, opened 2020-06-12
fix(ui): update emojis with full set

juliendonck:
Since we're using twemoji now, we can add some emoji's that wouldn't render on all devices before.


xla:
The tests fail, because the generated emojis don't match anymore. Maybe you can pair with someone to fix them, otherwise happy to do it.

juliendonck:
🙇

PR #489, opened 2020-06-15
style(ui): sidebar visual update

juliendonck:
Update the sidebar to represent the latest state of the design: 
- Reorder profile / orgs / wallet / etc
- Allow for orgs to overflow

Screenshot 2020-06-15 at 10 52 57


ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -20,11 +20,39 @@
...
     flex-direction: column;
     align-items: center;
+    justify-content: space-between;
+  }
+
+  .top {
+    overflow-y: scroll;
+    padding-bottom: 36px;
+    padding-top: 16px;
+  }
```
rudolfs, line 34:
```suggestion
  }

```

juliendonck:
@rudolfs I removed Search from the sidebar so that's why the routing test fails (but I don't know how to fix it) but I don't know why the other one does. Could you take a look? Or can I just merge?

rudolfs:
> Could you take a look? Or can I just merge?

I'll have a look when I have some free time. We shouldn't merge broken code.

ui/DesignSystem/Component/Sidebar.svelte
juliendonck: Fixed

juliendonck:
@brandonhaslegs good catch, should be fixed now

PR #490, opened 2020-06-15
chore(ci): improve proxy steps

FintanH:
* Separated out the clippy and fmt commands so that they can run in a separate CI step. That allows them to run in parallel.
* Added a doc check.
* Fixed some links that were breaking the docs.

### DIFF
--- a/.buildkite/run.sh
+++ b/.buildkite/run.sh
@@ -68,12 +68,6 @@ echo "--- Set custom git config"
 cp .buildkite/.gitconfig "$HOME/"
 cat "$HOME/.gitconfig"
 
-echo "--- Run cargo fmt"
-(cd proxy && time cargo fmt --all -- --check)
-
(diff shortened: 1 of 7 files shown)


.buildkite/run.sh
```diff
@@ -68,15 +68,12 @@ echo "--- Set custom git config"
...
-(cd proxy && time cargo fmt --all -- --check)
-
-echo "--- Run proxy lints"
-(cd proxy && time cargo clippy --all --all-features --all-targets)
-
 echo "--- Run proxy tests"
 (cd proxy && time cargo test --all --all-features --all-targets)
 
+echo "--- Run proxy docs"
+(cd proxy && time cargo doc --no-deps)
```
xla, line 75: Why don't we bundle this one with fmt and clippy?

FintanH: Ya, can do! :)

PR #493, opened 2020-06-15
chore(proxy): Drop the custom Hash struct

NunoAlexandre:
In https://github.com/radicle-dev/radicle-registry/issues/512,  we considered moving the `struct Hash` defined in this repo into the registry and adopting it client-side. We have [a draft PR](https://github.com/radicle-dev/radicle-registry/pull/517) introducing this change.

However, by implementing these changes the question became whether it's worth introducing it all. The registry hash type already implements `Serialize` and `Deserialize`, so the custom struct wrapper only serves to implement the custom implementation that hex encodes the hash. 

Comparing the output of the default implementation with the custom, hex-encoded implementation reveals that it's not really worth it:

- Default impl / with `#[derive(Serialize, Deserialize]`

```
0xc6e42491b3497438bd559219bbea2cc1badd7340393ce3231a553f750759dfd9
```

- Custom impl / with hex encode

```
c6e42491b3497438bd559219bbea2cc1badd7340393ce3231a553f750759dfd9
```

The only difference being the `0x` prefix. I and @geigerzaehler agree that the simplest approach, as of now, is to drop this `struct Hash` altogether and use the underlying type directly. In the UI, the `0x` prefix is ignored.

### DIFF
--- a/proxy/src/http/transaction.rs
+++ b/proxy/src/http/transaction.rs
@@ -58,18 +58,13 @@ mod handler {
         input: super::ListInput,
     ) -> Result<impl Reply, Rejection> {
         // TODO(xla): Don't panic when trying to convert ids.
-        println!("handler::list input.ids {:?}", input.ids.clone());
         let tx_ids = input
             .ids
             .iter()
             .map(|id| {
-                println!("handler::list will try to convert id {}", id);
                 let string = format!("\"{}\"", id);
-                println!("handler::list wrapped id with quotes: {}", string);
-                let deserialized = serde_json::from_str::<registry::Hash>(&string)
-                    .expect("unable to get hash from string");
-                println!("handler::list id now deserialized {:?}", deserialized);
-                deserialized
+                serde_json::from_str::<registry::Hash>(&string)
+                    .expect("unable to get hash from string")
             })
             .collect();
         let txs = cache.read().await.list_transactions(tx_ids)?;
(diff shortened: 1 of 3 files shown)


NunoAlexandre:
@xla dropping this struct and thus use the default underlying hash de(serialize) implementation doesn't work as it cannot deserialize transaction cache keys. See https://github.com/radicle-dev/radicle-upstream/pull/493/commits/971c1f4017b86e6fe7d38bf2a94831edf503231c.

In https://github.com/radicle-dev/radicle-upstream/pull/493/commits/837288c86a79f377c79188a8e757dec3061453eb I'd expect it to work, as I no longer hex-encode the key before storing it but `serde_json::to_string` it. Then, when deserializing, I do `serde_json::from_str`, which - looking at the unit test I added for reference, should work. Do you understand why it doesn't? Is there another implicit (de)serializing step I am not seeing?

NunoAlexandre:
@xla, in https://github.com/radicle-dev/radicle-upstream/pull/493/commits/b7e9fc078fedf45258258dd40d709c5cb46ed996 I identified the issue when deserializing the hash cache key. When the key is serialized, it looks like `"\"0x88788f187f40a12f88aad8e3dd6ba494452305ed0ced83367c9dcd997dd026c2\""`, while the ids (`String`) received in `handler::lost` given to convert to `Hash` look like `"0x88788f187f40a12f88aad8e3dd6ba494452305ed0ced83367c9dcd997dd026c2"` (note the missing `\"`). First formatting the ids with `format!("\"{}\"", id);` before passing it to `serde_json::from_str::<registry::Hash>(` does the trick.

However, these are still some tests that fail caused by the same problem, probably in a place where this fix is not in place.

Now, I wonder if it's worth fixing this in all the places it needs to be addressed or to drop this PR all together and just keep the `struct Hash` here in the upstream. [Having it in the registry](https://github.com/radicle-dev/radicle-registry/pull/517) adds overhead without any benefit to the registry client, dropping it from the upstream all together doesn't seem like such a smooth step, so maybe the best is to just leave it as is.

Let me have your thoughts :skateboard:

xla:
@NunoAlexandre Thanks for the thorough deep-dive. I agree that further work on this seems wasteful for little return at this point. Let's drop it for now and review this at a later point in time.

Issue #492, opened 2020-06-15
Fix validation bug that causes tests on CI to randomly fail


rudolfs:
Happens quite often on CI, can be reproduced locally if we switch the cypress browser to a simulated "Slow 3G" connection.

The underlying issue seems to be that the order in which remote validations are processed is not taken into account.

Might be related to #317.

![bug](https://user-images.githubusercontent.com/158411/84661734-0e78d880-af1b-11ea-9074-00b92b9d26ee.gif)

![org registration -- validations -- prevents the user from registering an invalid org id (failed)](https://user-images.githubusercontent.com/158411/84657865-3bc28800-af15-11ea-8a10-746e585cd1c8.png)

sarahscott:
Part of #219

sarahscott:
This is going to be an issue for validations across the board; should be addressed with the overhaul @MeBrei has suggested.

PR #497, opened 2020-06-15
fix(ci): source env file

FintanH:
Thinking that sourcing the whole environment set up could help fix the issue on `master`


.buildkite/pipeline.yaml
```diff
@@ -6,7 +6,7 @@ steps:
       platform: "linux"
     timeout_in_minutes: 60
     env: &build-docker
-      DOCKER_IMAGE: "gcr.io/opensourcecoin/radicle-upstream:0.4.0"
+      DOCKER_IMAGE: "gcr.io/opensourcecoin/radicle-upstream"
```
xla, line 9: This won't work as we don't symlink `latest` in registry.

FintanH: Ah. I'm trying to just figure out what's going on here because `docker: Error response from daemon: error while mounting volume '': File exists (os error 17).` is not enlightening :smile:

Do you think it's worth just reverting back to doing it all in one script? :/

xla: It seems something is contending on certain state on disk, we should find out what that is and isolate it per step. Looked into other repos with parallel execution but couldn't find anything obvious.

FintanH: Ya, fair. I don't want this work to suck up my cycle time though. I'm going to put the calls back and create an issue for follow up work on this.

FintanH: https://github.com/radicle-dev/radicle-upstream/issues/500

FintanH:
@xla the org test failure is failing. Are we just merging over that?

rudolfs:
> Are we just merging over that?

I think we should disable that test until the [issue](https://github.com/radicle-dev/radicle-upstream/issues/492) is fixed as it get's too much in the way.

xla:
Let's merge in master and see it become greeeeen!

PR #498, opened 2020-06-15
chore(proxy): Drop duplicated registry types

NunoAlexandre:
Closes #470, expect the `Hash` type which is handled in #https://github.com/radicle-dev/radicle-upstream/pull/493

Drops the duplicated registry types by moving their trait implementations to the respective types in the registry.


NunoAlexandre:
@xla this one is ready for your review. The registry revision needs to be touched still to make it compile for wasm but no changes are expected here for the upstream.

PR #503, opened 2020-06-16
chore(ui): fix header scroll behavior

juliendonck:
- give a background to the action bar
- Make the large header scroll, not be fixed
- Remove large top padding from onboard components


ui/Screen/Org.svelte
```diff
@@ -80,9 +80,9 @@
   $: fetch({ id: params.id });
 </script>
 
-<SidebarLayout dataCy="org-screen" style="margin-top: var(--bigheader-height);">
+<SidebarLayout dataCy="org-screen" style="margin-top: 0;">
```
rudolfs, line 83: Do we even need the style override here?

rudolfs: Was just wondering, can leave it as is.

PR #504, opened 2020-06-16
feat: apply project registration permission

MeBrei:
This should cover all current use cases of `session.permissions.registerProject` as mentioned in #371

### DIFF
--- a/ui/Screen/Project.svelte
+++ b/ui/Screen/Project.svelte
@@ -87,7 +87,7 @@
 
   let registerProjectMenuItem;
 
-  if (session.permissions.registerProject && session.identity.registered) {
+  if (session.permissions.registerProject) {
     registerProjectMenuItem = {
       dataCy: "register-project",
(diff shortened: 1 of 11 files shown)


ui/Screen/Project.svelte
```diff
@@ -87,7 +87,7 @@
 
   let registerProjectMenuItem;
 
-  if (session.identity.registered) {
+  if (session.permissions.registerProject && session.identity.registered) {
```
xla, line 90: This logic should be behind the bool returned from the proxy, it should only be true if the project can be registered and the session has a registered user handle. I assume this will come once we have permissions per entity?

xla: Looking at the proxy part this condition is alreayd checked, so we can drop the identity check here.

MeBrei: removed in 8092bbc

PR #505, opened 2020-06-16
chore(ui): align sidebar width & topbar height

juliendonck:
By redesigning the sidebar, we noticed some inconsistencies between the sidebar width (68px) & topbar height (61px). We decided to align these to be both 64px, which meant we had to update some of the icon size etc.

Also took this opportunity to make a global variable for our min and max width of the app and used it everywhere we defined a min/max width.

Screenshot 2020-06-16 at 16 05 35


juliendonck:
> AddOrg button hover is broken for disabled state (it also used to be a dashed line, was the change to a solid line intended?):
> 
> hover

Good catch @rudolfs. Had to do with unregistered state. Had to make it all a bit smarter, but made it much much MUCH better ;)

PR #507, opened 2020-06-16
Keystore

### LINKED ISSUE #479: Store SecretKey upon first start up
The `SecretKey` should be created and stored on disk upon first start-up of the application. If the key does exist we read it from disk.

FintanH:
Closes #479 

Introducing a key store so that we can persist keys in the application.
It persists both `librad` and `registry` keys.

The application, if started in non-test mode, will prompt the user for their password.

I also moved some of the configuration logic for `coco` into its own module.

I think the next step would be to have the storage fetch the key off disk where needed. For registry, this means removing the calls to `from_str_legacy`. For librad, I think it would mean not having the `SecretKey` live in the `PeerApi` anymore.


proxy/src/main.rs
```diff
@@ -54,42 +57,30 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 
-        coco::Peer::new(config)
-            .await
-            .expect("failed to create /tmp user peer")
+        (pw, args.path)
     };
 
+    let paths_config = path
+        .map(coco::config::PathsConfiguration::FromRoot)
+        .unwrap_or(coco::config::PathsConfiguration::Default);
```
xla, line 74: Can't this be shortened `unwrap_or_default`?

proxy/src/main.rs
```diff
@@ -1,20 +1,22 @@
...
+use proxy::keystore;
 use proxy::registry;
 
 /// Flags accepted by the proxy binary.
 struct Args {
     /// Host name or IP for the registry node to connect to. If the special value "emulator" is
     /// provided the proxy will not connect to a node but emulate the chain in memory.
     registry: String,
+    /// Choosing a base path for Peer configuration. Uses a default location if not provided.
+    path: Option<std::path::PathBuf>,
```
xla, line 19: The param itself is quite generally named and doesn't really communicate the function. Also we don't really have a use for it at the moment, so I'd drop it until we need it.

proxy/src/main.rs
```diff
@@ -54,42 +57,30 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        // TODO(finto): read in from config or passed as param
-        let paths = paths::Paths::new()?;
-        let config = net::peer::PeerConfig {
-            key,
-            paths,
-            listen_addr,
-            gossip_params,
-            disco,
-        };
+        let pw = pinentry::Prompt::new("please provide your Radicle passphrase: ")
```
xla, line 65: Does that mean the startup is currently halted until the user entered a passphrase?

proxy/src/keystore.rs
```diff
@@ -0,0 +1,211 @@
...
+
+use secstr::{SecStr, SecUtf8};
+
+use librad::keys;
+use librad::paths;
+use radicle_keystore::{
+    crypto::{Pwhash, SecretBoxError},
+    file, FileStorage, Keystore, SecretKeyExt,
+};
+use radicle_registry_client::{ed25519, CryptoError, CryptoPair};
```
xla, line 14: As we start to depend on `ed25519` beyond full-filling registry related functionality we should move to explicitly depend on it down the road. Maybe worth capturing in a TODO or a follow-up issue. Alternatively the keystore crate exposes it.

xla: To expand on the comment above, it seems a common utility we need and we should have a unified way to handle keys, which in turn will benefit auditability and correctness. I know there is a strong resistance to blur the boundaries between the protocols, but this seems like a @radicle-dev/engineering concern.

xla: Maybe I was to quick with my ideas here, looks like we only use it for the registry key. Why do we use different keys? Thought we could tap into `ed25519` for both of them.

proxy/src/keystore.rs
```diff
@@ -0,0 +1,211 @@
...
+pub struct Keystorage {
+    librad_store: LibradStore,
+    registry_store: RegistryStore,
+}
+
+impl Keystorage {
+    /// Create a new `CocoStore` for storing your [`keys::SecretKey`].
+    #[must_use = "must use CocoStore to put/get a key"]
+    pub fn new(paths: &paths::Paths, pw: SecUtf8) -> Result<Keystorage, Error> {
+        let path = paths.keys_dir();
```
xla, line 29: It feels invasive to use the librad owned paths here, esp. as we control the name of the keys and therefore can't assume overlap with other librad usages outside of the application. We have our own data tree which we use for our session store already and should expand with a keys sub directory:

https://github.com/radicle-dev/radicle-upstream/blob/0ada292486946746fe5eef9046f7b280ae70d899/proxy/src/main.rs#L105

proxy/src/keystore.rs
```diff
@@ -0,0 +1,211 @@
...
+    crypto::{Pwhash, SecretBoxError},
+    file, FileStorage, Keystore, SecretKeyExt,
+};
+use radicle_registry_client::{ed25519, CryptoError, CryptoPair};
+
+const LIBRAD_KEY: &str = "librad.key";
+const REGISTRY_KEY: &str = "registry.key";
+
+/// Storage for putting and getting the necessary cryptographic keys.
+pub struct Keystorage {
```
xla, line 20: Uncertain about a bundled `Keystorage`. I envision this to be injected wherever we need keys, but then again most places depend on either of them and seldomly on both together. Do you see downsides to have `keystore::Coco` and `keystore::Registry` instead?

proxy/src/keystore.rs
kim: I would be surprised if registry would also use `libsodium`. I agree, however, that it is a constant battle to not depend on `n > 1` implementations of the same signature algorithm, because crates tend to not abstract over it (and instead depend on a specific implementation).

The solution I would propose is that both `librad` and `radicle-registry-client`, instead of passing around secret key material, standardise on a trait like this:

```rust
pub trait Signer {
    type PublicKey;
    type Signature;

    fn public_key(&self) -> Self::PublicKey;
    fn sign(&self, data: &[u8]) -> Self::Signature;
}
```

Implementations of this trait could, for example, delegate signing to an `ssh-agent`, which could greatly improve usability.

Note that the dependency of `radicle-keystore` on `libsodium` is unaffected by this, and not in itself a problem -- the keystore doesn't perform signing, it uses key derivation and "sealing" acc. to the NaCl bestest practices.

geigerzaehler: I second the proposal to use a `Signer` trait to do the signing. This should be straightforward to implement on the registry client side.

proxy/src/keystore.rs
kim: The problem is that upstream can't own these key paths either -- otherwise all other implementations (eg. CLIs) will have to hijack the `upstream` namespace for their XDG needs.

I'm not entirely happy with the fact that `librad`, being a library, defines a fixed config location, but I also don't see a better solution -- those locations are, in fact, global. Iow, I don't know a better way than to just acknowledge the existence of the trinity, and standardise on three different namespaces:

    xyz.radicle.upstream
    xyz.radicle.link
    xyz.radicle.registry-client

It would, however, be sufficient if `librad` would implement `TryFrom<ProjectDirs> for Paths` instead of hardcoding the namespace. The namespacing scheme just needs to be documented, so alternative implementations can share it.

proxy/src/main.rs
kim: Yeah... this is not ideal. There isn't a secure way to supply the passphrase to `main`, however: if you pass it as a CLI argument or read it from an environment variable, everyone can see it.

For now, I would acknowledge that the `proxy` is not run as a standalone daemon, and simply start it up in "locked mode". That is, it needs to be "unlocked" via an API request. As a second step, we can think about some kind of vault interface, enabling secure standalone operations.

proxy/Cargo.toml
```diff
@@ -21,6 +21,7 @@ nonempty = "0.2"
 pico-args = "0.3"
 pretty_env_logger = "0.3"
 rand = "0.7"
+secstr = "0.3"
```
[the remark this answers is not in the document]
FintanH: :shushing_face: 

@kim This actually brings up a question for me: should re-export `SecUtf8` since we're implementing `Pinentry` on it. That means the versioning here has to match. That sound reasonable to you?

proxy/src/coco/config.rs
```diff
@@ -0,0 +1,66 @@
...
+pub enum PathsConfiguration {
+    /// Select the default [`paths::Paths`] for configuration.
+    Default,
+    /// Use [`paths::Paths::from_root`] for configuration.
+    FromRoot(std::path::PathBuf),
+}
+
+impl PathsConfiguration {
+    /// Get the [`paths::Paths`] for this configuration.
+    pub fn to_paths(&self) -> Result<paths::Paths, error::Error> {
```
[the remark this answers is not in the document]
FintanH: Nah, just having the optionality. Can remove.

proxy/src/coco/config.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn try_from(config: PathsConfiguration) -> Result<Self, Self::Error> {
+        config.to_paths()
+    }
+}
+
+/// Configure a [`super::Peer`].
+pub async fn configure(
+    paths: paths::Paths,
+    key: keys::SecretKey,
+) -> Result<coco::Peer, error::Error> {
```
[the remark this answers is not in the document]
FintanH: Ah ya, good point.

proxy/Cargo.toml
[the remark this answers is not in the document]
kim: Yeah right... crate hell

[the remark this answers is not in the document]
FintanH: :fire: :smiling_imp: :fire:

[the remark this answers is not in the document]
FintanH: https://github.com/radicle-dev/radicle-link/pull/184

proxy/src/main.rs
FintanH: I'm not sure I understand this "locked" and "unlocked" API request business :thinking:

You need key storage to be able to configure and create a Peer. And the Peer gets propagated to all the API handlers. So I'm not sure, personally, what this would like.

xla: I'd suggest we use a fake pw for now as we do for `args.test` and lock at that password problem as follow-up.

kim: Nothing which requires access to the key works until the `proxy` process is able to decrypt the key, right? Storing the passphrase somewhere where `proxy` can read it non-interactively is a no-go, because then you could store they unencrypted in the first place. Interactively requiring a passphrase to start up `proxy` is also not ideal, because, do you actually have a terminal?

So, either you have a `Pinentry` impl which _somehow_ shows a dialog box to the user (who started `upstream` by clicking on an icon) _before_ all of the backend and frontend is loaded. Or, you start `proxy` with all state `None` which would require a key to be initialised (and handle that in the API), and provide an endpoint to which the frontend can submit the passphrase, which in turn loads and decrypts the key and initialises all state which depends on it.

xla: > So, either you have a `Pinentry` impl which _somehow_ shows a dialog box to the user (who started `upstream` by clicking on an icon) _before_ all of the backend and frontend is loaded. Or, you start `proxy` with all state `None` which would require a key to be initialised (and handle that in the API), and provide an endpoint to which the frontend can submit the passphrase, which in turn loads and decrypts the key and initialises all state which depends on it.

As I pointed out this is something we need to align with @radicle-dev/product to see where and when we want the prompt to happen. We already exercise replacing the `Peer` and other resources we need to full-fill requests for testing purposes, so a similar strategy work there as well. A question in that context do you need the key for all coco operations or only for mutations/writes?

proxy/src/keystore.rs
xla: Great summary and I think the proposal is sound. Can we capture the follow-ups?

proxy/src/keystore.rs
xla: > The problem is that upstream can't own these key paths either -- otherwise all other implementations (eg. CLIs) will have to hijack the `upstream` namespace for their XDG needs.

This makes a lot of sense, if we want that then librad should dictate/own the entire location of the key including the file name, or am I missing something?

proxy/src/main.rs
FintanH: > I'd suggest we use a fake pw for now as we do for args.test and lock at that password problem as follow-up.

I already have that in place. It's just when you run it without `--test` that you provide a password.

===============

> A question in that context do you need the key for all coco operations or only for mutations/writes?

Not currently because the `PeerApi` allows you to get a `SecretKey`, but that means the `SecretKey` lives in memory within the `PeerApi`. That's no bueno for :tinfoilhat: reasons. So a follow-up to this would be looking at removing the `SecretKey` from `PeerApi`.

===============

But alas! Judging from this conversation, I suppose I should remove this, create a follow-up ticket, and put in a fake password for now?

proxy/src/keystore.rs
FintanH: You want to do a follow-up or I make these changes here?

proxy/src/keystore.rs
FintanH: Should this mean that registry should also dictate where their key lives?

xla: > Should this mean that registry should also dictate where their key lives?

Not as I'm aware of. Maybe @radicle-dev/registry can shed some light on how keys are usually handling in other tools like the cli for registry.

FintanH: Well, I'd say that it's a good idea to do the same thing. Since the whole concept is key safety, where we can read the encrypted key from storage at the point where it's necessary, and then drop the key again so that it doesn't live in memory.

But ya, would be nice to get the input from registry on this. I can follow up with any changes if they're needed.

proxy/src/main.rs
xla: > But alas! Judging from this conversation, I suppose I should remove this, create a follow-up ticket, and put in a fake password for now?

Yeah, to avoid this blocking this changeset and have a focused exploration around pw handling in the app.

proxy/src/keystore.rs
xla: From my understanding this would require follow-ups in both coco and registry, followed by integration of these follow-ups in the app. Happy to look into it.

geigerzaehler: I have some follow-up questions on the design and how we want to tackle this consistently between the code bases. If we had a place to talk about this, @xla. I don’t care where exactly.

proxy/src/keystore.rs
FintanH: I suppose this `Signer` business would cover this too? :thinking:

PR #509, opened 2020-06-17
ProjectDomain -> ProjectRegistrant

sarahscott:
Depends on https://github.com/radicle-dev/radicle-registry/pull/528


cloudhead:
I'm confused :confused: "registrant" means "the one who registers". This would be the user in the case of registering a project.

sarahscott:
Registrant means the entity to whom a registered record belongs, i.e. a user or an org.

xla:
@sarahscott Can you provide some basis for that? Most sources refer to the person that registers.

cloudhead:
Yeah my understanding, which is confirmed by what I find online, is that a registrant is a *person*. It so happens that the owner of the record and the person submitting it is often the same person (eg. a user registering a project under their user handle), but in the case of an org, the registrant is not the owner of the record per se.

sarahscott:
In a registration context, the registrant is a person, company, or organization that wants to register something (e.g. a domain name, access to an event, or a trademark/patent), and to whom the record belongs. Here are several examples:

### SEC Regulations
The SEC maintains a registry for keeping track of corporate events that shareholders should know about. Companies are called "registrants" in this context:
[source](https://www.sec.gov/fast-answers/answersform8khtm.html)
[wiki](https://en.wikipedia.org/wiki/SEC_filing)
[example SEC filing](https://www.sec.gov/Archives/edgar/data/1577882/000159406218000178/form8k.htm)

### Domain names
> The person, company or entity who owns or holds a domain name. When corporations and companies register a domain name, the registrant should be the company name (not an individual employee within the company) to ensure the business maintains ownership of the domain name.
[source](https://www.webopedia.com/TERM/D/domain_name_registrant.html)

> The Registrant is the licensee of the domain name - this is the individual or company who has the right to use, sell or destroy a domain name. The Administrative Contact is the licensee's appointed agent for the functions above and any other purpose. In the case that the Registrant is a company the Administrative Contact should be an employee, director, manager of the company. This person has full authority for all changes in our system so care should be taken in selecting the appropriate person to nominate for this role.
[source](https://vcgcorporate.zendesk.com/hc/en-us/articles/214821983-What-is-a-registrant-administrative-technical-billing-contact-)

### Cars

> “Registrant” means a person, firm or corporation in whose name or names a vehicle is properly registered.
[source](https://www.lawinsider.com/documents/g6BvyMB8vne#registrant)

### Drugs

> 80. "Registrant" shall mean a pharmacy or other entity which is registered with the Mississippi Board of Pharmacy to buy, sell, destroy or maintain controlled substances.
[source](https://www.lawinsider.com/documents/hN9NoBWcrKh#registrant)

### Hunting animals
> “Registrant” means a person or organization that holds an interest in land including a title holder that registers annually to possess a whitetail hunting preserve operator’s registration issued by the Iowa department of natural resources.
[source](https://www.lawinsider.com/documents/j9Uuey12MIw#registrant)


I think the question we need to answer is what we should call the entity that maintains ownership of a project - if a user registers a project to their own handle, they become its registrant. If they register a project on behalf of an org with the intent to share ownership with other members of that org, they are more like an administrative contact. The registrant in this case is the org because the project in question belongs to a collective entity, just as  a company's domain name belongs to the company (the registrant) and not to the particular employee who submitted the registration.

cloudhead:
I think what’s confusing here and perhaps different, is that the “registrant” is a (sub) registry in itself. The org isn’t owning an entry in some external registry, it’s owning an entry within itself — so the ownership is implied.

cloudhead:
In other words, “registrant” is not incorrect, just misleading, because it doesn’t highlight the most important attribute of an org within this context: that it is a namespace in the registry.

xla:
The outcome of this seems to be inconclusive. Closing for now. Let's follow-up with a fresh change-set if need be.

Issue #289, opened 2020-04-16
Repository selector


cloudhead:
One of the things missing from the current project browsing is the ability to switch repository by user.

Each user has the ability to publish their own copy of a project (a repository). These repositories should be browsable by other users who track these copies.

Hence, all source/commit browsing should *always* be in the context of a repository, ie. a `{project, user}` pair.

This has repercussions in the UI, but also in the routing, which will have an additional route element (the user).

Though for now this is a blocker for code browsing, the underlying changes are necessary for Code Sharing II as well, and that is the milestone it belongs to.

Spec here: https://www.notion.so/monadic/Repository-Selector-75e01aa8abb944eabb576cb0e4130348

* [x] Project page UI refresh
* [x] Repository/revision selector UI
  - [x] adjust repo selector visuals: https://github.com/radicle-dev/radicle-upstream/issues/570
* [x] Repository selector hooked up with backend
* [x] Commit, Source and Blob views are in the context of a user/repository

PR #510, opened 2020-06-18
feat: integrate repository selector

rudolfs:
Backend part of: #289.

### DIFF
--- a/proxy/src/http/source.rs
+++ b/proxy/src/http/source.rs
@@ -348,19 +348,22 @@ mod handler {
         let revs = remotes
             .into_iter()
             .map(|(remote, refs)| {
-                let hash = Hash::hash(remote.as_bytes());
+                // TODO(finto): Can we do this by not going through string?
+                let hash = Hash::hash(remote.to_string().as_bytes());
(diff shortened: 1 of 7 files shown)


proxy/src/http/source.rs
```diff
@@ -344,26 +344,40 @@ mod handler {
...
+            let refs = if remote == &peer_id {
+                let browser = repo.browser(project.default_branch()).unwrap();
+                coco::local_branches(&browser).unwrap()
+            } else {
+                let refs = storage.rad_refs_of(&project.urn(), remote.clone()).unwrap();
+                refs.heads.keys().cloned().map(coco::Branch).collect()
+            };
+
+            (remote, refs)
+        });
```
cloudhead, line 365: @kim when implementing this, we ran into the fact that `rad_refs().remotes` includes your own PeerId in the list of remotes. We work around this by checking whether it's us, and if so, getting local branches from surf. Is there a better way to use this API?

rudolfs:
Superseded by: https://github.com/radicle-dev/radicle-upstream/pull/510.

PR #511, opened 2020-06-18
feat(proxy): User Creation

### LINKED ISSUE #478: Create User
We want to have a function in proxy that creates a `User`.

FintanH:
Depends on https://github.com/radicle-dev/radicle-link/pull/186

Fixes #478 

This solves the ability to create and get a user in the monorepo, as well as linking it up with the identity create and get API.

Also threw in changes for get project, since the change we rely on in `radicle-link` allows us to fix that.

### DIFF
--- a/cypress/integration/user_registration.spec.js
+++ b/cypress/integration/user_registration.spec.js
@@ -1,12 +1,12 @@
 context("user registration", () => {
   before(() => {
     cy.nukeAllState();
-    cy.nukeCache();
     cy.registerAlternativeUser("nope");
     cy.createProjectWithFixture();
   });
 
   beforeEach(() => {
+    cy.nukeCocoState();
     cy.nukeSessionState();
     cy.createIdentity();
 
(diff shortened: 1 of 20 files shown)


proxy/src/coco.rs
```diff
@@ -550,14 +550,15 @@ pub fn default_config(
...
+#[allow(clippy::panic)]
 mod test {
     use librad::keys::SecretKey;
 
     use crate::error::Error;
 
     #[tokio::test]
     async fn test_can_create_user() -> Result<(), Error> {
-        let tmp_dir = tempfile::tempdir().unwrap();
+        let tmp_dir = tempfile::tempdir().expect("failed to create temdir");
```
xla, line 561: You can avoid this by adding `clippy::unwrap_used`.

FintanH: Ah the error messages are useful at the same time :shrug:

cypress/integration/user_registration.spec.js
```diff
@@ -90,23 +90,23 @@ context("user registration", () => {
     before(() => {
       // Clear everything again so transaction center is empty
       cy.nukeAllState();
-      cy.nukeCache();
       cy.createProjectWithFixture();
     });
 
-    it("shows the correct transaction details for confirmation", () => {
+    it("shows the correct transaction details for confirmation", async () => {
```
rudolfs, line 96: We should never use `async` in cypress specs as all commands are async by default and Cypress manages the queueing:
https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Commands-Are-Asynchronous

Besides, there is no `await` in the body.

PR #523, opened 2020-06-19
fix(ui): check id availability in org and user namespaces

### LINKED ISSUE #492: Fix validation bug that causes tests on CI to randomly fail

### LINKED ISSUE #519: Adjust org and user id availability check
When checking if an id is available to register an org or a user with, we check if there is any entity of the respective type already registered with said id.

However, we are consolidating user and org ids in the same namespace so we have to adjust the validations here to follow that requirement.

NunoAlexandre:
Closes #519
Fixes #492


NunoAlexandre:
@xla we should add tests covering the cases where an entity of the other type is registered with a given id. Now we only test within the same entity. Could you enlighten me on how?

ui/Screen/UserRegistration/PickHandleStep.svelte
```diff
@@ -3,7 +3,8 @@
   import validatejs from "validate.js";
 
   import * as notification from "../../src/notification.ts";
-  import * as user from "../../src/user.ts";
+  import * as id from "../../src/id.ts";
```
xla, line 6: Let's order imports alphanumerically.

ui/src/id.ts
```diff
@@ -0,0 +1,10 @@
+import * as org from "./org";
+import * as user from "./user";
+
+// Check if the given id is already taken
+export const isTaken = (id: string): Promise<boolean> =>
+  org.getOrg(id).then(org => !!org || user.get(id).then(user => !!user));
```
NunoAlexandre, line 6: @xla fyi, first I did:

```
org.getOrg(id).then(org => !!org)  || user.get(id).then(user => !!user);
```

While it worked, the pre-commits complained that I was using a promise in a bool condition, which made sense.

ui/src/id.ts
```diff
@@ -0,0 +1,10 @@
...
+
+// Check if the given id is available
+export const isAvailable = (id: string): Promise<boolean> =>
```
xla, line 9: We should consolidate these two and use `isAvailable` everywhere.

ui/src/id.ts
xla: You should chain it with `then` calls. Happy to pair on that.

xla:
> @xla we should add tests covering the cases where an entity of the other type is registered with a given id. Now we only test within the same entity. Could you enlighten me on how?

We use cypress for our assertions for user flows. You can find the specs in the cypress directory and info over on their [homepage](https://www.cypress.io/). Also happy to pair on that one.

ui/src/id.ts
NunoAlexandre: Would be nice :) I'd expect what I tried and what I have to be "chain it with 'then' calls", look forward to pairing. thanks

ui/Screen/UserRegistration/PickHandleStep.svelte
NunoAlexandre: Is there a CI check that can run for these?

ui/src/id.ts
NunoAlexandre: If we can use `isAvailable` in a readable way in both site calls, sure! Thing is that at the moment, in one site call we need to check for `isAvailable` and in the other for `isTaken`. Having both is, in my view, nice given their current usage, it offers a more readable API for both scenarios, with no logic duplication. Happy to discuss it further!

NunoAlexandre: @xla in https://github.com/radicle-dev/radicle-upstream/pull/523/commits/b9d89bae19ec76676f1d8ed5688e7c0135a9c4f2 I have dry and clean this up by unifying the validation stores that were inconsistent across `OrgRegistration.svelte` and `PickHandleStep.svelte`. The caveat is that it also unifies terminology, now always mentioning "id". We can either make that depend on some input or we can simplify the language to be agnostic. For instance, we could:

- `Sorry, this id is already taken` -> `Sorry, this one is already taken` or `Sorry, it is already taken`
 - `Id is required` -> `This field is required`
- `Id should match (...)` -> `It should match (...)`

NunoAlexandre:
I am having an issue with the UI tests.

1. The output I get running `yarn test` locally differs from the output I get from the CI

    * Example with revision https://github.com/radicle-dev/radicle-upstream/pull/523/commits/c08161897e3f60f0f8e608ed7ece4f02d65c3bd3

differ-ci-tests


   * Example with passing revision on the CI https://github.com/radicle-dev/radicle-upstream/pull/523/commits/cd4244791ae7e0b812bdc02118d1199d76019d16 (see above), failing locally now as well.


2. `yarn test` and `yarn test:debug` produce different results. With `yarn test` I see just one error (see screenshot above) while with `yarn test:debug` I get _many_ failing tests.

Following the test case manually through the app reveals the expect, correct behaviour.

xla:
@NunoAlexandre Merged master into this branch, seems all green now.

NunoAlexandre:
> @NunoAlexandre Merged master into this branch, seems all green now.

That's a wonderful way to wake up right there!

NunoAlexandre:
@xla I see that the problematic tests were `skip`ped. Adding the new ones we wanted to add here will make it fail again. I am adding them but `skip` them too. With that, could you have another look?

ui/src/id.ts
NunoAlexandre: Done in https://github.com/radicle-dev/radicle-upstream/pull/523/commits/b4f7bd6ea5f86d56963211605f6420d5c7818932

NunoAlexandre:
@xla thanks for the fix. did you follow some process to get to it? I'd like to know how you to got there.
As for this PR, I think we have everything now. Want to have a look and approve? Thanks :hand:

xla:
> @xla thanks for the fix. did you follow some process to get to it? I'd like to know how you to got there.
> As for this PR, I think we have everything now. Want to have a look and approve? Thanks hand

@NunoAlexandre :hand: - What I did is pretty simple in the end, I added a reactive print for the current value of the handle and the validation store: `$: console.log(handle, $validation)`, which would show transition to `Success` after the errors already rolled in. This indicated out of order validations, likely caused by asynchronous calls returning, as `runValidation` is invoked on every input change. The reason we don't see that when using the app is, that we don't type fast enough to out run the response time from proxy. I then guarded callbacks to `update` with a check if the input this call was made for is also the one latest stored in the `inputStore`. While being a stop-gap solution until we properly serialise the access, it will do the trick. Learning I guess is that even in a single-threaded environment like this it's not always easy to reason about state and order.

From there it was only some cypress touch-ups, setting up the correct state for the tests and hitting the right selectors for elements in the dom. Check 14ecee2 and 002ab23.

PR #415, opened 2020-05-22
refactor(ui): move to screen based layout

xla:
*work in progress*

PR #524, opened 2020-06-19
style(ui): add transaction center animations

juliendonck:
animation

### DIFF
--- a/ui/App/TransactionCenter.svelte
+++ b/ui/App/TransactionCenter.svelte
@@ -1,5 +1,4 @@
 <script>
-  import { fly } from "svelte/transition";
   import { push } from "svelte-spa-router";
 
   import * as path from "../src/path.ts";
@@ -23,7 +22,7 @@
 
 <Remote {store} let:data={transactions}>
   {#if transactions.length > 0}
-    <div class="center" in:fly={{ delay: 240, duration: 320, x: 500 }}>
+    <div class="center">
       <Center on:select={select} summary={$summary} {transactions} />
     </div>
   {/if}
--- a/ui/App/TransactionCenter.svelte
+++ b/ui/App/TransactionCenter.svelte
@@ -1,4 +1,5 @@
 <script>
+  import { fly } from "svelte/transition";
   import { push } from "svelte-spa-router";
 
   import * as path from "../src/path.ts";
@@ -9,10 +10,23 @@
   import Center from "../DesignSystem/Component/Transaction/Center.svelte";
 
   const select = event => push(path.transactions(event.detail));
+
+  $: centerIn = { delay: 240, duration: 320, x: 500 };
 </script>
+
+<style>
+  .center {
+    bottom: 32px;
+    position: fixed;
+    right: 32px;
+    z-index: 900;
+  }
+</style>
 
 <Remote {store} let:data={transactions}>
   {#if transactions.length > 0}
-    <Center on:select={select} summary={$summary} {transactions} />
+    <div class="center" in:fly={centerIn}>
+      <Center on:select={select} summary={$summary} {transactions} />
+    </div>
   {/if}
 </Remote>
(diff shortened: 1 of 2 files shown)


ui/App/TransactionCenter.svelte
```diff
@@ -9,10 +10,23 @@
...
+    position: fixed;
+    right: 32px;
+    z-index: 900;
+  }
+</style>
+
 <Remote {store} let:data={transactions}>
   {#if transactions.length > 0}
-    <Center on:select={select} summary={$summary} {transactions} />
+    <div class="center" in:fly={centerIn}>
```
rudolfs, line 28: You can inline the params like so:
```suggestion
    <div class="center" in:fly={{ delay: 240, duration: 320, x: 500 }}>
```

then you don't need the ` $: centerIn = { delay: 240, duration: 320, x: 500 };` above.

ui/DesignSystem/Component/Transaction/Center.svelte
```diff
@@ -62,8 +65,14 @@
...
   data-cy="transaction-center">
-  <div class="list-wrapper" class:hidden>
-    <List on:select {transactions} />
-  </div>
+  {#if !hidden}
+    <div
+      class="list-wrapper"
+      class:hidden
+      in:slide={listIn}
+      out:slide={listOut}>
```
rudolfs, line 73: Params can be inlined here as well.

ui/App/TransactionCenter.svelte
xla: The inlining doesn't really help with readability, having them in assignments also lends it well to be extracted into  library code.

ui/DesignSystem/Component/Transaction/Center.svelte
xla: Not in favour of this.

xla:
> There's a weird double sidebar popping up on the left side when I click on the transaction:

> star The changes look good to me, but is it intentional that the animation of the transaction center happens whenever I switch screens? I would expect it to happen only if a (new) transaction occurs or only on the first transaction.

I ran into that as well when originally building the animations out. Therefore I started #415 which will address both the double layout and the reoccurrence of the animations. In the meantime we should be able to live with it for now.

ui/App/TransactionCenter.svelte
rudolfs: In that case the `$:` reactive statement is not necessary.

ui/DesignSystem/Component/Transaction/Center.svelte
```diff
@@ -22,19 +23,21 @@
     }
   };
 
+  $: listIn = { duration: 360 };
+  $: listOut = { duration: 240 };
```
rudolfs, line 27: This could be just `let listIn` and `let listOut`, right? (In case you decide against inlining).

juliendonck:
@rudolfs @xla even though I agree with xla that it would be nice to abstract them out in to general variables we can reuse all over the app, I think for now it's actually easier having them inline to replace them later on. Since I'll probably own this in the long run, I'll abstract them out once we have defaults all over the app.

juliendonck:
> There's a weird double sidebar popping up on the left side when I click on the transaction:
> weird

@rudolfs I was only able to reproduce that on the empty profile page. It seems to have to do something with the `Notfound` route being called when navigating to the transaction. I wonder if we even need a notfound here. When would that happen?

xla:
We'll address the layout bug in #415 as a follow-up.

brandonhaslegs:
Also the transaction list should animate into the viewport from the bottom instead of the right, so all the animations are going the same direction.

juliendonck:
@brandonhaslegs @MeBrei decided to remove the slide in animation, until we have what @xla was working on here #415 in place. The animation repeating all the time looked broken.

xla:
#540

PR #526, opened 2020-06-19
feat: diff integration

cloudhead:
Diff integration with backend.

For now, created files only show up as paths, due to how libgit2 works. Later, we can show them as modified files with the full diff.

I've renamed `Commit` to `CommitHeader`. `Commit` is now the header plus the stats and the `Diff`.

2020-06-22-151853_1278x1414_scrot


proxy/src/coco/source.rs
```diff
@@ -42,8 +42,27 @@ pub struct Person {
...
+}
+
+/// Representation of a changeset between two revs.
 pub struct Commit {
+    /// The commit header.
+    pub header: CommitHeader,
+    /// The change statistics for this commit.
+    pub stats: CommitStats,
+    /// The changeset introduced by this commit.
+    pub diff: surf::diff::Diff,
```
FintanH, line 61: So this is always the `Diff` between this `Commit` and its parent?

proxy/src/coco/source.rs
```diff
@@ -42,8 +42,27 @@ pub struct Person {
...
+}
+
+/// Representation of a code commit.
+pub struct CommitHeader {
```
FintanH, line 65: I'm not sure the renaming makes sense to me here. I think the `Commit` struct above that contains the `Diff` could be named `Changeset` instead. A `Commit` is the details of some change made, whereas the above is capturing the differences between two of those changes.

proxy/src/coco/source.rs
```diff
@@ -260,18 +281,66 @@ pub fn local_state(repo_path: &str) -> Result<LocalState, error::Error> {
...
+    let diff = if let Some(parent) = commit.parents.first() {
+        browser.diff(*parent, oid)?
+    } else {
+        browser.initial_diff(oid)?
+    };
+
+    let mut deletions = 0;
+    let mut additions = 0;
+
+    for file in diff.modified.iter() {
```
FintanH, line 322: This could probably be expressed with some `Iterator` combinators, but not sure if it's worth it :thinking:

proxy/src/coco/source.rs
```diff
@@ -260,18 +281,66 @@ pub fn local_state(repo_path: &str) -> Result<LocalState, error::Error> {
...
-    Ok(Commit::from(commit))
+    Ok(CommitHeader::from(commit))
+}
+
+/// Retrieves a [`Commit`].
+///
+/// # Errors
+///
+/// Will return [`error::Error`] if the project doesn't exist or the surf interaction fails.
+pub fn commit<'repo>(browser: &mut Browser<'repo>, sha1: &str) -> Result<Commit, error::Error> {
```
FintanH, line 306: We could have a general case between any two commits as well right? Where this is the specialised case of comparing to the parent?

proxy/src/coco/source.rs
```diff
@@ -42,8 +42,27 @@ pub struct Person {
...
 }
 
-/// Representation of a code commit.
+/// Commit statistics.
+#[derive(Serialize)]
+pub struct CommitStats {
+    /// Additions.
+    pub additions: u64,
+    /// Deletions.
+    pub deletions: u64,
```
rudolfs, line 51: `u64` ought to be enough for anybody. 😬

ui/DesignSystem/Component/SourceBrowser/FileDiff.svelte
```diff
@@ -2,6 +2,39 @@
   import { Title, Icon } from "../../Primitive";
 
   export let file = null;
+
+  const lineNumL = line => {
```
rudolfs, line 6: Nit: I'd prefer non-abbreviated naming in the UI code.

Eg.: `lineNumL` -> `lineNumberLeft`, `lineNum` -> `lineNumber`, etc.

rudolfs:
Oh, and you should probably look into the test failures before merging.

proxy/src/coco/source.rs
cloudhead: Yeah, exactly, since it's on a `Commit` object.

proxy/src/coco/source.rs
cloudhead: Yeah I know, but don't think it's worth it heh.

proxy/src/coco/source.rs
cloudhead: Yeah, that's actually what I had previously, but then realized I would have to change the API endpoints to match :cry:, because it's currently `/commit/<sha>`. I was also thinking that within the scope of the UI, calling it a commit is fine. GitHub does the same, eg. https://github.com/radicle-dev/radicle-upstream/commit/8bf7ea37d10b9e9c8352abc42e3e3d4d9398dd19, even though this isn't technically "just" the commit.

proxy/src/coco/source.rs
cloudhead: Yeah, though we won't be needing it until we introduce revisions, down the line..

proxy/src/coco/source.rs
FintanH: But what do _we_ want?! If it makes sense to change the endpoints now I'd say we go for it.

proxy/src/coco/source.rs
FintanH: Well how about we make a private function that takes two `Oid`s and express `commit` in terms of that?

proxy/src/coco/source.rs
cloudhead: I think from the UI perspective, a commit is everything that is shown on the commit page -- ie. it doesn't have to mean the same thing as from surf's perspective.

FintanH: Cool, that makes sense.

proxy/src/coco/source.rs
cloudhead: So `pub fn diff<'repo>(browser: &mut Browser<'repo>, left: Option<&str>, right: &str) -> Result<Diff, error::Error>`?

FintanH: I'd be more inclined to do `Oid` instead of `&str` because when you get the `Commit`'s parent it will be an `Oid` instead of `&str` and then you'll end up doing `to_string()` and inside there `from_str`. Goad I hate strings.

cloudhead: It's a bit ugly, it's also not clear where the stats code goes then.

FintanH: Ah, I think I see what you mean. Nevermind, we can think about it again :)

proxy/src/coco/source.rs
FintanH: You could use `usize` and avoid disabling the integer arithmetic and its warning of overflow.

PR #527, opened 2020-06-19
List Entities

FintanH:
Allows us to list projects and users via the `Peer` and means that project and user retrieval is persistent across runs.


FintanH:
Emmm... I think the basing off the previous branch has messed this up.

PR #528, opened 2020-06-19
feat(proxy): List Entities

FintanH:
Allows us to list projects and users via the Peer and means that project and user retrieval is persistent across runs.

### DIFF
--- a/proxy/src/coco.rs
+++ b/proxy/src/coco.rs
@@ -83,7 +83,10 @@ impl Peer {
     ///
     /// The function will error if:
     ///   * A lock was poisioned. See [`Self::with_api`].
-    #[allow(clippy::wildcard_enum_match_arm, clippy::match_wildcard_for_single_variants)]
+    #[allow(
+        clippy::wildcard_enum_match_arm,
+        clippy::match_wildcard_for_single_variants
+    )]
     pub fn list_projects(&self) -> Result<Vec<project::Project<entity::Draft>>, error::Error> {
         self.with_api(|api| {
             let storage = api.storage();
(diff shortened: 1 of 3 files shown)


FintanH:
Dependent on patch mentioned in https://github.com/radicle-dev/radicle-link/pull/190#discussion_r442916915

proxy/src/coco.rs
```diff
@@ -83,7 +60,6 @@ impl Peer {
         let (api, _futures) = peer.accept()?;
         Ok(Self {
             api: Arc::new(Mutex::new(api)),
-            projects: Arc::new(Mutex::new(HashMap::new())),
```
xla, line 86: :heart_eyes_cat:

proxy/src/coco.rs
```diff
@@ -101,15 +77,58 @@ impl Peer {
...
+    ///
+    /// # Errors
+    ///
+    /// The function will error if:
+    ///   * A lock was poisioned. See [`Self::with_api`].
+    #[allow(
+        clippy::wildcard_enum_match_arm,
+        clippy::match_wildcard_for_single_variants
+    )]
+    pub fn list_users(&self) -> Result<Vec<user::User<entity::Draft>>, error::Error> {
```
xla, line 117: Are we using this anywhere?

FintanH: Not yet, but I'm sure it'll be useful for feed-like features.

FintanH:
This has a bug in it, given that I introduced it in `radicle-link` (see comment above). It might be worth waiting for https://github.com/radicle-dev/radicle-link/pull/197 to land, or else we merge and patch after. What do you think @xla?

proxy/src/coco.rs
xla: Either we cover it with a test or we add when we need it, otherwise just feels like a liability.

xla:
> This has a bug in it, given that I introduced it in `radicle-link` (see comment above). It might be worth waiting for [radicle-dev/radicle-link#197](https://github.com/radicle-dev/radicle-link/pull/197) to land, or else we merge and patch after. What do you think @xla?

Sound plan, wanna make this PR draft n the meantime?

FintanH:
I don't think you can go back from a "Ready for Review" to "Draft". Maybe reject it and I'll fix it up. The radicle-link PR has been merged now.

PR #538, opened 2020-06-22
fix(ui): show ProfileProjects page on startup

rudolfs:
This fixes a bug where the profile menu items wouldn't show up when the app is started.

Before:
Screenshot 2020-06-22 at 16 19 52

After:
Screenshot 2020-06-22 at 16 20 18


rudolfs:
> +1 -3

@xla it's the little things in lyfe...

PR #541, opened 2020-06-22
chore(proxy): crush build times

xla:
By boxing all our filters.

### DIFF
--- a/proxy/src/http.rs
+++ b/proxy/src/http.rs
@@ -21,6 +21,34 @@ mod source;
 mod transaction;
 mod user;
 
+/// Helper to combine the multiple filters together with Filter::or, possibly boxing the types in
+/// the process. This greatly helps the build times for `ipfs-http`.
+macro_rules! combine {
+    ($x:expr, $($y:expr),+) => {
+        {
+            let filter = boxed_on_debug!($x);
+            $(
+                let filter = boxed_on_debug!(filter.or($y));
+            )+
+            filter
+        }
+    }
+}
+
+#[cfg(debug_assertions)]
+macro_rules! boxed_on_debug {
+    ($x:expr) => {
+        $x.boxed()
+    };
+}
+
+#[cfg(not(debug_assertions))]
+macro_rules! boxed_on_debug {
+    ($x:expr) => {
+        $x
+    };
+}
+
 /// Main entry point for HTTP API.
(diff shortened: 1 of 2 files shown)


rudolfs:
I wonder what the run-time and debugging implications of all the boxing are.

If we bump into any issues, we could also do the boxing dynamically for debug builds via a macro like in this PR: https://github.com/rs-ipfs/rust-ipfs/commit/ae3306686209afa5911b1ad02170c1ac3bacda7c (found [via](https://github.com/seanmonstar/warp/issues/507#issuecomment-615974062)).

cloudhead:
> I wonder what the run-time and debugging implications of all the boxing are.

I'll buy everyone in the office a donut if it makes a difference!

xla:
> I wonder what the run-time and debugging implications of all the boxing are.

The runtime implications are an increase in vtable lookups, we won't see a difference.

> If we bump into any issues, we could also do the boxing dynamically for debug builds via a macro like in this PR: [rs-ipfs/rust-ipfs@ae33066](https://github.com/rs-ipfs/rust-ipfs/commit/ae3306686209afa5911b1ad02170c1ac3bacda7c) (found [via](https://github.com/seanmonstar/warp/issues/507#issuecomment-615974062)).

I applied the macro instead as it isolates the change better, thanks for the reminder.

cloudhead:
This is great, what kind of build time reduction are we looking at?

I would be in favor of always boxing, as it may reduce binary size quite a bit too (we should check).

xla:
> This is great, what kind of build time reduction are we looking at?

After cargo clean, from 11min to 2:20. Incremental builds of the proxy binary from 7/8mins down to 20s.

NunoAlexandre:
Super awesome!

proxy/src/http.rs
```diff
@@ -21,6 +21,23 @@ mod source;
 mod transaction;
 mod user;
 
+/// Helper to combine the multiple filters together with Filter::or, possibly boxing the types in
```
NunoAlexandre, line 24:
```suggestion
/// Helper to combine the multiple filters together with Filter::or, boxing the types in
```
?

rudolfs: Whoops, I got too excited about the PR and forgot to apply this change : |

PR #551, opened 2020-06-23
style(ui): Improve profile & org header

### LINKED ISSUE #543: Profile header fixes
- [x] Should have an @ in front of the big username
- [x] Should have a [ Register your handle ] button after the username
image
---
- [x] the ID should be truncated. right now it just flies off the viewport. We should just always format it like this: `username` + `@rad:git:` + `20 characters of the unique Radicle ID` + `…` since it's just so long and awkward to display the full ID
image
---
- [x] "Radicle ID" label breaks in a weird way
image
---
- [x] Radicle ID should be click-to-copy
- [x] on hover of the Radicle ID, the `icon-mini-copy` icon should appear
image

juliendonck:
closes #543


ui/DesignSystem/Component/BigHeader.svelte
```diff
@@ -1,9 +1,26 @@
...
+  import { Copyable } from "../../DesignSystem/Component";
+
+  const dispatch = createEventDispatcher();
 
   export let style = null;
   export let entity = null;
   export let variant = null; // profile | org
+
+  const onRegisterAction = () => {
+    dispatch("registerAction");
```
rudolfs, line 13: I'd call the function `onRegisterHandle` and the event `registerHandle`.

PR #555, opened 2020-06-23
feat(proxy): start up peer run loop

### LINKED ISSUE #455: Implement Peer Run Loop
We're discarding the event loop in https://github.com/radicle-dev/radicle-upstream/pull/414. In the next steps, we should implement the handling of this loop and allow the peer to subscribe to events. There are a couple of small examples of this behaviour in `radicle-link/tests/propogation_basic.rs`.

https://github.com/radicle-dev/radicle-upstream/blob/600f29ba57f69601979402e9978b60fab8d0610e/proxy/src/coco.rs#L79-L85

Follow-up to #414

xla:
Closes #455


FintanH:
I'd suggest trying to add some tests, one similar to the test case in `radicle-link` might make sense. I'd like to convince myself that it works :smile:

xla:
> I'd suggest trying to add some tests, one similar to the test case in `radicle-link` might make sense. I'd like to convince myself that it works smile

This is pending on the extraction of the testnet code in link, would not make that a prerequisite to this change.

FintanH:
I kind of don't see the full relevance of this PR then. I understand that we'd need to implement this, but if we can't test it and we're not sure if we can build on top of it, then what's the use of merging it yet?

proxy/src/coco.rs
```diff
@@ -55,9 +57,30 @@ impl Peer {
...
+        let api_notifications = subscriber.await.for_each(|notification| {
+            trace!("API: {:?}", notification);
+
+            futures::future::ready(())
+        });
+        tokio::spawn(api_notifications);
+
+        tokio::spawn(async move {
+            run_loop.await;
+        });
```
cloudhead, line 82: Async golf :golf: -- is this still necessary?

proxy/src/coco.rs
```diff
@@ -55,9 +57,30 @@ impl Peer {
...
-        // TODO(finto): discarding the run loop below. Should be used to subsrcibe to events and
-        // publish events.
-        let (api, _futures) = peer.accept()?;
+
+        let (api, run_loop) = peer.accept()?;
+
+        let protocol = api.protocol();
+        let protocol_subscriber = protocol.subscribe().await;
+        let protocol_notifications = protocol_subscriber.for_each(|notification| {
+            trace!("PROTOCOL: {:?}", notification);
```
cloudhead, line 66: If we leave these in, maybe we can be more helpful, eg. `trace!("Received protocol notification: {:?}, notif)`

xla:
Outdated, is being addressed somewhere else.

Issue #443, opened 2020-06-03
Integrate repository selector


rudolfs:
This issue tracks e2e integration of the repository selector.

- [ ] it shows the actual project collaborators
  - [ ] all source browsing endpoints take an additional parameter - the collaborator
- [ ] it shows the branches of each collaborator

<img width="50%" alt="Screenshot 2020-06-03 at 09 31 14" src="https://user-images.githubusercontent.com/158411/83608327-05802280-a57d-11ea-9b98-5d382875f24e.png">

Implements UI of #289.

Depends on:
- [x] https://github.com/radicle-dev/radicle-surf/milestone/2
- [ ] https://github.com/radicle-dev/radicle-link/issues/95
- [ ] https://github.com/radicle-dev/radicle-link/issues/142

cloudhead:
One question is: since remotes are based on the `PeerId`, and a user can have multiple of PeerIds, what exactly do we show in the repository selector: users, or peers?

xla:
@cloudhead We going to show users check radicle-dev/radicle-link#181

FintanH:
This makes me wonder what the behaviour of a user will be. If I join Radicle and have three machines, would it be best practice to add 4 user profiles: one for each machine and a unified profile that is both of those? Because if I'm one machine I'd like to easily differentiate the three machines's state of the repository. Presumably, the default one would be this machine's monorepo, but then how do I tell the other two apart from one another than giving a name to them.

Then following that my peers who track me would probably want to look at a default machine too. 

Maybe it's something to think about later.

cloudhead:
Yeah, the problem with showing users is the case where a user has multiple PeerIds, and thus repositories. In the end, the repo is mapped to PeerId, not user.

FintanH:
Putting a note here that names should be `handle@<peer_id>` or some prefix of the `<peer_id>` in case you're tracking a storm of cloudheads that all have their own copy of the code.

It would be good for `librad`'s future CLI to have a consistent output when using `git branch -a` :)

PR #564, opened 2020-06-23
refactor(ui): move all repository browsing state into store

rudolfs:
So far part of the source browsing state was in the router and another part in the global store. With this PR all the state will live in the Typescript/Svelte store.

We don't gain anything by having this state in the browser's window.location as it isn't presented to the user. And if we will want to present it to the user it's going to be easier to just grab it from the store.

This is in preparation of the UI changes for https://github.com/radicle-dev/radicle-upstream/issues/443.

### DIFF
--- a/ui/DesignSystem/Component/SourceBrowser/File.svelte
+++ b/ui/DesignSystem/Component/SourceBrowser/File.svelte
@@ -1,20 +1,8 @@
 <script>
-  import { link } from "svelte-spa-router";
-
-  import * as path from "../../../src/path.ts";
-  import {
-    currentPath,
-    currentRevision,
-    ObjectType,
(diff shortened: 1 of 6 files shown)


rudolfs:
> What I would flag is that this will make it not possible to "deep-link" to a specific path in a proejcts source tree. 

Why wouldn't it work? Instead of using `<a href={path.projectSource(...) use:link}>` you'd just do:
```
<button on:click={
    updateParams({
      path: ...,
      projectId: ...,
      revision: ...,
      type: ...,
    })}>deep link</button>
```

You can call `updateParams` from anywhere, like a callback to when a user pastes a link into the app or if the app receives a notification.

PR #565, opened 2020-06-24
feat(ui): real project stats ✨

### LINKED ISSUE #348: Real project stats
Currently we only return fake stats, we would like to show the real deal. This might be well suited be delivered by **surf**. @FintanH How would you like to see this moved forward with a downstream dependency?

### LINKED ISSUE #546: Project Stats UI Display
Once #348 is in the proxy we can serve up the real stats in the UI. Maybe there's no work to do here but creating this ticket for completion's sake.

sarahscott:
- [ ] Get https://github.com/radicle-dev/radicle-link/pull/203 & https://github.com/radicle-dev/radicle-surf/pull/144 merged
- [ ] Finish unit & integration tests

Screen Shot 2020-06-23 at 8 35 27 PM

Closes #348 
Closes #546


proxy/src/coco/source.rs
```diff
@@ -188,6 +188,10 @@ pub struct TreeEntry {
     pub path: String,
 }
 
+/// Representation of high-level project [`Stats`] - its known number of branches,
+/// commits, and contributors on the default branch
+pub struct Stats(pub surf::vcs::git::Stats);
```
FintanH, line 193: Something I realised is that since this is just for the `ToDocumentedType` nonsense, we could define it in `http/source.rs` like:
```rust
struct DocumentStats;

impl ToDocumentType DocumentStats { /* etc */ }
```

Because really it has nothing to do with `surf::Stats`

proxy/src/http/project.rs
```diff
@@ -184,11 +184,6 @@ mod handler {
                 shareable_entity_identifier,
                 metadata: meta.into(),
                 registration: None,
-                stats: project::Stats {
```
FintanH, line 187: So judging from these, maybe we made the wrong call by _adding_ and endpoint for `get_stats` here. Maybe we should actually have added the call to `browser.get_stats` where we're fetching a project. Which makes sense really. Sorry, this was my bad :hand:

proxy/Cargo.toml
```diff
@@ -32,8 +32,7 @@ url = "2.1"
 warp = { git = "https://github.com/radicle-dev/warp", branch = "openapi", features = [ "openapi" ] }
 
 [dependencies.librad]
-git = "https://github.com/radicle-dev/radicle-link.git"
-rev = "e2e5b1ed4b1aa03f718da0c3b8c5cb414df1c328"
+path = "../../radicle-link/librad"
```
xla, line 35: That doesn't work for us. :smile_cat:

proxy/src/coco.rs
```diff
@@ -88,20 +90,32 @@ impl Peer {
...
+                    .map_err(error::Error::from)
+                    .and_then(|repo| {
+                        repo.browser(project.default_branch())
+                            .map_err(error::Error::from)
+                            .and_then(|browser| {
+                                browser
+                                    .get_stats()
+                                    .map_err(error::Error::from)
+                                    .map(|stats| Project::from_project_stats(project, stats))
+                            })
```
xla, line 115: I can see @FintanH style all over it. While it's smart, it's a bit dense and hard to comprehend, maybe we can break it up in smaller bits.

proxy/src/http/org.rs
```diff
@@ -16,830 +16,830 @@ use crate::registry;
 
 /// Prefixed filters.
 pub fn routes<R>(
-    peer: Arc<Mutex<coco::Peer>>,
-    registry: http::Shared<R>,
-    subscriptions: notification::Subscriptions,
+  peer: Arc<Mutex<coco::Peer>>,
+  registry: http::Shared<R>,
+  subscriptions: notification::Subscriptions,
```
xla, line 21: These and others look like miss-configured rustfmt config, we have one for the project which should be respected by editors.

proxy/src/coco.rs
FintanH: It's not my fault I can't use `?` in fucking `map` :weary: I was saying to @sarahscott that this is showing a code smell around my `with_*` code because I was running into deadlocks.

FintanH: I think I have an idea to clean this up and avoid deadlocks

proxy/src/http/org.rs
```diff
@@ -702,12 +702,12 @@ mod test {
...
                 },
                 "registration": Value::Null,
-                "shareableEntityIdentifier": urn.to_string(),
+                "shareableEntityIdentifier": format!("%{}", urn),
                 "stats": {
-                    "branches": 11,
-                    "commits": 267,
-                    "contributors": 8,
-                },
+                  "branches": 2,
```
FintanH, line 707: An issue we have is that the `surf::Stats` has `*Count` as the names. We should either update here, or we update `surf`.

proxy/src/coco.rs
```diff
@@ -72,10 +74,10 @@ impl Peer {
     /// [`std::sync::Mutex::lock`] for further details.
     pub fn with_api<F, T>(&self, f: F) -> Result<T, error::Error>
     where
-        F: FnOnce(&net::peer::PeerApi) -> T,
```
FintanH, line 75: This caused more headaches than it solved, so I opted for returning `Result<T, error::Error>` and avoided `flatten` everywhere

proxy/src/coco.rs
```diff
@@ -88,20 +90,24 @@ impl Peer {
...
+            });
+            project_meta
+                .map(|project| {
+                    Self::with_browser(api, &project.urn(), |browser| {
+                        let stats = browser.get_stats()?;
+                        Ok(Project::from_project_stats(project, stats))
                     })
                 })
-                .collect())
+                .collect::<Result<Vec<Project>, error::Error>>()
```
FintanH, line 109: I think this turbo fish (`::<>`) type declaration can go

proxy/src/coco.rs
```diff
@@ -251,27 +251,24 @@ impl Peer {
...
-
-                if storage.has_urn(&urn)? {
-                    return Err(error::Error::EntityExists(urn));
-                } else {
-                    let _repo = storage.create_repo(&user)?;
-                }
+        let user = self.with_api(|api| {
+            let key = api.key();
+
+            // Create the project meta
```
FintanH, line 257: Could change the comment while we're here :sweat_smile:  Or just remove it

proxy/src/http/project.rs
```diff
@@ -204,23 +195,10 @@ mod handler {
...
-                registration: None,
-                stats: project::Stats {
-                    branches: 11,
-                    commits: 267,
-                    contributors: 8,
-                },
-            })
-            .collect::<Vec<project::Project>>();
+        let peer = peer.lock().await;
+        // TODO(sos): Is there a better way to do this? Slicker use of move? Not locking the peer
```
FintanH, line 199: Can remove this comment now.

proxy/src/http/source.rs
```diff
@@ -296,7 +298,9 @@ mod handler {
     ) -> Result<impl Reply, Rejection> {
         let urn = project_urn.parse().map_err(Error::from)?;
```
FintanH, line 299: If we move the `urn` parsing into the `with_api` block we can actually avoid the `Error::from` which is sort of nice :) I did it in a few places but missed others.

proxy/src/http/source.rs
```diff
@@ -1217,13 +1247,15 @@ mod test {
         let prefix = "src";
 
         let default_branch = "master".to_string(); // TODO(finto): need to change this
```
FintanH, line 1249:
```suggestion
        let default_branch = platinum_project.default_branch();
```
While we're here :)

ui/DesignSystem/Component/Stats.svelte
```diff
@@ -1,13 +1,19 @@
 <script>
   import { Icon, Numeric } from "../Primitive";
 
-  export let stats = [
-    { icon: Icon.Commit, count: -1 },
-    { icon: Icon.Branch, count: -1 },
-    { icon: Icon.Member, count: -1 },
-  ];
+  export let stats = {
+    contributors: -1,
```
FintanH, line 5: Why is `-1` the default here? :thinking: Just to differentiate it from a project with a genuine `0` is it?

ui/Screen/Project/Source.svelte
```diff
@@ -188,11 +188,7 @@
           <div class="repo-stat-item">
             <Icon.Commit />
             <Text style="margin: 0 8px;">
-              <!-- Nb. We don't `use:link` here because it doesn't work due to
-              some reactivity bug. -->
-              <a
-                href={`#${path.projectCommits(project.id, $currentRevision)}`}
-                data-cy="commits-button">
```
FintanH, line 195: Why is this removed for this change? :thinking:

proxy/src/http/org.rs
sarahscott: See https://github.com/radicle-dev/radicle-surf/pull/145; will monitor https://github.com/radicle-dev/radicle-link/pull/112 to figure out the best time for bumping the surf version

FintanH: FYI: #112 isn't the issue. It's what's currently on master that I think could cause issues. You could try and see by picking the head commit. Maybe it'd be good to see if you can fix anything that shows up 😁 I'm also just speculating that there's compat issues, I haven't confirmed personally

proxy/src/coco.rs
sarahscott: turbofish

i will vanquish him

ui/DesignSystem/Component/Stats.svelte
sarahscott: I believe that's the reasoning. I think `null` might make more sense for error-handling.

ui/src/project.ts
```diff
@@ -23,8 +24,8 @@ export interface Project {
   id: string;
   shareableEntityIdentifier: string;
   metadata: Metadata;
-  registration: string; // TODO(rudolfs): what will this type be?
   stats: Stats;
+  registration?: org.Org | user.User; // TODO(rudolfs): what will this type be?
```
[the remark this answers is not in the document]
sarahscott: I copied the types from `proxy` so I assume those should be the ones used here - any objection @rudolfs ?

ui/Screen/Project/Source.svelte
sarahscott: It's different on master now

proxy/src/coco/source.rs
sarahscott: Done ✨

proxy/src/http/org.rs
sarahscott: Ok I've confirmed the errors 😄 they look like: `method not found in &librad::net::peer::PeerApi`. Will continue digging into them on Monday

sarahscott:
I believe my force push broke github... those merge conflicts aren't reproducible locally. I'll fix it; just noting so others do not repeat my mistakes

ui/src/project.ts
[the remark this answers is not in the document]
rudolfs: No objection : P

proxy/src/http/org.rs
FintanH: I'm fixing this in another branch :) Sorry you've gotten caught between a rock and a hard place currently. But we should be able to avoid this if https://github.com/radicle-dev/radicle-upstream/pull/609 is accepted :+1:

Issue #512, opened 2020-06-18
Project creation feels slow when adding a large existing project


juliendonck:
When adding a large project, there's a weird break after I click the button. See gif

![slow-project-creation](https://user-images.githubusercontent.com/2326909/85044405-db8a4b00-b18d-11ea-9ca6-02cf3af6b0c7.gif)

rudolfs:
Refs: https://github.com/radicle-dev/radicle-upstream/issues/912.

PR #566, opened 2020-06-24
feat(ui): add a loading state to create project button

sarahscott:
This isn't a permanent solution because the loading time for a big project can be very tedious. We should work toward something on the project screen itself, or maybe pop to the user profile and add some kind of loading indicator to the project within the list. This PR at least disables the button so the user can't click it a bunch of times.

Related to #512 (doesn't totally fix it though)

createproj


juliendonck:
@sarahscott Was chatting with @brandonhaslegs about this. Maybe we can just grey out the button but keep the text (no spinner in the button) and just change the cursor to `waiting`.

rudolfs:
> Maybe we can just grey out the button but keep the text (no spinner in the button) and just change the cursor to waiting.

@juliendonck you mean the cursor when you hover the button?

juliendonck:
No the whole time it's loading.

xla:
@juliendonck Can you have another look?

rudolfs:
@xla I think @juliendonck meant the cursor should switch to `waiting` if you're anywhere on the screen, not only when hovering the button.

juliendonck:
Yeah i actually think this won't work. Cause I can still click on other things while it's loading. We'd need to lock the UI while loading and if that's the case then I think we're better of trying to speed this up or putting the user on the next page and doing the loading of the content there.

xla:
> Yeah i actually think this won't work. Cause I can still click on other things while it's loading. We'd need to lock the UI while loading and if that's the case then I think we're better of trying to speed this up or putting the user on the next page and doing the loading of the content there.

Agreed with that sentiment.

xla:
We still track the issue in #512 - this is not an adequate fix. Closing. :<

PR #574, opened 2020-06-24
Fix commits view

cloudhead:
* Fixes an issue with branch names that have a `/` in them not being able to be loaded.
* Fixes a reactivity bug with `svelte-spa-router`, that broke the commits link

### DIFF
--- a/ui/Screen/Project/Source.svelte
+++ b/ui/Screen/Project/Source.svelte
@@ -188,6 +188,8 @@
           <div class="repo-stat-item">
             <Icon.Commit />
             <Text style="margin: 0 8px;">
+              <!-- Nb. We don't `use:link` here because it doesn't work due to
+              some reactivity bug. -->
               <a href={`#${path.projectCommits(project.id, $currentRevision)}`}>
                 Commits
               </a>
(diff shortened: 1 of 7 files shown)


ui/Screen/Project/Source.svelte
```diff
@@ -189,9 +188,7 @@
           <div class="repo-stat-item">
             <Icon.Commit />
             <Text style="margin: 0 8px;">
-              <a
-                href={path.projectCommits(project.id, $currentRevision)}
-                use:link>
+              <a href={`#${path.projectCommits(project.id, $currentRevision)}`}>
```
rudolfs, line 191: We should add a comment about `link` not being able to handle reactivity, so this doesn't get refactored away accidentally.

rudolfs: Actually, scrap that. Upgrading to the latest `svelte-spa-router` should fix this problem:
> You can now pass a parameter to the use:link action (e.g. use:link={val}) which will take precedence over the tag's href value. You can use this to create links whose target changes reactively, for example. (Thanks to @cianx!)

https://github.com/ItalyPaleAle/svelte-spa-router/releases

If you go for the upgrade, could also fix it in this place too:
https://github.com/radicle-dev/radicle-upstream/blob/master/ui/DesignSystem/Component/HorizontalMenu/MenuItem.svelte#L45

cloudhead: Can do it in a separate ticket.

PR #580, opened 2020-06-24
feat(ui): add back button on commit page

juliendonck:
- update to the commit detail page header to bring them in line with the issues and revisions detail pages.
- adds a back button on the commit page, this will be implemented on the pages mentioned above

back-button


cloudhead:
(It also updates the header design)

PR #583, opened 2020-06-24
style(ui): abstract header as a component

juliendonck:
- Changed `BigHeader` component to `Header.Large` component

Screenshot 2020-06-24 at 22 57 15

- Abstracted the back button header out of the commit page `Header.Back`

Screenshot 2020-06-24 at 22 59 04

- Use `Header.Back` on the issue page

Screenshot 2020-06-24 at 22 59 33


juliendonck:
> Also I feel like we're moving towards using the `variant` prop for defining different variants of components.

@rudolfs Made it a component. Still like using the `Header.Something` thing as I think it's more elegant here for readability and code base navigation.

ui/DesignSystem/Component/Header/Back.svelte
```diff
@@ -0,0 +1,44 @@
+<script>
+  import { pop } from "svelte-spa-router";
```
xla, line 2: Components shouldn't touch routing, instead we should just `dispatch` an event when the back buton is clicked. With that the screen can react to that event and decide what suppose to happen.

juliendonck: Yeah I thought about that but the back button will always just pop. So I think it makes sense to contain it within that. Otherwise we need to duplicate that on all the pages you use it and that will lead to it breaking.

PR #586, opened 2020-06-25
Update Stats PR

FintanH:
I decided to create a PR against yours to highlight the changes I was considering making. We can go over it on a call if that would help :)

### DIFF
--- a/proxy/src/coco.rs
+++ b/proxy/src/coco.rs
@@ -19,14 +19,16 @@ use librad::surf::vcs::git::git2;
 use librad::uri::RadUrn;
 
 use crate::error;
+use crate::project::Project;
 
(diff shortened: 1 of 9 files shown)


proxy/src/coco.rs
```diff
@@ -88,20 +90,32 @@ impl Peer {
         clippy::wildcard_enum_match_arm,
         clippy::match_wildcard_for_single_variants
     )]
-    pub fn list_projects(&self) -> Result<Vec<project::Project<entity::Draft>>, error::Error> {
-        self.with_api(|api| {
-            let storage = api.storage();
```
FintanH, line 93: We have the api lock here. I ran into double-locking using `with_browser` but maybe we can make the change cleaner. Not sure :shrug:

PR #595, opened 2020-06-25
fix(ui): fix various routing issues

### LINKED ISSUE #406: Project README doesn't always show up
The README shows up when a project is first loaded in the `/source` page. However, if you navigate to a file, and then click on `Source` again, or click on the root folder, the readme doesn't show up.

### LINKED ISSUE #438: README is fetched with wrong project ID when switching
1. Have two projects with different README file names (eg. `README` and `README.md`).
2. Go to first one. Ok.
3. Go back to project list.
4. Go to second project. **Error: It tries to load a README with the path of the previous project.**

This might have been introduced in #335, I'm not sure.

rudolfs:
https://github.com/radicle-dev/radicle-upstream/pull/564 was a flawed idea. We **do** want to have the state in the router after all. Otherwise we won't be able to use the router history to navigate back and forth in the repo browser.

Basically @xla was right about "deep-linking" in: https://github.com/radicle-dev/radicle-upstream/pull/564#pullrequestreview-436443237.

Fixes #406
Fixes #438.

### DIFF
--- a/ui/DesignSystem/Component/SourceBrowser/File.svelte
+++ b/ui/DesignSystem/Component/SourceBrowser/File.svelte
@@ -2,19 +2,15 @@
   import { link } from "svelte-spa-router";
 
   import * as path from "../../../src/path.ts";
-  import {
-    currentPath,
-    currentRevision,
-    ObjectType,
-  } from "../../../src/source.ts";
+  import { ObjectType } from "../../../src/source.ts";
 
   import { Icon } from "../../Primitive";
 
   export let projectId = null;
   export let filePath = null;
   export let name = null;
-
-  $: active = filePath === $currentPath;
+  export let currentRevision = null;
+  export let active = false;
 </script>
 
 <style>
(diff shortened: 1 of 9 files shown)


ui/Screen/Project/Source.svelte
```diff
@@ -220,7 +213,7 @@
             rootPath={path.projectSource(project.id)}
             projectName={project.metadata.name}
             projectId={project.id} />
-        {:else if object.info.objectType === ObjectType.Tree}
+        {:else if object.path === ''}
```
xla, line 216:
```suggestion
        {:else if object.path === ""}
```

ui/DesignSystem/Component/SourceBrowser/File.svelte
```diff
@@ -1,8 +1,20 @@
 <script>
+  import { link } from "svelte-spa-router";
```
xla, line 2: Not for this PR but for a future improvement, we should have a `Link` component which hides away that we use this specific package for that concern.

Issue #397, opened 2020-05-15
Make associated entity IDs required


rudolfs:
In order to make things more unambiguous we should make the coco attestation attributes mandatory and stick to a certain naming scheme.

Identity attestation should be named `cocoIdentityId` and project attestation should be `cocoProjectId`:

- Register a new unique Org
  - [ ] response `POST /v1/orgs`: `members/maybeEntityId` → `members/maybeURN`
- Find Org by ID
  - [ ] response `GET /v1/orgs/{id}`: `members/maybeEntityId` → `members/maybeURN`
- Lists all Projects of the Org
  - [ ] response `GET /v1/orgs/{org_id}/projects`: `maybeProjectId` → `maybeURN`
- Find Project for Org
  - [ ] response `GET /v1/orgs/{org_id}/projects/{project_name}`: `maybeProjectId` → `maybeURN`
- Register a Project on the Registry
  - [ ] request `POST /v1/projects/register`: `maybeCocoId` → `URN`
- Fetch current Session
  - [ ] response `GET /v1/session`: `orgs/members/maybeEntityId` → `orgs/members/URN`
- Register a handle on the Registry
  - [ ] request `POST /v1/users`: `maybeEntityId` → `URN`
- Fetch a User
  - [ ] response `GET /v1/users/{handle}`: `maybeEntityId` → `maybeURN`
- List all orgs the user is a member of
  - [ ] response`GET /v1/users/{handle}/orgs`: `members/maybeEntityId` → `members/maybeURN`

FintanH:
To add to this: `RadUrn` has a `Deserialize` impl so we can use the type in the payload. DEATH TO STRINGS!

xla:
We removed registry related functionality from the proxy.

PR #597, opened 2020-06-25
refactor(proxy): improve api ergonomics around project registration

### LINKED ISSUE #474: Improve API ergonomics regarding project registration
In #453 we introduce the ability to register a project under a user in addition to the ability to register a project under an org.

Such a change has led @xla to suggest improving the API ergonomics regarding project registration.

@xla, my understanding of your suggestion is that now we need the API request is, in pseudo-code:

```
POST /projects/register 
    {   domain_type: Org|User, 
        domainId: a1b2, 
        projectName: upstream, 
        .. 
    }
```

So when you speak about bad ergonomics here, you might be referring to having to pass the `domainType` and `domainId`, where those are already known from the context of the call site triggering this request.

The idea would, therefore, to have:

A dedicated `orgs` endpoint:

```
POST /orgs/{org_id}/projects/register 
    {
        projectName: upstream, 
        .. 
    }
```

and a dedicated `users` endpoint:

```
POST /users/{user_id}/projects/register 
    {
        projectName: upstream, 
        .. 
    }
```

For both of these endpoints, we could then build the respective `ProjectDomain`, where the variant is known from the context and the id is obtained from the URI.

### Open questions

**Q1**: Is that the idea?
**Q2**: If so, to help me understand your point of view, can you explain how you find the suggestion a solution with better ergonomics?
   I wonder since this approach introduces duplication, where each project domain entity will need to implement the same endpoint for project registration, build its `ProjectDomain` variant, call the same handler in the `registry`. That duplication seems to introduce maintenance efforts that don't seem to outweigh an issue with ergonimics (again, as far as I see).

NunoAlexandre:
Closes #474 

Moves away from `POST projects/register/` with an intertwined payload to have `POST users/<handle>/projects/<name>` and `POST orgs/<id>/projects/<name>`. The payload, `RegisterProjectInput` and the HTTP handler `register_project` are shared and DRY for both endpoints.

### Note

1. When I introduced the new endpoints under `/users`, I ran into this issue where the result of the request was a `500`:

```
'res is Response { status: 500, version: HTTP/1.1, headers: {"content-type": "text/plain; charset=utf-8"}, body: b"Request body consumed multiple times" }', src/http/user.rs:560:9
```

Searching for `Request body consumed multiple times` led me to https://github.com/seanmonstar/warp/issues/291. So, the order in which the endpoint "filters" are combined either makes it or breaks it. That's odd. Have you seen this before?

2. I noticed that if I try to register a project twice from the app, I get the same error as reported above. I have no clue why that would happen yet. The UI should not allow a user to register a project twice anyway.

### DIFF
--- a/proxy/src/http.rs
+++ b/proxy/src/http.rs
@@ -27,7 +27,6 @@ mod project;
 mod session;
 mod source;
 mod transaction;
-
 mod user;
 
 /// Helper to combine the multiple filters together with Filter::or, possibly boxing the types in
--- a/proxy/src/http.rs
+++ b/proxy/src/http.rs
@@ -1,9 +1,17 @@
 //! HTTP API delivering JSON over `RESTish` endpoints.
 
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::convert::Infallible;
+use std::convert::TryFrom;
+use std::str::FromStr;
 use std::sync::Arc;
 use tokio::sync::{Mutex, RwLock};
-use warp::{path, Filter, Rejection, Reply};
+use warp::http::StatusCode;
+use warp::{
+    document::{self, ToDocumentedType},
+    path, reply, Filter, Rejection, Reply,
+};
 
 use crate::coco;
 use crate::registry;
@@ -19,6 +27,7 @@ mod project;
 mod session;
 mod source;
 mod transaction;
+
 mod user;
 
 /// Helper to combine the multiple filters together with Filter::or, possibly boxing the types in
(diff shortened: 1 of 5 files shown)


proxy/src/http.rs
```diff
@@ -156,6 +165,82 @@ pub fn with_subscriptions(
...
 
+/// Bundled input data for project registration, shared
+/// between users and orgs.
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RegisterProjectInput {
+    /// User specified transaction fee.
+    transaction_fee: registry::Balance,
+    /// Optionally passed coco id to store for attestion.
+    maybe_coco_id: Option<String>,
```
FintanH, line 175: Say if we changed this to `RadUrn`, would it be turned from a string into the correct type using `Deserialize`? :thinking: I'm wondering if it would be better to have stronger types on input.

rudolfs: We should also make the coco_id required, but that's another issue: https://github.com/radicle-dev/radicle-upstream/issues/397.

NunoAlexandre: @rudolfs can we move this discussion to that ticket then?

PR #598, opened 2020-06-26
feat(ui): remove application menu-bar

cloudhead:
Auto-hides the annoying menu bar. Tested on Linux only.


juliendonck:
is this a linux thing? What bar are we talking about?

xla:
@juliendonck See the bar at the top, this is for linux and windows afaik.

2020-06-26-115610_1918x2105_scrot

PR #608, opened 2020-06-29
refactor(proxy): coco peer

### LINKED ISSUE #488: Cover coco module with tests
We currently lack any coverage for the coco module.

Follow-up to #414

FintanH:
The main goal here is to move from using Peer, which wrapped the PeerApi
in an Arc + Mutex, to just having functions that take &PeerApi.
I created a new module to house these PeerApi operations -- similar in
vein to source housing Browser operations.
This means that the http endpoints take PeerApi wrapped in Arc + Mutex
and avoids the indirection.
This also meant that we need to avoid using PeerApi in functions that
involve a async function being called. If that's necessary we need to
pass an Arc + Mutex PeerApi.

The result of this is that we avoid this locking inside our peer
operations and can freely use it.

Closes #488


proxy/src/coco.rs
```diff
@@ -517,185 +48,9 @@ impl entity::Resolver<user::User<entity::Draft>> for FakeUserResolver {
 // TODO(xla): Transform into Peer::create_user.
 /// Constructs a fake user to be used as an owner of projects until we have more permanent key and
 /// user management.
-pub async fn fake_owner(peer: &Peer) -> User {
-    let key = peer
-        .with_api(|api| api.key().clone())
-        .expect("failed to get key");
+pub async fn fake_owner(key: keys::SecretKey) -> User {
```
xla, line 51: Why is this one left in here? Wouldn't it be better to have it with the peer or config module?

proxy/src/coco/peer.rs
```diff
@@ -0,0 +1,617 @@
...
+
+/// Returns the list of [`project::Project`]s for your peer.
+///
+/// # Errors
+///
+/// The function will error if:
+///   * The retrieving the project entities from the store fails.
+#[allow(
+    clippy::wildcard_enum_match_arm,
+    clippy::match_wildcard_for_single_variants
```
xla, line 47:
```suggestion
    clippy::match_wildcard_for_single_variants,
    clippy::wildcard_enum_match_arm
```

proxy/src/coco/peer.rs
```diff
@@ -0,0 +1,617 @@
...
+
+/// Returns the list of [`user::User`]s known for your peer.
+///
+/// # Errors
+///
+/// The function will error if:
+///   * The retrieving the project entities from the store fails.
+#[allow(
+    clippy::wildcard_enum_match_arm,
+    clippy::match_wildcard_for_single_variants
```
xla, line 70:
```suggestion
    clippy::match_wildcard_for_single_variants,
    clippy::wildcard_enum_match_arm
```

proxy/src/coco/peer.rs
```diff
@@ -0,0 +1,617 @@
...
+    if storage.has_urn(&urn)? {
+        return Err(error::Error::EntityExists(urn));
+    } else {
+        let _repo = storage.create_repo(&user)?;
+    }
+
+    Ok(user)
+}
+
+/// Verify a user using a fake resolver that resolves the user to itself.
```
xla, line 216: Should we leave a TODO and/or a follow up issue for this?

proxy/src/coco/peer.rs
```diff
@@ -0,0 +1,617 @@
...
+            "refs/heads/{}:{}/heads/{}",
+            default_branch, namespace_prefix, default_branch
+        )],
+        None,
+    )?;
+
+    Ok(())
+}
+
+// TODO(xla): Move into coco::source.
```
xla, line 298: Any strong reason to not resolve this TODO in the context of this PR?

proxy/src/coco/peer.rs
```diff
@@ -0,0 +1,617 @@
...
+            if *name != "master" {
+                platinum_repo.branch(name, &commit, false)?;
+            }
+        }
+    }
+
+    Ok(())
+}
+
+// TODO(xla): Move into control module.
```
xla, line 338: Could we do this as part of this PR as well?

proxy/src/coco/peer.rs
```diff
@@ -0,0 +1,617 @@
...
+            ],
+            None,
+        )?;
+    }
+
+    // Init as rad project.
+    Ok(meta)
+}
+
+// TODO(xla): Move into control module.
```
xla, line 404: Same as above.

proxy/src/coco/peer.rs
```diff
@@ -0,0 +1,617 @@
...
+        // std::fs::create_dir_all(path.clone())?;
+        replicate_platinum(peer, owner, info.0, info.1, info.2)?;
+    }
+
+    Ok(())
+}
+
+/// Acting as a fake resolver where a User resolves to itself.
+/// This allows us to check the history status of a single User.
+/// TODO(finto): Remove this once Resolvers are complete.
```
xla, line 442: :new_moon_with_face:

proxy/src/coco/peer.rs
```diff
@@ -0,0 +1,617 @@
...
+        _uri: &RadUrn,
+        _revision: u64,
+    ) -> Result<user::User<entity::Draft>, entity::Error> {
+        Ok(self.0.clone())
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::panic)]
+mod test {
```
xla, line 462: :black_heart: :black_heart:

proxy/src/coco.rs
FintanH: I think it should be removed altogether. I'm working on the `rad/self` update and we need the certifier of a `Project` to exist in the monorepo. So that means our first order of business is to create the `User` in the monorepo. Meaning we should change all the calls to `fake_owner` to `peer.init_user(key, "cloudhead")`.

Long story short, I suggest we leave it for now and it'll be deleted in a future PR.

proxy/src/coco/peer.rs
FintanH: The follow-up issue is to integrate real resolvers. I can leave a TODO :ok_hand:

proxy/src/coco/peer.rs
FintanH: Can do for both of these requests :ok_hand: :point_down:

FintanH: All three in fact

proxy/src/coco/control.rs
```diff
@@ -0,0 +1,148 @@
...
+use librad::meta::project;
+use librad::net::peer::PeerApi;
+use radicle_surf::vcs::git::git2;
+
+use crate::coco::peer::{init_project, User};
+use crate::error;
+
+/// This function exists as a standalone because the logic does not play well with async in
+/// `replicate_platinum`.
+fn clone_platinum(
```
xla: This should go in source as the original TODO mentioned as it only operates on repos on disk.

FintanH: Ya, but it's only used in `replicate_platinum` so I thought that it living beside its `pub fn` would make more sense.

PR #615, opened 2020-06-30
fix(ui): fix error display

cloudhead:
Introduce `Error` component for displaying full-width error messages.

Before:
2020-06-30-124737_1074x706_scrot

After:
2020-06-30-124345_766x706_scrot


cloudhead:
> Could have also used our existing error notifications:
> 
> ```
> import * as notification from "../../src/notification.ts";
> 
> notification.error(error.message);
> ```

Aren't those ephemeral?

PR #616, opened 2020-06-30
chore(proxy): update librad

FintanH:
Before integrating the `rad/self` feature we need to deal with the fact we're not able to get a `SecretKey` anymore from the `PeerApi`. This PR achieves that by adding the `SecretKey` as a parameter where needed. Then to get hold of the key we use the `Keystorage` to access it.

We need to have `User` in the monorepo when we create a project. This is so we can add the owner as a certifier, which _must_ be resolvable to an entity in the monorepo. This means we get rid of `fake_owner` and instead use `init_user` + `verify_user` instead.

### DIFF
--- a/proxy/src/http/control.rs
+++ b/proxy/src/http/control.rs
@@ -32,8 +32,8 @@ where
         })
         .untuple_one()
         .and(
-            create_project_filter(Arc::clone(&peer), keystore, Arc::clone(&owner))
-                .or(nuke_coco_filter(peer, owner))
+            create_project_filter(Arc::clone(&peer), Arc::clone(&keystore), Arc::clone(&owner))
+                .or(nuke_coco_filter(peer, keystore, owner))
                 .or(nuke_registry_filter(Arc::clone(&registry)))
                 .or(register_user_filter(registry)),
         )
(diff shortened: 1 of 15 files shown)


proxy/src/http/control.rs
```diff
@@ -170,14 +187,30 @@ mod handler {
...
+
+        let mut peer = peer.lock().await;
+        let mut keystore = keystore.write().await;
+
+        let mut owner = owner.write().await;
+        let new_owner = if let Some(old_owner) = &*owner {
+            let new_owner = coco::init_user(&new_peer, key, old_owner.name())?;
+            Some(coco::verify_user(new_owner).await?)
+        } else {
+            None
```
xla, line 208: Shouldn't we panic for now in this branch.

FintanH:
> The object graph we have to construct for even trivial test cases becomes quite large, we could be well served by some form of context object our fitlers can expect with a constructor just for tests. e.g. http::Context::default().

Ya agreed, the fixture setup is quite verbose and repetitive. I think it can be captured in some sort of common setup. I'm thinking of doing it with the actual use of `rad/self` integration because the setup of the peer with a default user should go in its own function I would say.

proxy/src/http/control.rs
FintanH: I'm inclined to leave it like this because it could be the case that we're testing the user not being set :thinking: I'm also unsure of how this will _exactly_ change with the `default_rad_self` change when I do that next.

I'm aiming for compatibility in this PR :)

FintanH:
:ghost: error

PR #618, opened 2020-06-30
feat: syntax highlighting

### LINKED ISSUE #461: Syntax highlighting
It would be nice to have syntax highlighting for the B1.

There's https://github.com/trishume/syntect if we want to do it in the backend (which we should).

cloudhead:
Closes #461 

This implements syntax highlighting for source code, in the backend. It handles light/dark mode correctly.

2020-06-30-162940_1381x706_scrot

### DIFF
--- a/proxy/src/coco/source.rs
+++ b/proxy/src/coco/source.rs
@@ -8,7 +8,11 @@ use radicle_surf::{
     diff, file_system,
     vcs::git::{self, git2, BranchName, Browser},
 };
+
+use syntect::easy::HighlightLines;
+use syntect::highlighting::ThemeSet;
 use syntect::parsing::SyntaxSet;
+use syntect::util::LinesWithEndings;
 
 use crate::error;
 use crate::session::settings::Theme;
@@ -235,13 +239,26 @@ pub fn blob(
         .last_commit(commit_path)?
         .map(|c| CommitHeader::from(&c));
     let (_rest, last) = p.split_last();
-    let content = match (std::str::from_utf8(&file.contents), theme) {
+
+    let content = blob_content(path, file.contents, theme);
+
+    Ok(Blob {
+        content,
+        info: Info {
+            name: last.to_string(),
+            object_type: ObjectType::Blob,
+            last_commit,
+        },
+        path: path.to_string(),
+    })
+}
+
+/// Return a [`BlobContent`] given a file path, content and theme. Attempts to perform syntax
+/// highlighting when the theme is `Some`.
+fn blob_content(path: &str, content: Vec<u8>, theme: Option<&Theme>) -> BlobContent {
+    match (std::str::from_utf8(&content), theme) {
         (Ok(content), None) => BlobContent::Ascii(content.to_owned()),
         (Ok(content), Some(theme)) => {
-            use syntect::easy::HighlightLines;
-            use syntect::highlighting::ThemeSet;
-            use syntect::util::LinesWithEndings;
-
             let syntax = std::path::Path::new(path)
                 .extension()
                 .and_then(std::ffi::OsStr::to_str)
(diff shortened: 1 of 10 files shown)


proxy/src/coco/source.rs
```diff
@@ -8,8 +8,17 @@ use radicle_surf::{
...
 };
+use syntect::parsing::SyntaxSet;
 
 use crate::error;
+use crate::session::settings::Theme;
+
+lazy_static::lazy_static! {
+    // The syntax set is slow to load (~30ms), so we make sure to only load it once.
+    // It _will_ affect the latency of the first request that uses syntax highlighting,
+    // but this is acceptable for now.
```
xla, line 19: Is there no way to warm it up?

proxy/src/coco/source.rs
```diff
@@ -217,9 +235,43 @@ pub fn blob(
...
     let (_rest, last) = p.split_last();
-    let content = match std::str::from_utf8(&file.contents) {
-        Ok(content) => BlobContent::Ascii(content.to_string()),
-        Err(_) => BlobContent::Binary,
+    let content = match (std::str::from_utf8(&file.contents), theme) {
+        (Ok(content), None) => BlobContent::Ascii(content.to_owned()),
+        (Ok(content), Some(theme)) => {
+            use syntect::easy::HighlightLines;
+            use syntect::highlighting::ThemeSet;
+            use syntect::util::LinesWithEndings;
```
xla, line 243: Let's avoid inline `use` statements, we don't use them anywhere else in the codebase. Normally clippy should complain. :thinking:

proxy/src/session.rs
```diff
@@ -140,6 +140,17 @@ pub fn set_settings(store: &kv::Store, settings: settings::Settings) -> Result<(
     set(store, KEY_CURRENT, sess)
 }
 
+/// Gets the [`settings::Settings`] in the current session.
+///
+/// # Errors
+///
+/// Errors if access to the session state fails.
+pub fn get_settings(store: &kv::Store) -> Result<settings::Settings, error::Error> {
```
xla, line 148: This will produce an incorrect session, check the existing one, which we should use for now:

https://github.com/radicle-dev/radicle-upstream/blob/0c3e457757988d0a05feab43727421c709f06778/proxy/src/session.rs#L59-L97

proxy/src/coco/source.rs
```diff
@@ -217,9 +235,43 @@ pub fn blob(
         .last_commit(commit_path)?
         .map(|c| CommitHeader::from(&c));
     let (_rest, last) = p.split_last();
-    let content = match std::str::from_utf8(&file.contents) {
-        Ok(content) => BlobContent::Ascii(content.to_string()),
-        Err(_) => BlobContent::Binary,
+    let content = match (std::str::from_utf8(&file.contents), theme) {
```
xla, line 238: Can we break this out into its own function?

xla: This way we could also test it in isolation at some point.

proxy/src/coco/source.rs
```diff
@@ -217,9 +235,43 @@ pub fn blob(
...
+            use syntect::util::LinesWithEndings;
+
+            let syntax = std::path::Path::new(path)
+                .extension()
+                .and_then(std::ffi::OsStr::to_str)
+                .and_then(|ext| SYNTAX_SET.find_syntax_by_extension(ext));
+
+            let ts = ThemeSet::load_defaults();
+            let theme = match theme {
+                Theme::Light => ts.themes.get("base16-ocean.light"),
```
FintanH, line 252: Should this be configurable?

proxy/src/session.rs
cloudhead: This requires the registry to be passed in though?

proxy/src/coco/source.rs
cloudhead: Perhaps eventually, but I think for now it might make sense to create our own radicle theme..

proxy/src/coco/source.rs
cloudhead: Not unless we run it in some other thread or something..

proxy/src/session.rs
xla: For now that's the case yes, if we want a consistent view on the session that is what it takes. We have some work to do to streamline the object graph necessary to full-fill our requests.

cloudhead:
@xla could you take it from here? It's just the session comment that needs addressing.

PR #620, opened 2020-06-30
feat(ui): scope repository selector by peer

rudolfs:
feat

* Move source browsing to a querystring implementation
* Add mock default_peer attribute to Project
* Add peer scope to source browsing
* Remove revisions title
* Move all source browsing state into the Source screen
* Align peer and branch in revision selector
* Fetch tree data just in time when it's needed
* Make File component leaner
* Select mock peer in revision selector
* Add spec for repo browser when switching between projects

Closes: #443.

### DIFF
--- a/cypress/integration/project_source_browsing.spec.js
+++ b/cypress/integration/project_source_browsing.spec.js
@@ -198,7 +198,7 @@ context("source code browsing", () => {
     });
 
     context("when switching between projects", () => {
-      it("opens the selected project on the default branch", () => {
+      it("opens the selected project on the default repository and branch", () => {
         cy.createProjectWithFixture("gold");
         cy.pick("revision-selector").click();
         cy.get('[data-branch="dev"][data-repo-handle="cloudhead"]').click();
--- a/cypress/integration/project_source_browsing.spec.js
+++ b/cypress/integration/project_source_browsing.spec.js
@@ -6,7 +6,7 @@ before(() => {
 
 beforeEach(() => {
   cy.visit("./public/index.html#/profile/projects");
-  cy.contains("Monadic").click();
+  cy.contains("platinum").click();
   cy.contains("Source").click();
 });
 
@@ -140,18 +140,18 @@ context("source code browsing", () => {
             cy.contains(".i-am-well-hidden").click();
           });
           cy.pick("project-screen", "file-source").contains(
-            "Monadic / .i-am-well-hidden"
+            "platinum / .i-am-well-hidden"
           );
-          cy.pick("project-screen", "file-source").contains("Monadic").click();
+          cy.pick("project-screen", "file-source").contains("platinum").click();
           cy.pick("project-screen", "file-source").contains("README.md");
 
           cy.pick("source-tree").within(() => {
             cy.contains(".i-too-am-hidden").click();
           });
           cy.pick("project-screen", "file-source").contains(
-            "Monadic / .i-too-am-hidden"
+            "platinum / .i-too-am-hidden"
           );
-          cy.pick("topbar", "project-avatar").contains("Monadic").click();
+          cy.pick("topbar", "project-avatar").contains("platinum").click();
           cy.pick("project-screen", "file-source").contains("README.md");
 
           // Switching between different revisions shows the correct README
@@ -194,6 +194,17 @@ context("source code browsing", () => {
           '.revision-dropdown [data-tag="v0.5.0"][data-repo-handle="cloudhead"]'
         ).click();
         cy.contains("test-file-deletion.txt").should("not.exist");
+      });
+    });
+
+    context("when switching between projects", () => {
+      it("opens the selected project on the default branch", () => {
+        cy.createProjectWithFixture("gold");
+        cy.pick("revision-selector").click();
+        cy.get('[data-branch="dev"][data-repo-handle="cloudhead"]').click();
+        cy.pick("sidebar", "profile").click();
+        cy.pick("project-list", "project-list-entry-gold").click();
+        cy.pick("revision-selector").contains("rudolfs master");
       });
     });
   });
--- a/cypress/integration/project_source_browsing.spec.js
+++ b/cypress/integration/project_source_browsing.spec.js
@@ -204,7 +204,7 @@ context("source code browsing", () => {
         cy.get('[data-branch="dev"][data-repo-handle="cloudhead"]').click();
         cy.pick("sidebar", "profile").click();
         cy.pick("project-list", "project-list-entry-gold").click();
-        cy.pick("revision-selector").contains("rudolfs master");
+        cy.pick("revision-selector").contains("cloudhead master");
       });
     });
   });
--- a/cypress/integration/project_source_browsing.spec.js
+++ b/cypress/integration/project_source_browsing.spec.js
@@ -204,7 +204,8 @@ context("source code browsing", () => {
         cy.get('[data-branch="dev"][data-repo-handle="cloudhead"]').click();
         cy.pick("sidebar", "profile").click();
         cy.pick("project-list", "project-list-entry-gold").click();
-        cy.pick("revision-selector").contains("cloudhead master");
+        cy.pick("revision-selector").contains("cloudhead@");
+        cy.pick("revision-selector").contains("master");
       });
     });
   });
--- a/proxy/src/coco/source.rs
+++ b/proxy/src/coco/source.rs
@@ -10,7 +10,20 @@ use radicle_surf::{
     vcs::git::{self, git2, BranchName, Browser},
 };
 
+use syntect::easy::HighlightLines;
+use syntect::highlighting::ThemeSet;
+use syntect::parsing::SyntaxSet;
+use syntect::util::LinesWithEndings;
+
 use crate::error;
+use crate::session::settings::Theme;
+
+lazy_static::lazy_static! {
+    // The syntax set is slow to load (~30ms), so we make sure to only load it once.
+    // It _will_ affect the latency of the first request that uses syntax highlighting,
+    // but this is acceptable for now.
+    static ref SYNTAX_SET: SyntaxSet = SyntaxSet::load_defaults_newlines();
+}
 
 /// Branch name representation.
 #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd, Deserialize, Serialize)]
@@ -161,6 +174,12 @@ impl Blob {
     pub fn is_binary(&self) -> bool {
         self.content == BlobContent::Binary
     }
+
+    /// Indicates if the content of the [`Blob`] is HTML.
+    #[must_use]
+    pub fn is_html(&self) -> bool {
+        matches!(self.content, BlobContent::Html(_))
+    }
 }
 
 /// Variants of blob content.
@@ -168,6 +187,8 @@ impl Blob {
 pub enum BlobContent {
     /// Content is ASCII and can be passed as a string.
     Ascii(String),
+    /// Content is syntax-highlighted HTML.
+    Html(String),
     /// Content is binary and needs special treatment.
     Binary,
 }
@@ -202,6 +223,7 @@ pub fn blob(
     default_branch: &str,
     maybe_revision: Option<String>,
     path: &str,
+    theme: Option<&Theme>,
 ) -> Result<Blob, error::Error> {
     browser.revspec(&maybe_revision.unwrap_or_else(|| default_branch.to_string()))?;
 
@@ -219,10 +241,8 @@ pub fn blob(
         .last_commit(commit_path)?
         .map(|c| CommitHeader::from(&c));
     let (_rest, last) = p.split_last();
-    let content = match std::str::from_utf8(&file.contents) {
-        Ok(content) => BlobContent::Ascii(content.to_string()),
-        Err(_) => BlobContent::Binary,
-    };
+
+    let content = blob_content(path, &file.contents, theme);
 
     Ok(Blob {
         content,
@@ -233,6 +253,45 @@ pub fn blob(
         },
         path: path.to_string(),
     })
+}
+
+/// Return a [`BlobContent`] given a file path, content and theme. Attempts to perform syntax
+/// highlighting when the theme is `Some`.
+fn blob_content(path: &str, content: &[u8], theme: Option<&Theme>) -> BlobContent {
+    match (std::str::from_utf8(content), theme) {
+        (Ok(content), None) => BlobContent::Ascii(content.to_owned()),
+        (Ok(content), Some(theme)) => {
+            let syntax = std::path::Path::new(path)
+                .extension()
+                .and_then(std::ffi::OsStr::to_str)
+                .and_then(|ext| SYNTAX_SET.find_syntax_by_extension(ext));
+
+            let ts = ThemeSet::load_defaults();
+            let theme = match theme {
+                Theme::Light => ts.themes.get("base16-ocean.light"),
+                Theme::Dark => ts.themes.get("base16-ocean.dark"),
+            };
+
+            match (syntax, theme) {
+                (Some(syntax), Some(theme)) => {
+                    let mut highlighter = HighlightLines::new(syntax, theme);
+                    let mut html = String::with_capacity(content.len());
+
+                    for line in LinesWithEndings::from(content) {
+                        let regions = highlighter.highlight(line, &SYNTAX_SET);
+                        syntect::html::append_highlighted_html_for_styled_line(
+                            &regions[..],
+                            syntect::html::IncludeBackground::No,
+                            &mut html,
+                        );
+                    }
+                    BlobContent::Html(html)
+                },
+                _ => BlobContent::Ascii(content.to_owned()),
+            }
+        },
+        (Err(_), _) => BlobContent::Binary,
+    }
 }
 
 /// Given a project id to a repo returns the list of branches.
--- a/proxy/src/coco/source.rs
+++ b/proxy/src/coco/source.rs
@@ -366,10 +366,7 @@ pub fn commit_header<'repo>(
 /// # Errors
 ///
 /// Will return [`error::Error`] if the project doesn't exist or the surf interaction fails.
-pub fn commit<'repo>(
-    browser: &mut Browser<'repo>,
-    sha1: &str,
-) -> Result<Commit, error::Error> {
+pub fn commit<'repo>(browser: &mut Browser<'repo>, sha1: &str) -> Result<Commit, error::Error> {
     let oid = git::Oid::from_str(sha1)?;
     browser.commit(oid)?;
 
--- a/proxy/src/coco/source.rs
+++ b/proxy/src/coco/source.rs
@@ -388,6 +388,7 @@ pub fn tags<'repo>(browser: &Browser<'repo>) -> Result<Vec<Tag>, error::Error> {
 /// TODO(fintohaps): default branch fall back from Browser
 pub fn tree<'repo>(
     browser: &mut Browser<'repo>,
+    _peer_id: Option<&peer::PeerId>,
     default_branch: &str, // TODO(finto): This should be handled by the broweser surf#115
     maybe_revision: Option<String>,
     maybe_prefix: Option<String>,
--- a/proxy/src/coco/source.rs
+++ b/proxy/src/coco/source.rs
@@ -4,6 +4,7 @@ use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::str::FromStr;
 
+use librad::peer;
 use radicle_surf::{
     diff, file_system,
     vcs::git::{self, git2, BranchName, Browser},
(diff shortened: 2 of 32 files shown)


ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
```diff
@@ -59,7 +60,7 @@
     <Icon.Commit style="fill: var(--color-secondary)" />
     <a
       class="commit-sha"
-      href={path.projectCommit(projectId, commitSha)}
+      href={path.projectCommit(projectId, userId, commitSha)}
```
xla, line 63: Why does this depend on user information?

ui/src/source.ts
```diff
@@ -118,18 +112,20 @@ enum Kind {
...
   FetchCommits = "FETCH_COMMITS",
   FetchRevisions = "FETCH_REVISIONS",
-  Update = "UPDATE",
+  FetchObject = "FETCH_OBJECT",
 }
 
 interface FetchCommit extends event.Event<Kind> {
   kind: Kind.FetchCommit;
   projectId: string;
+  userId: string;
```
xla, line 121:
```suggestion
  userUrn: string;
```

ui/src/source.ts
```diff
@@ -118,18 +112,20 @@ enum Kind {
...
   sha1: string;
 }
 
 interface FetchCommits extends event.Event<Kind> {
   kind: Kind.FetchCommits;
   projectId: string;
+  userId: string;
```
xla, line 128:
```suggestion
  userUrn: string;
```

ui/src/source.ts
```diff
@@ -138,9 +134,10 @@ interface FetchRevisions extends event.Event<Kind> {
   projectId: string;
 }
 
-interface Update extends event.Event<Kind> {
-  kind: Kind.Update;
+interface FetchObject extends event.Event<Kind> {
+  kind: Kind.FetchObject;
   path: string;
+  user: string;
```
xla, line 140:
```suggestion
  userUrn: string;
```

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
cloudhead: I think it's the `PeerId`?

ui/src/source.ts
cloudhead: I don't think that's correct, it's the `PeerId` if I'm not mistaken. But not sure that's what we want to call it in the UI components..

proxy/src/project.rs
```diff
@@ -50,21 +52,26 @@ pub struct Project {
     pub registration: Option<Registration>,
     /// High-level statistics about the project
     pub stats: coco::Stats,
+    /// The user to which all source browsing queries are scoped to.
+    pub default_user: identity::Identity,
```
cloudhead, line 56: Does this belong here? I would think it's part of the Metadata (since that's where we get it from)?

xla: The metadata of a project is the bucket for user defined input during project creation and update. It's not derived or generated data.

ui/Screen/Project/Source.svelte
```diff
@@ -221,8 +238,8 @@
         {#if object.info.objectType === ObjectType.Blob}
           <FileSource
             blob={object}
-            path={$currentPath}
-            rootPath={path.projectSource(project.id)}
+            path={currentObjectPath}
+            rootPath={path.projectSource(project.id, project.default_user.id)}
```
cloudhead, line 242: So I'm guessing this PR doesn't scope files to the currently selected peer? Just commits?

ui/src/path.ts
```diff
@@ -37,24 +32,41 @@ export const projectIssues = (id: string): string => `/projects/${id}/issues`;
...
-export const projectCommit = (id: string, hash: string): string =>
-  `/projects/${id}/commit/${hash}`;
-export const projectCommits = (id: string, branch: string): string =>
-  `/projects/${id}/commits/${encodeURIComponent(branch)}`;
+
+export const projectCommit = (id: string, user: string, hash: string): string =>
+  `/projects/${id}/${user}/commit/${hash}`;
+export const projectCommits = (
+  id: string,
+  user: string,
```
cloudhead, line 67: Let's figure out what this "user" is: If it's a URN, eg. `rad:git:12389asd8`, then `user` is appropriate. If it's a PeerID on the other hand, we should call this `peer`.

ui/Screen/Org/Projects.svelte
```diff
@@ -22,7 +22,12 @@
   const select = event => {
     const orgProject = event.detail;
     if (orgProject.maybeProject) {
-      push(path.projectSource(orgProject.maybeProject.id));
+      push(
+        path.projectSource(
+          orgProject.maybeProject.id,
+          orgProject.maybeProject.default_user.id
```
cloudhead, line 28: So here it looks like the project source indeed depends on a user. But I was under the impression that source code is tied to the device/peer, not the user. How do we go from `user -> device -> repo`?

proxy/src/project.rs
cloudhead: Ah I see, so it's not the project identity.

ui/Screen/Org/Projects.svelte
xla: Same concern as in https://github.com/radicle-dev/radicle-upstream/pull/620#pullrequestreview-440583382

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
rudolfs: Yeah, it'll be the `PeerId` for now, but I thought that we'd be able to resolve them to user's soon, is that not correct?

rudolfs:
> Looks like we're almost there, but like @xla I'm a little confused about the user being so closely tied into this, given that if I'm not mistaken, source code is tied to peers.

My thinking on this was that since we'd want to resolve `PeerId`s to users in the UI, I'd only handle that in the UI and expect the resolution to happen in the backend. But maybe that's premature since the whole peer <-> user resolution is still being worked out.

Happy to switch out everything for a `PeerId`. @cloudhead should we have a short chat about this first?

ui/Screen/Project/Source.svelte
rudolfs: Indeed, changing the remaining endpoints to accept a `userId` (or `peerId`) is still TBD. Wanted to get the UI routing merged first before @xla rewrites the whole router...

xla:
> My thinking on this was that since we'd want to resolve `PeerId`s to users in the UI, I'd only handle that in the UI and expect the resolution to happen in the backend. But maybe that's premature since the whole peer <-> user resolution is still being worked out.

While the resolution indeed should happen transparently, there still is the reality that there is a 1:N to relationship between `Project` and `Repository`/source tree, where the later is tight to a `PeerID`. In turn that means that we can end up with many repositories from the same user, but from different peers. @FintanH Correct me if I'm wrong here.

> Happy to switch out everything for a `PeerId`. @cloudhead should we have a short chat about this first?

For now let's simplify and just use peerids in the first iteration and make sure to inform the data model in the proxy from the hiereachy I outline above.

ui/src/path.ts
FintanH: I think `userUrn` would be more helpful. `user: String` doesn't portray much to me, it could just be the handle for all I know (*cough* types *cough*). So I think `userUrn` or `peerId` would more indicative.

FintanH:
> While the resolution indeed should happen transparently, there still is the reality that there is a 1:N to relationship between Project and Repository/source tree, where the later is tight to a PeerID. In turn that means that we can end up with many repositories from the same user, but from different peers. @FintanH Correct me if I'm wrong here.

Yup, this is in similar vein to my comment here https://github.com/radicle-dev/radicle-upstream/pull/620#discussion_r448214355. We have 1:N for User devices (technically you could have multiple User's per device but let's not get into _that_ :)). This means that you may want to look at `fintohaps@yfijf8575` _or_ `fintohap@iuhd8485` because they use two machines and they both show up in the selector.

ui/Screen/Project/Source.svelte
rudolfs: The remaining endpoints are taken care of in:
https://github.com/radicle-dev/radicle-upstream/pull/620/commits/e2c4a7f0d339a145ca4c94b8b18cd57116b897f8
https://github.com/radicle-dev/radicle-upstream/pull/620/commits/d4f044e705b391b371409455a496b9941fedb075
https://github.com/radicle-dev/radicle-upstream/pull/620/commits/f25eb586bf32e7ddc3abd70fce13cdd2466af24c
https://github.com/radicle-dev/radicle-upstream/pull/620/commits/f25eb586bf32e7ddc3abd70fce13cdd2466af24c

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
rudolfs: @cloudhead addressed in https://github.com/radicle-dev/radicle-upstream/pull/620/commits/d5b10c171dddbf5b93c4a3b1493d25c1689799f0

@FintanH no, this just constructs a link to the single commit screen, when you click the link the screen will fetch the respective commit and use these params as scope.

Screenshot 2020-07-01 at 16 21 40

ui/src/source.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/d5b10c171dddbf5b93c4a3b1493d25c1689799f0

ui/src/source.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/d5b10c171dddbf5b93c4a3b1493d25c1689799f0

ui/src/source.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/d5b10c171dddbf5b93c4a3b1493d25c1689799f0

proxy/src/project.rs
rudolfs: Had a chat with @xla IRL, we decided that the ordering of the `GET /source/revisions` response (which also contains peers) will determine the default_peer. It's going to be the first in the list.

ui/src/path.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/d5b10c171dddbf5b93c4a3b1493d25c1689799f0

ui/Screen/Org/Projects.svelte
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/d5b10c171dddbf5b93c4a3b1493d25c1689799f0

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
FintanH: > I think it is fine to be able to jump to any commit, as long as some metadata is computed which shows in which projects + branches it appears. The only ambiguous case is when you jump to a commit in the current project, but that commit doesn't appear in that project anywhere.

As Kim said, it might not be that we scope it. I would suggest we remove this change until we're sure what the behaviour will be.

rudolfs: Alright.

rudolfs: @FintanH https://github.com/radicle-dev/radicle-upstream/pull/620/commits/cc7bc8a3819ab9ab419487c80f0758ca56898462

proxy/src/http/source.rs
```diff
@@ -274,16 +276,28 @@ mod handler {
 
     /// Fetch a [`coco::Blob`].
     pub async fn blob(
-        peer: Arc<Mutex<coco::PeerApi>>,
+        api: Arc<Mutex<coco::PeerApi>>,
```
FintanH, line 279: Was this @xla changing the naming? xD

proxy/src/coco/source.rs
```diff
@@ -197,6 +198,7 @@ pub struct TreeEntry {
 /// Will return [`error::Error`] if the project doesn't exist or a surf interaction fails.
 pub fn blob(
     browser: &mut Browser,
+    _peer_id: Option<&peer::PeerId>,
```
FintanH, line 201: I'm unsure if we need the `PeerId` at this scope, but that can be determined in another PR.

proxy/src/http/source.rs
```diff
@@ -795,6 +835,7 @@ mod test {
         let want = coco::with_browser(&*peer.lock().await, &urn, |mut browser| {
             coco::blob(
                 &mut browser,
+                None,
```
FintanH, line 838: Could use `peer.peed_id` or the peer id derived from the key. Maybe for a later change.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -108,10 +114,11 @@
   hidden={expanded}>
   <div class="selector-avatar">
     <Avatar
-      title={revisions[0].identity.metadata.handle}
-      avatarFallback={revisions[0].identity.avatarFallback}
+      title={currentSelectedUser.identity.metadata.handle}
```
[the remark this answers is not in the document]
FintanH: This says it's outdated but I don't see the change I was expecting :eyes:

proxy/src/http/source.rs
rudolfs: Indeed, he was : D

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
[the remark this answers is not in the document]
rudolfs: It says outdated, because the underlying code changed in the user -> peer rename commit. Haven't yet had time to address this comment.

rudolfs:
> The concept of a "default user" for a project seems odd. What is a non-default user? What is a user in general in the context of a project? Isn't that we have multiple source trees for a project, namespaced by peerids which can resolve to user urns?

@xla https://github.com/radicle-dev/radicle-upstream/pull/620/commits/06e9bdacf9c31c800b28cf6779cb1559d4323175

proxy/src/project.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/06e9bdacf9c31c800b28cf6779cb1559d4323175

proxy/src/project.rs
```diff
@@ -81,10 +88,29 @@ pub enum Registration {
...
+    }
+}
+
 /// Fetch the project with a given urn from a peer
 pub fn get(peer: &coco::PeerApi, project_urn: &uri::RadUrn) -> Result<Project, error::Error> {
     let project = coco::get_project(peer, project_urn)?;
     let stats = coco::with_browser(peer, project_urn, |browser| Ok(browser.get_stats()?))?;
 
-    Ok((project, stats).into())
+    Ok((project, stats, get_fake_default_user()).into())
```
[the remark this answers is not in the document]
rudolfs: The default_user got removed. I think we should tackle the proxy end of this in https://github.com/radicle-dev/radicle-upstream/pull/510.

cypress/integration/project_source_browsing.spec.js
```diff
@@ -196,6 +196,17 @@ context("source code browsing", () => {
         cy.contains("test-file-deletion.txt").should("not.exist");
       });
     });
+
+    context("when switching between projects", () => {
+      it("opens the selected project on the default branch", () => {
```
xla, line 201:
```suggestion
      it("opens the selected project on the default repository and branch", () => {
```

cypress/integration/project_source_browsing.spec.js
```diff
@@ -6,7 +6,7 @@ before(() => {
 
 beforeEach(() => {
   cy.visit("./public/index.html#/profile/projects");
-  cy.contains("Monadic").click();
+  cy.contains("platinum").click();
```
xla, line 9: Did we change the fixtures?

proxy/src/coco/source.rs
xla: @FintanH We want to scope by namespace (peer_id, "repository") whenever doing source queiries. Under what condition would we not need to be explicit about that scope when fetching a blob?

proxy/src/http/source.rs
xla: Too many peers in the house. May th real peer please stand up!

proxy/src/http/source.rs
xla: I think the heuristic is a bit more complicated and depends on the state of the local monorepo: https://www.notion.so/monadic/Overview-4969e31627984cc1938a6603d7d33738#3f74974beb934d81a911fc1eed1f8a88

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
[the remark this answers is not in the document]
rudolfs: So I tried with the shareableIdentityIdentifier, but I feel the current design doesn't have enough space to give the user much meaningful information in the collapsed version:

Screenshot 2020-07-02 at 10 21 08

Slightly better in the expanded version:
Screenshot 2020-07-02 at 10 21 12

Isn't it enough that the avatar is different, hence it's derived from the PeerId?

proxy/src/http/source.rs
rudolfs: Here `peer` is `librad::net::peer::PeerApi`.

cypress/integration/project_source_browsing.spec.js
rudolfs: I did. We created a project called "Monadic" before, which would make more sense for an org, so I changed it to something that resembles a project.

xla: Thanks for the clarification.

proxy/src/coco/source.rs
FintanH: Discussed on IRC. I didn't provide enough context here. Basically what I was trying to say that I'm not sure whether `blob` et al. should be aware of `PeerId` or if we make a `Browser` remote aware (where the `PeerId` is the remote). I need to evaluate whether this is useful for surf (my feeling is that it is) and then see what changes we can make. This is can be left as a reminder to do that work ;)

proxy/src/http/source.rs
FintanH: Ya, but we're in tests so passing in `peer.peer_id`, knowing that this `peer` created the projects, seems like a reasonable test to me. But it was just a suggestion :)

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -108,10 +131,11 @@
   hidden={expanded}>
   <div class="selector-avatar">
     <Avatar
-      title={revisions[0].identity.metadata.handle}
-      avatarFallback={revisions[0].identity.avatarFallback}
+      title={currentSelectedPeer.identity.metadata.handle}
```
FintanH, line 132: It's still `handle` :grimacing:

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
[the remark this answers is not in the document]
xla: We should just shortne the peer_id. Is there a way to have a shortform similiar to how Oids have a 7-digiti version @FintanH?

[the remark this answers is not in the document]
FintanH: What if we have a shortened version of the shareable identifier that does this?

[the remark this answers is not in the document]
rudolfs: Ok, I'll add the full id for now, but I feel that this is a bit out of scope for this issue.
Would like to address the shortening in a follow-up once we have more of the real backend. Perhaps we'll also need to adjust the design of this dropdown anyway.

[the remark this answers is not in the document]
FintanH: Ok, can you make a follow-up ticket for this?

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/b7fe61d0bc3f0547a378b2c4d6b15657cb0ec679

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/b7fe61d0bc3f0547a378b2c4d6b15657cb0ec679

proxy/src/http/source.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/105b9d048bd8d1b6bb8573f496246a6f00b048fb

rudolfs:
@FintanH right, addressed here: https://github.com/radicle-dev/radicle-upstream/pull/620/commits/e819df8b04c9b1c33c8af7272c511b805ac427f1.

PR #622, opened 2020-07-01
refactor(ui): View based routing

xla:
Here is the first part of the new routing/navigation setup. At the core is the `View` which has a component to render and can be parameterised with props when navigated to. When mutliple of these are bundled into a `Map` the `ViewRouter` can based on a standardised `Navigation` interface display the currently active one. Conceptually this work for navigation of and between the main Screens and the same infrastructure can be used to build "sub-navigations" inside of Screens to route between their fragments. To achieve this we will allow a `View` to have an optional `Navigation` so that it can decided which fragment to show when navigated to. Addtionallly a couple of other concerns will be addressed before this change is ready for review, they arose from shortcomings we ran into in the past and are captured in todo items below.

Being the first building block for a new global display and routing strategy in follow-up changes we are going to introduce new overlays which can live on-top of screens and not replace them and a way to globally handle dropodowns so we avoid double dropdown.

Furthermore I'm torn if we should the integration/swap in this PR or should just merge the actual core implementations. Also curious to know if there are ideas how to effectively test `view.ts` and `View/Router.svelte`. I for now left the changes in `App.svelte` to showcase how the differetn pieces fall into place and are combined to set up a view based routing scheme.

**TODO**
* [ ] support prefetching/preloading per `View` - probably in the first iteration based on the `remote.Store`, but could easily be a promised based solution where the resolution of said Promise is left to the component (encapsulation ya know)
* [ ] support nested `Navigation` per  `View` - so that a `Screen` can express the existence of fragemnts and decide early (on render) which fragment to show
* [ ] implement navigation helpers - so that we can easily render nav items and highlight the currently active
* [ ] update `DEVELOPMENT.md` with new architecture and explain how all the pieces fit together.

***
First part of #754


ui/App.svelte
```diff
@@ -9,45 +10,49 @@
...
+  /*   "/projects/:id/*": Project, */
+  /*   "/design-system-guide": DesignSystemGuide, */
+  /*   "/help": Help, */
+  /*   "/user-registration": UserRegistration, */
+  /*   "/transactions/:id": TransactionDetails, */
+  /*   "*": NotFound, */
+  /* }; */
+
+  const views = {};
+  views[navigation.Screen.Blank] = { component: Blank };
```
sarahscott, line 55: this syntax looks a little funny - maybe ergonomics would be improved if this functionality happened inside an exported method from `history.ts` or `router.ts`

MeBrei:
Thank you for putting this together!

The current changes are still quite abstract. An example implementation (eg sidebar + org page + modal) would be helpful to understand how it will work, what the benefits of this implementation are and how much work it will be to integrate it. The app currently only shows a blank screen for me, testing the new navigation would be nice.
In the current stage I can't immediately see, how this approach will help eg with the double dropdown.

> Furthermore I'm torn if we should the integration/swap in this PR or should just merge the actual core implementations.

I would be in favor of doing most of the integration in this PR. On the one hand to see how it works, but also to avoid having two different systems at the same time and dragging out the completion switch.

Furthermore I would be interested in hearing a bit why you chose this approach, ie if there was no feasible way to build on the current routing.

rudolfs:
Great initiative!

I second the need to see the proposed implementation in action to fully evaluate if it's the right way to go.

As the new routing functionality is a completely separate thing, it shouldn't be impacted by changes happening in the master branch (i.e. you won't have many merge conflicts until you try to integrate it), so it would be nice to keep it in a PR until it's more fleshed out.

xla:
@MeBrei 

> The current changes are still quite abstract. An example implementation (eg sidebar + org page + modal) would be helpful to understand how it will work, what the benefits of this implementation are and how much work it will be to integrate it. The app currently only shows a blank screen for me, testing the new navigation would be nice.
> In the current stage I can't immediately see, how this approach will help eg with the double dropdown.

Starting the integration will mean we have to migrate the entire routing in the context of this PR. While doable this will delay the merge of a subset of the changes and extend the scope quite a bit. To address double dropdown the View navigation does not immediately help but the overall refactor tracked in #754. Including a solution for that in this PR will further increase the scope and increase the chance it's dragged out for longer.

> Furthermore I would be interested in hearing a bit why you chose this approach, ie if there was no feasible way to build on the current routing.

Will update the tracking issue with the problem statement and challenges we faced with the current setup.

@rudolfs 

> As the new routing functionality is a completely separate thing, it shouldn't be impacted by changes happening in the master branch (i.e. you won't have many merge conflicts until you try to integrate it), so it would be nice to keep it in a PR until it's more fleshed out.

As routing is deeply embedded in all parts of the app any integration work will ultimately be subject to changes in master, so I don't see how this can be done in isolation.

@MeBrei @rudolfs I updated the `App.svelte` and `Profile.svelte` to showcase it in action with working nav to some screens/fragments. Lemme know what else you wanna see touched.

xla:
We don't want to invest in this for now as we don't have urgent need to see the new overlays land. This feature is well scoped and designed in #754 - which we will turn to, once we wanna update the navigation and layout concerns. For now the gains woul dbe too marginal to dump more engineering time into it.

PR #626, opened 2020-07-02
feat: implement nuke command

### LINKED ISSUE #394: Wire up "clear coco data" button in Settings
Clicking the button should be the same as:
```rm -rf ~/Library/Preferences/xyz.radicle.radicle && rm -rf ~/Library/Application\ Support/xyz.radicle.radicle```

There's probably a more elegant way via the library we use for preference storage.

Screenshot 2020-05-15 at 09 13 26

### PR DESCRIPTION
**Pull request by xla**
Developers often are unable to start up the app because the state on disk is not compatible with code versions or it has been corrupted in other ways. To remedy that and provide a drastic tool to get uninitated unstuck without the knowledge of the on disk fs layout, this change-set introduces `yarn nuke`.

Closes #394

### DIFF
--- a/proxy/src/bin/nuke.rs
+++ b/proxy/src/bin/nuke.rs
@@ -1,0 +1,41 @@
+use std::fs::remove_dir_all;
+use std::io::ErrorKind;
+use std::process::exit;
+
+use log::{info, trace};
+
+use proxy::coco::control;
+use proxy::config;
+use proxy::env;
+
+fn main() {
+    env::set_if_unset("RUST_BACKTRACE", "full");
+    env::set_if_unset("RUST_LOG", "info");
+    pretty_env_logger::init();
+
+    info!("Nuking application state...");
+    match remove_dir_all(config::dirs().data_dir()) {
+        Ok(_) => info!("done"),
+        Err(err) => {
+            if err.kind() == ErrorKind::NotFound {
+                info!("already gone");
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: proxy/src/bin/nuke.rs [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+                info!("already gone");
+            } else {
+                trace!("{:?}", err);
+                exit(1);
+            }
+        },
+    };
+
+    info!("Nuking coco state...");
+    match control::nuke_monorepo() {
```
FintanH, line 30: I _think_ the above `remove_dir_all` would mean the monorepo is already gone no?

xla: No the above one nukes the directory managed by Upstream for its state.

FintanH: Ah, I suppose it's the case they're technically the same. But could be different. Great point!

xla: Nah they live under their own directories, so they are not the same. Upstream constructs its `ProjectDirs` from `xyz.radicle.radicle-upstream` and coco its from `xyz.radicle.radicle`. On linux that means the following two:

```
~/.local/share/radicle
~/.local/share/radicle-upstream
```

FintanH: Even better point!

PR #628, opened 2020-07-02
feat(proxy): integrate rad/self

### LINKED ISSUE #548: feat(proxy): User Start Up
When starting up the application there are two cases:

1. This is the first time and/or there's no previous state
2. The user has started the application before and the state is preserved

In the first case, the user needs to create an `Identity`, and thus create a `librad::User`. This means that the API should only be presenting an initial set of endpoints.

In the second case, the existing state should be loaded giving us access to this `librad::User`.

### LINKED ISSUE #576: Project list is kept across identities
If you create a project with identity **A**, then clear your session, and create identity **B**, you will see in your profile **A**'s project.

FintanH:
Depends on https://github.com/radicle-dev/radicle-link/pull/229

Fixes #576 
Closes #548

This integrates the use of `rad/self` in the monorepo. This means when we create an identity, we set the default `rad/self` in the peer to that `User`.

We can now filter the projects by the `rad/self` of the `Project` is the same as the default `rad/self`. This is demonstrated in [test_list_projects](https://github.com/radicle-dev/radicle-upstream/compare/fintan/rad-self?expand=1#diff-8b447e0828f494b1a01fea3e55d2925eR497)

### DIFF
--- a/proxy/src/coco/peer.rs
+++ b/proxy/src/coco/peer.rs
@@ -8,6 +8,7 @@ use librad::meta::project;
 use librad::meta::user;
 use librad::net::discovery;
 pub use librad::net::peer::{PeerApi, PeerConfig};
+use librad::git::storage;
 use librad::uri::RadUrn;
 use radicle_surf::vcs::git::{self, git2};
 
@@ -202,6 +203,7 @@ pub fn init_project(
             return Err(error::Error::EntityExists(urn));
         } else {
             let _repo = storage.create_repo(&meta)?;
+            storage.set_rad_self(&owner.urn(), storage::RadSelfSpec::Default)?;
         }
         Ok(meta)
     };
--- a/proxy/src/coco/peer.rs
+++ b/proxy/src/coco/peer.rs
@@ -1,6 +1,8 @@
 use std::net::SocketAddr;
+use std::sync::Arc;
 
 use async_trait::async_trait;
+use tokio::sync::Mutex;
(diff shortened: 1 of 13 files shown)


proxy/src/coco/peer.rs
```diff
@@ -444,15 +496,32 @@ mod test {
...
+        control::setup_fixtures(peer, key.clone(), &user)?;
+
+        let kalt = super::init_user(peer, key.clone(), "kalt")?;
+        let kalt = super::verify_user(kalt).await?;
+        let fakie = super::init_project(
+            peer,
+            key,
+            &kalt,
+            &repo_path,
+            "fakie-nose-kickflip-backside-180-to-handplant",
```
rudolfs, line 519: lol

proxy/src/http/identity.rs
```diff
@@ -296,9 +301,15 @@ mod test {
 
         let store = &*store.read().await;
         let registry = &*registry.read().await;
-        let session = session::current(peer, registry, store).await?;
+        let session = session::current(Arc::clone(&peer), registry, store).await?;
         let urn = session.identity.expect("failed to set identity").id;
 
+        // Assert that we set the default owner ans it's the same one as the session
```
rudolfs, line 307:
```suggestion
        // Assert that we set the default owner and it's the same one as the session
```

`:CocInstall coc-spell-checker` : P

PR #632, opened 2020-07-03
fix(ui): fix repository selector regressions

rudolfs:
Fixes regressions introduced in https://github.com/radicle-dev/radicle-upstream/pull/620:
- source tree file highlighting
- topbar `Source` menu item active state
- source tree expanded state persistance

Adds tests for all the regressions.

### DIFF
--- a/ui/src/path.ts
+++ b/ui/src/path.ts
@@ -37,7 +37,7 @@ export const projectSource = (
   projectId: string,
   peerId?: string,
   revision?: string,
-  objectType: string = ObjectType.Tree,
+  objectType?: string,
   objectPath?: string
 ): string => {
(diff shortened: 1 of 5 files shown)


rudolfs:
Added @kim as a 2nd reviewer because everyone else is vacationing...

ui/src/path.ts
```diff
@@ -50,14 +50,17 @@ export const projectSource = (
 
 export const parseProjectSourceLocation = (
   querystring: string,
-  defaultRevision: string
+  defaultPeerId: string,
```
sarahscott, line 53: I think for now this makes sense but in the future it might be better to fetch these source-related `defaults` in `source.ts` instead of passing them around through `path`

PR #629, opened 2020-07-03
feat(ui): discovery ui

sarahscott:
- [x] seed config in settings
Screen Shot 2020-07-08 at 2 53 46 PM
Screen Shot 2020-07-08 at 2 53 58 PM

- [x] project discovery ui components
discovery

- [x] polish
- [x] tests (unit & integration)

Closes #591 ; closes ui side of #594

Issue #613, opened 2020-06-29
discovery: Proxy API


cloudhead:
An API the frontend can query to get a set of unique projects with metadata, to be displayed in #591 

This API should be high-level, and expose a deduplicated, ordered set with all the necessary metadata embedded within.

sarahscott:
Note: once this API is integrated with the existing UI, what should happen when a user clicks on an org they are not a part of? Will there be a public-facing org page (similar to viewing another user's profile) or are some orgs completely private for non-members?

cc @cloudhead @juliendonck

cloudhead:
So I think the easiest is to just not link them if they aren't "known", ie. in your sidebar already. We can revisit the question when we tackle org public profiles.

PR #638, opened 2020-07-06
feat(proxy): seeds api

### LINKED ISSUE #594: Add section in Settings page, relating to coco network configuration
Textarea that simply lists seed hosts

sarahscott:
- [x] add `CoCo` settings to `session`
- [x] connect coco settings to ui (after #629 )

Related to https://github.com/radicle-dev/radicle-upstream/issues/613
Closes #594


FintanH:
Lemme know if you need some sort of input here yet. Otherwise, I'll probably just wait til you flip the switch from draft :+1:

proxy/src/session.rs
```diff
@@ -222,4 +224,11 @@ pub mod settings {
             Self::Emulator
         }
     }
+
+    /// p2p config parameters subject to user preferences
+    #[derive(Debug, Default, Deserialize, Serialize, PartialEq)]
+    pub struct Ecosystem {
```
xla, line 230: We refer to the p2p domain as coco throughout the codebase, `Ecosystem` is quite general and could refer to many aspects.

ui/App.svelte
```diff
@@ -73,9 +74,27 @@
...
+  }
+</style>
+
 <Hotkeys />
 <NotificationFaucet style="margin-top: calc(var(--topbar-height) + 11px)" />
 <Theme />
 <Remote {store} let:data={session} context="session">
   <Router {routes} />
+
+  <div slot="error" class="error">
```
sarahscott, line 94: This error was bugging me every time I made a change to the session so now it's easier to `clear` it 😄 

Screen Shot 2020-07-16 at 5 36 40 PM

proxy/src/session.rs
```diff
@@ -228,4 +230,23 @@ pub mod settings {
             Self::Emulator
         }
     }
+
+    /// `CoCo` config parameters subject to user preferences
+    #[derive(Debug, Deserialize, Serialize, PartialEq)]
+    pub struct CoCo {
```
FintanH, line 236: Not sure `CoCo` naming makes sense for me. What do we think of calling it `Seeds` instead? Or `Garden`? :smirk:

sarahscott: No @FintanH that would be too cute and fun.

sarahscott: No @FintanH, that would be too cute and fun.

FintanH: We need more puns tbqhwy 😅 but ya CoCo doesn't cut it for me. Still to general for me

sarahscott: I think if the only coco-related configuration we're doing right now is an array of `seeds`, we could call it `Seeds`. Going forward though, aren't there going to be other coco/p2p settings?

FintanH: Let's call it `Seeds` then :) 

> Going forward though, aren't there going to be other coco/p2p settings?

Could be, but let's not pre-optimise. Renaming things is easier with a compiler to catch things for you and we're not some backwards-compat lib, so let's take our liberties and convey the current intent of this structure.

xla: Strongly disagree with this change. If we built up namespace to group and bucket configuration paremeters `seeds.seeds` is not communicating at all what is referred to. Instead we should be true to the fact, that these are the seeds used for the p2p code-collab protocol. There might be more settings related to that in the future and if we don't have an adequate alternative by way of names, I'd like to see this changed back to `CoCo` as this is the nomenclatue in the rest fo the codebase as well.

We can debate to adopt a different naming in a separate conversation and outside of this change-set.

xla:
This seems to address #594

proxy/src/session.rs
FintanH: Well I'd say make it `pub struct Seed(Vec<String>)` in this case.

I'm not willing to die on a hill over naming :) Sorry for misleading you in this case @sarahscott

sarahscott: kein problem 😄  it happens, happy to get this one merged. we can change the name to `p2p` or `peer` down the line if we get tired of `coco`

proxy/src/session.rs
```diff
@@ -224,4 +226,23 @@ pub mod settings {
...
+        /// Sources of feeds
+        pub seeds: Vec<String>,
+    }
+
+    // TODO(sos): should these come  from librad?
```
FintanH, line 237: These will come from some DNS lib https://github.com/radicle-dev/radicle-link/issues/249

proxy/src/session.rs
```diff
@@ -224,4 +226,23 @@ pub mod settings {
...
+    impl Default for CoCo {
+        fn default() -> Self {
+            Self {
+                seeds: ["seed.radicle.xyz", "194.134.54.13"]
+                    .iter()
```
FintanH, line 242: You can probably do `into_iter` and not have to do `(*s)` below.

ui/App.svelte
rudolfs: Tagging: @juliendonck, @brandonhaslegs, so they're aware of it.

rudolfs: Wondering if we should keep this, as we already have `yarn nuke` and clearing the session doesn't clear the registry state (can't create the same identity twice).

proxy/src/http/session.rs
```diff
@@ -294,6 +294,12 @@ mod test {
                     "appearance": {
                         "theme": "light",
                     },
+                    "coco": {
+                        "seeds": [
+                            "seed.radicle.xyz",
+                            "194.134.54.13"
```
cloudhead, line 300: Just a note: IP addresses won't work as seeds because they can't carry the PeerId which is needed for the TLS handshake.

xla: Could you expand how the PeerID will be encoded in DNS entries? Also if this is the case that we can't accept IP addresses, we should find a way to encode that so that we never are exposed to the risk of constructing a peer "address" incorrectly.

cloudhead: PeerIds are planned to be stored in a CNAME entry.

xla: To expand here a bit more, the idea is that for a given DNS entry the client is suppose to do two lookups:

* `CNAME` which gives the `PeerId` as the value
* `A` which gives you the `IP` address

Which will result in `<peer_id>@<ip>` as the true location identification of the peer. Common practice as far as identifiers go that depend on some form of node id. Is there spec/docs somewhere?

sarahscott: @xla there's a [ticket here](https://github.com/radicle-dev/radicle-upstream/issues/693) for DNS discussion

cloudhead: Since these seeds aren't actual seeds I don't see any issue besides potentially confusing the reader.

sarahscott: I can remove the IP address one; it's only there because it was in the design spec.

PR #639, opened 2020-07-07
chore(proxy): fetch owner on request

xla:
This removes the brittle handling of holding onto an owner, instead we read the necessary on per request basis.

### DIFF
--- a/proxy/src/http.rs
+++ b/proxy/src/http.rs
@@ -51,7 +51,6 @@ macro_rules! combine {
 /// Main entry point for HTTP API.
 pub fn api<R>(
     peer: coco::PeerApi,
-    owner: Option<coco::User>, // None if the default owner was not found
     keystore: keystore::Keystorage,
     registry: R,
     store: kv::Store,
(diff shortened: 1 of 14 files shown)


proxy/src/registry/transaction.rs
```diff
@@ -1,15 +1,18 @@
-#![allow(clippy::integer_arithmetic)]
 //! Abstractions and types to handle, persist and interact with transactions.
 
+#![allow(clippy::integer_arithmetic)]
```
FintanH, line 3: ohno

proxy/src/http.rs
```diff
@@ -140,17 +141,36 @@ where
 /// Asserts presence of the owner and reject the request early if missing. Otherwise unpacks and
 /// passes down.
 #[must_use]
-pub fn with_owner_guard(maybe_owner: Shared<Option<coco::User>>) -> BoxedFilter<(coco::User,)> {
+pub fn with_owner_guard<R>(
```
FintanH, line 144: I seeeee. This makes sense from an application point of view, ya :+1:

PR #653, opened 2020-07-09
feat(proxy): transfer endpoint

### LINKED ISSUE #659: Transfer funds endpoint
Create a new `transfer` endpoint that can be used to send funds from the own user or an org to other users or orgs.

MeBrei:
Closes #659

**Note** Account ids are currently not exposed in the API yet (#658), so the implemented tests have to suffice for this PR. Once the account ids are exposed, the transaction can be made via the API playground.


NunoAlexandre:
I think that we should follow the same principle that drove #474 and have this single endpoint under their respective entities.
So instead of having one transfer endpoint that receives the necessary data in the payload so that we know what to do, to have two endpoints that clearly show that structure:

- `orgs/<id>/transfer`
- `users/<handle>/transfer`

Then we can share the payload, which should contain the recipient's account and the amount.

proxy/src/http/transfer.rs
```diff
@@ -0,0 +1,240 @@
...
+        Ok(reply::with_status(reply::json(&tx), StatusCode::CREATED))
+    }
+}
+
+/// Bundled input data for transfer.
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Input {
+    /// Id if sent from an Org.
+    maybe_org: Option<String>,
```
FintanH, line 108: I think because `Id` has a `Deserialize` instance we could use its type here instead of `String`. The `try_from` will be called when trying to deserialize the input then and you don't have to do it above.

proxy/src/registry.rs
```diff
@@ -255,6 +255,33 @@ pub trait Client: Clone + Send + Sync {
...
+    /// Will return `Err` if a protocol error occurs.
+    async fn transfer_from_user(
+        &self,
+        author: &protocol::ed25519::Pair,
+        recipient: protocol::ed25519::Public,
+        value: Balance,
+        fee: Balance,
+    ) -> Result<Transaction, error::Error>;
+
+    /// Transfer tokens from an orgto the recipient.
```
FintanH, line 271:
```suggestion
    /// Transfer tokens from an org to the recipient.
```

proxy/src/registry.rs
```diff
@@ -595,6 +622,69 @@ impl Client for Registry {
...
+    async fn transfer_from_org(
+        &self,
+        author: &protocol::ed25519::Pair,
+        org_id: Id,
+        recipient: protocol::ed25519::Public,
+        value: Balance,
+        fee: Balance,
+    ) -> Result<Transaction, error::Error> {
+        // Prepare and submit transfer transaction.
+        let transfer_message = protocol::message::TransferFromOrg {
```
FintanH, line 664: It seems like there's duplication between these two functions where the only difference is only the messages. Could we pass those in as parameters and have a generic `transfer` function?

Could probably have a `From<TransferFromOrg> for Message` and `From<Transfer> for Message` impls to do the conversion in the call to `confirmed`.

FintanH: To give an update we tried having:
```rust
pub enum Transfer {
  User(protocol::message::Transfer),
  Org(protocol::message::TransferFromOrg),
}
```

The issue was that we couldn't implement `protocol::Message` for this, nor could we return a `Box<dyn protocol::Message>` because one of the methods didn't take `self` -- `result_from_events`. We did our best :(

proxy/src/http/transfer.rs
```diff
@@ -0,0 +1,235 @@
+//! Transfer funds between orgs and users
+
+// use serde::ser::SerializeStruct as _;
```
FintanH, line 3: Some errant commenting out here :)

MeBrei:
> We should follow the same principle that drove #474 and have this single endpoint under their respective entities.
> So instead of having one transfer endpoint that receives the necessary data in the payload so that we know what to do, to have two endpoints that clearly show that structure:
> 
> * orgs//transfer
> * users//transfer
> 
> Then we can share the payload, which should contain the recipient's account and the amount.

@NunoAlexandre I restructured the endpoints, but added separate `TransferInput`s that match the naming of the transaction messages.

NunoAlexandre:
> > We should follow the same principle that drove #474 and have this single endpoint under their respective entities.
> > So instead of having one transfer endpoint that receives the necessary data in the payload so that we know what to do, to have two endpoints that clearly show that structure:
> > 
> > * orgs//transfer
> > * users//transfer
> > 
> > Then we can share the payload, which should contain the recipient's account and the amount.
> 
> @NunoAlexandre I restructured the endpoints, but added separate `TransferInput`s that match the naming of the transaction messages.

Ah yes, makes sense!

proxy/src/http/org.rs
```diff
@@ -244,6 +252,36 @@ fn register_member_filter<R: registry::Client>(
...
+    registry: http::Shared<R>,
+    subscriptions: notification::Subscriptions,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone
+where
+    R: registry::Client,
+{
+    http::with_shared(registry)
+        .and(http::with_subscriptions(subscriptions))
+        .and(warp::post())
+        .and(document::param::<String>("id", "Unique ID of the Org"))
```
NunoAlexandre, line 266:
```suggestion
        .and(document::param::<registry::Id>("id", "Unique ID of the Org"))
```

proxy/src/http/org.rs
```diff
@@ -387,6 +425,35 @@ mod handler {
 
         Ok(reply::with_status(reply::json(&tx), StatusCode::CREATED))
     }
+
+    /// Transfer funds to the given `recipient`.
+    pub async fn transfer<R: registry::Client>(
+        registry: http::Shared<R>,
+        subscriptions: notification::Subscriptions,
+        id: String,
```
NunoAlexandre, line 433:
```suggestion
        id: registry::Id,
```

proxy/src/http/user.rs
```diff
@@ -152,6 +164,39 @@ fn register_project_filter<R: registry::Client>(
...
+    registry: http::Shared<R>,
+    subscriptions: notification::Subscriptions,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone
+where
+    R: registry::Client,
+{
+    http::with_shared(registry)
+        .and(http::with_subscriptions(subscriptions))
+        .and(warp::post())
+        .and(document::param::<String>(
```
NunoAlexandre, line 178:
```suggestion
        .and(document::param::<registry::Id>(
```

proxy/src/http/user.rs
```diff
@@ -239,6 +284,33 @@ mod handler {
         )
         .await
     }
+
+    /// Transfer funds to the given `recipient`.
+    pub async fn transfer<R: registry::Client>(
+        registry: http::Shared<R>,
+        subscriptions: notification::Subscriptions,
+        _handle: String,
```
NunoAlexandre, line 292:
```suggestion
        _handle: registry::Id,
```

proxy/src/http/user.rs
```diff
@@ -305,14 +377,53 @@ impl ToDocumentedType for RegisterInput {
...
+        document::DocumentedType::from(props).description("Input for User registration")
+    }
+}
+
+/// Bundled input data for transfer.
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TransferInput {
+    /// Account id of the recipient.
+    recipient: radicle_registry_client::ed25519::Public,
```
NunoAlexandre, line 389: I guess that in the future we wouldn't need to have this here since we will have the handle on the URN already and we will be able to look up the account id based on the handle?

MeBrei: In the ui, either `handle` or `account_id` can be passed into the send-funds form. Depending on where/how we validate those, the input here could be also one or the other.

NunoAlexandre: Or, from the UI, if a handle is given we look up the address, that'd be another way of solving it. It would be nice to cut the duplication of having the handle in the UI and an address in the payload, which must refer to the same resource. We can think about this in a next step!

PR #654, opened 2020-07-09
chore: update surf to 0.4.0

### LINKED ISSUE #573: User Repository Browser
The browsing of source trees currently just browses the local user's view. i.e `namespaces/<project>/refs/heads/*`.
We want to be able to select a remote's branch. This means that the endpoints should take the `PeerId` of the user so that we can look at the branch `namespaces/<project>/remotes/<peer>/refs/heads/*`.

My thinking around this is that the parameter would be optional and so if it isn't provided we would surf our own `heads`.

FintanH:
Fixes #573 

This means that we have structured revisions now. For some of the
endpoints, this is a minor change. What's passed in can be converted to
a Rev. For blob and tree, we need a way to get the structured input for
the revision -- instead of guessing from a string. The structure can be
a Tag, Sha, or a Branch (with an optional remote PeerId).


proxy/src/http/source.rs
```diff
@@ -1003,6 +1028,71 @@ mod test {
...
+        let query = super::BlobQuery {
+            path: path.to_string(),
+            peer_id: None,
+            revision: Some(revision),
+            highlight: Some(false),
+        };
+
+        let path = format!("/blob/{}?{}", urn, serde_qs::to_string(&query).unwrap());
+
+        println!("PATH: {}", path);
```
rudolfs, line 1085:
```suggestion
```

FintanH: :kissing::ok_hand:

ui/Screen/Project/Source.svelte
```diff
@@ -29,18 +29,30 @@
...
+    const parsed = path.parseProjectSourceLocation($querystring);
+
+    currentPeerId = parsed.peerId || null;
+    currentObjectType = parsed.objectType || ObjectType.Tree;
+    currentObjectPath = parsed.objectPath || null;
+
+    if (parsed.revision) {
+      if (JSON.stringify(currentRevision) !== JSON.stringify(parsed.revision)) {
+        currentRevision = parsed.revision;
+      }
```
cloudhead, line 47: :eyes:

cloudhead: Since they are simple objects, maybe better to check the fields manually, or implement a `compare` or `eq` method on the type.

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/654/commits/a12ee4e64a06243d566e5522b7bec83f0d021e87

PR #656, opened 2020-07-10
Update registry client

geigerzaehler:
The requires the following necessary changes

* Use `runtime_transaction_version` instead of `runtime_spec_version` for `TransactionExtra`
* `Client::block_header` returns `Option<Header>` instead of returning an error when a block is missing. We add a `BlockMissing` error to the proxy and use that in the `None` case.

We also extract common logic into `Client::submit_transaction` to simplify the change.


NunoAlexandre:
Done in #663

Issue #641, opened 2020-07-08
feat(ui): Extend id validation store


NunoAlexandre:
Now that we have radicle-dev/radicle-registry#544, the id validation store should be extended to ask the registry about a specific id availability instead of trying to get an org or a user with said id and check for their existence.

So now, an id might be available (:heavy_check_mark:), taken (currently in use by another user or org) (:heavy_check_mark:), or retired (:new:). We need a better error message for the latter case, or else users will be confused that they can't use an id they don't see in use by any org or user.

It entails:

- Integrating with the latest registry (after https://github.com/radicle-dev/radicle-registry/issues/578)
- Adding a new HTTP endpoint `GET id/{id}/availability` (open for discussion)
- Adding types on the UI side that map the id availability response type (enum `Available, Taken, Retired`)
- Adjusting the Id validation store to switch on this response and build the appropriate error message

The validation store now is binary though, it operates on a boolean validation fn and shows the specified error message when appropriate. We need to see how we can extend that. @xla any thoughts on this?

PR #663, opened 2020-07-13
chore(proxy): Bump registry client

NunoAlexandre:
To go ahead with https://github.com/radicle-dev/radicle-upstream/issues/641, we need to adjust to the latest breaking changes presented by the radicle registry client.

@xla the biggest chunk here is moving to `Optional<BlockNumber>`. Could you have a detailed look at how I've done it? I didn't introduce those changes in the Registry and it's not 100% clear to me if I am doing the right thing in all affected lines of code.

### DIFF
--- a/proxy/src/registry.rs
+++ b/proxy/src/registry.rs
@@ -307,10 +307,10 @@ impl Registry {
         let nonce = self.client.account_nonce(&author.public()).await?;
         let runtime_spec_version = self.client.runtime_version().await?.spec_version;
         let extra = protocol::TransactionExtra {
-            nonce,
             genesis_hash: self.client.genesis_hash(),
+            nonce,
+            runtime_spec_version,
             fee,
-            runtime_transaction_version: runtime_spec_version,
         };
         Ok(protocol::Transaction::new_signed(author, message, extra))
     }
@@ -375,7 +375,7 @@ impl Client for Registry {
         let block = self.client.block_header(applied.block).await?;
         let tx = Transaction::confirmed(
             Hash(applied.tx_hash),
-            block.map(|b| b.number),
+            block.number,
             Message::OrgRegistration { id: org_id.clone() },
             fee,
         );
@@ -406,7 +406,7 @@ impl Client for Registry {
 
         Ok(Transaction::confirmed(
             Hash(applied.tx_hash),
-            block.map(|b| b.number),
+            block.number,
             Message::OrgUnregistration { id: org_id },
             fee,
         ))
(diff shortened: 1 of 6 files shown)


proxy/src/registry.rs
```diff
@@ -375,7 +375,7 @@ impl Client for Registry {
         let block = self.client.block_header(applied.block).await?;
         let tx = Transaction::confirmed(
             Hash(applied.tx_hash),
-            block.number,
+            block.map(|b| b.number),
```
FintanH, line 378: Maybe `block` could have a helper `to_number: Block -> Option<Number>`

proxy/src/registry/transaction.rs
```diff
@@ -293,7 +293,9 @@ where
                 State::Confirmed {
                     block, timestamp, ..
                 } => {
-                    let target = block.checked_add(BLOCK_BOUND).unwrap_or(MIN_CONFIRMATIONS);
+                    let target = block.map_or(0, |b| {
```
FintanH, line 296: Can I ask why `0` is safe here?

NunoAlexandre: I am not certain it is, thus my comment on the description. My idea is that the block height of a tx starts in 0, so if there is no block for a given tx, treating its height as `0` seems appropriate. @CodeSandwich could you have a look at this as well?

proxy/src/registry.rs
NunoAlexandre: The problem is that it's `block` that is `Option`al.

proxy/src/registry/transaction.rs
```diff
@@ -96,7 +96,7 @@ impl Transaction {
     #[must_use]
     pub fn confirmed(
         id: registry::Hash,
-        block: protocol::BlockNumber,
+        block: Option<protocol::BlockNumber>,
```
cloudhead, line 99: In what cases do we not have a number?

proxy/src/registry/transaction.rs
cloudhead: This will result in txs always being settled if there is no block number provided..

CodeSandwich: I think that finding `None` in this context is a nasty error and should be treated as such. If I undrestand correctly, this comes from change in the `client::ClientT` API, where `block_header` changed its return type from `Result<BlockHeader>` to `Result<Option<BlockHeader>>`. `None` is returned when the given block does not exist. It was lifted from a variant of the `Error` to a separate layer, because not in all contexts it's an error. In this one I think it is, we're 100% sure that this block should exist.

NunoAlexandre: > this comes from change in the client::ClientT API, where block_header changed its return type from Result<BlockHeader> to Result<Option<BlockHeader>>

That is correct, @CodeSandwich. I will add an error layer here then and revert the change to using Options.

NunoAlexandre: Done, thanks for the feedback!

PR #664, opened 2020-07-13
fix(ui): fix sidebar button hover

cloudhead:
Currently, when a sidebar button is hovered, it's only clickable if the cursor is over the icon. This change makes the whole width clickable, which matches the hover indicator.

2020-07-13-124058_170x77_scrot


sarahscott:
Running into GH shenanigans with the commit history 🤔

cloudhead:
Fixed in master.

PR #666, opened 2020-07-14
feat(proxy): Feed API

sarahscott:
- [x] update `projects/` endpoint to have `routes`
- [x] add a `discover/` path
- [x] return a mocked feed & integrate with ui


proxy/src/project.rs
```diff
@@ -47,25 +47,30 @@ pub struct Project {
     /// Informs if the project is present in the Registry and under what top-level entity it can be
     /// found.
     pub registration: Option<Registration>,
+    /// Maintainers of the project.
+    pub maintainers: Vec<coco::User>,
     /// High-level statistics about the project
     pub stats: coco::Stats,
 }
 
 /// Construct a Project from its metadata and stats
```
sarahscott, line 56: (and `maintainers`, potentially?)

proxy/src/project.rs
```diff
@@ -85,5 +90,14 @@ pub fn get(peer: &coco::PeerApi, project_urn: &coco::Urn) -> Result<Project, err
...
 
-    Ok((project, stats).into())
+    let maintainers: Vec<coco::User> = project
+        .maintainers()
+        .iter()
+        .map(|user_urn| {
+            let draft = coco::get_user(peer, user_urn);
+            draft.and_then(coco::verify_user)
+        })
+        .collect::<Result<_, _>>()?;
```
sarahscott, line 100: Turbofish is banned right? What is a prettier way to do this?

ui/src/project.ts
```diff
@@ -24,7 +24,7 @@ export interface Project {
   id: string;
   shareableEntityIdentifier: string;
   metadata: Metadata;
-  stats: Stats;
+  stats?: Stats;
```
sarahscott, line 27: Optionality here depends on whether or not we use a different `type` for `Discovery` projects.

proxy/src/project.rs
FintanH: I use it all the time. Pretty sure it's not banned.

cloudhead: Turbofish is great :fish:

cloudhead:
The maintainers are part of the radicle-link project metadata document, which we want to include in the feed if I'm not mistaken (cc @massimiliano-mantione), so we can retrieve it from there. As for the stats, we should include them inline in the feed item. I expect there to be a separate time for a "Project Event" that includes potentially some additional data that we can display (like the stats).

cloudhead:
> Is there a way to get project maintainers without a browser? If not, should we remove maintainers & stats for now?

Can you explain why we can't use a browser? Is it because it requires the full source code?

cloudhead:
The maintainers can be gotten from the radicle-link project though: `librad::meta::project::Project::maintainers`

sarahscott:
`librad::meta::project::Project::maintainers` returns a vector of `urn`'s but not fully-formed `User`'s; maybe it makes sense to have the `project::Project` return a vector of maintainer `urn`'s as well instead of fetching each user individually.

FintanH:
> Can you explain why we can't use a browser? Is it because it requires the full source code?

We're getting stats through the browser, which in turn needs a valid reference. I can't remember... Will we have any of the tracking peer's references?

> librad::meta::project::Project::maintainers returns a vector of urn's but not fully-formed User's; maybe it makes sense to have the project::Project return a vector of maintainer urn's as well instead of fetching each user individually.

If a `Project` is in the monorepo that would have to mean we have `rad/ids/*` which means we have the `User`s in the monorepo as well.

cloudhead:
> We're getting stats through the browser, which in turn needs a valid reference. I can't remember... Will we have any of the tracking peer's references?

Right -- in this case though, the stats have to be included by the seed node into the event stream, since they can't be derived from untracked projects (yet).

FintanH:
> Right -- in this case though, the stats have to be included by the seed node into the event stream, since they can't be derived from untracked projects (yet).

Right, that would make sense. I'm sort of fuzzy on the details of how the discovery will work. I'll try to do some spelunking of the proejct boards and see what I can figure out. But iirc, there's no spec right?

proxy/src/coco/peer.rs
```diff
@@ -10,6 +10,7 @@ pub use librad::net::peer::{PeerApi, PeerConfig};
 use librad::uri::RadUrn;
 use radicle_surf::vcs::git::{self, git2};
 
+use crate::discovery::{get_feed, Feed};
```
xla, line 13: This goes against our dependency direction, where inner modules (i.e. protocol wrappers like `coco` and `registry`) are consumed by outer ones (e.g. `project`). There should be a way to wrap whatever coco offfers into our domain understanding. An example for this kind of exercise is `coco::source`, it's easy to imagine a `coco::feed` module to handle this concern.

proxy/src/discovery.rs
```diff
@@ -0,0 +1,34 @@
+//! Types and functionality related to Discovery
+
+use serde::Serialize;
+
+/// Representation of an untracked project
+#[derive(Serialize)]
+pub struct FeedItem {
```
xla, line 7: The fields in this struct are all project related, yet the name indicates a general purpose type. What is the to be expected needs for items in a feed?

proxy/src/coco/peer.rs
xla: So does the import on `project::Project` below.

proxy/src/http/discovery.rs
```diff
@@ -0,0 +1,52 @@
+//! Endpoints for discovery feed.
+
+use warp::document::{self};
+use warp::{path, Filter, Rejection, Reply};
+
+/// `GET /discovery`
```
xla, line 6: If this is the endpoint to discover projects it should live under `projects` and doesn't require its own mount pont.

sarahscott: I think that could make sense - I guess the argument for having a `discovery/` endpoint is that the discovery process might extend to other users or non-project entities in the future. @cloudhead what do you think?

proxy/src/coco/peer.rs
sarahscott: Cool cool makes sense

sarahscott:
I think #686 makes a lot of really nice improvements that affect this, so I'm going to table it and focus on some housekeeping in the meantime.

proxy/src/http/discovery.rs
xla: That will probably be the case, which would in turn mean that we have `/discover` endpoints under the respective mount points, e.g. `users` and so on. We should think along the lines of resources and how endpoints expose these resources. If one wants to discover a certain type of entity the least amount of surprise will be to find the endpoint to do so under the namespace of that entity.

xla:
Trying to finish up #686 today.

ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -11,6 +11,9 @@
 
   const dispatch = createEventDispatcher();
 
+  // TODO(sos): remove this feature flag when we're ready
+  const showDiscovery = true;
```
rudolfs, line 15: We could expose `process.env.NODE_ENV === "development"` to the renderer via `native/preload.js` and show "preview-features" only when the app is started in dev mode.

proxy/src/http/project.rs
```diff
@@ -11,26 +11,40 @@ use crate::http;
...
+    path("projects").and(
+        get_filter(ctx.clone())
+            .or(create_filter(ctx.clone()))
+            .or(discover_filter(ctx.clone()))
+            .or(list_filter(ctx)),
+    )
+}
+
+/// Combination of all project filters.
+#[cfg(test)]
```
rudolfs, line 28: What does this do?

ui/Screen/Discovery/Project.svelte
```diff
@@ -109,16 +115,18 @@
...
 
   <div class="bottom">
     <Stats
       branches={project.stats.branches}
       commits={project.stats.commits}
       contributors={project.stats.contributors} />
-    <Tooltip value={project.maintainers[0].handle}>
-      <Avatar avatarFallback={project.maintainers[0].avatar} size="small" />
-    </Tooltip>
+    {#if project.maintainers && project.maintainers.length}
```
rudolfs, line 126:
```suggestion
    {#if project.maintainers && project.maintainers.length > 0}
```

proxy/src/coco/peer.rs
```diff
@@ -14,10 +14,10 @@ pub use librad::net::peer::{PeerApi, PeerConfig};
 use librad::paths;
 use librad::peer::PeerId;
 use librad::uri::RadUrn;
-use radicle_surf::vcs::git::{self, git2};
+use radicle_surf::vcs::git::{self, git2, Stats};
 
 use crate::error;
-use crate::project::Project;
+use crate::project::{DiscoveryItem, Metadata, Project};
```
xla, line 20: Continue to importing this is violating our domain boundaries. This module should not be aware of the internals of the project/discovery domain. I'd like to see this addressed for the reason that if we don't do it now it will start to creep and we further blurr these boundaries. I understand that there is a clear formaulation of how these boundaries look like and will work to provide appropriate documentation for it.

proxy/src/coco/peer.rs
```diff
@@ -307,6 +307,34 @@ impl Api {
         Ok(meta)
     }
 
+    /// Returns a feed of untracked [`project::Project`]s
+    ///
+    /// # Errors
+    ///
+    ///   * Retrieving the feed fails
+    #[allow(clippy::unused_self)]
+    pub fn discover_projects(&self) -> Result<Vec<DiscoveryItem>, error::Error> {
```
xla, line 316:
```suggestion
    pub fn discover_projects() -> Result<Vec<DiscoveryItem>, error::Error> {
```
Instead of introducing this exception we can convert this method into an [associated function](https://doc.rust-lang.org/reference/items/associated-items.html).

proxy/src/coco/peer.rs
```diff
@@ -307,6 +307,34 @@ impl Api {
         Ok(meta)
     }
 
+    /// Returns a feed of untracked [`project::Project`]s
```
xla, line 310:
```suggestion
    /// Returns a feed of untracked [`project::Project`]s.
```

xla: We should flag that this is currently only dealing with stub data.

proxy/src/coco/peer.rs
xla: To the comment above about domain boundaries this method should live in the `project` module.

proxy/src/http/project.rs
xla: The [`cfg`](https://doc.rust-lang.org/stable/rust-by-example/attribute/cfg.html) attribute helps to conditionally compile code. In this case it guards this function to be only available in the context of tests. The reason for this indirection is that we in our module tests when we assert endpoint behaviour are not interested in the routing prefix/namespace. This oddity will be addressed in a small refactor of the http stack.

proxy/src/project.rs
```diff
@@ -87,3 +87,22 @@ pub fn get(api: &coco::Api, project_urn: &coco::Urn) -> Result<Project, error::E
 
     Ok((project, stats).into())
 }
+
+/// Controversial placeholder struct for `EventStream` projects. It's cumbersome to create
+/// a new `RadUrn` otherwise and the `EventStream` will likely include fields the `Project`
+/// does not (e.g. maintainers)
```
xla, line 93:
```suggestion
/// does not depend on (e.g. maintainers).
```

proxy/src/project.rs
```diff
@@ -87,3 +87,22 @@ pub fn get(api: &coco::Api, project_urn: &coco::Urn) -> Result<Project, error::E
...
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DiscoveryItem {
+    /// Unique identifier of the project in the network.
+    pub id: String,
```
xla, line 98: We should be adamant to use more concrete types wherever possible. While it introduces more impedence initially the gains over time are outweighing that. It would make @FintanH very happy.

ui/DesignSystem/Component/Sidebar.svelte
xla: Introducing a vital configuration dimension like feature flags like this will not set us up to properly address the need for it. If we feel the urgent need for it we should spend time to do proper requirement engineering and build up a solution.

Yet I'd argue we don't need to feature guard this at the moment, we don't have a diverse audience which could be confused by this.

xla: Another question in this context similar to the ongoing discussions around capabilities, permissions and validations is where should the source of truth live for this.

ui/src/discovery.ts
```diff
@@ -1,133 +1,12 @@
...
-      },
-    ],
-    tracked: false,
-  },
-];
+import * as api from "./api";
+import * as project from "./project";
+import * as remote from "./remote";
+
+const feedStore = remote.createStore<project.Project[]>();
```
xla, line 5: If this is just holding projects it's fine to name it accordingly, we can move to a general feed store once we have proper feed items.

ui/DesignSystem/Component/Sidebar.svelte
sarahscott: Let's not overcomplicate things; sounds like feature flags aren't necessary at this stage.

proxy/src/project.rs
sarahscott: @FintanH or @xla what's the best way to create a stubbed `urn` without going through the whole dance of a stubbed `api`?

proxy/src/coco/peer.rs
sarahscott: For interacting with the `peer_api`, would you suggest something like
```
    pub fn get_entity(&self, urn: &RadUrn) -> Result<entity::Entity, error::Error> {
        let api = self.peer_api.lock().expect("unable to accquire lock");
        let storage = api.storage().reopen()?;

        Ok(storage.metadata(urn)?)
    }
```

instead of the existing `get_project` and `get_user` functions? Or would it make more sense to expose the `peer_api` & `storage` to `project`/`user`/other proxy modules?

proxy/src/project.rs
FintanH: This would make me happy :)

There's an example in `librad`. I'll post it here to save you looking for it this time:

```rust
use librad::{
    hash::Hash,
    uri::{Path, Protocol, RadUrn},
};

let urn = RadUrn::new(
    Hash::hash(b"geez"),
    Protocol::Git,
    Path::parse("rad/issues/42").unwrap(),
);

assert_eq!(
    "rad:git:hwd1yredksthny1hht3bkhtkxakuzfnjxd8dyk364prfkjxe4xpxsww3try/rad/issues/42",
    urn.to_string()
)
```

proxy/src/coco.rs
```diff
@@ -1,7 +1,8 @@
 //! Abstractions and utilities for git interactions through the API.
 
-pub use librad::uri::rad_urn::ParseError;
-pub use librad::uri::RadUrn as Urn;
+pub use librad::hash::Hash;
+pub use librad::meta::project::Project;
+pub use librad::uri::{rad_urn::ParseError, Path, Protocol, RadUrn as Urn};
```
FintanH, line 5: I wouldn't expose the `ParseError` at the top level. There are multiple error types that are called `ParseError` in librad. I would probably say the same for `Path` and `Protocol`, as well.

Why did you decide to export them here?

Note that this was ok for `RadUrn` because it's unique in the namespace.

proxy/src/http/project.rs
```diff
@@ -59,6 +60,7 @@ where
     http::with_context(ctx)
         .and(warp::get())
         .and(document::param::<String>("id", "Project id"))
+        .and(path::end())
```
FintanH, line 63: What's this? :thinking:

proxy/src/project.rs
```diff
@@ -87,3 +86,47 @@ pub fn get(api: &coco::Api, project_urn: &coco::Urn) -> Result<Project, error::E
 
     Ok((project, stats).into())
 }
+
+/// Returns a list of `Project`s for your peer.
+pub fn list_projects(api: &coco::Api) -> Result<Vec<Project>, error::Error> {
```
FintanH, line 91: Thanks for fixing my module boundary mistake :bow:

proxy/src/project.rs
```diff
@@ -87,3 +86,47 @@ pub fn get(api: &coco::Api, project_urn: &coco::Urn) -> Result<Project, error::E
...
+        })
+        .collect()
+}
+
+/// Returns a stubbed feed of `Project`s
+pub fn discover() -> Result<Vec<Project>, error::Error> {
+    let urn = coco::Urn::new(
+        coco::Hash::hash(b"hash"),
+        coco::Protocol::Git,
+        coco::Path::parse("rad/issues/42")?,
```
FintanH, line 110: This should be empty for a project. An empty path defaults to `rad/id`.

proxy/src/project.rs
```diff
@@ -87,3 +86,47 @@ pub fn get(api: &coco::Api, project_urn: &coco::Urn) -> Result<Project, error::E
...
+    );
+
+    let projects = vec![
```
FintanH, line 113: I would say add another project. Just to make sure that things look ok in the frontend when there's more than one datum.

proxy/src/coco.rs
sarahscott: Ah ok. I think I was unsure about why we import from `librad` in some instances but export `librad` entities from `coco` elsewhere. The `ParseError` was already exported here, should it be removed?

proxy/src/project.rs
sarahscott: good call

proxy/src/http/project.rs
sarahscott: idk i think it snuck in during a merge, will remove it if it's not necessary

proxy/src/coco.rs
FintanH: We can do a safe middle ground and export `self` so they can be accessed via `coco::uri`

proxy/src/http/project.rs
FintanH: I'd have to check. Not sure what it's for. Can you check the docs and see if it's necessary?

sarahscott: Yeah I'll take a look at the docs and try to summarize findings. If everything works & the tests pass without it I'll get rid of it

sarahscott: It seems like it's just more specific about where exactly the `path` should end... this endpoint functions properly with or without it. @xla any feelings about using `end`? From the [docs](https://docs.rs/warp/0.1.22/warp/filters/path/fn.end.html): 

> Note that not including end() may result in shorter paths like /math unintentionally matching /math/sum.

Tbh I'm fine with either; I see we're using `path::end` in some other places so we should pick a convention and stick with it.

xla: The `end` is significant, as it prevents matches in situations where you have a route for `/` and a route for `/foo/`. The latter might never be triggered as the first one consumes the path until the first slash. So if we want an explicit match on `GET /` we should always add the `/`.

proxy/src/http/project.rs
```diff
@@ -103,6 +104,22 @@ where
...
+fn discover_filter<R>(
+    ctx: http::Ctx<R>,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone
+where
+    R: registry::Client + 'static,
+{
+    path("discover")
+        .and(warp::get())
+        .and(http::with_context(ctx))
+        .and(document::document(document::description(
```
xla, line 117: We do return a list of projects here we should describe the body in more detail. e.g.:

https://github.com/radicle-dev/radicle-upstream/blob/64371b53e20caa9f67803842f2e0136e13fff40e/proxy/src/http/project.rs#L93-L102

proxy/src/http/project.rs
```diff
@@ -505,4 +532,26 @@ mod test {
...
+        let key = ctx.keystore.get_librad_key()?;
+        let owner = ctx.peer_api.init_owner(key.clone(), "cloudhead")?;
+
+        coco::control::setup_fixtures(&ctx.peer_api, key, &owner)?;
+
+        let feed = project::discover()?;
+        let res = request().method("GET").path("/discover").reply(&api).await;
+
+        http::test::assert_response(&res, StatusCode::OK, |have| {
+            assert_eq!(have, json!(feed));
```
xla, line 552: It would be nice to assert not only the self-referencing test here but also that we expect a populated json response.

proxy/src/http/project.rs
FintanH: Ah cool, thanks for enlightening me. Damn ambiguous grammars :)

PR #669, opened 2020-07-14
feat(proxy): expose the account id for user and org

### LINKED ISSUE #658: Expose public keys for org and user
For the send and receive funds functionality we need to expose the public key of users and orgs in the API responses. 
Screenshot 2020-07-10 at 12 23 48
Screenshot 2020-07-10 at 12 24 04

MeBrei:
Closes #658

### DIFF
--- a/proxy/src/http/user.rs
+++ b/proxy/src/http/user.rs
@@ -243,7 +243,7 @@ mod handler {
 
 impl ToDocumentedType for registry::User {
     fn document() -> document::DocumentedType {
-        let mut props = HashMap::with_capacity(2);
+        let mut props = HashMap::with_capacity(3);
         props.insert(
             "handle".into(),
             document::string()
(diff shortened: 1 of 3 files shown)


proxy/src/http/user.rs
```diff
@@ -346,23 +352,33 @@ mod test {
             .await
             .unwrap();
 
-        let api = super::filters(registry, store, subscriptions);
+        let api = super::filters(Arc::clone(&registry), store, subscriptions);
```
sarahscott, line 355: Just wondering for my own knowledge - why should this be `clone`d here?

MeBrei: I introduced a new place were `registry` is being used [L.362](https://github.com/radicle-dev/radicle-upstream/pull/669/files/ffe70a3beb5dfc071244b406e244c9cd85452033#diff-adc9819afa99f02dc58458363a1e459bR362), so to use `registry` there, I need to clone it here.

PR #670, opened 2020-07-14
chore(ui): Adjust id validation

### LINKED ISSUE #641: feat(ui): Extend id validation store

NunoAlexandre:
Closes #641 

It introduces a new endpoint `ids/<id>/status` that fetches the availability status of a given id in the registry, that is either `available | retired | taken`. In the UI, we now call this endpoint and validate the availability of an id based on its status being `available`. The error message is changed from "Sorry, this one is already taken" to "Sorry, this one is no longer available", which better represents the semantics behind an id not being available.

To make testing possible, I added support to unregister a user from the registry in the proxy only.

**Note**: In a follow up issue I would move the `avatars/<id>` endpoint under `ids`, as `ids/<id>/avatar`.

### DIFF
--- a/proxy/src/http/id.rs
+++ b/proxy/src/http/id.rs
@@ -61,7 +61,6 @@ mod handler {
     use crate::http;
     use crate::registry;
 
-    #[allow(clippy::all)]
     /// Get the status for the given `id`.
     pub async fn get_status<R: registry::Client>(
         registry: http::Shared<R>,
--- a/proxy/src/http/id.rs
+++ b/proxy/src/http/id.rs
@@ -1,42 +1,21 @@
 //! Endpoints for Id.
 
 use serde::{Deserialize, Serialize};
-use std::sync::Arc;
 use warp::document::{self, ToDocumentedType};
 use warp::{path, Filter, Rejection, Reply};
 
 use crate::http;
 use crate::registry;
 
-/// Prefixed filters.
-pub fn routes<R>(
-    registry: &http::Shared<R>,
-) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone
-where
-    R: registry::Client,
-{
-    get_status_filter(Arc::clone(registry))
-}
-
-/// Combination of all ids routes.
-#[cfg(test)]
-fn filters<R>(
-    registry: &http::Shared<R>,
-) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone
-where
-    R: registry::Client,
-{
-    get_status_filter(Arc::clone(registry))
-}
-
-/// `GET /<id>/status`
-fn get_status_filter<R: registry::Client>(
+/// `GET ids/<id>/status`
+pub fn get_status_filter<R: registry::Client>(
     registry: http::Shared<R>,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone
 where
     R: registry::Client,
 {
-    http::with_shared(registry)
+    path("ids")
+        .and(http::with_shared(registry))
         .and(warp::get())
         .and(document::param::<String>(
             "id",
@@ -44,8 +23,9 @@ where
         ))
         .and(path("status"))
         .and(path::end())
+        .and(document::document(document::tag("Id")))
         .and(document::document(document::description(
-            "Get the status for the given id",
+            "Fetch the availability status of the given id",
         )))
         .and(document::document(
             document::response(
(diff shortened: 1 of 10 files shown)


ui/src/id.ts
```diff
@@ -1,10 +1,23 @@
...
+  Retired = "retired",
+}
+
+// const getStatus = (id: string): Promise<Status> =>
+//   api.get<Status>(`ids/${id}/status`);
+
 // Check if the given id is available
 const isAvailable = (id: string): Promise<boolean> =>
-  org.getOrg(id).then(org => !org && user.get(id).then(user => !user));
+  api.get<string>(`ids/${id}/status`).then(status => status == "available");
```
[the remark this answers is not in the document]
NunoAlexandre: Good point!

NunoAlexandre:
> Really nice change-set!! 

Thank you! As well for your thoughtful review.

> 
> The only thing that would be nice to improve is naming. To me `id` is very non-descriptive and generic, it doesn't communicate what it actually is. It could become confusing especially because there's `IDs` on the coco side as well. I think the most straightforward solution would be to just namespace it under `registry:
> 
> `proxy/src/http/id.rs` -> `proxy/src/http/registryId.rs`
> `ui/src/id.ts` -> `ui/src/registryId.ts`
> `GET ids/<id>/status` -> `GET /registry/ids/<id>/status`
> 
> But I'm also open for other suggestions.

I feel you on this, "id" is indeed super generic and it can take multiple meanings or sources here.
Having that said, I don't think that prefixing this new endpoint is the way to go, given opportunities such as #671 to make the API more cohesive and structured, where `id` can take different meanings but still be sensible to organize such endpoints under a common nomenclature. For instance, when we register a user in the registry, `POST /users/<handle>/register`, we don't do that under a `registry` prefix for that same reason. So, I think all in all it will work well to have this "ids" endpoint, as it will reduce the scatteredness around ids and have a more structured look at them at the API level.

Let me know what you think?

> 
> We should also urge to add a couple integration tests for the UI, see `cypress/integration`, we should at least cover the following cases:
> 
> * register user handle, then try registering org with the same name;
> * register org, try registering another org with the same name.

I thought I had added these already in a previous PR? Will have a look!

NunoAlexandre:
> > Let me know what you think?
> 
> Yeah, I agree, name-spacing the endpoint under `registry` is also not optimal. I'm still concerned that this will cause confusion with coco stuff later. But I guess we can burn the bridge when we get there ;P

Cool :)
> 
> > I thought I had added these already in a previous PR? Will have a look!
> 
> You're right, found them. Never mind.

No worries, better safe than sorry!

proxy/src/http/id.rs
```diff
@@ -0,0 +1,266 @@
...
+    use warp::{reply, Rejection, Reply};
+
+    use crate::error::Error;
+    use crate::http;
+    use crate::registry;
+
+    /// Get the status for the given `id`.
+    pub async fn get_status<R: registry::Client>(
+        registry: http::Shared<R>,
+        id_string: String,
```
FintanH, line 67: Pretty sure you could do `registry::Id` here because of its `Deserialise` instance

proxy/src/registry.rs
```diff
@@ -624,6 +648,31 @@ impl Client for Registry {
...
     }
 
+    async fn unregister_user(
+        &self,
+        author: &protocol::ed25519::Pair,
+        handle: Id,
+        fee: Balance,
+    ) -> Result<Transaction, error::Error> {
+        // Prepare and submit user unregistration transaction.
+        let unregister_message = protocol::message::UnregisterUser {
```
FintanH, line 658: Heh... So _all_ these functions look the same! Recalling back to @MeBrei's PR yesterday. I'm _highly_ interested to see if we could find the right abstraction :eyes:

NunoAlexandre: Yeah, agree. Could you maybe move that discussion into a dedicated issue?

proxy/src/http/id.rs
NunoAlexandre: Good point, thanks for keeping on bringing that up.

proxy/src/registry.rs
FintanH: #672

PR #673, opened 2020-07-15
fix(ui): show correct commit branch

cloudhead:
Removes the "master" stub.

### DIFF
--- a/proxy/src/coco/source.rs
+++ b/proxy/src/coco/source.rs
@@ -73,7 +73,7 @@ pub struct Commit {
     /// The changeset introduced by this commit.
     pub diff: diff::Diff,
     /// The branch this commit belongs to.
-    pub branch: String,
+    pub branch: Branch,
 }
 
 /// Representation of a code commit.
@@ -429,10 +429,12 @@ pub fn commit<'repo>(browser: &mut Browser<'repo>, sha1: &str) -> Result<Commit,
     let branch = branches.first();
 
     // Known commits always have at least one branch. If this isn't the case, it's a bug.
-    let branch = branch
-        .expect("known commits must be on a branch")
-        .name
-        .to_string();
+    let branch = Branch(
+        branch
+            .expect("known commits must be on a branch")
+            .name
+            .to_string(),
+    );
 
     Ok(Commit {
         header: commit.into(),
--- a/proxy/src/coco/source.rs
+++ b/proxy/src/coco/source.rs
@@ -72,6 +72,8 @@ pub struct Commit {
     pub stats: CommitStats,
     /// The changeset introduced by this commit.
     pub diff: diff::Diff,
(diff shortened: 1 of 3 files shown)


proxy/src/coco/source.rs
```diff
@@ -72,6 +72,8 @@ pub struct Commit {
     pub stats: CommitStats,
     /// The changeset introduced by this commit.
     pub diff: diff::Diff,
+    /// The branch this commit belongs to.
+    pub branch: String,
```
FintanH, line 76: I would say we could use the `Branch` type defined in this module

proxy/src/coco/source.rs
```diff
@@ -419,12 +421,26 @@ pub fn commit<'repo>(browser: &mut Browser<'repo>, sha1: &str) -> Result<Commit,
...
+    let branches = browser.revision_branches(sha1)?;
+
+    // If a commit figures in more than one branch, there's no real way to know
+    // which branch to show without additional context. So, we choose the first
+    // branch.
+    let branch = branches.first();
+
+    // Known commits always have at least one branch. If this isn't the case, it's a bug.
+    let branch = branch
+        .expect("known commits must be on a branch")
```
FintanH, line 433: I'd prefer we added an error variant than `expect`

cloudhead: Is this an error we want to handle though? If a commit doesn't have a branch, I'm not sure what we're supposed to do?

FintanH: Well, our error variant gets resurfaced as a `warp::Rejection`. It's more to have a structure to the error and we're already in the `Result<_, error::Error>`

cloudhead:
We should have a broader conversation about unwrap vs. Result. I have my own heuristic for deciding which to use, but I bet it's different from yours.

FintanH:
> We should have a broader conversation about unwrap vs. Result. I have my own heuristic for deciding which to use, but I bet it's different from yours.

I pretty much put my thoughts here: https://github.com/radicle-dev/radicle-decisions/blob/master/proposals/0001.md :smile: I think it could be refined, but I generally avoid panic related code. Error documentation is useful to me as a developer too :)

FintanH:
> Hmm there's no mention of panic/unwrap/expect there though, so doesn't answer the question of what to do in cases where an error is unrecoverable..

Well unrecoverable errors would be the `dyn Error` case. But also, I'm pretty sure none of our errors in `error::Error` are recoverable. And in the future we could use it to have the correct error responses. I think in the case of panicing it will always just be `500`.

cloudhead:
A correct error response would only make sense if you expect the error to happen.  In this case the code is saying that this state cannot happen unless there is a bug in the underlying library, or in git, in which case you want to blow up, because an invariant was broken.

Issue #570, opened 2020-06-24
Have a better way of handling overflows on repo/branch selector


cloudhead:
Perhaps we could for example hide the user handle and only show the avatar and branch name, when the dropdown isn't expanded. In addition, we probably don't want to wrap the branch name to the next line?

![2020-06-24-114208_507x455_scrot](https://user-images.githubusercontent.com/40774/85532184-c723d380-b60f-11ea-8f2d-6a3c64f0963c.png)

Something like this could work, when the dropdown is collapsed (notice the handle is no longer there):

![2020-06-24-114722_383x219_scrot](https://user-images.githubusercontent.com/40774/85533065-7e204f00-b610-11ea-863c-ebff0d2e0d43.png)

juliendonck:
I think that's a good solution

rudolfs:
<img width="306" alt="Screenshot 2020-06-24 at 14 15 00" src="https://user-images.githubusercontent.com/158411/85554360-1cb6ab00-b625-11ea-87d0-d7b629a8647c.png">

I already worked on it a bit in my feature branch: https://github.com/radicle-dev/radicle-upstream/pull/510/commits/9204309775a832e49c0ac6ba30819cf4f37ea51a.

cloudhead:
> this

Yes, this is broken in master

PR #677, opened 2020-07-15
style(ui): tweak repo selector

rudolfs:
Based on feedback from @cloudhead:
  - show only selected revision, remove handle/peer
  - make opened dropdown fit longer revisions, limit size with max-width

Based on feedback from @juliendonck:
  - remove `PeerId` from expanded dropdown -- we can already distinguish peers based on the icon

Closes: https://github.com/radicle-dev/radicle-upstream/issues/570.

New (NB: there is a hover background color which doesn't show in the gif):
rev

Old:

before


cloudhead:
> remove PeerId as we can already distinguish peers based on the icon

We can't actually. The icon is derived from the RadUrn :)

juliendonck:
> > remove PeerId as we can already distinguish peers based on the icon
> 
> We can't actually. The icon is derived from the RadUrn :)

You're saying it will only be the randomHashId? No avatar and no handle?

cloudhead:
> @cloudhead You're saying it will only be the randomHashId? No avatar and no handle?

Not at all, I'm saying we can't distinguish peers based on the icon alone.

juliendonck:
> > @cloudhead You're saying it will only be the randomHashId? No avatar and no handle?
> 
> Not at all, I'm saying we can't distinguish peers based on the icon alone.

But in the dropdown you'd see the icon and handle. Which should be more than enough.

rudolfs:
Here's the updated version after IRL feedback from @cloudhead & @juliendonck (note: gif still doesn't show hover background, but it is there):
v2

FintanH:
I think I missed something... Why's there an issue with keeping the PeerId hash as part of the name? As mentioned, it would uniquely distinguish a user's machine if they had multiple.

The icons will surely overlap at some stage too right? Which is not as likely with a hash :)

rudolfs:
> I think I missed something... Why's there an issue with keeping the PeerId hash as part of the name? As mentioned, it would uniquely distinguish a user's machine if they had multiple.

> The icons will surely overlap at some stage too right? Which is not as likely with a hash :)

@FintanH: @juliendonck, @cloudhead and I had a conversation offline and settled on only showing `PeerId`s if there are multiple peers for a user. If there's only one peer per user we want to hide the `PeerId` so the UI looks cleaner.

For β1 there won't be support for multiple peers per user in Upstream, hence the formatting was split out in a separate issue: https://github.com/radicle-dev/radicle-upstream/issues/683.

FintanH:
> For β1 there won't be support for multiple peers per user in Upstream, hence the formatting was split out in a separate issue: #683.

I don't see how you could prevent this 🤔 that's stopping a person using Upstream on more than one machine

rudolfs:
> I don't see how you could prevent this 🤔 that's stopping a person using Upstream on more than one machine

Will the functionality for transferring keys to another machine to set up the same identity be already integrated for β1?

FintanH:
> Will the functionality for transferring keys to another machine to set up the same identity be already integrated for β1?

Transferring keys has nothing to do with it. It would be two or more machines attesting that they're the same identity and reaching a quorum. If all goes well that should be possible given the discovery and replication work.

cloudhead:
> Transferring keys has nothing to do with it. It would be two or more machines attesting that they're the same identity and reaching a quorum. If all goes well that should be possible given the discovery and replication work.

This would require manual work from the user though at this stage. Upstream doesn't support updating an identity with an extra device currently, so the only way you'd end up in that situation is if you really know what you're doing.

FintanH:
> This would require manual work from the user though at this stage. Upstream doesn't support updating an identity with an extra device currently, so the only way you'd end up in that situation is if you really know what you're doing.

Pesky super users!

PR #678, opened 2020-07-15
feat(proxy): peer branches

FintanH:
Moving the revisions function into source so that it only operates over
a Browser. The peer logic gets moved to handler and the information gets
passed down.

As well as this, we extend the branches endpoint so that it can take a
PeerId if we want to look at a peer's list of branches. We test this in the revisions test.


FintanH:
I left this as a draft because I built on top of the other PR. But it's pretty much good to review.

proxy/src/http/source.rs
```diff
@@ -349,10 +354,12 @@ mod handler {
     pub async fn branches(
         peer: Arc<Mutex<coco::PeerApi>>,
         project_urn: coco::Urn,
+        super::PeerQuery { peer_id }: super::PeerQuery,
     ) -> Result<impl Reply, Rejection> {
+        log::info!("peer.id = {:?}", peer_id);
```
rudolfs, line 359:
```suggestion
```

FintanH: I mean... I'm sort of interested in logging in the endpoints. But maybe I'll create a ticket for it.

proxy/src/http/source.rs
```diff
@@ -101,10 +101,14 @@ fn branches_filter(
     path("branches")
         .and(warp::get())
         .and(http::with_peer(peer))
-        .and(document::param::<String>(
+        .and(document::param::<coco::Urn>(
             "project_id",
             "ID of the project the blob is part of",
         ))
+        .and(warp::filters::query::query::<PeerQuery>())
```
cloudhead, line 108: Shouldn't it be called `BranchesQuery`? Since it's the query string for `/branches`?

proxy/src/coco/source.rs
```diff
@@ -237,6 +242,18 @@ impl TryFrom<Revision> for Rev {
...
 }
 
+/// Bundled response to retrieve both branches and tags for a user repo.
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UserRevisions {
+    /// Owner of the repo.
+    pub(crate) identity: identity::Identity,
+    /// List of [`git::Branch`].
+    pub(crate) branches: Vec<Branch>,
```
cloudhead, line 252: Should branches be `NonEmpty`? As far as I know you must have a commit and therefore a branch for the application to display a repo.

cloudhead: I realize this wasn't added in this pr, so feel free to ignore :p

proxy/src/coco/source.rs
```diff
@@ -237,6 +242,18 @@ impl TryFrom<Revision> for Rev {
     }
 }
 
+/// Bundled response to retrieve both branches and tags for a user repo.
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UserRevisions {
+    /// Owner of the repo.
+    pub(crate) identity: identity::Identity,
```
sarahscott, line 250: Just curious - why are these only visible in the `crate`?

sarahscott: Is this a best practice we should be following elsewhere too?

proxy/src/http/source.rs
FintanH: Ya fair :+1:

proxy/src/coco/source.rs
FintanH: _ignoring_ :joy: 

Surf doesn't return `NonEmpty`. So maybe you can create a ticket there :eyes:

proxy/src/coco/source.rs
FintanH: They were private, so I went one level up. I suppose since `proxy` is actually the only consumer of this library it could just be `pub`.

What's your stance on this @xla?

proxy/src/coco/source.rs
FintanH: https://github.com/radicle-dev/radicle-surf/issues/151

proxy/src/coco/source.rs
```diff
@@ -237,6 +242,18 @@ impl TryFrom<Revision> for Rev {
     }
 }
 
+/// Bundled response to retrieve both branches and tags for a user repo.
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UserRevisions {
```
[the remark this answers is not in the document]
FintanH: We have `Revision`, which is an enum over `Tag`, `Branch` or `Sha`. So I wouldn't say that this is multiple of those, technically. Not sure what would capture this concept better :shrug: Maybe `Revisions` with docs

proxy/src/coco/source.rs
```diff
@@ -14,7 +17,9 @@ use syntect::highlighting::ThemeSet;
 use syntect::parsing::SyntaxSet;
 use syntect::util::LinesWithEndings;
 
+use crate::coco::User;
 use crate::error;
+use crate::identity;
```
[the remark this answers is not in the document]
FintanH: I forgot about generics. No more modules :)

FintanH:
WIP Update: I parameterised but the output isn't exactly what we want for the front end. Will finish up tomorrow.

FintanH:
@xla: should have a cleaner bound of separation now. Lemme know what you think :)

proxy/src/coco/source.rs
```diff
@@ -213,7 +213,7 @@ pub enum Revision {
         /// Name of the branch.
         name: String,
         /// The remote peer, if specified.
-        peer_id: Option<peer::PeerId>,
+        peer_id: Option<P>,
```
xla, line 216: Guess unrelated to the changes at hand, but when can the `peer_id` be `None`?

proxy/src/http/source.rs
```diff
@@ -399,33 +402,51 @@ mod handler {
...
+            .map_err(Error::from)?
+            .map(|peer_id| {
+                repo.get_rad_self_of(peer_id.clone())
+                    .map(|user| (peer_id.clone(), user))
+                    .map_err(Error::from)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        let peer_id = api.peer_id().clone();
+        let revisions: Vec<super::Revisions> = coco::with_browser(api, &project_urn, |browser| {
+            let owner = owner.to_data().build()?; // TODO(finto): downgraded verified user, which should not be needed.
```
xla, line 425:
```suggestion
            // TODO(finto): downgraded verified user, which should not be needed.
            let owner = owner.to_data().build()?;
```

proxy/src/coco/source.rs
FintanH: When we're not browsing a peer's Branch, i.e. just local browsing

Issue #667, opened 2020-07-14
Add "accounts" endpoint


NunoAlexandre:
Add two endpoints necessary for the wallets work:

- accounts/<account_id>/exists
 Check with the registry whether an account with a given id exists  (Added in #675)

- accounts/<account_id>/balance
  Fetch the balance of an account from the registry

PR #681, opened 2020-07-16
feat(proxy): accounts endpoint group

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-upstream/issues/667.

Note: This PR is branched off https://github.com/radicle-dev/radicle-upstream/pull/675 for now, which is waiting on a registry PR to be merged.


proxy/src/http/account.rs
```diff
@@ -0,0 +1,245 @@
...
+        .and(document::param::<registry::AccountId>("id", "Account id"))
+        .and(path("balance"))
+        .and(path::end())
+        .and(document::document(document::tag("Account")))
+        .and(document::document(document::description(
+            "Fetch the balance of the account from the Registry",
+        )))
+        .and(document::document(
+            document::body(http::RegisterProjectInput::document()).mime("application/json"),
+        ))
```
MeBrei, line 73: `RegisterProjectInput` not needed here

proxy/src/http/account.rs
```diff
@@ -0,0 +1,245 @@
...
+            radicle_registry_client::ed25519::Pair::from_legacy_string("//Cloudhead", None);
+
+        let res = request()
+            .method("GET")
+            .path(&format!("/{}/balance", author.public().to_string()))
+            .reply(&api)
+            .await;
+
+        http::test::assert_response(&res, StatusCode::OK, |have| {
+            assert_eq!(have, json!(0), "Account was expected to exist on chain");
```
MeBrei, line 240:
```suggestion
            assert_eq!(have, json!(0), "Account was not expected to exist on chain");
```

NunoAlexandre:
> Looks good so far. I inlined 2 small comments.
> 
> I tried to test both requests in the API playground. The balance request returned `Response Status: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body. (CORS or Network Issue) ` and the account exists request said that the account did not exist, though I used the account id returned from the fetch user request.

My bad, I am sorry for the poor quality check. Thanks for your thorough look.

One thing more to discuss: Currently, if an account does **not** exist on chain, its balance will be 0. Do we want to return an error or a 404 in such case? We could introduce that error at the proxy registry side and if that error occurs on the api side we return a 404?

proxy/src/http.rs
```diff
@@ -101,6 +103,7 @@ where
 
     let api = path("v1").and(combine!(
         avatar_filter,
+        account_filter,
```
NunoAlexandre, line 106: For reviewers: I added this here instead of before `avatar_filter` because doing the latter failed to compile. Since we will merge `avatar_filter` into `ids`, I just put it here for now.

MeBrei:
> One thing more to discuss: Currently, if an account does **not** exist on chain, its balance will be 0. Do we want to return an error or a 404 in such case? We could introduce that error at the proxy registry side and if that error occurs on the api side we return a 404?

I think we should return a 404. It is quite unlikely that it would happen with the current use cases (ie we only pass the account id we get for the user/org from registry) but especially then, we should make an error noticeable in the ui

NunoAlexandre:
> > One thing more to discuss: Currently, if an account does **not** exist on chain, its balance will be 0. Do we want to return an error or a 404 in such case? We could introduce that error at the proxy registry side and if that error occurs on the api side we return a 404?
> 
> I think we should return a 404. It is quite unlikely that it would happen with the current use cases (ie we only pass the account id we get for the user/org from registry) but especially then, we should make an error noticeable in the ui

Done :v:

proxy/src/http.rs
xla: Addressed in 8953b2f2a1466b2871e2810d128ff7926c73e88a

proxy/src/http/account.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+/// `GET /<id>/exists`
+fn exists_filter<R: registry::Client>(
+    registry: http::Shared<R>,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone
+where
+    R: registry::Client,
+{
+    http::with_shared(registry)
```
xla, line 28: We usually put the dependencies after the verb, path, etc.

proxy/src/http.rs
```diff
@@ -66,6 +67,7 @@ where
     let store = Arc::new(RwLock::new(store));
     let subscriptions = crate::notification::Subscriptions::default();
 
+    let account_filter = path("accounts").and(account::filters(&Arc::clone(&registry)));
```
MeBrei, line 70: I would keep `path("accounts")` in `account.rs` so it is consistent with the other declarations of the path

NunoAlexandre: That was @xla. I would prefer that too.

xla: We gonna move all of the other ones out. It's prefactoring, the great promise of a golden future to come.

proxy/src/http/id.rs
```diff
@@ -17,7 +17,7 @@ where
     path("ids")
         .and(http::with_shared(registry))
         .and(warp::get())
-        .and(document::param::<registry::Id>(
+        .and(document::param::<String>(
```
xla, line 20: Would be great if we had `try_param`. :>

proxy/src/http.rs
NunoAlexandre: Ok. That wold cut the duplication that we currently have setting filters for testing purposes :ok_hand:

proxy/src/http/id.rs
NunoAlexandre: @xla not sure what you are referring to :) could you elaborate? To avoid misunderstandings it's helpful to read this commit's description in case you haven't yet: https://github.com/radicle-dev/radicle-upstream/pull/681/commits/12529a5b31741cfd52b1f8226ddcb69d62cc080f

xla: My comment was in regards that `param` doesn't seem to support types which only implement `TryFrom`, which would be handy in this instance.

NunoAlexandre: ah I see! that's right, will follow up on that. As the diff shows, I moved this one back to `String` to have the docs playground working but I will check whether using `try_param::<registry::Id>` solves the issue. :+1:

NunoAlexandre: @xla, `document::try_param` is not a thing apparently?

xla: That's why I said it would be great if it existed. :dancers:

NunoAlexandre: [papoy](https://youtu.be/l0tOJ3gyokk?t=9)! I thought you were proposing to change it :b

NunoAlexandre:
Before we merge this, I want to fix the issues we are seeing regarding the interaction with this new endpoint.

More specifically, I have two issues I want to address.

### Issue 1 - Interacting with the endpoints - Good and Bad requests

Merle realized that playing with it via the docs playground didn't work, even though it works via the Rust unit tests calling the endpoint. I noticed that the `ids/<>/status` endpoint didn't work from the docs playground either, despite it working from the tests, ui, and the terminal. I fixed the ids case with https://github.com/radicle-dev/radicle-upstream/pull/681/commits/12529a5b31741cfd52b1f8226ddcb69d62cc080f.

Now, for these new `accounts` endpoint, I see it working correctly when I use a well-formed account id:

- exists

```
 ➜ curl http://127.0.0.1:8080/v1/accounts/5FA9nQDVg267DEd8m1ZypXLBnvN7SFxYwV7ndqSYGiN9TTpu/exists     
true%
```

- balance

```
 ➜ curl http://127.0.0.1:8080/v1/accounts/5FA9nQDVg267DEd8m1ZypXLBnvN7SFxYwV7ndqSYGiN9TTpu/balance     
1152921504606846976%
```


**However**, if we provide an ill-formed account id, it fails:

```
 ➜ curl http://127.0.0.1:8080/v1/accounts/foo/exists                                                  
{"message":"Resource not found","variant":"NOT_FOUND"}%
```

The respective error code being `500`:

```
➜ curl -I http://127.0.0.1:8080/v1/accounts/foo/exists                                               
HTTP/1.1 500 Internal Server Error
```


This is no good. For ill-formed input, the returned status code should be `400 Bad Request` and should contain a message describing what's wrong.

**How** I _think_ we could fix this by expecting the param to be a String and deserialize it within the handler. If it fails to deserialize, we then return a 400 explaining it received an invalid account id.


### Issue 2 - Getting `undefined` when automating testing these endpoints from the ui

See `ui/src/account.test.ts`. Both the tests fails with:

```
Expected: <value>
Received: undefined
```

Note they fail using well-formed input that works on the terminal. For a moment I thought I had forgotten to mock the api but that's done. Any idea? :thinking:

NunoAlexandre:
@MeBrei Alright, so I have dropped `account.test.ts` since it was pointless to test these calls against a mocked api. I have fixed the issues reported in issue #1, just waiting for the companion branch in the registry to be merged.

@xla It's weird to me that to fail with `404` one does `reject::not_found` but to fail with `400` I need to `Ok(reply...`. I couldn't find a way of return an Err to return a 400, which would help dry and clean the code around this a little. Do you have any tips? We can also address that in a follow up.

PR #686, opened 2020-07-17
refactor(proxy): simplify http dependencies

xla:
This change introduces the context object that is passed to all filter chains in our http stack. Expect an improvement in ergonomics and ease of setting up new filters.


proxy/src/coco/peer.rs
```diff
@@ -32,316 +39,400 @@ pub struct UserRevisions {
...
+    pub fn init_project(
+        &self,
+        key: keys::SecretKey,
+        owner: &User,
+        path: impl AsRef<std::path::Path> + Send,
+        name: &str,
+        description: &str,
+        default_branch: &str,
+    ) -> Result<project::Project<entity::Draft>, error::Error> {
+        let api = self.peer_api.lock().expect("unable to acquire lock");
```
MeBrei, line 349: maybe this could be extracted as well, also to make sure that the error message is consistent and to avoid repetition.

proxy/src/http/org.rs
```diff
@@ -288,81 +258,74 @@ where
...
         Ok(reply::json(&project))
     }
 
     /// Get all projects under the given org id.
-    pub async fn get_projects<R>(
-        registry: http::Shared<R>,
-        peer: Arc<Mutex<coco::PeerApi>>,
-        org_id: String,
-    ) -> Result<impl Reply, Rejection>
+    pub async fn get_projects<R>(ctx: http::Ctx<R>, org_id: String) -> Result<impl Reply, Rejection>
```
FintanH, line 318: I wonder if we should turn all these `String`s into types while we're doing this refactor?

proxy/src/http/org.rs
```diff
@@ -892,23 +817,19 @@ mod test {
...
-            let (client, _) = radicle_registry_client::Client::new_emulator();
-            Arc::new(RwLock::new(registry::Registry::new(client)))
-        };
-        let subscriptions = notification::Subscriptions::default();
+        let ctx = http::Context::tmp(&tmp_dir).await?;
+        let api = super::filters(ctx.clone());
 
+        let ctx = ctx.read().await;
+        let key = ctx.keystore.get_librad_key()?;
+        let owner = ctx.peer_api.init_user(key.clone(), "cloudhead")?;
```
FintanH, line 825:
```suggestion
        let owner = ctx.peer_api.init_owner(key.clone(), "cloudhead")?;
```

proxy/src/http/source.rs
```diff
@@ -318,18 +306,17 @@ mod handler {
...
     where
-        R: registry::Client,
+        R: registry::Client + 'static,
     {
-        let registry = registry.read().await;
-        let store = store.read().await;
-        let session = session::current(Arc::clone(&api), &*registry, &store).await?;
+        let ctx = ctx.read().await;
+
+        let session = session::current(&ctx.peer_api, &ctx.registry, &ctx.store).await?;
```
FintanH, line 313: I think it might make more sense for us to pass the theme down into this function.

proxy/src/http/source.rs
```diff
@@ -95,12 +79,15 @@ where
...
+fn branches_filter<R>(
+    ctx: http::Ctx<R>,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone
+where
+    R: registry::Client + 'static,
+{
     path("branches")
         .and(warp::get())
-        .and(http::with_peer(peer))
+        .and(http::with_context(ctx))
```
FintanH, line 90: I wonder if it would make sense to still have something like `http::with_peer_api` that passes down the `Api` to the handler functions. That way they don't take any extra data they don't need. Would that be possible?

proxy/src/coco.rs
```diff
@@ -6,22 +6,13 @@ pub use librad::uri::RadUrn as Urn;
...
-/// Module that captures all the functions for working with `librad`'s [`PeerApi`].
+pub mod config;
+pub mod control;
 mod peer;
-pub use peer::{
-    create_peer_api, default_owner, get_project, get_user, init_owner, init_project, init_user,
-    list_projects, list_users, revisions, set_default_owner, verify_user, with_browser, PeerApi,
-    User, UserRevisions,
-};
+pub use peer::{verify_user, Api, User, UserRevisions};
```
sarahscott, line 12: 👏

proxy/src/coco/peer.rs
```diff
@@ -32,316 +39,400 @@ pub struct UserRevisions {
...
+        }
+    }
 
-        let user = repo.get_rad_self_of(peer_id.clone())?;
+    /// Set the default owner for this `PeerApi`.
+    ///
+    /// # Errors
+    ///
+    ///   * Fails to set the default `rad/self` for this `PeerApi`.
+    pub fn set_default_owner(&self, user: User) -> Result<(), error::Error> {
```
sarahscott, line 123: Does this only happen upon initialization?

proxy/src/http.rs
```diff
@@ -60,44 +60,27 @@ pub fn api<R>(
...
-        keystore,
-        Arc::clone(&registry),
-        Arc::clone(&store),
-    );
-    let session_filter =
-        session::routes(Arc::clone(&peer), Arc::clone(&registry), Arc::clone(&store));
-    let source_filter = source::routes(peer, Arc::clone(&registry), Arc::clone(&store));
-    let transaction_filter = transaction::filters(Arc::clone(&registry));
-    let user_filter = user::routes(registry, store, subscriptions);
+    let control_filter = control::routes(enable_control, ctx.clone());
```
sarahscott, line 74: so much nicer 💯

proxy/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 //! Proxy serving a specialized API to the Upstream UI.
 
-#![deny(missing_docs, unused_import_braces, unused_qualifications, warnings)]
```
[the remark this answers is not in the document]
xla: Leftover from local development, will revert.

proxy/src/http/user.rs
```diff
@@ -675,37 +607,24 @@ mod test {
...
-        );
+    async fn transfer() -> Result<(), error::Error> {
+        let tmp_dir = tempfile::tempdir()?;
+        let ctx = http::Context::tmp(&tmp_dir).await?;
+        let api = super::filters(ctx.clone());
+
         let author = radicle_registry_client::ed25519::Pair::from_legacy_string("//Alice", None);
         let handle = registry::Id::try_from("alice")?;
 
+        let ctx = ctx.write().await;
```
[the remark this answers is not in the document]
xla: That's right, changed. :>

proxy/src/coco/peer.rs
xla: Definitely blood. The issue with extraction anything sync related is that these lock mechanisms hinge on drop semantics. And it is already hard enough to debug all these cases where we deadlock at the moment. For now it's safer to handle locking explicitly in the method bodies. We could maybe introduce an error for the consistent error messaging.

proxy/src/coco/peer.rs
```diff
@@ -1,14 +1,21 @@
-use nonempty::NonEmpty;
-use serde::Serialize;
+//! Utility to work with the peer api of librad.
+
 use std::convert::TryFrom;
 use std::net::SocketAddr;
+use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
```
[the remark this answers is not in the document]
xla: Tokios `Mutex` is future based and therefore works with `await`, but I didn't wont to force all methods which want to access the peer_api to by async as it trickles through everywhere.

proxy/src/coco/peer.rs
```diff
@@ -32,316 +39,400 @@ pub struct UserRevisions {
...
-        // We only list projects that are owned by the peer
-        if rad_self.urn() != owner.urn() {
-            return None;
-        }
+    /// Convenience method to trigger a reopen of the storage.
+    ///
+    /// # Errors
+    ///
+    /// When the underlying lock acquisition fails or opening the storage.
+    pub fn reopen(&self) -> Result<(), error::Error> {
```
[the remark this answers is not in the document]
xla: Yeah in `coco::control`.

proxy/src/coco/peer.rs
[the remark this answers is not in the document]
FintanH: Fair enough :+1:

proxy/src/http/source.rs
xla: It is possible, but I don't think it's worth the effort at the moment.

FintanH: Word :ok_hand:

proxy/src/coco/peer.rs
xla: We use it in many place where set up state.

proxy/src/http/org.rs
xla: Would do that in a follow-up, there are more improvements to come.

proxy/src/http/source.rs
xla: You mean into the call to `session::current`? The theme comes from the settings which stored in the session. A bit confused what you proposing.

FintanH: Well, the thing was that I started asking myself why we needed a `Registry` here and then I realised we needed it for the session, which in turn made me wonder why needed the session. But really, what we need is a `theme`. Maybe it'd be better as a follow-up ticket in a similar vein to what I said about `with_peer_api` :)

proxy/src/http.rs
```diff
@@ -147,67 +128,91 @@ where
...
 }
 
-/// Thread-safe container for threadsafe pass-through to filters and handlers.
-pub type Shared<T> = Arc<RwLock<T>>;
+/// Container to pass down dependencies into HTTP filter chains.
+pub struct Context<R> {
+    /// [`coco::Api`] to operate on the local monorepo.
+    peer_api: coco::Api,
+    /// Storage to manage keys.
+    keystore: keystore::Keystorage,
```
rudolfs, line 163: Since the context gets cloned around a bunch, I wonder how this plays together with the security measures that I mentioned before?

proxy/src/coco/peer.rs
```diff
@@ -32,316 +39,401 @@ pub struct UserRevisions {
...
+
+        Ok(storage.metadata(urn)?)
+    }
+
+    /// Get a repo browser for a project.
+    ///
+    /// # Errors
+    ///
+    /// The function will result in an error if the mutex guard was poisoned. See
+    /// [`std::sync::Mutex::lock`] for further details.
```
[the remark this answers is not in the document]
FintanH: Tis accurate. `self.monorepo` locks and `self.get_project` locks.

proxy/src/coco/peer.rs
```diff
@@ -32,316 +39,401 @@ pub struct UserRevisions {
...
+
+    /// Initialize a [`project::Project`] that is owned by the `owner`.
+    /// This kicks off the history of the project, tracked by `librad`'s mono-repo.
+    ///
+    /// # Errors
+    ///
+    /// Will error if:
+    ///     * The signing of the project metadata fails.
+    ///     * The interaction with `librad` [`librad::git::storage::Storage`] fails.
+    #[allow(clippy::needless_pass_by_value)] // We don't want to keep `SecretKey` in memory.
```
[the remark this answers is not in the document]
FintanH: Ideally, we should actually be using the `Signer` trait (see https://github.com/radicle-dev/radicle-link/issues/216). This will delegate the fetch of the secret key from the store and `link` will do it's thing.

So this owned key shenanigans was probably overly cautious in retrospect.

proxy/src/coco/peer.rs
```diff
@@ -32,316 +39,401 @@ pub struct UserRevisions {
...
+    pub fn init_project(
+        &self,
+        key: &keys::SecretKey,
```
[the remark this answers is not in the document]
FintanH: Answered in https://github.com/radicle-dev/radicle-upstream/pull/686/files#r458688223

proxy/src/http.rs
FintanH: `Keystorage` is fine. It's just a handle to the file system and unencrypts the key from disk when it needs the key. Again, the second part of the equation is using this https://github.com/radicle-dev/radicle-upstream/pull/686/files#r458688223

proxy/src/coco/peer.rs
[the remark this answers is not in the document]
xla: #704

PR #688, opened 2020-07-17
test(ui): add repository selector integration tests

rudolfs:
Adds tests covering different peer selection in repository selector.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/289.

### DIFF
--- a/proxy/src/coco/control.rs
+++ b/proxy/src/coco/control.rs
@@ -8,7 +8,7 @@ use librad::net::peer::PeerApi;
 use radicle_surf::vcs::git::git2;
 
 use crate::coco::config;
-use crate::coco::peer::{init_project, User};
+use crate::coco::peer::{init_project, init_user, User};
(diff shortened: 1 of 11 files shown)


proxy/src/coco/control.rs
```diff
@@ -71,7 +71,7 @@ pub fn setup_fixtures(
 /// the coco project.
 pub fn replicate_platinum(
     peer: &PeerApi,
-    key: keys::SecretKey,
+    key: &keys::SecretKey,
```
FintanH, line 74: I try and keep `SecretKey` as owned because we don't want to keep in memory.

rudolfs: Is that a security measure, if so, does it make sense since these functions should only be run by tests?

FintanH: Ya, fair :+1:

Issue #684, opened 2020-07-17
Show the actual account id on the wallet page


MeBrei:
#658 exposed the account ids for orgs and users in the responses. Once #662 is merged, we should show the actual account id in the `Receive` tab.

NunoAlexandre:
Just a note that we should be using the SS58 encoding format to display account ids. That's the advised procedure, used in the Registry CLI and the expected format in the new "accounts" endpoint as well.

MeBrei:
👍 I believe the account ids are already returned in the correct format by the user and org endpoints

PR #691, opened 2020-07-17
refactor(ui): show actual org accountID

juliendonck:
Shows the actual accountId for orgs

Screenshot 2020-07-17 at 17 04 31

*Tests will be written at a later stage*


ui/Screen/Org/Fund.svelte
```diff
@@ -13,9 +14,11 @@
   };
 </script>
```
sarahscott, line 15:
```suggestion
const accountId = $org.accountId;
</script>
```

ui/Screen/Org/Fund.svelte
```diff
@@ -13,9 +14,11 @@
...
-  <Wallet
-    transactions={orgTransactions(transactions)}
-    balance="3484"
-    address="hyda4fhdx8up8hhocagkdz3d41txb98stw4pkqe3uommo1p5m6s9oy" />
+<Remote store={txStore} let:data={transactions}>
+  <Remote store={orgStore} let:data={org}>
+    <Wallet
+      transactions={orgTransactions(transactions)}
+      balance="3484"
+      address={org.accountId} />
```
sarahscott, line 22:
```suggestion
      address={accountId} />
```

ui/Screen/Org/Fund.svelte
```diff
@@ -13,9 +14,11 @@
   };
 </script>
 
-<Remote {store} let:data={transactions}>
-  <Wallet
-    transactions={orgTransactions(transactions)}
-    balance="3484"
-    address="hyda4fhdx8up8hhocagkdz3d41txb98stw4pkqe3uommo1p5m6s9oy" />
+<Remote store={txStore} let:data={transactions}>
+  <Remote store={orgStore} let:data={org}>
```
sarahscott, line 18:
```suggestion
```

ui/Screen/Org/Fund.svelte
```diff
@@ -13,9 +14,11 @@
...
+    <Wallet
+      transactions={orgTransactions(transactions)}
+      balance="3484"
+      address={org.accountId} />
+  </Remote>
```
sarahscott, line 23:
```suggestion
```

ui/Screen/Org/Fund.svelte
```diff
@@ -1,5 +1,6 @@
 <script>
-  import { transactions as store } from "../../src/transaction.ts";
+  import { transactions as txStore } from "../../src/transaction.ts";
+  import { org as orgStore } from "../../src/org.ts";
   import { Wallet, Remote } from "../../DesignSystem/Component";
 
   export let params = null;
```
sarahscott, line 6: It'd be nice to move the `orgTransactions` function into either `org.ts` or `transaction.ts` so it can be typesafe & testable.

ui/Screen/Org/Fund.svelte
```diff
@@ -13,9 +14,11 @@
   };
 </script>
 
-<Remote {store} let:data={transactions}>
-  <Wallet
-    transactions={orgTransactions(transactions)}
-    balance="3484"
-    address="hyda4fhdx8up8hhocagkdz3d41txb98stw4pkqe3uommo1p5m6s9oy" />
+<Remote store={txStore} let:data={transactions}>
```
sarahscott, line 17: The big benefit of the `Remote` component is simplicity, which we lose if we nest them. Unless you're doing separate loading states for each of these entities, it makes more sense to use the reactive store for single values within `<script>`.

ui/Screen/Org/Fund.svelte
```diff
@@ -13,9 +14,11 @@
...
+  <Remote store={orgStore} let:data={org}>
+    <Wallet
+      transactions={orgTransactions(transactions)}
+      balance="3484"
```
sarahscott, line 21: Why isn't `balance` a number?

juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/691/commits/2b65cde4b2b5ab138d3aa919b9688d8144626f4e

ui/Screen/Org/Fund.svelte
juliendonck: Good point. https://github.com/radicle-dev/radicle-upstream/pull/691/commits/3899cd8c8fdda44ba868d618a78a9a20c485f9cc

ui/Screen/Org/Fund.svelte
juliendonck: Yes good catch. This has been done here: https://github.com/radicle-dev/radicle-upstream/pull/691/commits/3899cd8c8fdda44ba868d618a78a9a20c485f9cc
I also think i'll improve it by not having it match a string, as you mentioned in an other review.

ui/Screen/Org/Fund.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/691/commits/3899cd8c8fdda44ba868d618a78a9a20c485f9cc

ui/Screen/Org/Fund.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/691/commits/3899cd8c8fdda44ba868d618a78a9a20c485f9cc

ui/Screen/Org/Fund.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/691/commits/3899cd8c8fdda44ba868d618a78a9a20c485f9cc

ui/Screen/Org/Fund.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/691/commits/3899cd8c8fdda44ba868d618a78a9a20c485f9cc

ui/Screen/Org/Fund.svelte
```diff
@@ -1,21 +1,15 @@
...
   export let params = null;
-
-  const orgTransactions = transactions => {
-    return transactions.filter(tx => {
-      return "domainId" in tx.messages[0]
-        ? tx.messages[0].domainId == params.id
-        : null;
-    });
-  };
+  const accountId = $org.data.accountId;
```
xla, line 7:
```suggestion
  $: accountId = $org.data.accountId;
```

Shouldn't this be reactively assigned to be up to date with the value in the store?

ui/Screen/Org/Fund.svelte
```diff
@@ -1,21 +1,15 @@
...
 
-<Remote {store} let:data={transactions}>
+<Remote store={txStore} let:data={transactions}>
   <Wallet
-    transactions={orgTransactions(transactions)}
-    balance="3484"
-    address="hyda4fhdx8up8hhocagkdz3d41txb98stw4pkqe3uommo1p5m6s9oy" />
+    transactions={orgTransactions(transactions, params.id)}
+    balance={3484}
+    address={accountId} />
```
xla, line 14: Should we adjust the param of the `Wallet` component if we now refer to it as account id?

ui/src/org.ts
```diff
@@ -172,3 +173,12 @@ export const memberHandleValidationStore = (
     },
   ]);
 };
+
+export const orgTransactions = (
+  transactions: [transaction.Transaction],
+  id: string
+): [transaction.Transaction] | transaction.Transaction[] => {
```
xla, line 180: Such type system skills! :heart:

xla:
@juliendonck Does this close #684?

ui/Screen/Org/Fund.svelte
NunoAlexandre: yeah we should uniformalize the terminologies. An address, technically speaking, is the account id encoded in a specific way. We use the SS58 format as the address. The `accountId` value here is actually the address-encoded-like representation of the entity's account id, so we should just have it called `address` in throughout the whole UI. since we might need to do this in other places, I'd be ok with sticking to this for now and then make that change in one go in a follow up PR.

ui/Screen/Org/Fund.svelte
juliendonck: Nah this creates an infinite loop cause the component is already contained in a reactive fetch in the `org` component

ui/Screen/Org/Fund.svelte
juliendonck: I think the `params.id` is actually the orgId, not the accountId

juliendonck: ok

ui/src/org.ts
juliendonck: @rudolfs helped, I must admit :)

juliendonck:
> @juliendonck Does this close #684?

@xla only for the org wallet, not yet for the user wallet.

ui/Screen/Profile/Wallet.svelte
```diff
@@ -12,6 +12,6 @@
 <Remote {store} let:data={transactions}>
   <Wallet
     transactions={userTransactions(transactions)}
-    balance="374"
+    balance={3552}
```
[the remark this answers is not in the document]
juliendonck: The linter automatically wraps in in quotes if I don't wrap it in curly brackets.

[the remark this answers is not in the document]
NunoAlexandre: TIL.

ui/Screen/Org/Fund.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/691/commits/be8cf18abc7416c9281e19c1a5e92c062bfa71a9

ui/Screen/Org/Fund.svelte
juliendonck: No strings! https://github.com/radicle-dev/radicle-upstream/pull/691/commits/29a7c250a0a1f39cd8246c3e43af52d528730de0
🧨

Issue #698, opened 2020-07-21
Can't register project in org even if a project exists.


juliendonck:
### video
Check out the [video of the bug](https://res.cloudinary.com/juliendonck/video/upload/v1595322794/register_project_bug_m7npzi.mov)

### steps to reproduce:

1. create identity
2. register identity
3. register org
4. create project under registered user
5. go to org and try to register project.

6. if you refresh the app it works

### expected behavior

I can register an the project under the org right after I create it.

### possible solution

The button looks at `session.permissions.registerProject` to define the state. And when I refresh the app the session updates. So I'm assuming that if we refresh the session after project creation this might resolve this issue.

PR #699, opened 2020-07-21
fix(ui): project registration button permissions

rudolfs:
Closes: #698.


cypress/integration/project_creation.spec.js
```diff
@@ -60,7 +60,6 @@ const withPlatinumStub = callback => {
 beforeEach(() => {
   cy.nukeAllState();
   cy.createIdentity();
-  cy.createProjectWithFixture();
```
rudolfs, line 63: This was here because proxy would error out if we didn't have a project after nuking all state previously. It is now fixed and this workaround is not needed anymore.

PR #700, opened 2020-07-21
feat: Integrate registration fee and drop deposits

NunoAlexandre:
The integration for https://github.com/radicle-dev/radicle-registry/pull/600/commits/23699c95636f54a6d334e34c9412e029ed97f698

**Notes:** 

- The new registration fee drops the placeholder deposit and is only applicable to org and user registration
  The UI and the tests have been adjusted to reflect this accordingly :heavy_check_mark:


proxy/src/registry.rs
```diff
@@ -11,7 +11,7 @@ use std::str::FromStr;
 use radicle_registry_client::{self as protocol, ClientT, CryptoPair};
 pub use radicle_registry_client::{
     parse_ss58_address, AccountId, Balance, BlockHash, Id, IdStatus, ProjectDomain, ProjectName,
-    MINIMUM_FEE,
+    MINIMUM_TX_FEE, REGISTRATION_FEE,
```
MeBrei, line 14: Maybe `HANDLE_REGISTRATION_FEE` would be more descriptive, as we also call the transaction a "registration" for projects and members.

proxy/src/session.rs
```diff
@@ -27,8 +27,8 @@ pub struct Session {
     pub permissions: Permissions,
     /// User controlled parameters to control the behaviour and state of the application.
     pub settings: settings::Settings,
-    /// Transaction deposits.
-    pub transaction_deposits: registry::Deposits,
+    /// Registration fee for user and org registration in the Registry.
+    pub registration_fee: registry::Balance,
```
MeBrei, line 31: Same as above

ui/DesignSystem/Component/Transaction.svelte
```diff
@@ -36,35 +36,42 @@
...
 
-  <div slot="right">
-    <Rad rad={summary.depositRad} usd={summary.depositUsd} variant="deposit" />
-  </div>
-</Row>
+    <div slot="right">
+      <Rad
+        rad={summary.registrationFee.rad}
+        usd={summary.registrationFee.usd}
+        variant="deposit" />
```
MeBrei, line 61: Should we rename the `variant` now?

ui/Screen/Org/MemberRegistration.svelte
```diff
@@ -99,7 +99,7 @@
           {transaction}
           {subject}
           {payer}
-          transactionDeposits={$session.data.transactionDeposits} />
+          registrationFee={$session.data.registrationFee} />
```
MeBrei, line 102: I thought the registration fee is only for org and user registration 🤔

NunoAlexandre: I oversaw this one! Will fix it.

ui/DesignSystem/Component/Transaction.svelte
NunoAlexandre: hm I don't think so. This refers to the field representing a credit or a deposit.

juliendonck: Yes, we should remove it from the `<Rad />` component. And I think I'd add a total to the transaction view now, like this:
section amounts

juliendonck: @brandonhaslegs what do you think?

proxy/src/registry.rs
NunoAlexandre: I feel that. `HANDLE_REGISTRATION_FEE` is not appropriate since it's not common language in the registry neither for orgs. Since we call users and orgs "domains" we could call it `DOMAIN_REGISTRATION_FEE`? 

Having this said I would prefer to keep it called this way, shall we broaden the scope of this fee. I am fine with the notion that the registration fee only applies to user and org registration.

What do you think?

ui/DesignSystem/Component/Transaction.svelte
NunoAlexandre: I can drop the variants, @juliendonck. Could you address the UI change in a separate PR?

proxy/src/registry.rs
MeBrei: I am willing to accept the name, but what might actually be the problem here is that in the current implementation, the ui "decides" when to show the registration fee. I would think this is a registry or proxy concern. SO maybe we could actually keep the way `deposits` were split up by transaction type and this way pass to the ui which transaction types add a registration fee. This preserves being flexible for future adjustments.

NunoAlexandre: > I am willing to accept the name, but what might actually be the problem here is that in the current implementation, the ui "decides" when to show the registration fee. I would think this is a registry or proxy concern. SO maybe we could actually keep the way `deposits` were split up by transaction type and this way pass to the ui which transaction types add a registration fee. This preserves being flexible for future adjustments.

This is a very good observation. I will give it some thought, I think I have an idea already.

ui/DesignSystem/Component/Transaction.svelte
juliendonck: sure, if you create an issue so I don't forget.

proxy/src/registry.rs
NunoAlexandre: Let's quickly give a step back.

### Considerations

1. How do we adjust to changes to the scope of the registration fee?
    say, we want project registration to incur in the registration fee
2. How do we adjust in case we need a specific registration fee value per type of transaction?

We should keep these in mind when evaluating the current solution and different suggestions, though we do not see any changes in this field coming up.

### Now (Option 0)

- The Registry exposes `REGISTRATION_FEE`, a fee that is withdrawn when registering users and org (not projects or adding members to an org).
- Proxy provides `registrationFee` to the `ui` through `session`
- each registration screen (`UserRegistration`, `OrgRegistration`, `MemberAddition`, `ProjectRegistration` passes this universal `registrationFee` value to the `Transaction` component
- `Transaction` simply delegates the received universal `registrationFee` value to the `costSummary` function
- `costSummary` looks at the received transaction/message and decides to include the `registrationFee` in the returned `CostSummary` or not based on the message.
- `Transaction` shows or hides the `registration-fee` `Row` based on `registrationFee` being defined in the `summary`

**Regarding the considerations**

1. We would only need to change the `paysRegistrationFee`
2. We would need to change what we pass from the registration screens down to the `Transaction` component, possibly a struct with the breakdown of the fee value per tx type, as we had for the deposits. Then, in `costSummary`, we would look up the appropriate value. 
    - This is not something we see coming at all, to keep in mind.

**Observations**:

- Kind of confusing that we are passing the `registrationFee` value from the registration screens down to the `Transaction` component even for screens that don't incur into that fee.

### Option A

- (same) Registry exposes `REGISTRATION_FEE`, a fee that is withdrawn when registering users and org (not projects or adding members to an org).
- (same) Proxy provides `registrationFee` to the `ui` through `session`
- (**change**) each registration screen (`UserRegistration`, `OrgRegistration`, `MemberAddition`, `ProjectRegistration`  pass `session` to the `Transaction` component. `Transaction` then looks up the session's `registrationFee` value, passing it to `costSummary`
- (same)`costSummary` looks at the received transaction/message and decides to include the `registrationFee` in the returned `CostSummary` or not based on the message.
- (same) `Transaction` shows or hides the `registration-fee` `Row` based on `registrationFee` being defined in the `summary`

**Regarding the considerations**

1. We would only need to change the `paysRegistrationFee`
2. Same as for `Option 0`

**What does this improve?**

- Drops the confusion reported in `Option 0` where registration screens pass the `registrationFee` value even when unsuited.

### Option B

- (same) Registry exposes `REGISTRATION_FEE`, a fee that is withdrawn when registering users and org (not projects or adding members to an org).
- (**change**) Proxy provides `registrationFee` **struct** to the `ui` through `session`:

    ```rust
    struct RegistrationFee {
      user: Option<Balance>,
      org: Option<Balance>,
      memberAddition: Option<Balance>,
      project: Option<Balance>,
    }
    ```

As of now, it would be built as:

```rust
RegistrationFee {
  user: Some(registry::REGISTRATION_FEE),
  org: Some(registry::REGISTRATION_FEE),
  memberAddition: None,
  project: None,
}
```

- (**change**) each registration screen (`UserRegistration`, `OrgRegistration`, `MemberAddition`, `ProjectRegistration` passes their respective entry in the above `RegistrationFee` struct to the `Transaction` component, which represents whether that registration incurs in a registration fee or not and how much it is if yes.
- (**change**) `costSummary` looks at the received transaction/message and decides to include the `registrationFee` in the returned `CostSummary` based on the `registrationFee?` param
- (same) `Transaction` shows or hides the `registration-fee` `Row` based on `registrationFee` being defined in the `summary`

**Regarding the considerations**

1. We would only need to change the appropriate `RegistrationFee` instance field values in the proxy
2. We would need to expose more granular values from the Registry for each tx type and instantiate the `RegistrationFee` struct accordingly in the proxy. The UI is left untouched. 

**What does this improve?**

- It reduces the confusion around a registration screen passing the universal `registrationFee` value to `Transaction` even when the said screen does not incur in the registration fee.
- Injects data dependencies top-down
- Super flexible

- **Issues**
    - With this setup, we offer more flexibility than needed, which can be confusing and anti-YAGNI
    - making the view over which screens incur in the registration fee more implicit and spread (vs having a `paysRegistrationFee` that unifies it).
    - A screen can pass the wrong registration fee value down to `Transaction` (i.e, while having it unified in `paysRegistrationFee` ensures one source of truth)

### Conclusion

All considered I think that `Option A` would solve the confusion around screens that don't incur in the registration fee passing _the_ registration fee value down to `Transaction` without creating unnecessary machinery.

I believe `Option B` is what you (@MeBrei) were suggesting. I am ok with moving there, though I don't think it's a straightforward trade-off as it introduces more code that does not serve any current or foreseeable use case.

Let me know what you think :)

ui/Screen/Org/MemberRegistration.svelte
NunoAlexandre: Discussion moved to https://github.com/radicle-dev/radicle-upstream/pull/700#discussion_r458916091

NunoAlexandre:
@xla could you have a look as well? We are stuck at https://github.com/radicle-dev/radicle-upstream/pull/700#discussion_r458916091, would be great to have your fresh view on it :ringed_planet:

ui/DesignSystem/Component/Transaction.svelte
NunoAlexandre: just did it all :+1:

proxy/src/registry.rs
MeBrei: Thank you for the thorough comparison!

I would still prefer `Option B` as I think the ui should not have to have "insider" knowledge how transactions are handled in the registry. More specifically, if we ever want to show "historic" transactions, the registration fee should actually be part of the transaction data, as it might change over time.

BUT as a first iteration I think the current implementation could be merged to keep things moving

NunoAlexandre:
@xla @MeBrei so I have implemented option B with the addition of actually storing the registration fee in the transaction itself for the purpose of having it recorded and available to be shown in the history of the transaction.

There are two problems:

1. I need to figure out how to (de)serialize an `Optional<Balance>` value

2. The way things are set up, when we submit a transaction in the proxy > registry to the actual registry, we _then_ build `Transaction::confirmed`, which now expects an optional registration fee value. I made it pass the `REGISTRATION_FEE` we have on the linked radicle client dependency. In principle, that's fine because if the radicle client library is incompatible with the connected network's runtime, the call will fail in the first place, which _should_ guarantee that the `REGISTRATION_FEE` of the two counterparts is the same. **However**, it feels to me like an implicit and dangerous reliance. Shall that guarantee fail for some reason, we have a silently buggy system at a critical level (payments), where either our users will be charged more than they should have or they will see a withdrawn registration fee reported which really isn't the one that was actually withdrawn. The only sensible way I see forward is to have the runtime reporting back the registration it has (or not) withdrawn for a given transaction, to have a single source of truth. I need to think deeper about this.

NunoAlexandre:
@xla @MeBrei I have fixed point 1 above. For point 2, I think that now we can rely on a combination of a) knowing that the registry client we link to is guaranteed to be compatible with the registry it connects to and b) new unit tests I am adding that test the fees withdrew from an account when a tx is submitted.

I would like us to move forward with this revision and in the meantime I will think of a stricter solution that makes us less reliant on those two things for correctness.

NunoAlexandre:
@xla @MeBrei This PR is ready.

I have fixed point 1 above. Regarding point 2, for now, we can rely on a combination of:
   a) knowing that the registry client we link to is guaranteed to be compatible with the registry it connects to,
       meaning that no semantic differences can take place.
   b) extended unit tests that now check the fees withdrawn from an account when a tx is submitted. 
       Shall we bump to a newer version of the registry client that connects to a runtime with different semantics,
       the tests will fail and we know what to do.

I would like us to move forward with this revision and in the meantime I will think of a stricter solution that makes us less reliant on those two things for correctness. Until we connect to a real, public network, this is not a priority.

NunoAlexandre:
@xla @MeBrei This PR is ready.

I have fixed point 1 above. Regarding point 2, for now, we can rely on a combination of:
   
   - a) knowing that the registry client we link to is guaranteed to be compatible with the registry it connects to,
       meaning that no semantic differences can take place.
   - b) extended unit tests that now check the fees withdrawn from an account when a tx is submitted. 
       Shall we bump to a newer version of the registry client that connects to a runtime with different semantics,
       the tests will fail and we know what to do.

I would like us to move forward with this revision and in the meantime I will think of a stricter solution that makes us less reliant on those two things for correctness. Until we connect to a real, public network, this is not a priority.

ui/src/transaction.ts
```diff
@@ -542,55 +536,43 @@ export const summaryText = (counts: SummaryCounts): string => {
...
-    default:
-      throw `MessageType: ${messageType} not implemented`;
-      break;
-  }
+const paysRegistrationFee = (messageType: MessageType): boolean => {
+  return (
+    messageType === MessageType.OrgRegistration ||
+    messageType === MessageType.UserRegistration
+  );
+};
```
MeBrei, line 549: `paysRegistrationFee` should be decided on the values of `session.registrationFee`.

NunoAlexandre: forgot to delete that, it's no longer used!

Issue #689, opened 2020-07-17
Fix flaky integration test


rudolfs:
Keeps popping up on master:

![commit browsing -- commit history -- shows the commit history for the default branch (failed)](https://user-images.githubusercontent.com/158411/87794809-789cda00-c847-11ea-8a63-5a12e16c0c8b.png)

PR #702, opened 2020-07-22
fix(ui): race condition in commit navigation

rudolfs:
Turns out the test was not flaky, but the `navigate_on_ready` function had a race condition.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/689.

### DIFF
--- a/ui/Screen/Project/Source.svelte
+++ b/ui/Screen/Project/Source.svelte
@@ -5,6 +5,7 @@
 
   import * as path from "../../src/path.ts";
   import { project as projectStore } from "../../src/project.ts";
+  import * as remote from "../../src/remote.ts";
   import {
     fetchCommits,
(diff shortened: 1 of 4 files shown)


ui/Screen/Project/Source.svelte
```diff
@@ -71,13 +72,21 @@
     );
   };
 
+  let unsubscribe;
```
xla, line 75: We should leave a comment and/or a follow-up issue that we want to address this concern generally in the app by ways of encoding it in or navigation/routing infrastructure.

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/702/commits/082a9b637f19dbe62018a7e15c873f04c7b3a3aa

PR #703, opened 2020-07-22
refactor(proxy): dry submit transactions

### LINKED ISSUE #672: DRY registry messages
While reviewing some code in proxy, I came across some code that has the same pattern. Two examples of which I've pointed out here:
* https://github.com/radicle-dev/radicle-upstream/pull/653#discussion_r454294179
* https://github.com/radicle-dev/radicle-upstream/pull/670#discussion_r454925229

In the first link, I documented what I tried to do. In the end, I couldn't massage the type system to accept my shenanigans. I'd be interested in finding a way to clean this duplication up. Ideally, it would just be proxy oriented, but maybe we need to tweak something in the registry to make it work.

geigerzaehler:
We extract common logic into `Client::submit_transaction`. Closes #672

### DIFF
--- a/proxy/src/registry.rs
+++ b/proxy/src/registry.rs
@@ -389,14 +389,32 @@ impl Registry {
         fee: Balance,
     ) -> Result<protocol::Transaction<M>, error::Error> {
         let nonce = self.client.account_nonce(&author.public()).await?;
-        let runtime_spec_version = self.client.runtime_version().await?.spec_version;
+        let runtime_transaction_version = self.client.runtime_version().await?.transaction_version;
         let extra = protocol::TransactionExtra {
             nonce,
             genesis_hash: self.client.genesis_hash(),
+            runtime_transaction_version,
             fee,
-            runtime_transaction_version: runtime_spec_version,
         };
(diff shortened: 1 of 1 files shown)


xla:
@geigerzaehler Does this close #672?

proxy/src/registry.rs
```diff
@@ -389,15 +389,33 @@ impl Registry {
...
             fee,
-            runtime_transaction_version: runtime_spec_version,
         };
         Ok(protocol::Transaction::new_signed(author, message, extra))
     }
+
+    /// Sign a message, submit it to the chain and wait for it to be confirmed. Return the
+    /// transaction hash and the number of the block it was included in.
+    ///
+    /// Errors with [`error::Error::Runtime`] if applying the transction errors.
```
xla, line 405:
```suggestion
    /// # Errors
    ///
    /// Fails with [`error::Error::Runtime`] if applying the transction errors.
```

geigerzaehler:
> @geigerzaehler Does this close #672?

Depends on whether @FintanH see this as sufficient.

PR #705, opened 2020-07-22
feat(proxy): expose the account id in session

### LINKED ISSUE #690: Expose the account id in the session
Currently the account id is exposed in `User` and `Org`, but it is possible to have an account id and receive funds without being a registered user. Therefor the account id should be exposed as well outside of `User` in `Session`.

This blocks the profile wallet side of #684 and #685

MeBrei:
Closes #690


MeBrei:
> Everything looks good, but I think we should store the `address` format of the public key instead of the raw account id. The `accounts` API operates on the `address` format and wouldn't work with this `account_id` value (worth double-checking).
> 
> So my suggestion would be add `address` or `wallet_id`, which would be obtained in the proxy with `ed25519::Pair::from_legacy_string("//Alice", None).public().to_string()`, i.e, the SS58-formated account id.
> 
> Let me have your thoughts!

@NunoAlexandre Thank you for the feedback! I tested it in the API playground and I was able to take the account id that is returned in the session and use it successfully in the Account endpoints.

proxy/src/identity.rs
```diff
@@ -13,6 +13,8 @@ use crate::registry;
 
 pub use shared_identifier::SharedIdentifier;
 
+use radicle_registry_client::{ed25519, CryptoPair};
```
xla, line 16: This should go up next to the librad imports.

PR #709, opened 2020-07-22
chore(ui): update tools to support typescript

xla:
Most recently [typescript support](https://svelte.dev/blog/svelte-and-typescript) landed in Svelte. Along with it the consolidation of supporting tools. This change-set brings all these deps to the latest version and enables us to slowly transition across the codebase.


ui/DesignSystem/Component/Placeholder.svelte
```diff
@@ -2,7 +2,7 @@
   export let style = null;
 </script>
 
-<style>
+<style type="text/css">
```
sarahscott, line 5: why is this changed here and not elsewhere?

xla: Oops, this is a debug leftover. Gonna remove.

PR #712, opened 2020-07-23
feat(ui): send funds flow

juliendonck:
Flow that allows users to transfer money.

Screenshot 2020-07-23 at 11 28 56

### Todo:

- [x] validations @MeBrei 
- [x] submit as transaction @NunoAlexandre 
    - [x]  From user accounts
    - [x]  From org accounts


ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,14 @@
+<script>
+  import { pop } from "svelte-spa-router";
+  import { get } from "svelte/store";
+
+  import { fromStore, toStore, amountStore } from "../src/transfer.ts";
+
+  const fromAddress = get(fromStore);
```
sarahscott, line 7: The `get` helper is really only supposed to be used in odd cases like testing or when we can't subscribe to a store (see the [svelte docs on `get`](https://svelte.dev/docs#get)). For svelte components, you should subscribe to stores like this: `$fromStore`.

MeBrei: @sarahscott Thank you for pointing it out! I tried to address it in f32a9d5, let me know if it needs further change

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,256 @@
...
+    Input,
+    Text,
+    Title,
+  } from "../DesignSystem/Primitive";
+
+  const { identity } = getContext("session");
+  const { orgs } = getContext("session");
+
+  let fromAddress;
+  fromStore.subscribe(value => {
```
sarahscott, line 20: You can actually just delete these and subscribe to the values automatically when you need to access them, using `$`: https://svelte.dev/tutorial/auto-subscriptions

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,256 @@
...
+        <div slot="left">
+          <Text
+            variant="regular"
+            style="color:var(--color-foreground-level-6);">
+            Amount
+          </Text>
+        </div>
+
+        <div slot="right">
+          <Rad rad={amount} />
```
sarahscott, line 196:
```suggestion
          <Rad rad={$amountStore} />
```

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,256 @@
...
+        title: orgs[i].id,
+        avatarFallback: orgs[i].avatarFallback,
+        imageUrl: null,
+      },
+    });
+  }
+
+  let step = 1;
+
+  const nextStep = () => {
```
sarahscott, line 58: Instead of using magic steps, we can use enumerated states so it's easier to reason about:
```
step === TransactionSubmission
```` 

means more than 
```
step === 1
```

Check out some of the other flows like https://github.com/radicle-dev/radicle-upstream/blob/master/ui/Screen/OrgRegistration.svelte for examples

File: ui/Screen/SendFunds.svelte (hunk 3 above)
Comment (MeBrei) [reply ui/Screen/SendFunds.svelte:196]: Will this work if the amount get's changed in the form?

Comment (sarahscott) [reply ui/Screen/SendFunds.svelte:196]: Yeah it's reactive so if the component is subscribed to it, it should update automatically. The `$` prefix indicates the subscription. Happy to look through this and push an example commit if that would be helpful 😊

File: ui/Screen/SendFunds.svelte (hunk 4 above)
Comment (juliendonck) [reply ui/Screen/SendFunds.svelte:63]: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/e2d6bce27f3b132de219c3eaf121c97744e9bbac

File: ui/Screen/SendFunds.svelte (hunk 3 above)
Comment (juliendonck) [reply ui/Screen/SendFunds.svelte]: Nice one. Wondering if we should rename the variables. https://github.com/radicle-dev/radicle-upstream/pull/712/commits/059b135f36003ee31ca3b3d725e96732a61fe2bc

File: ui/Screen/SendFunds.svelte (hunk 2 above)
Comment (juliendonck) [reply ui/Screen/SendFunds.svelte]: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/059b135f36003ee31ca3b3d725e96732a61fe2bc

**Comment by juliendonck** [comment]
@MeBrei I responded to @sarahscott's comments and finished the UI. I added some todo's above that I'd like your help with. Let me know if it makes sense.

File: proxy/src/http/user.rs [hunk 5]
```diff
@@ -254,7 +254,7 @@ mod handler {
     /// Transfer funds to the given `recipient`.
     pub async fn transfer<R>(
         ctx: http::Ctx<R>,
-        _handle: registry::Id,
+        _handle: String,
```
sarahscott, line 257: Why are we changing this from a `registry::Id` to a `String`?

ui/Screen/SendFunds.svelte
sarahscott: Just pushed a commit with what I think is a svelty way to do the reactivity. Let me know if you have any questions!

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,258 @@
...
+        title: orgs[i].id,
+        avatarFallback: orgs[i].avatarFallback,
+        imageUrl: null,
+      },
+    });
+  }
+
+  let state = TransferState.Preparation;
+
+  const nextStep = () => {
```
sarahscott, line 55:
```suggestion
  const reviewTx = () => {
```

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,258 @@
...
+      <Title style="color: var(--color-foreground-level-6); padding: 0.5rem;">
+        From
+      </Title>
+      <!-- TODO(julien): shouldn't identity.accountId be the same as fromAddress -->
+      <Dropdown
+        placeholder="Select wallet you want to use"
+        bind:value={payer}
+        options={dropdownOptions} />
+      <div class="submit">
+        <Button on:click={() => nextStep()}>Review transfer</Button>
```
sarahscott, line 159:
```suggestion
        <Button on:click={() => reviewTx()}>Review transfer</Button>
```

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,258 @@
...
+    });
+  }
+
+  let state = TransferState.Preparation;
+
+  const nextStep = () => {
+    state = TransferState.Confirmation;
+  };
+
+  const previousStep = () => {
```
sarahscott, line 59:
```suggestion
  const cancel = () => {
```

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,258 @@
...
+                size="small"
+                title={option.avatarProps.title}
+                variant={option.avatarProps.variant}
+                avatarFallback={option.avatarProps.avatarFallback} />
+            {/if}
+          {/each}
+        </div>
+      </Row>
+      <div class="submit">
+        <Button variant="transparent" on:click={() => previousStep()}>
```
sarahscott, line 251:
```suggestion
        <Button variant="transparent" on:click={cancel}>
```

ui/Screen/SendFunds.svelte
sarahscott: or `submit`; anything that suggests we're submitting a form

ui/Screen/SendFunds.svelte
sarahscott: `back` is maybe fine too

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,248 @@
...
+      variant: "avatar",
+      value: orgs[i].accountId,
+      avatarProps: {
+        variant: "square",
+        title: orgs[i].id,
+        avatarFallback: orgs[i].avatarFallback,
+        imageUrl: null,
+      },
+    });
+  }
```
sarahscott, line 51:
```suggestion
const dropdownOptions = [
    {
      variant: "avatar",
      value: identity.accountId,
      avatarProps: {
        variant: "circle",
        title: identity.metadata.handle,
        avatarFallback: identity.avatarFallback,
        imageUrl: identity.imageUrl,
      },
    },
    ...orgs.map(org => ({
      variant: "avatar",
      value: org.accountId,
      avatarProps: {
        variant: "square",
        title: org.id,
        avatarFallback: org.avatarFallback,
        imageUrl: null,
      },
    })),
  ];
```

sarahscott: We're doing this in a few different places now; it might make sense to create a `payers`/`wallets` function in `transaction.ts` to get a list of all possible wallets. That might be beyond the scope of this PR but I think it'd be a lot easier than copying and pasting this every time the user wants to pay for something.

It could even be a `proxy` query for available wallets, and then `transaction.ts` just formats for display.

ui/DesignSystem/Component/Wallet/SendReceive.svelte
```diff
@@ -110,7 +125,7 @@
         placeholder="Optional message"
         style="flex: 1; padding-bottom: 1rem;" /> -->
       <div class="submit">
-        <Button disabled={true}>Send transaction</Button>
+        <Button on:click={() => openSendModal()}>Send transaction</Button>
```
sarahscott, line 131:
```suggestion
        <Button on:click={openSendModal}>Send transaction</Button>
```

ui/DesignSystem/Component/Wallet/SendReceive.svelte
```diff
@@ -1,8 +1,21 @@
...
+
   import { Button, Input, Icon, Title } from "../../Primitive";
   import Receive from "./Receive.svelte";
 
   export let accountId = null;
+  let toAddress;
+  let amount;
+
+  const openSendModal = () => {
+    fromStore.set(accountId);
```
sarahscott: There are two ways to do store updates with forms:

1. You can `set` the `store` upon form submission, so nothing is saved until the user decides to click `submit`. That's what is going on here - the store doesn't change until `openSendModal` is called.

2. You can `set` the `store` reactively every time the user changes the input value, as seen below in `SendFunds` (e.g. `$: amountStore.set(amount);`). With this approach, everything is automatically saved while the user types - see [docs on reactive declarations here](https://svelte.dev/tutorial/reactive-declarations)

Both are valid and have their pros and cons; it's a question of UX.

ui/DesignSystem/Component/Wallet/SendReceive.svelte
```diff
@@ -1,10 +1,26 @@
...
   import { Button, Input, Icon, Title } from "../../Primitive";
   import Receive from "./Receive.svelte";
 
   export let accountId = null;
+  let recipient,
+    amount,
+    currentlyActiveSend = true;
 
-  $: currentlyActiveSend = true;
+  const openSendModal = () => {
```
sarahscott, line 18: There are two ways to do store updates with forms:

You can `set` the `store` upon form submission, so nothing is saved until the user decides to click a submit button. That's what is going on here - the store doesn't change until `openSendModal` is called.

You can `set` the `store` reactively every time the user changes the input value, as seen below in `SendFunds`:
```
$: amountStore.set(amount); 
```
With this approach, everything is automatically saved while the user types - see [docs on reactive declarations here](https://svelte.dev/tutorial/reactive-declarations).

Both are valid and have their pros and cons; it's a question of UX.

ui/Screen/SendFunds.svelte
NunoAlexandre: I like `previousStep` better here since it makes it explicit that we are not cancelling the whole transfer, we are just going one step back:

``` rust
const previousStep = () => {
    state = TransferState.Preparation;
};
```

NunoAlexandre:
I found another issue, now that we have a `Transfer` message (and will have `TransferFromOrg`): `costSummary`, from `transaction.ts`, will need to take the amounts of those transfers into account when calculating the total. @juliendonck shall I tackle that?

ui/Screen/SendFunds.svelte
NunoAlexandre: That's a nice suggestion indeed. I propose we create an issue for that that we follow up in the next iteration since this whole wallets is already dragging as is as we have other concrete problems to solve first.

NunoAlexandre:
@MeBrei @juliendonck I will be working on the 2nd To-Do, submitting as transaction.

proxy/src/http/user.rs
MeBrei: I changed it so the API playground works again. It doesn't convert obkects correctly. @NunoAlexandre Did you find another way around this?

MeBrei:
@NunoAlexandre I was wondering if we should change the `fromAddress` to `fromId` and pass the id/handle of the user/org the form was submitted from. We do not need the senders address, but need the id/handle for the correct path to post the transfer

NunoAlexandre:
Yes, we should do that. We also need to know whether we are handling an org or a user to call the right API endpoint.

ui/Screen/SendFunds.svelte
juliendonck: nice one. Could one of you open an issue for this since this is outside this PR? Here's your suggestion, nice way of writing that :) https://github.com/radicle-dev/radicle-upstream/pull/712/commits/02a5bf1b159b69b0ab5d0056c2d251a93c40c6fe

ui/Screen/SendFunds.svelte
juliendonck: Makes sense to name them more explicitly: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/6b4d7fb0eb344641d5bd1f575446057ce0fc6ea4

ui/Screen/SendFunds.svelte
juliendonck: Makes sense to name them more explicitly: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/6b4d7fb0eb344641d5bd1f575446057ce0fc6ea4

ui/Screen/SendFunds.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/6b4d7fb0eb344641d5bd1f575446057ce0fc6ea4

ui/Screen/SendFunds.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/6b4d7fb0eb344641d5bd1f575446057ce0fc6ea4

NunoAlexandre:
@juliendonck just leaving an update for you to know what's left:

@NunoAlexandre
 - [ ] Add TransferFromOrg Message
 - [ ] Make sure Org wallet shows org-originated transactions
 - [ ] Fix/Correct transaction detail view

@MeBrei
  - Validate the Send `to` and `amount` fields

Ending with refactoring on all work done. @juliendonck will you address the review we have now? Let us know if there's anything in this list above you would like to work on yourself.

proxy/src/http/user.rs
NunoAlexandre: @MeBrei nop. The issue with using the `Id` type is that `warp` will try and use its default deserializer to parse a request and match it to and endpoint/handler. If it fails to do that, the whole thing fails with a 500 instead of a 400. I.e, say we do `users/123/transfer`. We know this should fail because `123` is not a valid `Id`. However, what warp sees is that there is no endpoint that expects a number so it fails completely. So yeah, better work with primitive numbers and avoid this implicit stuff taking place underneath and taking the control away from us.

juliendonck:
> @juliendonck will you address the review we have now? Let us know if there's anything in this list above you would like to work on yourself.

I've addressed everything that I can resolve. What's still open now touches part of the code I'm not familiar with

Also I've done a naive implementation of disabling the buttons [here](https://github.com/radicle-dev/radicle-upstream/pull/712/commits/d26fa8de73a8064ff1de8b4209c08f20444f99ba). We'll have to add or replace it with the validation work that's still outstanding.

ui/DesignSystem/Component/Wallet/SendReceive.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/d26fa8de73a8064ff1de8b4209c08f20444f99ba

proxy/src/http/user.rs
juliendonck: addressed https://github.com/radicle-dev/radicle-upstream/pull/712/commits/059b135f36003ee31ca3b3d725e96732a61fe2bc

ui/Screen/SendFunds.svelte
juliendonck: addressed https://github.com/radicle-dev/radicle-upstream/pull/712/commits/059b135f36003ee31ca3b3d725e96732a61fe2bc

ui/Screen/SendFunds.svelte
MeBrei: new issue: #728

ui/DesignSystem/Component/Wallet/SendReceive.svelte
MeBrei: For this form, I would go for option 1 as at this stage the store is set only to pass on the values to the modal. No complex validations happen here. In the modal the store is set reactively.

juliendonck:
@NunoAlexandre why is the amount in microRad. That seems weird to me. If I type `123 Rad`, should just be `123`, no?
Screenshot 2020-07-30 at 15 21 16

cypress/integration/send_funds.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    cy.pick("Wallet").click();
+    cy.pick("send-receive").should("exist");
+    cy.pick("send-tab").click();
+    cy.pick("send").should("exist");
+    cy.pick("recipient-input").should("exist");
+    cy.pick("recipient-input").type(
+      "5FA9nQDVg267DEd8m1ZypXLBnvN7SFxYwV7ndqSYGiN9TTpu"
+    );
+    cy.pick("amount-input").should("exist");
+    cy.pick("amount-input").type("123");
```
MeBrei, line 54: We don't need the input checks here, as we only test closing the modal.

cypress/integration/send_funds.spec.js
```diff
@@ -0,0 +1,140 @@
...
```
MeBrei, line 71: Same as above

cypress/integration/org_wallet.spec.js
```diff
@@ -57,6 +57,22 @@ context("navigation", () => {
       cy.pick("amount-input").type("123");
       cy.pick("send-transaction-button").click();
     });
+    it("tests the button is disabled when there is no amount is filled in", () => {
+      cy.pick("send-tab").click();
+      cy.pick("send").should("exist");
+      cy.pick("recipient-input").should("exist");
+      cy.pick("recipient-input").type(
+        "5FA9nQDVg267DEd8m1ZypXLBnvN7SFxYwV7ndqSYGiN9TTpu"
+      );
```
MeBrei, line 66: I think it is not necessary to check if the input exists and then enter the input. If you enter it directly, the test will fail anyway because it can't find the data-cy.
(Applies to the other tests added here as well)

MeBrei:
@juliendonck I would add tests, that the modal is opened with the correct sender pre-filled depending on where you open the modal from

MeBrei:
> @NunoAlexandre @MeBrei Also it seems that there is no validation seeing if you have enough funds in your wallet, or am I missing something?

No, it is missing. I'll think about how we can add it, as it is a more complex validation, ie depends on the wallet/sender you selected.

cypress/integration/send_funds.spec.js
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/0f58c678731a91801d6d0bc6f01096cc72045ef3

cypress/integration/send_funds.spec.js
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/0f58c678731a91801d6d0bc6f01096cc72045ef3

cypress/integration/org_wallet.spec.js
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/712/commits/0f58c678731a91801d6d0bc6f01096cc72045ef3

MeBrei:
> > @NunoAlexandre @MeBrei Also it seems that there is no validation seeing if you have enough funds in your wallet, or am I missing something?
> 
> No, it is missing. I'll think about how we can add it, as it is a more complex validation, ie depends on the wallet/sender you selected.

@sarahscott Do you have an idea, how to use the existing `validation.ts` functions to accommodate a validation that depends on another input? So far I only used it for passing a fixed value when creating the validation store.

MeBrei:
Created a follow up issue for the balance validation #730

juliendonck:
@NunoAlexandre @MeBrei if I'm sending money to myself, shouldn't I see 2 transactions. One outgoing and one incoming?
Screenshot 2020-07-31 at 09 19 39

NunoAlexandre:
> @NunoAlexandre @MeBrei if I'm sending money to myself, shouldn't I see 2 transactions. One outgoing and one incoming?
> Screenshot 2020-07-31 at 09 19 39

Well, we shouldn't let this be a thing I think. The user only wastes money doing so. So, for now, we either show it as incoming, or we don't allow it to happen in the first place. Showing both is not trivial and it doesn't seem to deliver any good value.

juliendonck:
> Well, we shouldn't let this be a thing I think. The user only wastes money doing so. So, for now, we either show it as incoming, or we don't allow it to happen in the first place. Showing both is not trivial and it doesn't seem to deliver any good value.

@NunoAlexandre sure, you're right. It would be a waste, but I don't think we should stop it. I would still expect to see outgoing transactions though.

NunoAlexandre:
> > Well, we shouldn't let this be a thing I think. The user only wastes money doing so. So, for now, we either show it as incoming, or we don't allow it to happen in the first place. Showing both is not trivial and it doesn't seem to deliver any good value.
> 
> @NunoAlexandre sure, you're right. It would be a waste, but I don't think we should stop it. I would still expect to see outgoing transactions though.

I still think we should help the user not making such an action/mistake which adds really no value to anyone and it's frustrating.
All financial apps I have used have this validation in place, even when there are no costs associated. Not sure about other wallet apps but I can't think of one good reason to not validate it :thinking:

cypress/integration/user_wallet.spec.js
```diff
@@ -78,7 +82,73 @@ context("navigation", () => {
...
+        .eq(0)
+        .get(".description")
+        .pick("subject")
+        .contains("from 5FA9nQDVg267DEd8m1ZypXLBnvN7SFxYwV7ndqSYGiN9TTpu");
+      cy.pick("transactions")
+        .children(".item")
+        .eq(0)
+        .get(".description")
+        .pick("subject")
+        .contains("from 5FA9nQDVg267DEd8m1ZypXLBnvN7SFxYwV7ndqSYGiN9TTpu");
```
MeBrei, line 123: the same test twice

ui/src/transfer.ts
```diff
@@ -57,7 +57,7 @@ const validateSufficientBalance = (fee: number, payerAccountId: string) => (
 ): Promise<boolean> => {
   return account
     .getBalance(payerAccountId)
-    .then(balance => balance >= +amount + fee);
+    .then(balance => currency.microRadToRad(balance) >= +amount + fee);
```
NunoAlexandre, line 60: I think it should be 
```suggestion
    .then(balance => balance >= currency.radToMicroRad(amount) + fee);
```

`balance` is obtained in `microRad`, `fee` is specified in `microRad` as well, only `amount` (specified by the user) is specified in Rad and thus needs to be converted to microRad here.

MeBrei: I understood that the `fee` is defined in `SendFunds` as Rad (see [here](https://github.com/radicle-dev/radicle-upstream/pull/712/files#diff-2a6d3c83a0fe02f639bc62d73afa75dcR43) as it is then shown in the transaction view as `0.000001`

NunoAlexandre: hm no. it is defined as:

```
// TODO(nuno): receive this from the user once possible
const transactionFee = 1;
```

This is in microRad. It is then converted and shown in the UI as RAD, thus the `0.000001`.

MeBrei: yes, I got confused. Fixed it in 6c70378

ui/src/transaction.ts
```diff
@@ -481,6 +611,19 @@ export const formatDate = (timestamp: number, option: string): string => {
   return `${time.toLocaleString(undefined, options)}`;
 };
 
+export const nFormatter = (num: number): string => {
```
sarahscott, line 614: We name all of the other formatting functions as verbs: `formatPayer`, `formatSubject`, I think this would make more sense as `formatNumber`

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,354 @@
...
+    margin-top: 1rem;
+  }
+
+  .from {
+    display: flex;
+    justify-content: flex-end;
+  }
+</style>
+
+<ModalLayout dataCy="page">
```
sarahscott, line 188: This component is feeling reeally big, it might be nice to break it up into some smaller reusable components.

ui/src/transaction.ts
NunoAlexandre: fixed, renamed it to `formatRad` :+1:

ui/Screen/SendFunds.svelte
NunoAlexandre: Indeed. How do you feel about accepting this as a technical debt of this big feature for which we create a follow-up issue?

ui/src/transaction.ts
```diff
@@ -323,12 +376,54 @@ interface Payer {
...
+    name: org.id,
+    type: PayerType.Org,
+    avatarFallback: org.avatarFallback,
+  };
+};
+
+// Get the payer of a transaction.
+// Note: It now look the payer up based on the local session, whereas
+// in the future we want to look it up on the network.
+export const getPayer = (msg: Message, session: Session): Payer | undefined => {
```
NunoAlexandre, line 398: A comment for peers and reviewers: This function was motivated by a bug found where we were _always_ passing a payer value that `Transaction.svelte` expects from the `session.identity`. I.e, we were saying the payer was always the identity stored on the machine. 

For screens _creating_ a transaction, such as `ProjectRegistration`, `OrgRegistration`, etc, we implicitly know who pays for it (in the registry). However, in `TransactionDetails` we have no context, only having an transaction `id` to work with, with we then use to fetch said transaction. Here, we need to figure who paid for that transaction, and we do that here. 

So since we _must_ use `getPayer` in `TransactionDetails`, I figured best to use it everywhere to have one source of truth and logic. That's the trade-off behind it.

ui/src/transaction.ts
```diff
@@ -323,12 +376,54 @@ interface Payer {
...
+    name: org.id,
+    type: PayerType.Org,
+    avatarFallback: org.avatarFallback,
+  };
+};
+
+// Get the payer of a transaction.
+// Note: It now looks the payer up based on the local session, whereas
+// in the future we want to look it up on the network.
+export const getPayer = (msg: Message, session: Session): Payer | undefined => {
```
NunoAlexandre, line 398: A comment for peers and reviewers: This function was motivated by a bug found where we were _always_ passing a payer value that `Transaction.svelte` expects from the `session.identity`. I.e, we were saying the payer was always the identity stored on the machine. 

For screens _creating_ a transaction, such as `ProjectRegistration`, `OrgRegistration`, etc, we implicitly know who pays for it (in the registry). However, in `TransactionDetails` we have no context, only having an transaction `id` to work with, with we then use to fetch said transaction. Here, we need to figure who paid for that transaction, and we do that here. 

So since we _must_ use `getPayer` in `TransactionDetails`, I figured best to use it everywhere to have one source of truth and logic. That's the trade-off behind it :blush: :seedling:

ui/Screen/SendFunds.svelte
sarahscott: Makes sense!

NunoAlexandre:
@xla @rudolfs we'd appreciate your review. This PR went kinda nuts, covering a lot of different things we encountered on the way and it was a collective effort between @juliendonck @MeBrei and me. I suggest that we focus on major issues found and for other things that can be seen as acceptable technical debt or future improvements we create follow up issues. Thanks already

ui/Screen/SendFunds.svelte
rudolfs: +1

ui/Screen/SendFunds.svelte
```diff
@@ -0,0 +1,354 @@
...
+          <div slot="left" style="display: flex;">
+            <Icon.Currency
+              size="big"
+              style="fill: var(--color-foreground-level-6)" />
+          </div>
+        </Input.Text>
+        <Title style="color: var(--color-foreground-level-6); padding: 0.5rem;">
+          From
+        </Title>
+        <!-- TODO(julien): shouldn't identity.accountId be the same as fromAddress -->
```
rudolfs, line 228: Is this TODO still relevant?

ui/src/org.ts
```diff
@@ -176,15 +176,54 @@ export const memberHandleValidationStore = (
...
-    return tx.messages[0].type ===
-      transaction.MessageType.ProjectRegistration &&
-      tx.messages[0].domainType === project.Domain.Org
-      ? tx.messages[0].domainId == id
-      : null;
-  });
+  return transactions.filter(tx => involvesOrg(tx.messages[0], id, accountId));
+};
+
+export const fallback = (): Org => {
```
rudolfs, line 214: What is a fallback org?

NunoAlexandre: We should document this. When we look at transfer records involving an org that is no longer around (on the local machine), we need data to display the avatar component of such entities. To reproduce, register an org, transfer something from it to the (registered) user, restart the app, and now look at the `Incoming Transfer` in the user wallet. The org is gone (flushed on restart). It will show it as:

Screenshot from 2020-08-03 10-25-37

ui/Screen/SendFunds.svelte
NunoAlexandre: I don't think so!

NunoAlexandre:
> LGTM 
> 
> Inlined some small suggestions and a question. 

:heavy_check_mark: all addressed.

> Feel like some of the bigger components should be broken up in a follow-up refactoring.

I already created #739 for splitting `SendFunds`. Let me know if there are others in your mind :+1: 

> 
> Also noticed a truncation issue:
> 
> Screenshot 2020-08-03 at 09 20 52

:heavy_check_mark:  Fixed.

> 
> When the balance is below 1 Rad, it's not possible to spend it anymore:
> 
> Screenshot 2020-08-03 at 10 17 59
> 
> Screenshot 2020-08-03 at 10 17 49
> 
> Screenshot 2020-08-03 at 10 19 31
> 
> But this can also be addressed in a follow-up as there are bigger changes coming to registry anyway.

Yeah, indeed I would leave this one for now.


Thanks for the thorough look, @rudolfs :ok_hand:

ui/Screen/SendFunds.svelte
juliendonck: @sarahscott I think @xla is looking at this already in his layout PR's

cypress/integration/org_wallet.spec.js
```diff
@@ -38,14 +45,103 @@ context("navigation", () => {
...
+      cy.pick("modal-amount-input").type("112");
+      cy.pick("review-transfer-button").click();
+
+      // 3. Now in the final review step, submit
+      cy.pick("review-step").should("exist");
+      cy.pick("transfer-amount", "amount").contains("-112");
+      cy.pick("transaction-fee", "amount").contains("-0.000001");
+      cy.pick("total", "amount").contains("-112.000001");
+      cy.pick("funding-source").contains("coolorg");
+      cy.pick("submit-tranfer-button").click();
```
MeBrei, line 101:
```suggestion
      cy.pick("submit-transfer-button").click();
```

cypress/integration/send_funds.spec.js
```diff
@@ -0,0 +1,186 @@
...
+    cy.pick("send-transaction-button").click();
+    cy.pick("send-funds-modal").should("exist");
+    cy.pick("page", "preperation-step").should("exist");
+    cy.pick("review-transfer-button").should("exist");
+    cy.pick("review-transfer-button").click();
+    cy.pick("page", "review-step").should("exist");
+    cy.pick("back-to-review-button").should("exist");
+    cy.pick("back-to-review-button").click();
+    cy.pick("preperation-step").should("exist");
```
MeBrei, line 88:
```suggestion
    cy.pick("preparation-step").should("exist");
```

cypress/integration/send_funds.spec.js
```diff
@@ -0,0 +1,186 @@
...
+    );
+    cy.pick("amount-input").type("123");
+    cy.pick("send-transaction-button").click();
+    cy.pick("send-funds-modal").should("exist");
+    cy.pick("page", "preperation-step").should("exist");
+    cy.pick("review-transfer-button").should("exist");
+    cy.pick("review-transfer-button").click();
+    cy.pick("page", "review-step").should("exist");
+    cy.pick("back-to-review-button").should("exist");
+    cy.pick("back-to-review-button").click();
```
MeBrei, line 87: We should try to remove the duplication of checks... `cy.pick("back-to-review-button").click();` should be enough, as it also checks if it exists.

Also applies to other areas of the tests

NunoAlexandre: Done

PR #715, opened 2020-07-23
feat(ui): Show actual free balance on wallet screens

### LINKED ISSUE #685: Show the actual free balance on the wallet page
#681 adds an endpoint to query the free balance for orgs and users. Once #662 is merged, we should show the actual free balance on the wallet page.

MeBrei:
Closes #685

### DIFF
--- a/ui/src/account.ts
+++ b/ui/src/account.ts
@@ -1,11 +1,21 @@
 import * as api from "./api";
+import * as remote from "./remote";
 
 /* Offers commodities to query useful information regarding accounts on chain. */
 
 // Get the balance of the account with the given id
-export const balance = (id: string): Promise<number> =>
+export const getBalance = (id: string): Promise<number> =>
   api.get<number>(`accounts/${id}/balance`);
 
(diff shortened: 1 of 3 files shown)


ui/src/account.ts
```diff
@@ -1,11 +1,21 @@
...
 
 // Check whether an account with the given id exists on chain
 export const exists = (id: string): Promise<boolean> =>
   api.get<boolean>(`accounts/${id}/exists`);
+
+// State
+const balanceStore = remote.createStore<number>();
+export const balance = balanceStore.readable;
+
+export const updateBalance = (id: string) => {
```
NunoAlexandre, line 18: maybe this is common terminology, but I find "update" here confusing as it suggests a write operation. how about "fetchBalance"?

MeBrei: It "updates" the `balanceStore` so technically if it would be called again by the component it would overwrite the store.

NunoAlexandre: right I see.

Issue #606, opened 2020-06-26
Integrate "project checkout" into UI


rudolfs:
The UI will provide a path where the checkout should happen to the proxy.
If possible the proxy should provide a response upon successful checkout.

Depends on: https://github.com/radicle-dev/radicle-upstream/issues/604 and https://github.com/radicle-dev/radicle-upstream/issues/601.

<img width="227" alt="Screenshot 2020-06-26 at 17 11 26" src="https://user-images.githubusercontent.com/158411/85873798-ef5b3000-b7d1-11ea-932c-01c4b3ee19f6.png">
<img width="336" alt="Screenshot 2020-06-26 at 17 11 30" src="https://user-images.githubusercontent.com/158411/85873803-eff3c680-b7d1-11ea-9a5b-196267a988c9.png">

Issue #601, opened 2020-06-26
Create Working Copy


FintanH:
Integration endpoint of https://github.com/radicle-dev/radicle-link/issues/138. If there is a project in the monorepo that doesn't have a working copy then we should be able to create one.

How will we know if there is a working copy or not or do we even care and it should always be possible? :thinking:

rudolfs:
At what point, starting from clicking the "Track" button on a project, will start showing the "checkout" button in the UI?

rudolfs:
Would the working copy be per project or per peer in the project?

rudolfs:
The UI will provide a path where the checkout should happen.
If possible we should provide back a response upon successful checkout.

<img width="227" alt="Screenshot 2020-06-26 at 17 11 26" src="https://user-images.githubusercontent.com/158411/85872654-52e45e00-b7d0-11ea-8105-38ecc97d7545.png">
<img width="336" alt="Screenshot 2020-06-26 at 17 11 30" src="https://user-images.githubusercontent.com/158411/85872659-537cf480-b7d0-11ea-861a-e79f8f630bb9.png">

FintanH:
> At what point, starting from clicking the "Track" button on a project, will start showing the "checkout" button in the UI?

As soon as there's a copy of the project in the monorepo, I would say.

>  Would the working copy be per project or per peer in the project?

Per project. You have your local view and every peer would look like another server essentially. You could checkout their branches as a local copy and edit them, but they would have to choose to integrate those changes once you announce them. Does that make sense?

rudolfs:
> Does that make sense?

👍

PR #722, opened 2020-07-24
feat: project checkout

rudolfs:
As discussed with @cloudhead: this PR implements project checkout via shelling out to `git` from `proxy` and using the `git-remote-rad` helper.

checkout

The default remote of the checkout will be created manually by `proxy` it should be derived from the `PeerId` which is selected in the UI. The default checked out branch should also come from the repository selector UI.

Next steps:
  - [x] [implement notification per design](https://www.figma.com/file/dnYlebm0ehgyJyT3Po9Kcl/Upstream?node-id=10309%3A75124) (there should be a button that opens the folder in finder when clicked)
  - [x] change proxy to create folder for us instead of checking out into folder
  - [x] pass the checkout path back from proxy to the frontend
  - [x] clean up proxy & ui code -> add missing types, etc
  - [ ] ~~set up remotes in working dir to point to monorepo~~ -> https://github.com/radicle-dev/radicle-upstream/issues/766
  - [ ] ~~add instructions for user on how to set up `rad-remote-helper`~~ -> https://github.com/radicle-dev/radicle-upstream/issues/763
  - [ ] ~~copy `rad-remote-helper` to a stable location (xdg/bin?) on app start (AppImage contents are only available while the app runs)~~ -> https://github.com/radicle-dev/radicle-upstream/issues/763

Closes: https://github.com/radicle-dev/radicle-upstream/issues/601
Closes: https://github.com/radicle-dev/radicle-upstream/issues/606


FintanH:
>  change the git-remote-rad helper to accept a SECRET_KEY_FILE_PATH env variable and pass the correct key path down to git as an environment variable

This doesn't need to be done because `RAD_HOME` is the environment variable that needs to be set.

>  figure out how to pass the key encryption password safely to the git-remote-rad helper

This should be done once the `Signer` work is complete.

FintanH:
I've successfully cloned a repository from my disk to my disk :tada: I'll try figure out how to add a remote test as well. And work on the clean-up items such as error handling/reporting.

FintanH:
> The default checked out branch should also come from the repository selector UI.

@rudolfs & @cloudhead: Would this not be the default branch defined by the `Project`? So if we get the `Project` via the `RadUrn` + `PeerId` we can look at what the default branch is.

proxy/src/http/project.rs
```diff
@@ -341,6 +394,30 @@ impl ToDocumentedType for CreateInput {
...
 }
 
+/// Bundled input data for project checkout.
+#[derive(Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CheckoutInput {
+    /// Location on the filesystem where the working copy should be created.
+    path: PathBuf,
+    /// Which peer are we checking out from. If it's `None`, we're checking out our own project.
+    peer_id: Option<coco::PeerId>,
```
FintanH, line 404: Note: I'm actually wrong about this. The `LocalUrl` is just the `Hash` of the project. Picking which `HEAD` is another issue: https://github.com/radicle-dev/radicle-link/issues/234

rudolfs:
> @rudolfs & @cloudhead: Would this not be the default branch defined by the Project? So if we get the Project via the RadUrn + PeerId we can look at what the default branch is.

My understanding was that we should check out the branch that is selected in the repo selector, not the default branch.

FintanH:
> My understanding was that we should check out the branch that is selected in the repo selector, not the default branch.

I'm not entirely sure what the behaviour is here, but is it possible that this branch becomes the "default branch" for the working copy? Is that really what we want?

proxy/src/project.rs
```diff
@@ -80,14 +86,138 @@ pub enum Registration {
...
+    ///
+    ///   * We couldn't resolve the executable path.
+    ///   * The checkout process failed.
+    pub fn run(self) -> Result<(), error::Error> {
+        let bin_path = match self.bin_path {
+            Some(path) => Ok(path),
+            None => Self::default_bin_path(),
+        }?;
+
+        let mut child_process = Command::new("git")
```
FintanH, line 177: Instead of naming the `remote = "origin"` we should start off by naming it after `PeerId`. But we should probably work towards this: https://github.com/radicle-dev/radicle-link/issues/235.

We should create a follow-up ticket to figure out how to work with this new remote and the rad-remote-helper.

We also have this question which is related: https://github.com/radicle-dev/radicle-link/issues/235#issuecomment-669297448

proxy/src/project.rs
```diff
@@ -80,14 +86,138 @@ pub enum Registration {
...
+            .arg(&self.path.as_ref().as_os_str())
+            .env("PATH", &bin_path)
+            .envs(std::env::vars().filter(|(key, _)| key.starts_with("GIT_TRACE")))
+            .spawn()?;
+
+        // TODO: Capture the error if any and respond
+        let result = child_process.wait()?;
+
+        if result.success() {
+            Ok(())
```
FintanH, line 193: Upon successfully cloning the repo to our directory we should set up the `remote = "rad"` which handles fetching and pushing from/to the monorepo. See `setup_remote` in `coco/peer.rs`

proxy/src/project.rs
```diff
@@ -125,6 +125,8 @@ where
     credential: Credential,
     /// The project URN.
     urn: coco::Urn,
+    /// The directory that will be created in path where the git repository will be checked out
+    directory_name: String,
```
FintanH, line 129: I think at this point we should just pass the `Project` in. Because that way we can get the `urn`, calculate the directory name, and get the branch`.

proxy/src/http/project.rs
```diff
@@ -221,9 +221,15 @@ mod handler {
         R: Send + Sync,
     {
         let ctx = ctx.read().await;
-        let project = ctx.peer_api.get_project(&urn, peer_id)?;
+        let default_branch = ctx
+            .peer_api
+            .get_project(&urn, peer_id)?
+            .default_branch()
+            .to_owned();
+        let directory_name = &project::get(&ctx.peer_api, &urn)?.metadata.name;
```
FintanH, line 229: We're getting the project twice here :)

proxy/src/project.rs
rudolfs: The remote helper already creates a remote for us that points to the monorepo, it just called it by default `origin`. Renamed it to `rad` in: https://github.com/radicle-dev/radicle-upstream/pull/722/commits/8651198e24cdaa89aa70a370373766be42bd9274.

cypress/integration/project_checkout.spec.js
```diff
@@ -0,0 +1,102 @@
+import { DIALOG_SHOWOPENDIALOG, OPEN_PATH } from "../../native/ipc.js";
+
+const withWorkspaceStub = callback => {
+  cy.exec("pwd").then(result => {
+    const pwd = result.stdout;
+    const checkoutPath = `${pwd}/cypress/workspace/checkout`;
+
+    // Clean up any left-overs from previous failed tests.
+    cy.exec(`rm -rf ${checkoutPath}`);
```
FintanH, line 9: Is there a way to specify setup and teardowns in general for cypress?

Scrolling down I see `beforeEach`. Should these commands live there?

cypress/integration/project_checkout.spec.js
```diff
@@ -0,0 +1,102 @@
...
+  context("happy path", () => {
+    it("checks out the project into a working directory", () => {
+      cy.pick("project-list-entry-platinum").click();
+      cy.pick("checkout-modal-toggle").click();
+
+      withWorkspaceStub(checkoutPath => {
+        cy.pick("choose-path-button").click();
+        // Make sure UI has time to update path value from stub,
+        // prevents this spec from failing on CI.
+        cy.wait(500);
```
FintanH, line 53: Is there no way to check the stub's condition or something?

cypress/integration/project_checkout.spec.js
```diff
@@ -0,0 +1,102 @@
...
+          expect(appWindow.mockOpenPathCalled).to.be.true;
+        });
+
+        // Make sure the notification gets closed after we open the folder in
+        // the OS file browser.
+        cy.contains("platinum checked out to").should("not.exist");
+
+        // Get project URN and check that the working directory has the
+        // appropriate rad:// remotes set up.
+        cy.pick("project-screen", "urn")
```
FintanH, line 85: Should the front end tests really be concerned with this? Surely it's more something we'd test in the proxy?

proxy/src/project.rs
```diff
@@ -80,14 +86,154 @@ pub enum Registration {
...
+                path.to_path_buf()
+            } else {
+                path.join(name)
+            }
+        } else {
+            path.join(&self.project.name().to_string())
+        };
+
+        let mut child_process = Command::new("git")
+            .arg("-c")
```
FintanH, line 192: It might help readability if we used the long names for these params, now that I think about it

proxy/src/project.rs
```diff
@@ -80,14 +86,154 @@ pub enum Registration {
...
+            .arg("rad")
+            .arg("-b")
+            .arg(self.project.default_branch())
+            .arg(LocalUrl::from(self.project.urn()).to_string())
+            .arg(project_path.as_os_str())
+            .env("PATH", &bin_path)
+            .envs(std::env::vars().filter(|(key, _)| key.starts_with("GIT_TRACE")))
+            .spawn()?;
+
+        // TODO: Capture the error if any and respond
```
FintanH, line 205: Do we think this is possible? :thinking: And/or is still valid?

ui/Screen/DesignSystemGuide.svelte
```diff
@@ -856,27 +856,51 @@
     </Section>
 
     <Section title="Notifications" subTitle="Info, Warnings and Errors">
+      <Swatch>
+        <Notification message="Snackbar" />
```
FintanH, line 860: Yum :yum:

ui/src/notification.ts
```diff
@@ -14,7 +14,10 @@ export enum Level {
 interface Notification {
   id: ID;
   level: Level;
+  showIcon: boolean;
   message: string;
+  actionText: string;
+  actionHandler: () => void;
```
FintanH, line 20: This action handler thing seems overly permissive. What kind of things should it be able to do? 

For example, I see that in one instance it can `openPath(path)` so should it be able to take some data and interpret it, such as `Path(string)` and interprets it as `Path(path) -> openPath(path)`.

ui/src/notification.ts
```diff
@@ -14,7 +14,10 @@ export enum Level {
 interface Notification {
   id: ID;
   level: Level;
+  showIcon: boolean;
   message: string;
+  actionText: string;
```
FintanH, line 19: Should `actionText` and `actionHandler` be paired as some kind of type `Action { message: string, action: Handler }`.

cypress/integration/project_checkout.spec.js
rudolfs: Possibly. I'm using the same pattern here as in https://github.com/radicle-dev/radicle-upstream/blob/master/cypress/integration/project_creation.spec.js.

At an earlier point I tried to extract this into a cypress command, but it didn't work as I imagined because of subtle differences on how Cypress is doing async: https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Commands-Are-Not-Promises.

Extracting this functionality into a wrapper function was the only way I could get it to work reliably. Not sure if sinking more time into making this prettier makes sense.

FintanH: > Not sure if sinking more time into making this prettier makes sense.

I wasn't thinking prettier, I was thinking more reliable. If commands that are needing during setup are grouped outside at least they're localised to one place and can be used if we needed other tests that would need to do the same.

But if it's a PITA, then now worries

cypress/integration/project_checkout.spec.js
rudolfs: I treat the cypress test suite more like high-level integration tests for app features. So anything that a user would be expected to be able to check, we'd check with the tests.

But if you think it's covered elsewhere, then I can remove it from here.

proxy/src/project.rs
rudolfs: Good point. The UI now shows an error notification if something goes wrong in proxy, but the error messaging could be improved.

This is what the user sees:
Screenshot 2020-08-10 at 18 00 38

And this is what is logged by proxy:
```
Cloning into '/Users/rudolfs/work/radicle-checkout/rx'...
Error: Signer key does not match key used at initialisation

Stack backtrace:
   0: std::backtrace::Backtrace::create
   1: std::backtrace::Backtrace::capture
   2: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   3: radicle_git_helpers::remote_helper::run
   4: git_remote_rad::main
   5: std::rt::lang_start::{{closure}}
   6: std::rt::lang_start_internal
   7: std::rt::lang_start
   8: main
 ERROR warp::filters::body   > request body already taken in previous filter
 ERROR proxy::http::error    > Rejection([MethodNotAllowed, BodyConsumedMultipleTimes, Checkout, MethodNotAllowed])
```

I'd create a follow-up issue for that to not increase the scope here any further, if you're okay with that.

cypress/integration/project_checkout.spec.js
FintanH: I'm not gonna say no to having more tests :) But it's definitely weird that we would test up here rather than closer to metal. It's a good test and I'm pretty sure we don't have it in the proxy :smile:

proxy/src/project.rs
FintanH: I'm perfectly ok with that :heart:

cypress/integration/project_checkout.spec.js
rudolfs: Not that I know of or wouldn't involve a lot of boilerplate, we have the same issue in project creation: https://github.com/radicle-dev/radicle-upstream/blob/master/cypress/integration/project_creation.spec.js#L270.

proxy/src/project.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/issues/767

ui/src/notification.ts
rudolfs: > What kind of things should it be able to do?

I think it could be anything. In this case we need it to fire the native method to open the OS file browser, but I could imagine it could be used for things like navigating to a certain screen in the app, etc.

Screenshot 2020-08-11 at 09 12 56

I'm not sure what we would gain by restricting it now, without knowing what other uses there will be in the future?

cypress/integration/project_checkout.spec.js
rudolfs: > I wasn't thinking prettier, I was thinking more reliable.

Sure. But yeah, let's do this another time.

ui/src/notification.ts
FintanH: Restricting it means we know what it does, extending the cases in the future is the [expression problem](https://en.wikipedia.org/wiki/Expression_problem).

Basically, we're giving this function too much power in the type and so it becomes harder to reason about what it does, because indeed it could do anything! Like open a file browser :scream_cat:

Again, none of my comments here are blockers but things to think about as we continue to write code and possibly make a concerted effort for the future :grin:

proxy/src/project.rs
rudolfs: I can't seem to get the credential helper to work with the `config` flag. Wanna show me how it's done?

ui/src/notification.ts
rudolfs: > Again, none of my comments here are blockers but things to think about as we continue to write code and possibly make a concerted effort for the future 😁

I can't merge without you accepting the PR though : P

FintanH: lol fair. I wanted to provoke the conversation before you got my thumbs up :stuck_out_tongue_closed_eyes: I'm no pushover ;)

rudolfs: lol

proxy/src/project.rs
FintanH: Ah maybe not this one, since the `key=value` is above in the `Credential`.

But `--origin` and `--branch` would be better :)

rudolfs: Good point!

https://github.com/radicle-dev/radicle-upstream/pull/722/commits/4c2e8ae98de6c19c0c1597ef990fb12df0680721

PR #723, opened 2020-07-24
chore(ui): converting one screen to typescript

sarahscott:
starting with `Org.svelte` & its children

### DIFF
--- a/ui/DesignSystem/Component/SidebarLayout.svelte
+++ b/ui/DesignSystem/Component/SidebarLayout.svelte
@@ -8,8 +8,8 @@
 
   import Sidebar from "./Sidebar.svelte";
 
-  export let dataCy = null;
-  export let style = null;
+  export let dataCy = "";
+  export let style = "";
 
   const session = getContext("session");
 </script>
--- a/ui/DesignSystem/Primitive/Icon/Verified.svelte
+++ b/ui/DesignSystem/Primitive/Icon/Verified.svelte
@@ -1,8 +1,8 @@
 <script>
   import IconWrapper from "./IconWrapper.svelte";
 
-  export let dataCy = null;
-  export let style = null;
+  export let dataCy = "";
+  export let style = "";
   export let size = "small"; // small | large
 </script>
 
--- a/ui/Screen/Org.svelte
+++ b/ui/Screen/Org.svelte
@@ -1,9 +1,9 @@
-<script>
+<script lang="ts">
   import { getContext } from "svelte";
   import Router, { push } from "svelte-spa-router";
 
-  import { fetch, org as store } from "../src/org.ts";
-  import * as path from "../src/path.ts";
+  import { fetch, org as store, Org } from "../src/org";
+  import * as path from "../src/path";
 
   import { Icon } from "../DesignSystem/Primitive";
   import {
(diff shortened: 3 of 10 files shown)


ui/DesignSystem/Component/SidebarLayout.svelte
```diff
@@ -8,8 +8,8 @@
 
   import Sidebar from "./Sidebar.svelte";
 
-  export let dataCy = null;
-  export let style = null;
+  export let dataCy = "";
+  export let style = "";
```
FintanH, line 12: I'm curious as to what this setup is for? Just for my self learning :)

ui/Screen/Org.svelte
```diff
@@ -30,14 +30,15 @@
...
+  import MenuItem from "../DesignSystem/Component/HorizontalMenu/MenuItem.svelte";
 
   const menuRoutes = {
     "/orgs/:id/": ProjectsMenu,
     "/orgs/:id/projects": ProjectsMenu,
     "/orgs/:id/members": MembersMenu,
   };
 
-  const topbarMenuItems = orgId => [
+  const topbarMenuItems = (orgId: string) => [
```
FintanH, line 41: Should `orgId` be a type `OrgId` to disambiguate it from other `string`s?

ui/src/identity.ts
```diff
@@ -19,6 +19,7 @@ export interface Identity {
   };
   registered?: string;
   avatarFallback: Avatar;
+  shareableEntityIdentifier?: string;
```
FintanH, line 22: I'd also consider pushing out the use of `string` everywhere and introducing types that have a better structure and/or allow us to disambiguate them from other `string`ly typed objects.

sarahscott: agreed

ui/DesignSystem/Component/SidebarLayout.svelte
sarahscott: Figuring out the vibe for type inference v type declaration in `props`; still under construction 🚧

ui/Screen/Org.svelte
sarahscott: ja i think that would make sense for all of the `id`s

FintanH: Sehr gut

ui/src/identity.ts
sarahscott: This might get a little awkward for components that render `string`s though, like `Title` or `Text`

FintanH: Oh ya, sure, They're literally strings. No argument there :)

ui/DesignSystem/Primitive/Icon/Verified.svelte
```diff
@@ -1,8 +1,8 @@
 <script>
   import IconWrapper from "./IconWrapper.svelte";
 
-  export let dataCy = null;
-  export let style = null;
+  export let dataCy = "";
+  export let style = "";
```
xla, line 5: There is some work around required/optional props in sveltejs/svelte#5183, sveltejs/language-tools#273 and sveltejs/language-tools#319.

ui/Screen/Org.svelte
```diff
@@ -58,7 +59,24 @@
     },
   ];
 
-  let registerProjectMenuItem;
+  interface MenuItem {
```
xla, line 62: Would be nice to push these into `src` code.

sarahscott:
Closing until we figure out the linting situation (#779 )

PR #727, opened 2020-07-28
fix(proxy): project init

### LINKED ISSUE #726: init_project creates git repo in the wrong path
Creating a new project doesn't build the full path to the git repository, instead, it creates the repository for the path provided: https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/src/coco/peer.rs#L269.

I would have expected the path to be the "base" path and the project name to be the destination folder. If this destination directory already exists then we don't do any directory creation.

FintanH:
Fixes #726

When using the application for **Creating a New Project** we're prompted for a directory and a project name.

I would think the UX would be, and correct me if I'm wrong, that:
* The directory is the base path
* The name is the folder to be created for the repository

In the case of the `directory.suffix() == name` I just leave the path as is.

I came across this when I was using the app and accidentally made my `/home/haptop` folder a git repository.


cloudhead:
If anything, it's always safer to create a sub-folder!

FintanH:
> If anything, it's always safer to create a sub-folder!

So do you think `.join(name)` is fine without my hoop-jumping to check the last component? I was thinking that the user might have prepped the folder before hand :thinking: I'm easy on the final implementation :shrug:

xla:
> I would think the UX would be, and correct me if I'm wrong, that:
> 
> * The directory is the base path
> * The name is the folder to be created for the repository

This has not been the assumption so far, we always assumed the user provides the actual path to be used. There was never the notion of a base path. In turn this works equally for existing and new repositories.

FintanH:
> This has not been the assumption so far, we always assumed the user provides the actual path to be used. There was never the notion of a base path. In turn this works equally for existing and new repositories.

I find it weird then to provide a name _and_ a directory. If I fucked up then I'm sure others will too. This is my attempt at remedying the experience, but I'm open to other suggestions.

cloudhead:
The name is the project name, not the dir name, which I agree is confusing

FintanH:
So the below is what I'm seeing. Now if I actually desired to create a repository from scratch I would have to:
1. Go to my **terminal**
2. Create `~/Developer/surfs-up` 
3. Navigate back to **Radicle Upstream**
4. Click **Create Project**.

That's quite a bit of friction to get the behaviour I'd expect.

image

xla:
@FintanH Can't you create a directory in the context of the dialog that opens when you click `Choose`?

FintanH:
> @FintanH Can't you create a directory in the context of the dialog that opens when you click Choose?

Ah, I found the dialog, but I didn't know it was there. I still find this weird though. Like, how would other tools do this? I feel like most of them use the name provided. Examples:

* `cargo new radicle-upstream`
* `stack new my-haskell-project`
* Unzipping a folder takes the prefix `<prefix>.zip`

Do you see what I'm getting at or am I just making a mountain out of a molehill?

sarahscott:
@FintanH FWIW I see where you're coming from and I think having a discrepancy between project name and dir name will lead to confusion down the road. 
```
npx degit sveltejs/template my-svelte-project
cd my-svelte-project
```

Svelte also create a new directory with the project name.

FintanH:
Not to dig my hole any deeper, but if we pick an existing working copy maybe picking the name doesn't make sense either :sweat_smile: 

So if I choose "Continue with an existing repository" the name would be the final component of the path. Maybe having a choice to override it if the user wanted.

FintanH:
Because the test failure -- image below -- is due to picking an existing repository and my changes are just creating a new repository underneath that. What do peeps think?

oopsie

xla:
We should get this addressed in the specs/designs and get consensus from @radicle-dev/product

FintanH:
> We should get this addressed in the specs/designs and get consensus from @radicle-dev/product

What's the process for that? :)

brandonhaslegs:
Not sure I follow 100% what needs to be decided here, but I do think the "Project name" field should match the directory name on your computer. So:
- When you choose "Continue with an existing repository", the "Project name" field should be auto-filled with the name of the directory you chose, and the field should be disabled and not editable.
- On the flip side, when you choose "Start with a new repository", the repository you are creating should be a directory **within** the directory you choose in the filepicker dialog. As the text says, "Choose where you’d like to create the repository". And the directory on your computer should have the same name as what you entered in the "Project name" field.

We definitely should change the text to make this all clearer. I can open a PR with some changes that might clear things up.

FintanH:
> When you choose "Continue with an existing repository", the "Project name" field should be auto-filled with the name of the directory you chose, and the field should be disabled and not editable.

Ya, that's exactly what I was thinking. Alternatively, we could just not have a field for the project name, but that's just a suggestion. Disabling the text box also makes sense :ok_hand: 

> On the flip side, when you choose "Start with a new repository", the repository you are creating should be a directory within the directory you choose in the filepicker dialog. As the text says, "Choose where you’d like to create the repository". And the directory on your computer should have the same name as what you entered in the "Project name" field.

Again, that's what I was thinking the behaviour _should_ be. So if I have `name: kickflip` `directory: /home/rodneymullen` the project should be created at `/home/rodneymullen/kickflip`.

FintanH:
Ok, I'm back to fixing this up :) I changed the logic on the backend but need to update stuff on the frontend to match.

rudolfs:
Some screen-grabs of the UI changes:

input-update
choose

ui/Screen/ProjectCreation.svelte
```diff
@@ -149,14 +149,13 @@
...
-          defaultBranch,
-        },
-        isNew ? newRepositoryPath : existingRepositoryPath
-      );
+      response = await create({
+        description,
+        defaultBranch,
+        repo: isNew
+          ? { type: NEW, name, path: newRepositoryPath }
+          : { type: EXISTING, path: existingRepositoryPath },
```
xla, line 157: These should come from an enum in typescript source file, otherwise the actual values gonna drift.

ui/src/project.ts
```diff
@@ -13,6 +13,19 @@ export interface Metadata {
   description?: string;
 }
 
+export interface New {
+  type: "new";
```
xla, line 17: As mentioned this should be an enum. For reference:

https://github.com/radicle-dev/radicle-upstream/blob/1485bba8d343f00e4cb87e3a644db22948ce72d2/ui/src/settings.ts#L2-L5

ui/Screen/ProjectCreation.svelte
FintanH: Ya, that makes sense :+1: Too much of a n00b around the JS stuff still

proxy/src/coco/peer.rs
```diff
@@ -249,55 +250,27 @@ impl Api {
     ///     * The signing of the project metadata fails.
     ///     * The interaction with `librad` [`librad::git::storage::Storage`] fails.
     #[allow(clippy::needless_pass_by_value)] // We don't want to keep `SecretKey` in memory.
```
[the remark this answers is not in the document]
FintanH: I'll just create a small PR for this after

PR #748, opened 2020-08-03
fix(ci): make clippy work on CI again

rudolfs:
This was first fixed in:
https://github.com/radicle-dev/radicle-upstream/pull/430.

But then later broken in:
https://github.com/radicle-dev/radicle-upstream/pull/721.


NunoAlexandre:
🌱

PR #750, opened 2020-08-04
fix(ci): add timezone to test:integration command

juliendonck:
Fixing an issue where running `yarn test:integration` would fail because of the timezone not being explicitly defined.


rudolfs:
> Would be good to keep track of what explicitly errored in the commit/merge message for future reference.

A bunch of time-zone sensitive tests were failing for @juliendonck locally.

juliendonck:
> Would be good to keep track of what explicitly errored in the commit/merge message for future reference.
> 
> 🖖 🕴 🔧 👥

@xla will do

Issue #741, opened 2020-08-03
Known identities API endpoints


MeBrei:
Create two endpoints, one that lists all "tracked" identities of the session identity and another one that lists all projects (ie projects locally stored) of the tracked identity. As explained in #740 for the first version, tracked identities are all identities involved with locally stored projects.

The endpoints should serve data to two places:
1. Tracking screen of the session identity profile
- list of identities with ✅ 
  - handle
  - avatar
  - URN
2. visitor view of the profile
- identity
  - handle
  - avatar
  - URN
- list of project names + description + stats + maintainer badge

xla:
> Create two endpoints, one that lists all "tracked" identities of the session identity and another one that lists all projects (ie projects locally stored) of the tracked identity. As explained in #740 for the first version, tracked identities are all identities involved with locally stored projects.

If we foresee that we want to get the tracked list of entities for other identitie than the one of the session, we should encode that in the endpoints. Wouldn't be more general to have these endpoints be parameterised so that the caller can get a filtered list of tracked entities, e.g. by passing a `RadUrn`?

MeBrei:
I updated the comment above, as I confused some things. What we want for this Issue, is still `identities/tracked` (we could do `identities/<id>/tracked` but there is currently no plan how we would implement that, ie if we need the `id` for it or the `peer_id`. 
For the lists of projects for a visited profile, we could do `/projects` and pass the peer_id (and urn ?) via json.

xla:
Great, thanks for the clarification. As we toss around terms like session identity I think it's important to highlight what the determining dimensions is that we narrow the query for tracked identities/projects is. In my limited understanding this would be per peer and not per librad user. Maybe @FintanH can clarify.

In any case I think it's more powerful if the endpoints expect such `peer_id` as this could serve in other scenarios, beyond the session use-case, e.g. show me all the projects a seed node is tracking.

FintanH:
> In my limited understanding this would be per peer and not per librad user. 

Well, to show any meaningful information we need to get the librad user this peer is using as their associated user. So if want to show the user's handle for the given peer id, we also need their profile.

Also, I could be sharing with `xla@peer1` and `xla@peer2`. And their projects might be disjoint. So that's an interesting case. We could attempt to correlate them using their advertised keys. But not sure we want to open that can of worms yet :upside_down_face: :bug:

xla:
[...]

PR #757, opened 2020-08-06
feat(proxy): tracked identities endpoint

MeBrei:
Partially solves #741 

Currently we only want to use this endpoint for the session identity

- [x] ~should the identity be passed? (currently `list_projects` uses `default_rad_self`)~ not necessary for this use case
- [x] ~should it also look at projects that are not owned by `default_rad_self`?~ not necessary for this use case
- [x] test `list_tracked_identities` on the proxy side`


proxy/src/http/identity.rs
```diff
@@ -80,6 +83,30 @@ where
         .and_then(handler::get)
 }
 
+/// `GET /tracked-identities`
+fn list_tracked_identities_filter<R>(
```
xla, line 87:
```suggestion
fn list_tracked_filter<R>(
```

proxy/src/http/identity.rs
```diff
@@ -126,6 +153,16 @@ mod handler {
...
         Ok(reply::json(&id))
     }
+
+    /// Retrieve the list of tracked identities.
+    pub async fn list_tracked_identities<R>(ctx: http::Ctx<R>) -> Result<impl Reply, Rejection>
+    where
+        R: Send + Sync,
+    {
+        let ctx = ctx.read().await;
+        let tracked_users = identity::list_tracked_identities(&ctx.peer_api)?;
```
xla, line 163:
```suggestion
        let tracked_users = identity::list_tracked(&ctx.peer_api)?;
```

proxy/src/identity.rs
```diff
@@ -85,6 +85,21 @@ pub fn get(api: &coco::Api, id: &coco::Urn) -> Result<Identity, error::Error> {
     Ok((api.peer_id(), user).into())
 }
 
+/// Retrieve the list of tracked identities.
+///
+/// # Errors
+pub fn list_tracked_identities(api: &coco::Api) -> Result<Vec<Identity>, error::Error> {
```
xla, line 91:
```suggestion
pub fn list_tracked(api: &coco::Api) -> Result<Vec<Identity>, error::Error> {
```

proxy/src/http/identity.rs
NunoAlexandre, line 163: It seems like this endpoint only operates on tracked users, while we can and will also support tracking projects.

Shouldn't we reflect that in the endpoint by having, for now, `GET /tracked/users` and later `GET /tracked/projects`?

@FintanH?

proxy/src/identity.rs
```diff
@@ -85,6 +85,21 @@ pub fn get(api: &coco::Api, id: &coco::Urn) -> Result<Identity, error::Error> {
...
+    let mut tracked_users = vec![];
+    for project in api.list_projects()? {
```
FintanH, line 93: A `flat_map` might be better here, as well as a `map` on the inner loop. Then we can just `collect`. Later if we decide to de-dupe then we can just say `collect` a `HashSet` instead of a `Vec` :tada:

proxy/src/http/identity.rs
FintanH: There's already a `/projects` endpoint. The next addition will be that we'll want something like `/projects/<user>`.

But I suppose this begs the question of "How should these endpoints be arranged?" I don't have any strong opinions or knowledge of organisation here :) cc @xla

MeBrei: I think what makes this confusing is that for the first version, we want:
1) `/projects` - all projects created by the session user
2) `tbd` - projects tracked by the session user
3) `tbd` - projects the passed identity is involved in from our local view

3 is just temporary. currently it is neither projects-the-identity-is-tracking nor all-projects-the-identity-created

FintanH: Ah ya of course, so:
1. Is what we have today
2. Is a project that you track but haven't created a working copy, i.e. `heads` is empty. (I always forget about this :facepalm:)
3. Is what we're trying to build in this new feature, right?

MeBrei: @FintanH yes to all three

proxy/src/identity.rs
```diff
@@ -88,16 +88,12 @@ pub fn get(api: &coco::Api, id: &coco::Urn) -> Result<Identity, error::Error> {
...
-        for peer in api.tracked(&project_urn)? {
-            tracked_users.push(peer.into())
-        }
-    }
-    // TODO(merle): Deduplicate `tracked_users`
-    Ok(tracked_users)
+pub fn list_tracked(api: &coco::Api) -> Result<Vec<Identity>, error::Error> {
+    let peers = api.list_projects()?.into_iter().flat_map(|project| {
+        api.tracked(&project.urn())
+            .expect("Peers could not be retrieved.")
```
FintanH, line 94: Ach! I always get annoyed when I remember iterators don't play well with `Result` :weary: I'd rather we avoid `expect` but not sure off the top of my head how you could do that nicely here.

MeBrei: yeah, I couldn't figure out a different way :/

xla: Alternativelly drop the iterators and go for a classic loop, which allows to tap into the normal error infrastructure. The only other way is for the closure to return a result, which then needs to be flattened again and so on.

FintanH: @MeBrei had a classic loop and I suggested the iterator change :'( My bad...

xla: Stop tempting the youth with your FP promises. Poor souls!

FintanH: Can't wait for the next time I realise this point again and get annoyed.

MeBrei:
I added a test to to the tracked identity. 
`list_projects` seems to return 2 projects, so that the tracked user is duplicated in the end. I kept it at that for now. My suspicion is that it is actually due to the test set up and I am wondering if we should keep it like that until we can test networking "by hand". If we deduplicate now we might be hiding a bug later on.

FintanH:
@MeBrei I think you're coming across this bug: https://github.com/radicle-dev/radicle-link/issues/250 :weary: Basically the `all_metadata` function is returning remote project references when it shouldn't. The `*` glob match isn't doing as we expected. There's a ticket to fix this in link:
* https://github.com/radicle-dev/radicle-link/issues/266

proxy/src/http/identity.rs
MeBrei: I summarized the paths in the [issue](https://github.com/radicle-dev/radicle-upstream/issues/741#issuecomment-672795977)

MeBrei:
> @MeBrei I think you're coming across this bug: [radicle-dev/radicle-link#250](https://github.com/radicle-dev/radicle-link/issues/250) 😩 Basically the `all_metadata` function is returning remote project references when it shouldn't. The `*` glob match isn't doing as we expected. There's a ticket to fix this in link:
> 
> * [radicle-dev/radicle-link#266](https://github.com/radicle-dev/radicle-link/issues/266)

Thank you for the link! Since that would mean the problem is not the test set up, I added a commit to de-duplicate `tracked_identities`.

proxy/src/http/identity.rs
```diff
@@ -18,7 +18,10 @@ pub fn filters<R>(ctx: http::Ctx<R>) -> BoxedFilter<(impl Reply,)>
 where
     R: registry::Client + 'static,
 {
-    get_filter(Arc::clone(&ctx)).or(create_filter(ctx)).boxed()
+    get_filter(Arc::clone(&ctx))
+        .or(create_filter(Arc::clone(&ctx)))
+        .or(list_tracked_filter(ctx))
```
xla, line 23:
```suggestion
        .or(list_filter(ctx))
```

proxy/src/http/identity.rs
```diff
@@ -80,6 +83,30 @@ where
         .and_then(handler::get)
 }
 
+/// `GET /`
+fn list_tracked_filter<R>(
```
xla, line 87:
```suggestion
fn list_filter<R>(
```

proxy/src/http/identity.rs
```diff
@@ -126,6 +153,16 @@ mod handler {
         let id = identity::get(&ctx.peer_api, &id)?;
         Ok(reply::json(&id))
     }
+
+    /// Retrieve the list of tracked identities.
+    pub async fn list_tracked<R>(ctx: http::Ctx<R>) -> Result<impl Reply, Rejection>
```
xla, line 158:
```suggestion
    pub async fn list<R>(ctx: http::Ctx<R>) -> Result<impl Reply, Rejection>
```

proxy/src/http/identity.rs
```diff
@@ -126,6 +153,16 @@ mod handler {
...
+    where
+        R: Send + Sync,
+    {
+        let ctx = ctx.read().await;
+        let tracked_users = identity::list_tracked(&ctx.peer_api)?;
```
xla, line 163:
```suggestion
        let users = identity::list(&ctx.peer_api)?;
```

proxy/src/http/identity.rs
```diff
@@ -342,4 +380,40 @@ mod test {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn list_tracked() -> Result<(), error::Error> {
```
xla, line 385:
```suggestion
    async fn list() -> Result<(), error::Error> {
```

proxy/src/identity.rs
```diff
@@ -85,6 +85,23 @@ pub fn get(api: &coco::Api, id: &coco::Urn) -> Result<Identity, error::Error> {
     Ok((api.peer_id(), user).into())
 }
 
+/// Retrieve the list of tracked identities.
+///
+/// # Errors
+pub fn list_tracked(api: &coco::Api) -> Result<Vec<Identity>, error::Error> {
```
xla, line 91:
```suggestion
pub fn listapi: &coco::Api) -> Result<Vec<Identity>, error::Error> {
```

proxy/src/identity.rs
```diff
@@ -85,6 +85,23 @@ pub fn get(api: &coco::Api, id: &coco::Urn) -> Result<Identity, error::Error> {
...
+    let mut tracked_users = vec![];
```
xla, line 92:
```suggestion
    let mut users = vec![];
```

proxy/src/identity.rs
```diff
@@ -85,6 +85,23 @@ pub fn get(api: &coco::Api, id: &coco::Urn) -> Result<Identity, error::Error> {
...
+    for project in api.list_projects()? {
+        let project_urn = project.urn();
+        for peer in api.tracked(&project_urn)? {
+            let user = peer.into();
+            if !tracked_users.contains(&user) {
+                tracked_users.push(user)
+            }
+        }
+    }
+    Ok(tracked_users)
```
xla, line 102:
```suggestion
    Ok(users)
```

MeBrei:
> Looks good, as said before as we might look into tracking as a public facing feature I'd remove all notion of it for now. That inside the coco domain the replication primitive is called `track` should no leak into our APIs.

Right! Removed the reference in a95ca6b

PR #764, opened 2020-08-10
chore(proxy): bump librad

kim:
Companion to radicle-dev/radicle-link#275 -- the brittle `cargo patch`ed `rustls` is no longer needed. Also, take both `librad` and `git-helpers` from the same git rev.


xla:
⏸ 🏩 🌶 📹

PR #768, opened 2020-08-11
feat(proxy): use Signer trait

### LINKED ISSUE #753: Parameterize Api on Signer

@FintanH in https://github.com/radicle-dev/radicle-upstream/pull/752/commits/4cde468a22fef61f8ab7408f008c73b9a2441aba

xla:
Closes #753


cypress/support/commands.js
```diff
@@ -1,30 +1,19 @@
...
-});
-
-Cypress.Commands.add("nukeRegistryState", async () => {
-  console.log("Nuking Registry state");
-  await fetch("http://localhost:8080/v1/control/nuke/registry");
-});
-
-Cypress.Commands.add("nukeSessionState", async () => {
-  console.log("Nuking Session state");
+Cypress.Commands.add("reseetSession", async () => {
```
sarahscott, line 6:
```suggestion
Cypress.Commands.add("resetSession", async () => {
```

package.json
```diff
@@ -131,7 +131,7 @@
     "prettier:check": "yarn prettier --check",
     "prettier:write": "yarn prettier --write",
     "lint": "eslint . --ignore-path .gitignore --ext .js,.svelte,.ts",
-    "nuke": "cd proxy && cargo run --bin nuke"
+    "reest": "cd proxy && cargo run --bin reset"
```
sarahscott, line 134:
```suggestion
    "resest": "cd proxy && cargo run --bin reset"
```

proxy/src/coco/signer.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// The underlying [`FileStorage`].
+    storage: Storage,
+}
+
+impl Store {
+    /// Sets up a new [`Store`] with a fresh key.
+    ///
+    /// # Errors
+    ///
+    /// * if the file system errors for any other reason than [`file::Error::NoSuckKey`]
```
sarahscott, line 59:
```suggestion
    /// * if the file system errors for any other reason than [`file::Error::NoSuchKey`]
```

rudolfs: lol

proxy/src/coco/signer.rs
```diff
@@ -0,0 +1,133 @@
...
+/// Synonym for an error when interacting with a store for [`librad::keys`].
+pub type Error = file::Error<crypto::SecretBoxError<Infallible>, keys::IntoSecretKeyError>;
+
+/// Parameterised [`FileStorage`] used in Store implementation.
+type Storage = FileStorage<
+    crypto::Pwhash<SecUtf8>,
+    keys::PublicKey,
+    keys::SecretKey,
+    <keys::SecretKey as SecretKeyExt>::Metadata,
+>;
```
rudolfs, line 43: What's up with the weird line break?

rudolfs (approved):
Sign all the things!!

There's one more reference to "nuke", perhaps it makes sense to rename it:
```
proxy/src/http/control.rs
242:    async fn create_project_after_nuke() -> Result<(), error::Error> {
254:        // Create project before nuke.
```

proxy/src/coco/signer.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        let key = match storage.get_key() {
+            Ok(key) => Ok(key.secret_key),
+            Err(err) => match err {
+                file::Error::NoSuchKey => {
+                    let key = keys::SecretKey::new();
+                    storage.put_key(key.clone())?;
+                    Ok(key)
+                },
+                file::Error::KeyExists
```
FintanH, line 74: Can we not do:

```rust
err => Err(err)
```

FintanH (commented):
Nice work! We're off to a great start with key management :)

proxy/src/coco/signer.rs
xla: Which line break are you referring to?

FintanH (dismissed):
Thank you for your time and work :heart:

proxy/src/coco/signer.rs
rudolfs: Ah, never mind.

proxy/src/coco/signer.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// * if the write of the newly generated key fails
+    fn reset(&mut self, paths: &paths::Paths, passphrase: SecUtf8) -> Result<(), Error>;
+}
+
+/// Synonym for an error when interacting with a store for [`librad::keys`].
+pub type Error = file::Error<crypto::SecretBoxError<Infallible>, keys::IntoSecretKeyError>;
+
+/// Parameterised [`FileStorage`] used in Store implementation.
+type Storage = FileStorage<
+    crypto::Pwhash<SecUtf8>,
```
kim, line 39: I suppose this is still temporary? Because, keeping the passphrase in memory is as good as keeping the secret key itself in memory.

rudolfs (approved):
Wondering if we should have a const for the fake encryption password: `let pw = signer::SecUtf8::from("radicle-upstream");`, as it pops up in multiple places.

kim (commented):
Did you consider storing the `Signer` as a trait object in `Api`? It seems quite a mouthful that almost the entire HTTP API needs to be parametrised over it.

I'm also not entirely convinced of conflating key storage and `Signer` impl: since you're keeping the passphrase in memory, there isn't really a benefit of reading the key from disk every time `sign` is called. I would have expected the secret key being read from disk at startup (or whenever the passphrase is available), and then passed around (as a `Signer`).

proxy/src/coco/signer.rs
```diff
@@ -0,0 +1,130 @@
...
+/// Filename of the key on disk.
+const KEY: &str = "librad.key";
+
+/// Blanket trait to use as our generic [`signer::Signer`].
+pub trait Signer: Clone + keys::AsPKCS8 + signer::Signer {}
+
+impl<T: Clone + keys::AsPKCS8 + signer::Signer> Signer for T {}
+
+/// Utility trait to manage state for an exisitng [`Signer`].
+pub trait Reset: Signer {
```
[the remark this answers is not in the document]
xla: It is a footgun, but we have the requirement for a reset in non-test operations. Where the cypress integration tests can reset the proxy state, including creating a new monorepo and so on. This is purely in support for the UI tests to assert proper behaviour in our user flows. Haven't found a better way to model this.

proxy/src/coco/signer.rs
```diff
@@ -0,0 +1,130 @@
...
+use radicle_keystore::crypto;
+use radicle_keystore::file;
+use radicle_keystore::sign::{self, Signer as _};
+use radicle_keystore::{FileStorage, Keystore, SecretKeyExt};
+
+pub use librad::keys::SignError;
+pub use radicle_keystore::pinentry::SecUtf8;
+
+/// Filename of the key on disk.
+const KEY: &str = "librad.key";
```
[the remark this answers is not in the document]
xla: :+1: Generally we should audit all paths under `~/.radicle` and see what needs to be shared so there is no accidental divergence or overlap.

xla:
> Did you consider storing the `Signer` as a trait object in `Api`? It seems quite a mouthful that almost the entire HTTP API needs to be parametrised over it.

I have not yet. And I agree this change is so invasive and had to touch too many places in unrelated places. Is there any obvious downside to use a **trait object** here instead of a type parameter?

> I'm also not entirely convinced of conflating key storage and `Signer` impl: since you're keeping the passphrase in memory, there isn't really a benefit of reading the key from disk every time `sign` is called. I would have expected the secret key being read from disk at startup (or whenever the passphrase is available), and then passed around (as a `Signer`).

That was my initial initution as well until @FintanH had me convinced that we don't want to keep the secret key in memory as it is an attack vector and can potentially leak the actual bytes of the key.

proxy/src/coco/signer.rs
xla: See https://github.com/radicle-dev/radicle-upstream/pull/768#issuecomment-676497331

kim:
> Is there any obvious downside to use a **trait object** here instead of a type parameter?

It is probably the case that the compiler wants you to specify a concrete type for `Signer::Error`, so you need to box that too. See: https://github.com/radicle-dev/radicle-link/blob/master/librad/src/signer.rs, perhaps this code can be used verbatim.

> we don't want to keep the secret key in memory as it is an attack vector and can potentially leak the actual bytes of the key.

Well, if that key is always just a `Signer`, the program shouldn't be able to accidentally leak it. In general, program memory is the safest place, really, modulo Rowhammer, Spectre, Meltdown and friends. If one can mount one of these, one can just try to get at the passphrase -- so we're screwed either way. Since typing your passphrase is also not tinfoil-hat proof, asking for it _every time_ the key is used doesn't help either (it would if we could rely on getting it via TouchID, as that's protected by a TPM ("secure enclave")).

So, typing your passphrase once, reading and decrypting the key from disk, and keeping it in memory is okay -- it's basically what the `ssh` or `gpg` agents are doing. To make it fancy, one could drop the key after some timeout, requiring the user to re-enter their passphrase.

FintanH:
Ah, then that's my fault for leading xla down the wrong path :sweat: I was under the impression having to unlock the store, at least periodically, made sense.

You can probably just use `SomeSigner` as the trait object.

kim:
> I was under the impression having to unlock the store, at least periodically, made sense

That’s ... what I said?

FintanH:
> That’s ... what I said?

Sorry, what I meant was that the user would periodically unlock the store throughout the lifetime of the program. Whereas, what you said was:

> So, typing your passphrase once, reading and decrypting the key from disk, and keeping it in memory is okay

Are we on the same page now? Or am I misunderstanding something?

xla:
Grossly outdated, lots of learnings and will be super-seeded by a more focused change-set

Issue #696, opened 2020-07-20
Fix outstanding eslint TS errors


rudolfs:
The [upgrade](https://github.com/radicle-dev/radicle-upstream/pull/694) to the latest versions of our UI dependencies made some typescript lints to be treated as errors.

[We marked some hard-to-fix ones as ignored](https://github.com/radicle-dev/radicle-upstream/commit/58e1a28a5d4a80598fe03471d6f8434f70d77d4c):

```
work/radicle-upstream » g eslint-disable                                                                                                                                                                                                               ‹rudolfs/ui-dep-upgrade›
ui/src/validation.ts
61:  // eslint-disable-next-line @typescript-eslint/unbound-method
72:    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
81:        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
116:              // eslint-disable-next-line @typescript-eslint/restrict-template-expressions, @typescript-eslint/no-unsafe-member-access

ui/src/api.ts
29:  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
37:  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
45:    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment

ui/src/session.ts
142:    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
150:    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment

ui/src/transaction.ts
185:  // eslint-disable-next-line @typescript-eslint/no-unsafe-return

ui/src/session.test.ts
48:    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
70:    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
74:    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
78:    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access

ui/src/remote.ts
51:  // eslint-disable-next-line @typescript-
```

These should be fixed properly.

geigerzaehler:
We don’t want to make a consolidated effort to fix this. We’ll fix things as we come across them.

PR #769, opened 2020-08-11
feat: set up remote helper

### LINKED ISSUE #763: Set up rad-remote-helper on app start
We need the `rad-remote-helper` binary to be installed locally for the checkout feature to work. It is not enough to have the binary be relative to the proxy binary as on Linux the AppImage package gets unmounted when the app is not running. And we need the `rad-remote-helper` binary to be in a stable path so users can interact with the monorepo even when Upstream is not running.

To provide a nice UX we'll copy the `rad-remote-helper` binary to a stable location: `~/.radicle/bin` upon first start of Upstream.

Scope of this issue:
1. copy helper into `~./radicle/bin` on proxy startup if it's not there, make sure it's `+x`
2. add hint to the checkout modal about how to set up your environment.
3. add hint to the project creation flow that instructs the user to set up the path.

Note: a simple check if the binary is in the path is sufficient for now, we'll think about upgrades when we get there.

Follow up for: https://github.com/radicle-dev/radicle-upstream/pull/722.

rudolfs:
hint

Screenshot 2020-08-14 at 16 16 32

Proxy startup when the rad helper is missing:
```
 INFO  proxy::coco::git_helper > Making sure git-remote-rad helper is set up
 DEBUG proxy::coco::git_helper > Copied git-remote-rad helper to: "/Users/rudolfs/.radicle/bin/git-remote-rad"
 INFO  proxy                   > Starting API
```

Proxy startup when the rad helper is in place:
```
 INFO  proxy::coco::git_helper > Making sure git-remote-rad helper is set up
 DEBUG proxy::coco::git_helper > Git helper already exists at: "/Users/rudolfs/.radicle/bin/git-remote-rad"
 INFO  proxy                   > Starting API
```

- [x] copy helper into ~./radicle/bin on proxy startup if it's not there, make sure it's executable
- [x] add hint to the checkout modal about how to set up your environment
- [x] add hint to the project creation flow that instructs the user to set up the path
- [x] remember dismissed hints
- [x] tests

Closes #763.


proxy/src/coco/git_helper.rs
```diff
@@ -0,0 +1,57 @@
...
+/// # Errors
+///
+///   * Could not get the path to directory where helper binaries should be stored.
+///   * Could not get the current working directory.
+///   * Could not create the path to binary directory.
+///   * Could not copy helper executable to the binary directory.
+pub fn setup() -> Result<(), error::Error> {
+    log::info!("Making sure git-remote-rad helper is set up");
+
+    let helper_binary_name = "git-remote-rad";
```
FintanH, line 40: This can be a `pub const` at the top of the file.

proxy/src/coco/git_helper.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let bin_dir = config::bin_dir()?;
+    let full_dest_path = bin_dir.join(helper_binary_name);
+
+    if full_dest_path.is_executable() {
```
FintanH, line 45: Wait, how does this confirm the file path exists?

FintanH: Or else the debug message below is confusing me

proxy/src/config.rs
```diff
@@ -1,9 +1,36 @@
...
 pub fn dirs() -> ProjectDirs {
     ProjectDirs::from("xyz", "radicle", "radicle-upstream").expect("couldn't build dirs")
 }
+
+/// Returns the path to a folder containing helper binaries.
+///
+/// # Errors
+///
+///   * Could not get the user home path from the HOME env variable
+pub fn bin_dir() -> Result<path::PathBuf, error::Error> {
```
FintanH, line 18: Nice :+1:

proxy/src/project.rs
```diff
@@ -214,12 +215,11 @@ where
 
     /// Set up the PATH env variable used for running the checkout.
     fn default_bin_path() -> Result<ffi::OsString, error::Error> {
```
FintanH, line 217: Maybe this could go in `config` too?

proxy/src/coco/git_helper.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Trait for extending the Path module.
+trait IsExecutable {
+    /// Returns true if it is an executable.
+    fn is_executable(&self) -> bool;
+}
+
+/// Implements isExecutable for Path.
+impl IsExecutable for Path {
+    fn is_executable(&self) -> bool {
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/769/commits/aefa346c5b1cf35fb3073ee029d80e24cb10ea67

proxy/src/coco/git_helper.rs
rudolfs: The existence of the path was checked in the `Err` arm of this match:
https://github.com/radicle-dev/radicle-upstream/pull/769/commits/25737acf7f801d7ff182e548b8e8e02b3bc8fb06#diff-9effc3f34cdc6f654db167f9ac8ebd2eR20


Improved the readability in:
https://github.com/radicle-dev/radicle-upstream/pull/769/commits/aefa346c5b1cf35fb3073ee029d80e24cb10ea67

proxy/src/coco/git_helper.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/769/commits/7c249eaa129c1cfd5996b6b76fcee1ffaf6a2046

proxy/src/coco/git_helper.rs
```diff
@@ -0,0 +1,46 @@
+//! git-remote-rad git helper related functionality.
+
+use crate::config;
+
+use crate::error;
+use std::fs;
+use std::os::unix::fs::PermissionsExt;
+
+/// Filename of the git helper binary.
+pub const HELPER_BINARY_NAME: &str = "git-remote-rad";
```
FintanH, line 10: Nit: I'd call it `GIT_REMOTE_RAD`. The name here is too generic.

proxy/src/coco/git_helper.rs
```diff
@@ -0,0 +1,46 @@
...
+    let full_dest_path = bin_dir.join(HELPER_BINARY_NAME);
+
+    if full_dest_path.exists() && is_executable(&full_dest_path)? {
+        log::debug!("Git helper already exists at: {:?}", full_dest_path);
+        return Ok(());
+    }
+
+    let proxy_path = config::proxy_path()?;
+
+    fs::create_dir_all(bin_dir)?;
```
FintanH, line 41: So what happens if it exists and is not executable? I imagine this will blow up.

proxy/src/coco/git_helper.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/769/commits/f9155636d1ef080844cb06df7c1ee1c3fbe4903f

proxy/src/coco/git_helper.rs
rudolfs: It works just fine:

"Recursively create a directory and all of its parent components **if they are missing**."
https://doc.rust-lang.org/std/fs/fn.create_dir_all.html

FintanH: Ah true, it won't blow up. But we don't actually have an executable :)

FintanH: Actually, I'd also think that the `copy` below would fail. Because if you `copy foo/existing_file bar/existing_file` it'll break, no?

proxy/src/project.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/769/commits/0bcf74272d1853758f9d9b0296eaad4304423b27

proxy/src/coco/git_helper.rs
rudolfs: I tested these cases manually and they all pass:
- no path exists
- no binary exists at path
- binary not executable at path

But I probably should just write tests for these...

FintanH: > binary not executable at path

How did you test this? Like would it mean that this code passes but we can't actually use the remote helper? :thinking: 

> But I probably should just write tests for these...

That'd be good :+1: You can add:
```rust
#[cfg(test)]
mod tests {
}
```
in this file :)

rudolfs: > How did you test this?

Manually changed permissions of the helper:
```
» cd ~/.radicle/bin
» ls -la git-remote-rad
-rwxr-xr-x  1 rudolfs  staff  8731832 Aug 17 10:29 git-remote-rad
» chmod -x git-remote-rad
-rw-r--r--  1 rudolfs  staff  8731832 Aug 17 10:29 git-remote-rad
```

then started the app via `yarn start`:

```
» RUST_LOG=proxy=debug yarn start                                                                                                                                                                                           INFO  proxy::coco::git_helper > Making sure git-remote-rad helper is set up
DEBUG proxy::coco::git_helper > Copied git-remote-rad helper to: "/Users/rudolfs/.radicle/bin/git-remote-rad"
```

and checked that the remote helper is executable after the app had started:
```
» ls -la git-remote-rad
-rwxr-xr-x  1 rudolfs  staff  8731832 Aug 17 10:29 git-remote-rad
```

> Like would it mean that this code passes but we can't actually use the remote helper? 🤔

It means the git remote helper is in place and working as intended.

FintanH: Ah cool :+1: Thanks for the details :)

proxy/src/coco/git_helper.rs
```diff
@@ -0,0 +1,46 @@
+//! git-remote-rad git helper related functionality.
+
+use crate::config;
+
+use crate::error;
+use std::fs;
+use std::os::unix::fs::PermissionsExt;
```
xla, line 7:
```suggestion
use std::fs;
use std::os::unix::fs::PermissionsExt;

use crate::config;
use crate::error;
```

We order `std -> third party crates -> radicle crates -> workspace crates -> crate modules`

proxy/src/coco/git_helper.rs
```diff
@@ -0,0 +1,46 @@
...
+///   * Could not get the current working directory.
+///   * Could not create the path to binary directory.
+///   * Could not copy helper executable to the binary directory.
+pub fn setup() -> Result<(), error::Error> {
+    log::info!("Making sure git-remote-rad helper is set up");
+
+    let bin_dir = config::bin_dir()?;
+    let full_dest_path = bin_dir.join(GIT_REMOTE_RAD);
+
+    if full_dest_path.exists() && is_executable(&full_dest_path)? {
```
xla, line 34: Should we beyond existance check also validate that it is the binary we expect?

proxy/src/config.rs
```diff
@@ -1,9 +1,56 @@
 //! Configuration vital to the setup and alteration of the application.
 
+use crate::error;
 use directories::ProjectDirs;
+use std::ffi;
+use std::path;
```
xla, line 6: Import ordering.

proxy/src/config.rs
```diff
@@ -1,9 +1,56 @@
...
+
+/// Returns the path to a folder containing helper binaries.
+///
+/// # Errors
+///
+///   * Could not get the user home path from the HOME env variable
+pub fn bin_dir() -> Result<path::PathBuf, error::Error> {
+    let home_dir = std::env::var("HOME").expect("Couldn't determine home dir from env");
+
+    Ok(path::Path::new(&home_dir).join(".radicle/bin"))
```
xla, line 22: As we occupy a shared namespace with other crates this should be extracted and treated as a shared resource. Similar to how we use the `librad` to get a the location of the monorepo.

ui/src/session.ts
```diff
@@ -151,6 +151,15 @@ export const updateAppearance = (appearance: Appearance): void =>
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     settings: { ...get(settings), appearance },
   });
+
+export const dismissRemoteHelperHint = (): void => {
+  updateAppearance({
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
xla, line 157: What does this exception do?

ui/DesignSystem/Component/RemoteHelperHint.svelte
```diff
@@ -0,0 +1,53 @@
+<script>
+  import { createEventDispatcher } from "svelte";
+
+  import { Icon } from "../../DesignSystem/Primitive";
+  import Copyable from "./Copyable.svelte";
+
+  const dispatch = createEventDispatcher();
+
+  export let style = null;
+  export let hidden = false;
```
xla, line 10: Instead of having a prop the calling side can decide to wrap the component in an if block and therefore not even invoke it.

ui/DesignSystem/Component/RemoteHelperHint.svelte
```diff
@@ -0,0 +1,53 @@
...
+    <a
+      style="color: var(--color-foreground-level-6);"
+      class="typo-link"
+      href="https://radicle.xyz/#upstream-faq">
+      Read more
+    </a>
+    .
+  </p>
+  <Copyable>
+    <p class="typo-text-small-mono">export PATH="~/.radicle/bin:$PATH"</p>
```
xla, line 51: Are we planning to only make it easy for users of `bash` and `zsh`? Or could we have a toggle for different shells with the line to copy chanigng accordingly?

FintanH:
> Dope! Generally we should aim to not depend on an extra binary for proxy operations, which is independent of the changes in this PR, just want to raise it and see if we have a follow-up issue for that.

So, I think @rudolfs and I strayed a bit when it came to understanding how the rad remote helper works. We don't actually need the binary in proxy, since we can just call `transport::register` and then it gets handled for us. The binary is necessary if you're doing actions outside of Upstream, i.e. `git push rad` from the CLI

xla:
> So, I think @rudolfs and I strayed a bit when it came to understanding how the rad remote helper works. We don't actually need the binary in proxy, since we can just call `transport::register` and then it gets handled for us. The binary is necessary if you're doing actions outside of Upstream, i.e. `git push rad` from the CLI

I understand that, but currently we use it as well when doing a checkout through upstream and this is what my comment was directed at.

FintanH:
> I understand that, but currently we use it as well when doing a checkout through upstream and this is what my comment was directed at.

Ya, so what I'm saying is that I think we might not have to do that. I'm doing an MVP for doing it through `register` as we speak

xla:
> Ya, so what I'm saying is that I think we might not have to do that. I'm doing an MVP for doing it through `register` as we speak

:black_heart:

proxy/src/coco/git_helper.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/769/commits/3fa334ad3f38b15e86e8433d6794c4bea4301b1d

proxy/src/coco/git_helper.rs
```diff
@@ -25,22 +24,117 @@ fn is_executable(path: &std::path::PathBuf) -> Result<bool, error::Error> {
...
+    use crate::coco;
+    use crate::error;
+    use std::fs;
+    use std::os::unix::fs::PermissionsExt;
+
+    // when the ~/.radicle/bin directory doesn't exist at all
+    #[tokio::test]
+    async fn setup_creates_destination_directory_if_none_exists() -> Result<(), error::Error> {
+        let tmp_src_dir = tempfile::tempdir().expect("failed to create source tempdir");
+        let src_git_helper_bin_path = tmp_src_dir.path().join(coco::git_helper::GIT_REMOTE_RAD);
```
FintanH, line 57: fyi: you can use `super` :)

proxy/src/coco/git_helper.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/769/commits/879708bcd37f25d6839658f1aeec2fec61992cb3

proxy/src/config.rs
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/769/commits/879708bcd37f25d6839658f1aeec2fec61992cb3

proxy/src/coco/git_helper.rs
rudolfs: I think for the first iteration it is sufficient to just make sure it's there and executable.

proxy/src/coco/git_helper.rs
rudolfs: Good point!

https://github.com/radicle-dev/radicle-upstream/pull/769/commits/9944006e19da6edde1c0754b73a1e364310b5568

ui/DesignSystem/Component/RemoteHelperHint.svelte
rudolfs: Good point!
https://github.com/radicle-dev/radicle-upstream/pull/769/commits/a12c2b846a35b815ad8bec0c945a1d1e9962eae0

ui/DesignSystem/Component/RemoteHelperHint.svelte
rudolfs: I think we should keep it simple initially and add different options later on, if there is demand for it.

xla: Personally I use `fish`. :>

Jokes aside, it's not worth the effort, full steam ahead.

rudolfs: > Personally I use fish. :>

I know :)

><(((º>

proxy/src/config.rs
rudolfs: You mean: `pub const RADICLE_BIN_HOME: &str = ".radicle/bin";`?

ui/src/session.ts
rudolfs: It's a leftover from our packaging upgrade. Should be tackled in: https://github.com/radicle-dev/radicle-upstream/issues/696.

Read more about it [here](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-member-access.md.).

proxy/src/config.rs
xla: More like `~/.radicle` is a shared directory across multiple crates. Flagging this here that this might be concern for the entire radicle-* ecosystem.

PR #779, opened 2020-08-15
chore(ui): attempt to get ts linted

sarahscott:
Currently, `eslint-plugin-svelte3` doesn't play nice with typescript. In order to use typescript in our components and also lint them, we need to preprocess before running through `eslint`.

see https://github.com/sveltejs/eslint-plugin-svelte3/pull/62


xla:
The committed `package.json` refers to a local location of `eslint-plugin-svelte`, therefore the tests fail and local checkout is also not helpful.

sarahscott:
@xla yup, it's still a draft. i'm working with a local version because [the branch with a potential fix can't be simply pointed to with npm](https://github.com/NicoCevallos/eslint-svelte3-preprocess-test/blob/master/package.json#L9).

I'll ensure this passes tests & works locally before marking as ready for review.

PR #783, opened 2020-08-17
doc: update test preparation in developer manual

geigerzaehler:


DEVELOPMENT.md
```diff
@@ -74,8 +74,11 @@ Before running UI end-to-end tests locally you'll need to check out the latest
 test fixtures which are included in this repository as a git submodule:
 
 ```sh
-git submodule update --init --remote
+git submodule update --remote --init --recursive
```
xla, line 77: What are the submodules of submoduels we have to check out?

rudolfs: You should know, as you added it:
https://github.com/radicle-dev/radicle-upstream/commit/bc9b818c74408d76870f57ec7ab20f832f59865e#diff-df1e26e5884898d9c34574e3c5ed313cR40

xla: :exploding_head:

xla: Think we don't need it `git-platinum` doesn't have any subs. `--init --remote` should suffice.

rudolfs: Then we should also update it in `.buildkite/run.sh`.

geigerzaehler: I don’t think there are any. I just copy pasted things from `.buildkite/run.sh` to have the least surprises.

xla: Would be cool to remove it from both then, sorry for the extra work.

PR #789, opened 2020-08-18
feat(proxy): filter projects by user URN

### LINKED ISSUE #741: Known identities API endpoints

geigerzaehler:
We add the `user` query parameter to the `GET /projects` endpoint. If provided only projects that are tracked by the given user are returned.

Closes #741

### DIFF
--- a/proxy/src/http.rs
+++ b/proxy/src/http.rs
@@ -1,7 +1,6 @@
 //! HTTP API delivering JSON over `RESTish` endpoints.
 
 use std::collections::HashMap;
-use std::convert::Infallible;
 use std::sync::Arc;
 
 use serde::{Deserialize, Serialize};
--- a/proxy/src/http.rs
+++ b/proxy/src/http.rs
@@ -1,6 +1,7 @@
 //! HTTP API delivering JSON over `RESTish` endpoints.
 
 use std::collections::HashMap;
+use std::convert::Infallible;
 use std::sync::Arc;
 
 use serde::{Deserialize, Serialize};
--- a/proxy/src/http/project.rs
+++ b/proxy/src/http/project.rs
@@ -701,6 +701,7 @@ mod test {
     }
 
     #[tokio::test]
+    #[allow(clippy::indexing_slicing)]
     async fn list_for_user() -> Result<(), error::Error> {
         let tmp_dir = tempfile::tempdir()?;
         let ctx = http::Context::tmp(&tmp_dir).await?;
@@ -711,7 +712,6 @@ mod test {
         let owner = ctx.peer_api.init_owner(&key, "cloudhead")?;
 
         coco::control::setup_fixtures(&ctx.peer_api, &key, &owner)?;
-        #[allow(clippy::indexing_slicing)]
         let project = &project::list_projects(&ctx.peer_api)?[0];
         let librad_project = ctx.peer_api.get_project(&project.id, None)?;
 
--- a/proxy/src/http/project.rs
+++ b/proxy/src/http/project.rs
@@ -125,7 +125,7 @@ where
     http::with_context(ctx)
         .and(warp::get())
         .and(path::end())
-        .and(http::query_default::<ListQuery>())
+        .and(http::with_qs_opt::<ListQuery>())
         .and(document::document(
             document::query("user", document::string())
                 .required(false)
@@ -245,11 +245,12 @@ mod handler {
     /// If [`super::ListUser::user`] is given we only return projects that this user tracks.
     pub async fn list<R>(
         ctx: http::Ctx<R>,
-        query: super::ListQuery,
+        opt_query: Option<super::ListQuery>,
     ) -> Result<impl Reply, Rejection>
     where
         R: Send + Sync,
     {
+        let query = opt_query.unwrap_or_default();
         let ctx = ctx.read().await;
 
         let projects = if let Some(user) = query.user {
--- a/proxy/src/http/project.rs
+++ b/proxy/src/http/project.rs
@@ -125,6 +125,12 @@ where
     http::with_context(ctx)
         .and(warp::get())
         .and(path::end())
+        .and(http::query_default::<ListQuery>())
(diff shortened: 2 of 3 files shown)


proxy/src/http/project.rs
```diff
@@ -137,6 +142,7 @@ where
             )
             .description("Creation succeeded"),
         ))
+        .and(http::query_default::<ListQuery>())
```
xla, line 145: This should go up to where the verb and path is defined.

proxy/src/http/project.rs
```diff
@@ -488,6 +522,13 @@ impl ToDocumentedType for MetadataInput {
     }
 }
 
+/// Query options for listing projects
+#[derive(Debug, Serialize, Deserialize, Default)]
```
xla, line 526:
```suggestion
#[derive(Debug, Default, Deserialize, Serialize)]
```

proxy/src/http/project.rs
```diff
@@ -488,6 +522,13 @@ impl ToDocumentedType for MetadataInput {
     }
 }
 
+/// Query options for listing projects
```
xla, line 525:
```suggestion
/// Query options for listing projects.
```

proxy/src/http/project.rs
```diff
@@ -676,6 +717,38 @@ mod test {
...
+        let tmp_dir = tempfile::tempdir()?;
+        let ctx = http::Context::tmp(&tmp_dir).await?;
+        let api = super::filters(ctx.clone());
+
+        let ctx = ctx.read().await;
+        let key = ctx.keystore.get_librad_key()?;
+        let owner = ctx.peer_api.init_owner(&key, "cloudhead")?;
+
+        coco::control::setup_fixtures(&ctx.peer_api, &key, &owner)?;
+        #[allow(clippy::indexing_slicing)]
```
xla, line 731: Let's keep attributes on the function.

proxy/src/http/project.rs
```diff
@@ -676,6 +717,38 @@ mod test {
...
+        let project = &project::list_projects(&ctx.peer_api)?[0];
+        let librad_project = ctx.peer_api.get_project(&project.id, None)?;
+
+        let fintohaps: identity::Identity =
+            coco::control::track_fake_peer(&ctx.peer_api, &key, &librad_project, "fintohaps")
+                .into();
+
+        let res = request()
+            .method("GET")
+            .path(&format!("/?user={}", fintohaps.urn))
```
xla, line 741: Are there no characters which need encoding in our URNs?

proxy/src/http.rs
```diff
@@ -246,12 +247,23 @@ where
         .boxed()
 }
 
+#[must_use]
+/// Similar to [`warp::filters::query::query`] but provides the default value if
+/// the query string cannot be parsed.
```
xla, line 252:
```suggestion
/// Similar to [`warp::filters::query::query`] but provides the default value if
/// the query string cannot be parsed.
#[must_use]
```

proxy/src/http.rs
```diff
@@ -246,12 +247,23 @@ where
         .boxed()
 }
 
+/// Similar to [`warp::filters::query::query`] but provides the default value if
+/// the query string cannot be parsed.
+#[must_use]
+pub fn query_default<T>() -> BoxedFilter<(T,)>
+where
+    for<'de> T: Deserialize<'de> + Default + Send + Sync + 'static,
```
FintanH, line 255: Instead of using `Default` can we pass in the default as a parameter? I find that more useful than (sometimes) arbitrary default impls.

proxy/src/http/project.rs
FintanH: Nope. We do this everywhere already as well.

proxy/src/http.rs
geigerzaehler: Fixed in #791

proxy/src/http/project.rs
xla: Still needs moving.

proxy/src/http/project.rs
```diff
@@ -235,12 +241,40 @@ mod handler {
...
+        super::ListQuery { user }: super::ListQuery,
+    ) -> Result<impl Reply, Rejection>
     where
         R: Send + Sync,
     {
         let ctx = ctx.read().await;
-        let projects = project::list_projects(&ctx.peer_api)?;
+        let all_projects = project::list_projects(&ctx.peer_api)?;
+
+        let projects = if let Some(user) = user {
```
xla, line 256: Usually finding large code blocks like this in handlers is a sign that the handler does too much, we want to treat them as light translations from the HTTP protocol into our core domain/implementations. This would be better suited to be in `project::list_projects`.

rudolfs (approved):

proxy/src/http.rs
FintanH: Should this not be removed because of #791?

proxy/src/http/project.rs
FintanH: Seems like this needs to be re-fixed

FintanH (approved):
Nice work :+1: The cleanness of these changes is something to be admired :bow:

rudolfs (approved):

geigerzaehler:
Another "final" fix :sweat:

FintanH (approved):

rudolfs (approved):

xla (approved):
⚗ 🍗 🍣 🎪

PR #791, opened 2020-08-18
refactor(proxy): Improve query string parsing

### LINKED ISSUE #788: Invalid query results in panic and empty server reply
An invalid query string for the `/tree` and `/blob` endpoints results in a panic of the handling task and immediately closes the connection without sending a reply. The expected behavior is for the server to reply with status code 400.

The cause for this behavior is the `http::with_qs` function. The problem does not seem easy to fix because we can’t really produce warp rejections with a payload on a 400 status code :(

geigerzaehler:
* Add a filter `with_qs_opt()` that handles optional query parameters. This will be used in #789
* Reject with routing error instead of panicking if query cannot be parsed. Fixes #788

### DIFF
--- a/proxy/src/http.rs
+++ b/proxy/src/http.rs
@@ -223,25 +223,63 @@ impl Context<registry::Cacher<registry::Registry>> {
     }
 }
 
-/// Deserialise a query string using [`serde_qs`]. This is useful for more complicated query
-/// structures that involve nesting, enums, etc.
+/// Parses an optional query string with [`serde_qs`] and returns the result.
+///
+/// If no query string is present (i.e. `?` is not included in the path) `None`
+/// is returned.
+///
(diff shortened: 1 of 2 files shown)


proxy/src/http.rs
```diff
@@ -223,26 +223,64 @@ impl Context<registry::Cacher<registry::Registry>> {
...
+/// If no query string is present (i.e. `?` is not included in the path) `None`
+/// is returned.
+///
+/// This filter is different from [`warp::filters::query::query`]. It is able to
+/// handle the absence of a query string and can deserialize more complex
+/// structures.
+///
+/// # Errors
+///
+/// If  the rejects with [`http::error::Routing::InvalidQuery`].
```
rudolfs, line 237: Could you rephrase that sentence?

Issue #785, opened 2020-08-17
Connect to seeds on startup


cloudhead:
On startup, the proxy should connect to the configured seed nodes, and announce its URNs. It should remain connected so that it can receive updates.

cloudhead:
Closed by #792

PR #792, opened 2020-08-18
feat: coco seeds on startup

### LINKED ISSUE #790: Turn off spell-check in some fields
Certain fields trigger spell-check, such as the "seeds list" textbox in settings. We should make sure to disable it in fields where it doesn't make sense.

cloudhead:
The seeds specified in settings are passed into coco, so that the peer can connect to something on startup.

Instead of using DNS, we make things more flexible by using a `<peer-id>@<host>:<port>` syntax. This allows seed nodes to use static IPs instead of DNS names.

Closes #790 #785


ui/DesignSystem/Primitive/Input/Textarea.svelte
```diff
@@ -87,6 +88,7 @@
   <textarea
     data-cy={dataCy}
     class:invalid={validation && validation.status === ValidationStatus.Error}
+    spellcheck={spellcheck ? 'true' : 'false'}
```
rudolfs, line 91:
```suggestion
{spellcheck}
```

proxy/src/error.rs
```diff
@@ -80,6 +80,10 @@ pub enum Error {
     #[error("the Project Name '{0}' is invalid")]
     InvalidProjectName(String),
 
+    /// Seed input is invalid.
+    #[error("the seed '{0}' is invalid: {:1}")]
+    InvalidSeed(String, Option<librad::peer::conversion::Error>),
```
FintanH, line 85: Again, I think having some separation like a `seed` module would be good where that module has an `Error` type. Trying to move away from dumping _all_ errors in here, but instead composing it of smaller error variants.

proxy/src/coco/config.rs
```diff
@@ -82,3 +82,72 @@ pub fn configure(
...
 }
+
+/// Resolve seed identifiers into `(PeerId, SocketAddr)` pairs.
+///
+/// The expected format is `<peer-id>@<host>:<port>`
+///
+/// # Errors
+///
+/// If any of the supplied seeds cannot be parsed, an error is returned.
+pub async fn resolve_seeds<T: AsRef<str> + Send + Sync>(
```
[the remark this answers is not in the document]
cloudhead: Hmm wouldn't it be a `Seeds` struct? Ie. if we want to avoid having `resolve_seeds`. Otherwise we can keep this function and call `Seed::from` for each seed..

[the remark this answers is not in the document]
FintanH: Ah so the `str` is the new-line separated seeds? I would probably break it down into `FromStr` for a single `Seed` then maybe have a `parse_seeds: AsRef<str> -> Result<Vec<Seed>, Error>` where it splits on new-line and parses individual lines.

proxy/Cargo.toml
```diff
@@ -33,7 +33,7 @@ serde_qs = "0.6"
 syntect = "4.2"
 tempfile = "3.1"
 thiserror = "1.0"
-tokio = { version = "0.2", features = [ "macros", "time" ] }
+tokio = { version = "0.2", features = [ "macros", "time", "dns" ] }
```
xla, line 36:
```suggestion
tokio = { version = "0.2", features = [ "dns", "macros", "time" ] }
```

proxy/src/session.rs
```diff
@@ -266,7 +276,7 @@ pub mod settings {
     /// `CoCo` config parameters subject to user preferences
     #[derive(Debug, Deserialize, Serialize, PartialEq)]
     pub struct CoCo {
-        /// Sources of feeds
+        /// Peers to connect to at startup.
```
xla, line 279: Is this any kind of peer or specifically seed nods?

proxy/src/coco/config.rs
```diff
@@ -82,3 +82,72 @@ pub fn configure(
...
+    seeds: &[T],
+) -> Result<Vec<(peer::PeerId, SocketAddr)>, error::Error> {
+    let mut resolved = Vec::with_capacity(seeds.len());
+
+    for seed in seeds.iter() {
+        let seed = seed.as_ref();
+
+        if let Some(ix) = seed.chars().position(|c| c == '@') {
+            let (peer_id, rest) = seed.split_at(ix);
+            #[allow(clippy::indexing_slicing)]
```
xla, line 103: Let's keep attributes on the function.

proxy/src/session.rs
cloudhead: Any peer.

sarahscott:
Looks good! Should we display validation errors to the user?

rudolfs (approved):

proxy/src/session.rs
xla: Then let's rename the field accordingly to reflect that.

proxy/src/coco/config.rs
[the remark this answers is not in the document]
cloudhead: `FromStr` won't work because the function is async. I can implement a resolve method though on `Seed`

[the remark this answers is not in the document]
FintanH: I see, because you need to look up the host. So maybe it should be:
```rust
struct Seed<Host> {
  peer_id: PeerId,
  host: Host,
}

impl FromStr for Seed<String> { /* split and parse */ }

impl Seed {
  pub fn resolve(&self) -> Seed<PeerId, SocketAddr>
}
```

What do you think?

proxy/src/session.rs
cloudhead: It already has the correct name, it was the "feed" part that was incorrect.

proxy/src/coco/config.rs
cloudhead: Done

proxy/src/coco/config.rs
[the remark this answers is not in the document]
cloudhead: We could do that, but I'm not sure what we gain. These functions are used internally and only in one place at the moment.

cloudhead:
> Looks good! Should we display validation errors to the user?

I think that's a good idea, but would require a lot of plumbing because currently these seeds are loaded at startup only. We'd have to think about this more.

proxy/src/coco/config.rs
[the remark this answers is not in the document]
FintanH: We'd gain the separation of each stage:
* Parse a string into peer and host
* Resolve the host
* And now we have a fully parsed and fully resolved seed :)

FintanH (approved):
I think it'd be nice to have my suggestion about the progressive states but won't block on it :) 

Nice work! The separation is much cleaner :grin:

rudolfs (approved):

proxy/src/session.rs
xla: The field is still called `seeds` and not `peers`.

PR #793, opened 2020-08-18
fix(ui): Accessible hotkeys for other layouts

geigerzaehler:
* Hotkeys dispatch uses `KeyboardEvent.key` instead of `KeyboardEvent.code` which does not respect the keyboard layout.
* Use Cmd/Ctrl+h for hotkey list instead of shift+/ which is hard to type on non-US layouts.
* Prevent fast toggling on repeat presses

### DIFF
--- a/ui/Hotkeys.svelte
+++ b/ui/Hotkeys.svelte
@@ -20,8 +20,8 @@
       return false;
     }
 
-    // To open help => OS modifier key + h
-    if (modifierKey && event.key === "h") {
+    // To open help => ?
+    if (event.key === "?") {
       toggle(path.shortcuts());
     }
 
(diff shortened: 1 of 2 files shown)


cloudhead:
`code` is the physical key, so it's layout dependent

geigerzaehler:
> Generally this makes sense, but could you expand how `code` doesn't respect keyboard layout.

[MDN docs say:](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)

> For example, the code returned is "KeyQ" for the Q key on a QWERTY layout keyboard, but the same code value also represents the ' key on Dvorak keyboards and the A key on AZERTY keyboards. That makes it impossible to use the value of code to determine what the name of the key is to users if they're not using an anticipated keyboard layout.



> I can't find the corresponding change for this point. It still is triggered by `?` judging by the code.

You’re right. I’ve reverted this change after @rudolfs pointed out that `cmd+h` is a system shortcut on mac

Issue #742, opened 2020-08-03
Show known identities


MeBrei:
Use the endpoint created in #741 to list known identities in the `Tracking` tab of the sessions identity profile view. The project list of the `Tracking` tab is implemented in #724 

**Design:** tbd

For more info see also #740

rudolfs:
While discussing the design we opted to add a link to the repo selector instead, for more detail see: https://github.com/radicle-dev/radicle-upstream/issues/741#issuecomment-675521865.

xla:
What does that mean for this issue? close, rephrease?

rudolfs:
@xla rephrased my comment above. I'll rephrase the issue.

PR #795, opened 2020-08-18
feat(ui): go to user profile aka visitor view

rudolfs:
This only implements the UI for going to a user profile.
Fetching of the user profile and styling the visitor view will be done in separate issues.

profile

To be able to test this with fake peers, start the app with `yarn start:test` and seed it with a fake project and peers:

```
curl -X POST "http://localhost:8080/v1/control/create-project" \
 -H "Accept: application/json" \
 -H "Content-Type: application/json" \
 -d '{"name":"rx","description":"pixel editor", "defaultBranch": "master", "fakePeers": ["ele", "abbey"]}' 
```

Closes: https://github.com/radicle-dev/radicle-upstream/issues/742


ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -1,9 +1,13 @@
 <script>
-  import { createEventDispatcher } from "svelte";
+  import { createEventDispatcher, getContext } from "svelte";
+  import { push } from "svelte-spa-router";
+
+  import * as path from "../../../src/path.ts";
```
xla, line 5: We can cluster the `ts` imports.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -162,14 +174,26 @@
...
+            variant="circle" />
+          <p class="typo-text-bold typo-overflow-ellipses">
+            {repo.identity.metadata.handle || repo.identity.shareableEntityIdentifier}
+          </p>
+        </div>
+        <Tooltip value="Go to profile" position="top">
+          <div
+            class="open-profile"
+            on:click={() => {
+              repo.identity.peerId === session.identity.peerId ? push(path.profileProjects()) : push(path.userProfile(repo.identity.peerId));
```
xla, line 192: Inlining this gets a bit unwieldy. The least it would help have this up in the script block.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -122,6 +130,10 @@
   .revision-dropdown ul:last-child li {
     border-radius: 0 0 3px 3px;
   }
+  .open-profile {
+    display: flex;
+    justify-content: center;
+  }
```
sarahscott, line 136:
```suggestion
      cursor: pointer;
  }
```

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/795/commits/b23b5b3135dd757c5fe4f566058dc012e401b910

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/795/commits/b23b5b3135dd757c5fe4f566058dc012e401b910

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/795/commits/b23b5b3135dd757c5fe4f566058dc012e401b910

rudolfs:
> Dope and very quick turnaround. The UserProfile concept is miss-leading as so far we referred to users as registered entries on the registry. Instead this is a general profile view, we should rework our profile and prepare it to work for the current identity and other coco identities.

I think there's going to be enough differences on how our own profile will be presented than somebody else's, that makes sense to have a separate screen for it. Otherwise you'll have one generic profile page with a lot of branching inside.

We should extract elements that perform the same function and look the same into components and reuse them in both screens obviously, but that's going to be handled in a different issue.

A compelling reason to go with this approach for me is the fact that for our own profile data will always be available, not so for a visitor view of somebody else's profile.

cloudhead (approved):
Looks good!

cloudhead:
> Dope and very quick turnaround. The `UserProfile` concept is miss-leading as so far we referred to users as registered entries on the registry. Instead this is a general profile view, we should rework our profile and prepare it to work for the current identity and other coco identities.

Not sure I follow, we also refer to radicle-link users as users..

geigerzaehler (approved):

PR #807, opened 2020-08-21
refactor(proxy): split into workspaces

FintanH:
* Removing `registry` is both a blessing and a curse
* coco compiles but a couple of tests are broken
* core compiles (haven't checked tests)
* http is not compiling and the error handling is a PITA
* I haven't even tried to compile api because it would be the last piece of the puzzle.


xla:
Super-seeded by #810

Issue #809, opened 2020-08-24
Migrate proxy to workspace


xla:
In order to have better isolation in our domains and ideally reduce compile times when working in one of the lower layers we want to migrate the proxy codebase to a Rust workspace with individual crates in it.

PR #810, opened 2020-08-24
refactor(proxy): extract coco into crate

xla:
First part of #809 

Mission proxy twerkspace under way.


cloudhead (approved):
:partying_face:

rudolfs (commented):
Looks good.

Will wait until CI passes before accepting.

proxy/Cargo.toml
```diff
@@ -1,64 +1,6 @@
...
-license = "GPL-3.0-only"
-authors = ["radicle team <dev@radicle.xyz>"]
-edition = "2018"
-description = "Proxy to serve a specialised API to the Upstream UI."
-readme = "https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/README.md"
-categories = [ "crypto", "p2p" ]
-keywords = [ "radicle-upstream", "proxy" ]
-repository = "git@github.com:radicle-dev/radicle-upstream.git"
-default-run = "proxy"
+[workspace]
```
geigerzaehler, line 1: How about moving this file to the repository root? That would allow us to run all cargo commands also from the root. Otherwise, nothing changes.

xla: The root is already quite crowded. We have three main components to manage: app container, ui and proxy. Not generally opposed to move, just a not convinced that we have quick ways to know what relevance a file has for which of our artefacts.

cloudhead (approved):

rudolfs (approved):

PR #811, opened 2020-08-25
feat(proxy): task queue

### LINKED ISSUE #803: Project request supervision
The story here is that as a user I have heard about a super cool project via [insert favourite social network here] and I have a `RadUrn`. I ask my network if anyone knows about this thing. This request ripples through the network until someone responds. The response includes a PeerId & Address. I can use this information to contact that peer and clone the project to my Monorepo. That's the happy path. The sad path is that the network may take a while to respond, it may not respond at all (for some given time), or I may fail to connect a peer that I get back. To keep track of requests that are waiting and their stages of the progress we need some kind of Waiting Area. Please take a ticket and we'll call you up when your order is ready, thank you.

More formally, we can split up the feature into two separate constituents. The first is the concept of the `Request`. Using the project request as the driving example, we see it takes some input -- the `RadUrn`. Some requirements that we consider sane are that we should be able to:
* Cancel a request
* Timeout a request
* Limit the number of attempts

These requirements, in turn, give us states of the request that are given for _any_ type of a request, but a request may also have its own states. For example, the project request would also have:
* Found a peer to clone from
* The project has been cloned

As well as this, there's the initial state where a request has simply been created. The final state is defined by a satisfied request, so in our case when a project has been cloned.

The second constituent is a `Supervisor` that manages a set of `Request`s. This component knows how to give us back information about `Request`s and their states. It should be able to:
* List the requests that it knows about
* Prevent double requests from being in-flight -- i.e. making the same request twice in quick succession is the same as making one request
* Exponentially back off requests with the same input
* Provide an overview of a request and its metadata (state, last timestamp of progress, etc.)
* Lookup a request given an identifier
* Provide the state changes for a given identifier -- useful for notifications

@xla foresees an event-driven interface for polling the `Supervisor` as useful. That is the `Supervisor` would have a `next` method which gives back the caller the next available request to execute (whether it's in a created and ready to go state or a state where it failed due to timing out -- or the like -- and it needs to be re-requested).

Another useful requirement would be to persist the state of a `Supervisor` and its `Request`s, but I believe this might fall out from being able to request the overview of the `Supervisor` mentioned above.

sarahscott:
This is a first sketch of a `Queue` for executing coco tasks & keeping `proxy` clients informed about a task's status, i.e. for something like a search for a Project of a given `urn`.

I'd like to use a lot of the excellent existing infrastructure by @xla for notifications and subscriptions even though we'll be removing a lot of the `registry`-specific code.

todos:
- [ ] Use something more specific than `String` for the task `id`
- [ ] Ensure this actually works with `coco::Api`
- [ ] More / better tests
- [ ] Timeouts
- [ ] Integration with `http` & `notification.rs`

Closes #803

### DIFF
--- a/proxy/src/lib.rs
+++ b/proxy/src/lib.rs
@@ -42,5 +42,6 @@ pub mod identity;
 pub mod keystore;
 mod notification;
 pub mod project;
+pub mod queue;
 pub mod registry;
 pub mod session;
--- a/proxy/src/queue.rs
+++ b/proxy/src/queue.rs
@@ -1,0 +1,258 @@
+//! A queue that manages & caches asynchronous tasks
+
+use std::time::{self, SystemTime};
+
+use kv::Codec as _;
+use serde::{Deserialize, Serialize};
+
+/// Possible messages a [`Task`] can carry.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase", tag = "type")]
+pub enum Message {
+    /// Find a [`Project`] of a given `urn`
+    FindProject {
+        /// The unique identifier of the desired [`Project`]
+        urn: String,
+    },
+}
+
+/// Possible states a [`Task`] can have. Useful to reason about the lifecycle and
+/// whereabouts of a given [`Task`].
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase", tag = "type")]
+pub enum State {
+    /// [`Task`] has completed successfully
+    Confirmed {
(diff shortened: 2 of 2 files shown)


xla (changes requested):
We might not be well served by a implementing a general task queue. As we have to generalise over concerns like life cycle states for all possible workloads. Instead we should solve the issue at hand that we want to keep track of finding projects on the network, if we find in the future to have more than one or exactly three use-cases for a queue system we can at that point build generalised infrastructure. For now we should spend on our focus to find the right design for the concept of finding projects.

For that we should determine the FSM for it and what information is important for us to make meaningful decisions, like for how long we have been looking for a project.

sarahscott:
On hold until we know more about project search

Issue #774, opened 2020-08-13
Broken Existing Repo Setup


FintanH:
If an existing repository doesn't contain any branch named after the passed in `default_branch`, the existing project creation flow breaks.

The monorepo gains an entry for the project you tried to create so this breaks EVERYTHING

rudolfs:
Not sure this is the same issue, but it leads to a very similar error message.

Steps to reproduce:

 - create a new project
    - name -> `new-project`
    - choose -> "Start with a new repository"
 - pick an empty directory `/Users/rudolfs/work/radicle-tests`
 - click "Create project" -> project is created without errors
 - create another new project
    - use the same name -> `new-project`
    - choose -> "Start with a new repository"
 - pick the same empty directory `/Users/rudolfs/work/radicle-tests`
 - click "Create project" -> Errors

```
 ERROR proxy::http::error      > Rejection([MethodNotAllowed, ProjectCreation(Git(Error { code: -15, klass: 11, message: "failed to create commit: current tip is not the first parent" })), MethodNotAllowed])
 ERROR proxy::http::error      > Rejection([MethodNotAllowed, MethodNotAllowed, Git(Git(Error { code: -3, klass: 4, message: "reference \'refs/namespaces/hwd1yrey85r1j1ppytntzr73hhystbmooj76dc8zj4uipozoxq1848w5fte/refs/heads/master\' not found" }))])
```
<img width="291" alt="Screenshot 2020-08-14 at 11 09 56" src="https://user-images.githubusercontent.com/158411/90233694-ce50ac00-de1e-11ea-92a2-9c0b3c8c2208.png">
<img width="300" alt="Screenshot 2020-08-14 at 11 09 59" src="https://user-images.githubusercontent.com/158411/90233697-cf81d900-de1e-11ea-9200-66e1bb96b894.png">
<img width="1363" alt="Screenshot 2020-08-14 at 11 10 03" src="https://user-images.githubusercontent.com/158411/90233698-cf81d900-de1e-11ea-9f56-6cbbe8ae33de.png">


This seems like a regression from https://github.com/radicle-dev/radicle-upstream/pull/727. There should be a validation that prevents creating a project with the same name.

rudolfs:
Fixed in https://github.com/radicle-dev/radicle-upstream/pull/825.

PR #813, opened 2020-08-25
fix(ui): improve error messaging in project creation

rudolfs:
The proxy does not allow to add the same repository twice. Creating another project from the same repository resulted in an undescriptive error.

Before:
Screenshot 2020-08-25 at 14 43 31

After:
Screenshot 2020-08-25 at 14 45 34

Stop-gap solution for https://github.com/radicle-dev/radicle-upstream/issues/774#issuecomment-673976707


xla (approved):
🕠 🐳 🌵 🔑

Issue #650, opened 2020-07-09
Clicking on "Create Project" twice leads to "Not Found" page


cloudhead:
If you create a project that is somewhat slow to create, and click on "Create Project" a second time while waiting, the app goes to a "Not Found" page, with an error saying the rad remote already exists.

The "issue" here is really that it goes to this "Not Found" page.

Related to #512

rudolfs:
Fixed in #815.

PR #815, opened 2020-08-25
feat(ui): lock screen during long backend activity

rudolfs:
This locks the project creation modal and the project checkout flow while waiting for proxy to finish the request.

It's a stop-gap solution for: https://github.com/radicle-dev/radicle-upstream/issues/650 and https://github.com/radicle-dev/radicle-upstream/issues/512.

lock2

lock

NB: the cursor looks broken only in the screengrab. It looks like this in reality:

Screenshot 2020-08-25 at 16 50 19

Inspired by: https://github.com/radicle-dev/radicle-upstream/pull/566.

### DIFF
--- a/ui/Screen/Project/Source.svelte
+++ b/ui/Screen/Project/Source.svelte
@@ -8,6 +8,7 @@
   import { checkout } from "../../src/project.ts";
   import * as notification from "../../src/notification.ts";
   import * as path from "../../src/path.ts";
+  import * as screen from "../../src/screen.ts";
   import { project as projectStore } from "../../src/project.ts";
   import * as remote from "../../src/remote.ts";
   import { Variant as IllustrationVariant } from "../../src/illustration.ts";
@@ -81,6 +82,7 @@
 
   const handleCheckout = async event => {
     try {
+      screen.lock();
       const path = await checkout(
         id,
         event.detail.checkoutDirectoryPath,
@@ -98,6 +100,8 @@
       );
     } catch (error) {
       notification.error(`Checkout failed: ${error.message}`, true);
+    } finally {
+      screen.unLock();
     }
   };
 
--- a/ui/Screen/Project/Source.svelte
+++ b/ui/Screen/Project/Source.svelte
@@ -101,7 +101,7 @@
     } catch (error) {
       notification.error(`Checkout failed: ${error.message}`, true);
     } finally {
-      screen.unLock();
+      screen.unlock();
     }
   };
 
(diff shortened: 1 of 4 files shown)


cloudhead (approved):

ui/src/screen.ts
```diff
@@ -0,0 +1,7 @@
+export const lock = (): void => {
+  document.documentElement.classList.add("lock-screen");
+};
+
+export const unLock = (): void => {
```
xla, line 5:
```suggestion
export const unlock = (): void => {
```

ui/src/screen.ts
```diff
@@ -0,0 +1,7 @@
+export const lock = (): void => {
+  document.documentElement.classList.add("lock-screen");
```
xla, line 2: Magical string appears twice, could be put in a const or the style applied directly.

ui/Screen/Project/Source.svelte
```diff
@@ -98,6 +100,8 @@
       );
     } catch (error) {
       notification.error(`Checkout failed: ${error.message}`, true);
+    } finally {
+      screen.unLock();
```
xla, line 104: Should we make sure that there is an `unlock` for the destruction of this component as well?

xla (changes requested):
Pretty neat, wonder how we could make this generally available across the app.

rudolfs:
> Pretty neat, wonder how we could make this generally available across the app.

I tried adding it to `api.ts` `request` and it works. But on larger projects both the proxy as well as the UI are slow, so it ends up only blocking the backend query. Ideally it should lock also while the UI is loading... We could probably integrate it into routing somehow in the future.

xla:
> I tried adding it to `api.ts` `request` and it works. But on larger projects both the proxy as well as the UI are slow, so it ends up only blocking the backend query. Ideally it should lock also while the UI is loading... We could probably integrate it into routing somehow in the future.

Good point, should be considered for navigation refactor.

ui/Screen/Project/Source.svelte
rudolfs: Probably doesn't make sense, because there is no automatic call to navigate anywhere after the backend request finishes and since the user can't click on anything, there is no way to close it. We'll have to rely on that `try ... catch ... finally`.

ui/src/screen.ts
rudolfs: I hear you. Not sure it's possible to apply this selector directly: `.lock-screen > *`. Would leave this as a follow up re-factor when we're not under time pressure.

xla (approved):
🛎 🎄 💙 ⚗

PR #816, opened 2020-08-25
feat(ui): visitor profile view

### LINKED ISSUE #743: Visitor view of profile
- [x] navigate from known identities #742 
- [x] refactor `Header.Large` for visitor view
- [x] show known project list
- [x] integrate data from tracked identity endpoint #741

For more info see also: #740

MeBrei:
Closes #743 

To set up a project with other users, execute the following after starting the app with `yarn start:test` and creating an identity

```
curl -X POST "http://localhost:8080/v1/control/create-project" \
 -H "Accept: application/json" \
 -H "Content-Type: application/json" \
 -d '{"name":"rx","description":"pixel editor", "defaultBranch": "master", "fakePeers": ["ele", "abbey"]}' 
```

The created `fakePeers` seem to not be set up correctly, as some fakePeers will have the same urn and thus listing projects that they are actually not involved in. This PR is only concerned with the ui though, so it should be handled in another PR.


xla (changes requested):
Looks good, the main concern I'd like to raise is that we muddy the hierarchy of our components with this change. Screens have a 1:N relationship to fragments, which neither are shared nor do screens import other screens. If we have the need to share components we do so by splitting them apart and moving them into the design system. I don't see why we can't move the projects list into the design system with it emitting select events that a fragment can listen on.

cypress/integration/user_profile.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    cy.contains("Source").click();
+
+    // Pick a user from the revision selector
+    cy.pick("revision-selector").click();
+    cy.get(".revision-dropdown").pick("abbey").click();
+
+    cy.pick("header").should("exist");
+  });
+
+  it("shows the right data", () => {
```
rudolfs, line 24: This test depends on the previous test. We should just combine them, otherwise it'll be brittle.

rudolfs (changes requested):
Functionality looks all good. 💯 

I'd also move the `Projects.svelte` component to `DesignSystem/Components/ProjectList.svelte`.

xla (approved):
💊 🐰 😂 😬

rudolfs (approved):

PR #817, opened 2020-08-25
refactor: remove registry

xla:
This change removes almost all registry related code in the proxy and unwires most of the functionality in the UI while keeping the components and code around.

Operation: RIPistry

Part of #809


cloudhead (dismissed):
Concept ACK :scream_cat:

cloudhead (approved):

PR #819, opened 2020-08-26
chore(ui): remove links to registry features

rudolfs:
Remove links to the following features
  - wallet
  - register handle button
  - add org button / orgs sidebar entries

Screenshot 2020-08-26 at 11 03 51

Screenshot 2020-08-26 at 11 04 08

Fix dropdown style where the bottom border was visible when there was only one entry.

Before:
Screenshot 2020-08-26 at 11 09 14

After:
Screenshot 2020-08-26 at 11 04 18
Screenshot 2020-08-26 at 11 04 27

### DIFF
--- a/ui/DesignSystem/Component/Sidebar.svelte
+++ b/ui/DesignSystem/Component/Sidebar.svelte
@@ -1,5 +1,4 @@
 <script>
-  import { createEventDispatcher } from "svelte";
   import { location, link } from "svelte-spa-router";
 
   import * as path from "../../src/path.ts";
@@ -8,13 +7,7 @@
   import Tooltip from "./Tooltip.svelte";
   import { Avatar, Icon } from "../Primitive";
 
-  import AddOrgButton from "./Sidebar/AddOrgButton.svelte";
-
-  const dispatch = createEventDispatcher();
-
   export let identity = null;
-  export let orgs = null;
-  export let registerOrgPermission = null;
 </script>
 
 <style>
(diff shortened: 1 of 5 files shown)


geigerzaehler:
I’d be ok with removing the code. We have no idea when or even if this will be used again. Since it’s not working anyways it just looks like dead code to me.

cloudhead (approved):

xla (changes requested):
If we really don't want it around, let's remove it. Instead of hiding it behind the `isDev` switch. Even in development one can't use these now anyway.

rudolfs:
@xla @geigerzaehler https://github.com/radicle-dev/radicle-upstream/pull/820.

rudolfs:
Ok, never mind, will remove only registry related stuff: registering, domains, etc. But will keep the transaction and wallet visuals.

ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -124,56 +118,20 @@
         </a>
       </Tooltip>
     </li>
-
-    {#each orgs as org}
+  </ul>
+  <ul class="bottom">
+    {#if isDev()}
```
xla, line 123: We use this check here before. Wouldn't introduce it in this PR.

xla (changes requested):
Looks good, wouldn't continue to introduce more uses of `isDev`.

ui/DesignSystem/Component/Sidebar.svelte
rudolfs: I thought we didn't want to show unfinished features, is that not the case?

xla (approved):
🐞 🎁 🎋 🍧

MeBrei (approved):

PR #820, opened 2020-08-26
chore(ui): remove registry UI

rudolfs:
Removes everything registry, transaction and wallet related.


rudolfs:
We wanna keep the visuals around, maybe we'll reuse them in the future.

Issue #821, opened 2020-08-26
Create project from existing repo fails if there is no master branch


rudolfs:
Steps to reproduce:
1. create a new git repo `mkdir fail && cd fail && git init . && touch README && git add . && git commit -m "FTW!"`
2. create a new branch `git branch never-gonna-let-me-down && git checkout never-gonna-let-me-down`
3. delete the master branch `git branch -D master`
4. create a new project from this existing repo

<img width="406" alt="Screenshot 2020-08-26 at 13 05 50" src="https://user-images.githubusercontent.com/158411/91296461-ed472a80-e79c-11ea-89ba-f37371c3f1ec.png">
<img width="1359" alt="Screenshot 2020-08-26 at 13 05 53" src="https://user-images.githubusercontent.com/158411/91296463-ef10ee00-e79c-11ea-8308-8498c6e69058.png">

```
Error: Internal Git error: Git(Error { code: -3, klass: 4, message: "reference \'refs/namespaces/hwd1yredufa83spg6yb6smdtjx86igebtz8f944g767r4ufj18bwbaghjsc/refs/heads/master\' not found" })
```

```
 INFO  librad::git::types    > creating symbolic ref refs/namespaces/hwd1yredufa83spg6yb6smdtjx86igebtz8f944g767r4ufj18bwbaghjsc/refs/rad/self -> refs/namespaces/hwd1yredbsi9sgn5e661h7kcfm8dhadf88i8nh4emy37r4pge5nqsfdurrc/refs/rad/id
 ERROR coco::project::create > error while trying to find default branch: Error { code: -3, klass: 4, message: "no reference found for shorthand \'master\'" }
 ERROR api::http::error      > Rejection([MethodNotAllowed, Coco(MissingDefaultBranch { repo_path: "/Users/rudolfs/work/radicle-tests/fail/.git/", branch: "master" }), MethodNotAllowed])
 INFO  proxy::http           > "POST /v1/projects HTTP/1.1" 400 55.679956ms
 ERROR api::http::error      > Rejection([MethodNotAllowed, MethodNotAllowed, Coco(SurfGit(Git(Error { code: -3, klass: 4, message: "reference \'refs/namespaces/hwd1yredufa83spg6yb6smdtjx86igebtz8f944g767r4ufj18bwbaghjsc/refs/heads/master\' not found" })))])
 INFO  proxy::http           > "GET /v1/projects HTTP/1.1" 400 11.256078ms
```

[...]

PR #825, opened 2020-08-26
fix(proxy): project creation from repo without master

rudolfs:
It failed because the surf `Browser` has to be initialised with a revision. This revision was set to "master" by default.
The workaround for now is to fall back to the first branch of the repository, if it doesn't have a "master" branch.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/821.

### DIFF
--- a/proxy/coco/src/source.rs
+++ b/proxy/coco/src/source.rs
@@ -513,8 +513,31 @@ pub struct LocalState {
 /// Will return [`Error`] if the repository doesn't exist.
 pub fn local_state(repo_path: &str) -> Result<LocalState, Error> {
     let repo = git::Repository::new(repo_path)?;
-    // TODO(finto): This should be the default branch of the project, possibly.
-    let browser = Browser::new(&repo, git::Branch::local("master"))?;
+
+    let repoo = git2::Repository::open(repo_path)?;
+    let first_branch = repoo
+        .branches(Some(git2::BranchType::Local))?
+        .filter_map(|branch_result| {
+            let branch = match branch_result {
+                Ok(s) => s,
+                Err(_) => return None,
+            };
+
+            match branch.0.name() {
+                Ok(Some(n)) => Some(n.to_owned()),
+                Ok(None) | Err(_) => None,
+            }
+        })
+        .min()
+        .expect("Could not find any branches.");
+
+    println!("first_branch: {:?}", first_branch);
+
+    let browser = match Browser::new(&repo, git::Branch::local("master")) {
+        Ok(browser) => browser,
+        Err(_) => Browser::new(&repo, git::Branch::local(&first_branch))?,
+    };
+
     let mut branches = browser
         .list_branches(Some(BranchType::Local))?
         .into_iter()
--- a/proxy/coco/src/source.rs
+++ b/proxy/coco/src/source.rs
@@ -512,10 +512,8 @@ pub struct LocalState {
 ///
 /// Will return [`Error`] if the repository doesn't exist.
 pub fn local_state(repo_path: &str) -> Result<LocalState, Error> {
-    let repo = git::Repository::new(repo_path)?;
-
-    let repoo = git2::Repository::open(repo_path)?;
-    let first_branch = repoo
+    let repo = git2::Repository::open(repo_path)?;
+    let first_branch = repo
         .branches(Some(git2::BranchType::Local))?
         .filter_map(|branch_result| {
             let branch = match branch_result {
@@ -531,7 +529,12 @@ pub fn local_state(repo_path: &str) -> Result<LocalState, Error> {
         .min()
         .expect("Could not find any branches.");
 
-    println!("first_branch: {:?}", first_branch);
+    log::debug!(
+        "The fallback branch for this repository is: {:?}",
+        first_branch
+    );
+
+    let repo = git::Repository::new(repo_path)?;
 
     let browser = match Browser::new(&repo, git::Branch::local("master")) {
         Ok(browser) => browser,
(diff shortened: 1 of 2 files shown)


proxy/coco/src/source.rs
```diff
@@ -513,8 +513,31 @@ pub struct LocalState {
...
+                Err(_) => return None,
+            };
+
+            match branch.0.name() {
+                Ok(Some(n)) => Some(n.to_owned()),
+                Ok(None) | Err(_) => None,
+            }
+        })
+        .min()
+        .expect("Could not find any branches.");
```
xla, line 532: Should we panic for this case?

proxy/coco/src/source.rs
```diff
@@ -513,8 +513,31 @@ pub struct LocalState {
...
+
+    println!("first_branch: {:?}", first_branch);
```
xla, line 534: If we want to keep this line, it should be a `log::debug`.

xla (changes requested):
Noice, two small things inline.

xla (approved):
🛰 🚓 ⌚️ 🏚

proxy/coco/src/source.rs
```diff
@@ -512,9 +512,35 @@ pub struct LocalState {
 ///
 /// Will return [`Error`] if the repository doesn't exist.
 pub fn local_state(repo_path: &str) -> Result<LocalState, Error> {
+    let repo = git2::Repository::open(repo_path)?;
+    let first_branch = repo
+        .branches(Some(git2::BranchType::Local))?
+        .filter_map(|branch_result| {
+            let branch = match branch_result {
+                Ok(s) => s,
```
sarahscott, line 520: This syntax looks a little funny but i'm also a rust noob so 🤷‍♀️

proxy/coco/src/source.rs
```diff
@@ -512,9 +512,35 @@ pub struct LocalState {
...
+        first_branch
+    );
+
     let repo = git::Repository::new(repo_path)?;
-    // TODO(finto): This should be the default branch of the project, possibly.
-    let browser = Browser::new(&repo, git::Branch::local("master"))?;
+
+    let browser = match Browser::new(&repo, git::Branch::local("master")) {
+        Ok(browser) => browser,
+        Err(_) => Browser::new(&repo, git::Branch::local(&first_branch))?,
```
sarahscott, line 541: I can see how this works out but should we always treat it as an `Error` when someone doesn't use `master` as the default branch?

ui/Screen/ProjectCreation.svelte
```diff
@@ -214,6 +214,9 @@
 
     try {
       localState = await getLocalState(path);
+      if (!localState.branches.includes(defaultBranch)) {
```
sarahscott, line 217: Would it make sense for `localState` to contain a `defaultBranch` instead of having the component set it to whatever branch exists at index 0?

proxy/coco/src/source.rs
rudolfs: When a user creates a project off of an existing repository, it's up to them to choose what the default branch should be. This request returns a list of all the branches of a repository (regardless whether it is already managed by radicle or not). We get all branches of a project via the surf `Browser`, but the API of `Browser` requires us to initialise it with a branch.

Now, we try to initialise it with "master" first and if that fails, we try with the first branch in the list, if there are no branches at all, the whole request should fail.

We could theoretically just use `Git2` for getting the branch list and not bother with `Browser`, but then we would need to add code here to determine whether the repo is already managed by radicle.

ui/Screen/ProjectCreation.svelte
rudolfs: But that is what triggered the bug in the first place, because the defaultBranch is set to "master", so when a repository doesn't have that branch it leads to an error.

What happens here is, we get back the list of branches and "master" is not part of that list, so the dropdown was pre-selected to "null" instead of whatever is the first branch in the list.

proxy/coco/src/source.rs
kim: The `?` operator actually works on `Option` (magically coercing `NoneError` back to `None` where applicable):

```rust
let (branch, _) = branch_result.ok()?;
let name = branch.name().ok()?;
name.map(String::from)
```

ui/Screen/ProjectCreation.svelte
rudolfs: Good point. I think we should refactor this bit and move the `DEFAULT_BRANCH_FOR_NEW_PROJECTS` constant to the backend and then pass it via session or something. But I'd like to do this in a follow up.

xla (approved):
(*☻-☻*)

cloudhead (approved):

proxy/coco/src/source.rs
kim: This does indeed seem a bit dangerous -- a freshly initialised repo (`git init`) will return an empty `Branches` iterator, so it's not really a should-never-happen situation.

How about this?:

```rust
 let default_branch = repo
        .branches(Some(git2::BranchType::Local))?
        .filter_map(|branch_result| {
            let (branch, _) = branch_result.ok()?;
            let name = branch.name().ok()?;
            name.map(String::from)
        })
        .min_by(|x, y| match (x.as_ref(), y.as_ref()) {
            ("master", _) => Ordering::Less,
            (_, "master") => Ordering::Greater,
            _ => x.cmp(y)
        }).
        .ok_or_else(Error::UninitialisedRepository(repo_path))?;
```

Now, "master" is returned if it exists, and otherwise the lexicographically smallest branch. Or none at all, which should be a non-fatal error.

PR #838, opened 2020-08-28
feat(proxy): coco announce

### LINKED ISSUE #602: Announcing Changes
We cannot always guarantee that the proxy is running when code is being changed. This means that the monorepo can change while the proxy is _not_ running. The question then is: "How do we advertise the changes to any of the refs?"

There are currently two ideas for this:
* Periodically announce the HEADs of the refs
* Keep a snapshot of all `ref -> HEAD` pairs and only update when there are differences between the monorepo state and the snapshot.

xla:
Closes #602


proxy/api/src/main.rs
```diff
@@ -83,15 +83,45 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    loop {
+        timer.tick().await;
+
+        let ctx = ctx.read().await;
+
+        // TODO(xla): get/load old state.
+
+        let old: Vec<announce::Announcement> = vec![];
+        let new = announce::build(&ctx.peer_api).expect("unable to build state");
+        let updates = announce::diff(&old, &new);
```
FintanH, line 117: I'm wondering if diff'ing is what we want. The scenario I have in mind is that: if someone was offline during an announcement then they would have to wait for the next change to be made until AND be online again to make sure they have the latest changes.

xla: Could you hash out the scenario more going into detail what the availabilities of peer `A` and `B` are?

Wouldn't we expect when a peer comes online again, that they do a full sync with other peers to get to the latest state instead of relying of delta updates of announcements? Or are you advocating to indiscriminately announce all heads every time this routine runs?

FintanH: Gonna resolve this since I'm wrong :)

proxy/api/src/main.rs
```diff
@@ -83,15 +83,51 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     warp::serve(api).run(([127, 0, 0, 1], 8080)).await;
 
     Ok(())
 }
+
+async fn announcement_watcher(ctx: context::Ctx) {
+    // TODO(xla): Take interval timings from config/settings.
+    let mut timer = tokio::time::interval(std::time::Duration::from_secs(10));
+
+    loop {
```
kim, line 110: Why is it safe to `unwrap` inside the loop? If one of these panics, the spawned future will die (probably silently).

proxy/api/src/main.rs
```diff
@@ -83,15 +83,51 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     let ctx = context::Ctx::from(context::Context {
         peer_api,
         keystore,
         store,
     });
+
+    {
+        let ctx = ctx.clone();
+        log::info!("Starting Announcement watcher");
+        tokio::task::spawn(async move { announcement_watcher(ctx).await });
```
cloudhead, line 95: I'm probably being a noob, but I would have thought that this was equivalent to:
```
tokio::task::spawn(announcement_watcher(ctx));
```

proxy/api/src/main.rs
```diff
@@ -83,15 +83,51 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+async fn announcement_watcher(ctx: context::Ctx) {
+    // TODO(xla): Take interval timings from config/settings.
+    let mut timer = tokio::time::interval(std::time::Duration::from_secs(10));
+
+    loop {
+        timer.tick().await;
+
+        let ctx = ctx.read().await;
+
+        let old = announce::load(&ctx.store).expect("unable to load cached announcemnts");
```
cloudhead, line 115:
```suggestion
        let old = announce::load(&ctx.store).expect("unable to load cached announcements");
```

proxy/api/src/main.rs
```diff
@@ -83,15 +83,51 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
+            ctx.peer_api
+                .with_protocol(|protocol| {
+                    Box::pin(async move { announce::announce(protocol, updates).await })
+                })
+                .await
+                .expect("announce failed");
+        }
+
+        let updates = announce::diff(&old, &new);
```
cloudhead, line 129: How come we diff twice?

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+/// # Errors
+///
+/// * if the announcemnet of one of the project heads failed
+pub async fn announce(
+    protocol: net::protocol::Protocol<net::peer::PeerStorage<keys::SecretKey>, net::peer::Gossip>,
+    updates: Vec<Announcement>,
+) -> Result<(), Error> {
+    for (project_urn, head, hash) in &updates {
+        let urn = uri::RadUrn::new(
+            project_urn.id.clone(),
```
cloudhead, line 35: Since we only use the `id` here, should `Announcement`  be `(Hash, String, Oid)`?

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+            uri::Protocol::Git,
+            uri::Path::parse(head)?,
+        );
+        let have = net::peer::Gossip {
+            urn,
+            rev: Some(net::peer::Rev::Git((*hash).into())),
+            origin: Some(protocol.peer_id().clone()),
```
kim, line 42: This should be `None`.

Sorry for this being confusing -- this field is actually meant to send a gossip message saying "I just fetched stuff from X (but I'm not X), go get it from me if you want". Iow, if I'm `Y`, and I'm interested in `X` (but not `Z`), but `Z` tells me they got some `X` stuff, I'll go look for `remotes/X/*` on `Z`. Or rather, I'll only bother fetching from `Z` if the update is about `X`. At least that's what's _supposed_ to happen.

proxy/coco/src/announce.rs
cloudhead: Actually you should probably leave as-is, but use `urn.protocol` below.

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+        };
+
+        protocol.announce(have).await;
```
cloudhead, line 45: Could be slow if lots of projects, perhaps something to run concurrently in the future..

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    Ok(())
+}
+
+/// Builds the latset list of [`Announcement`]s for the current state of the peer.
```
cloudhead, line 51:
```suggestion
/// Builds the latest list of [`Announcement`]s for the current state of the peer.
```

proxy/api/src/main.rs
```diff
@@ -83,15 +83,51 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        announce::save(&ctx.store, updates.clone()).expect("unable to save updated announcements");
```
cloudhead, line 130: We shouldn't forget to `save` when changes are announced during the lifecycle of the application, otherwise these will be-reannounced on startup!

cloudhead: Nevermind, this is not only run at startup!

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+        ];
+        let old = vec![
+            (
+                project0(),
+                "igor/zero-assertions".to_string(),
+                "72a78226".parse::<oid::Oid>()?,
+            ),
+            (
+                project0(),
+                "thoshol/remove".to_string(),
```
kim, line 204: :+1:

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+                "7dec3269".parse::<oid::Oid>()?,
+            ),
+            (
+                project0(),
+                "fintohaps/notations".to_string(),
+                "b4d3276d".parse::<oid::Oid>()?,
+            ),
+            (
+                project0(),
+                "kalt/eat-my-impls".to_string(),
```
kim, line 226: :ok:

proxy/coco/src/announce.rs
cloudhead: If for whatever reason announcing takes more than the tick rate, then you're gonna start to build up a bunch of latency

proxy/coco/src/peer.rs
```diff
@@ -323,6 +339,24 @@ impl Api {
         callback(&mut browser)
     }
 
+    /// Get the underlying [`librad::net::protocol::Protocol`].
+    ///
+    /// # Errors
+    ///
+    /// * if they `callback` errors
+    pub fn with_protocol<F, T>(&self, callback: F) -> BoxFuture<'static, Result<T, Error>>
```
kim, line 347: Why `BoxFuture`, actually? A generic parameter should work fine, and then the caller can choose to box or not.

proxy/api/src/main.rs
xla: Nope, I am. Just forgot to remove the closure during refactoring.

cloudhead: Ah cool, I think there's another one of those somewhere if I'm not mistaken :thought_balloon:

proxy/api/src/main.rs
```diff
@@ -83,15 +83,51 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
+        let old = announce::load(&ctx.store).expect("unable to load cached announcements");
+        let new = announce::build(&ctx.peer_api).expect("unable to build state");
+
+        {
+            let updates = announce::diff(&old, &new);
+
+            ctx.peer_api
+                .with_protocol(|protocol| {
+                    Box::pin(async move { announce::announce(protocol, updates).await })
```
cloudhead, line 123: This guy also maybe can be simplified :shrug:

proxy/api/src/main.rs
```diff
@@ -83,15 +83,51 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     let api = http::api(ctx, args.test);
 
     warp::serve(api).run(([127, 0, 0, 1], 8080)).await;
 
     Ok(())
 }
+
+async fn announcement_watcher(ctx: context::Ctx) {
+    // TODO(xla): Take interval timings from config/settings.
+    let mut timer = tokio::time::interval(std::time::Duration::from_secs(10));
```
FintanH, line 108: Maybe push this out one level and take `Duration` as a parameter?

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+use crate::oid::Oid;
+use crate::peer;
+
+/// Name for the bucket used in [`kv::Store`].
+const BUCKET_NAME: &str = "announcements";
+/// Key for the single value used as cache.
+const KEY_NAME: &str = "latest";
+
+/// An update and all the required information that can be announced on the network.
+pub type Announcement = (uri::RadUrn, String, Oid);
```
FintanH, line 22: Can we make it a struct to have field names?

proxy/coco/src/announce.rs
FintanH: I think adding a `set_path`/`with_path` method to `RadUrn` would generally be useful.

proxy/coco/src/announce.rs
FintanH: It would only announce if there's a diff though right?

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+    Ok(())
+}
+
+/// Builds the latset list of [`Announcement`]s for the current state of the peer.
+///
+/// # Errors
+///
+/// * if listing of the projects fails
+/// * if listing of the Refs for a project fails
+pub fn build(api: &peer::Api) -> Result<Vec<Announcement>, Error> {
```
FintanH, line 57: Why not build a `HashSet` here?

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+
+            Ok(list)
+        },
+    }
+}
+
+/// Computes the list of announcements based on the difference of the `new` and `old` state. An
+/// [`Announcement`] will be included if an entry in `new` can't be found in `old`.
+#[must_use]
+pub fn diff(old_state: &[Announcement], new_state: &[Announcement]) -> Vec<Announcement> {
```
FintanH, line 82: Then `HashSet`s can be passed here, and then `diff` is just [`difference`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.difference) :)

proxy/coco/src/error.rs
```diff
@@ -56,6 +56,10 @@ pub enum Error {
     #[error(transparent)]
     Io(#[from] io::Error),
 
+    /// Kv error.
+    #[error(transparent)]
+    Kv(#[from] kv::Error),
```
FintanH, line 61: Where did we land on sub-error enums? :eyes:

FintanH (commented):
Looking good :+1: Have some feedback, lemme know what you think

proxy/coco/src/announce.rs
cloudhead: Yeah true, and since you only have a diff on your own projects, it should be small enough.

NunoAlexandre (commented):
Looking good overall. Left small comments, it seems that bigger issues are already covered by other reviews :+1:

proxy/coco/src/error.rs
xla: We want them and we gonna address them in #850.

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+
+/// Update the cache with the latest announcements.
+///
+/// # Errors
+///
+/// * if the [`kv::Bucket`] can't be accessed
+/// * if the storage of the new updates fails
+pub fn save(store: &kv::Store, updates: Vec<Announcement>) -> Result<(), Error> {
+    let bucket = store.bucket::<&'static str, kv::Json<Vec<Announcement>>>(Some(BUCKET_NAME))?;
+    Ok(bucket.set(KEY_NAME, kv::Json(updates))?)
```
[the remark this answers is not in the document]
xla: What's the upside of that suggestion?

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
+//! Compute, track and announce noteworthy changes to the network.
```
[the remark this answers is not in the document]
xla: Yeah we could rename it, do you have a strong arguments for the noun case?

proxy/coco/src/announce.rs
```diff
@@ -0,0 +1,303 @@
...
+        Err(Error::Storage(librad::git::storage::Error::Config(_err))) => Ok(list),
+        Err(err) => Err(err),
+        Ok(projects) => {
+            for project in &projects {
+                let refs = api.list_project_refs(&project.urn())?;
+
+                for (head, hash) in &refs.heads {
+                    list.push((project.urn(), head.to_string(), Oid::from(*hash.deref())));
+                }
+            }
```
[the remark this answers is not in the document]
xla: How would that be improving things?

proxy/coco/src/error.rs
FintanH: Muchos gracias

proxy/api/src/main.rs
xla: This will happen once we address the `TODO`.

proxy/api/src/main.rs
xla: Unfortunately we run into borrow issues with `updates` if we get rid of the closure.

proxy/coco/src/peer.rs
xla: If we don't box here, the lock guard we hold for the lock we have around the internal API is crossing await boundaries. Didn't find a better way to solve that, open to suggestions.

proxy/coco/src/announce.rs
xla: It does make sense, guess it could be better named and/or different types for the different message constructions. As in have one where we craft a gossip msg for another peer and one where we build one for the current peer.

proxy/api/src/main.rs
xla: Indeed, not quite sure how to address this without channels and selects over the different tasks we spawn in main.

proxy/coco/src/announce.rs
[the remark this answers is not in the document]
NunoAlexandre: It separates two different things, the call `bucket.set` with the `?` and returning the succeeding value if the call hasn't failed. might be a matter of habit of opting for the suggested alternative in the registry, but I don't quite like having a call, the `?` operator and a return all on the same line.

proxy/coco/src/announce.rs
xla: @FintanH Do we want to capture a follow-up for this?

proxy/coco/src/announce.rs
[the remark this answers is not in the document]
NunoAlexandre: My argument is already stated, afaik module names should not be verbs but nouns, functions should be verbs. I'd swear you have pointed that out yourself before.

proxy/coco/src/announce.rs
FintanH: Can do :+1:

FintanH: Oh wait. The fields are public. I think you can do `RadUrn { path..old_urn }`

proxy/coco/src/announce.rs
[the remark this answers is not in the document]
xla: I follow your argument, if memory serves I was making a point about plural/singular. Going to change the module name.

proxy/coco/src/announce.rs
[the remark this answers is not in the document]
NunoAlexandre: Actually, my comment misregards the context, which is at the same level of abstraction as this how-to. Sorry for the noise.

proxy/coco/src/announce.rs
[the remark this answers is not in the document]
NunoAlexandre: That might have been it, yeah :thought_balloon: Cool :+1:

proxy/api/src/main.rs
FintanH: Can't you catch and log the errors and `continue` in the `loop`?

proxy/coco/src/announce.rs
kim: I'm open to suggestions, but note that the difference is not other vs current peer: if you set `origin`, you're saying that _you_ have `remotes/<origin>/*`.

proxy/api/src/main.rs
```diff
@@ -83,15 +83,57 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    log::info!("Starting API");
     let api = http::api(ctx, args.test);
 
     warp::serve(api).run(([127, 0, 0, 1], 8080)).await;
 
     Ok(())
 }
+
+async fn announcement_watcher(ctx: context::Ctx) {
+    // TODO(xla): Take interval timings from config/settings.
```
FintanH, line 107: Are you capturing todo's in follow up tickets?

proxy/api/src/main.rs
```diff
@@ -83,15 +83,57 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let mut timer = tokio::time::interval(std::time::Duration::from_secs(10));
+
+    loop {
+        timer.tick().await;
+
+        if let Err(err) = announce(ctx.clone()).await {
+            log::info!("Announcement watcher errored: {:?}", err);
```
FintanH, line 114: `log::error` or `log::warn` would be better no?

proxy/coco/src/announcement.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+/// Load the cached announcements from the [`kv::Store`].
+///
+/// # Errors
+///
+/// * if the [`kv::Bucket`] can't be accessed
+/// * if the access of the key in the [`kv::Bucket`] fails
+#[allow(clippy::or_fun_call)]
+pub fn load(store: &kv::Store) -> Result<HashSet<Announcement>, Error> {
```
FintanH, line 98: Ah! I'm just realising that the `kv` is crossing over the boundary to `coco`. I think a suggestion here would be to have a trait for this capability, i.e.

```rust
pub trait AnnouncementStore {
  type Error;
  fn save(&self, updates: HashSet<Announcement>) -> Result<(), Self::Error>;
  fn load(&self) -> Result<HashSet<Announcement>, Self::Error>;
}
```

I'd suggest doing that as a follow-up maybe, but what do you think?

FintanH (approved):
Trying to think in terms of capabilities would help with separation of concerns, changeability, and testing.

Otherwise, looks dope :+1:

cloudhead (approved):

proxy/api/src/main.rs
xla: Will be addressed as part of #854.

proxy/api/src/main.rs
xla: #855

proxy/coco/src/announcement.rs
xla: #857

proxy/api/src/main.rs
FintanH: That ticket doesn't mention auditing the levels of the logging messages

xla: > The proxy codebase should be overhauled and be equipped with a sound logging strategy

FintanH: A bit vague but I think I get where you're coming from

PR #843, opened 2020-08-29
refactor(proxy): extract context

xla:
As more components will rely on access to common dependencies we need a safe way to share them, beyond just the http subs-system. By extracting the context machinery into a top-level module it is not part of the http domain anymore, which not just depends on an already constructed `Ctx` instead of constructing one on the spot.

Part of #602

### DIFF
--- a/proxy/api/src/context.rs
+++ b/proxy/api/src/context.rs
@@ -28,6 +28,12 @@ pub struct Context {
 }
 
 impl Context {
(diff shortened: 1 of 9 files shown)


cloudhead (approved):

rudolfs (approved):

FintanH (approved):
Looks straight forward to me :+1:

PR #853, opened 2020-09-01
refactor(proxy): move Context to use Signer

### LINKED ISSUE #753: Parameterize Api on Signer

xla:
Part of #753


proxy/api/src/http/control.rs
```diff
@@ -40,16 +40,11 @@ mod handler {
 
     use librad::paths;
 
-<<<<<<< HEAD
```
rudolfs: 👀

proxy/coco/src/announcement.rs
```diff
@@ -96,15 +96,19 @@ mod test {
...
 
     #[tokio::test]
     async fn announce() -> Result<(), Error> {
         let tmp_dir = tempfile::tempdir().expect("failed to create temdir");
         let key = SecretKey::new();
-        let config = config::default(key.clone(), tmp_dir.path())?;
+        let signer = signer::BoxedSigner::new(signer::SomeSigner {
+            signer: key.clone(),
+        });
+        let config = config::default(key, tmp_dir.path())?;
```
FintanH, line 108: Shouldn't the `config` functions take the `BoxedSigner` too? :thinking:

proxy/coco/src/peer.rs
```diff
@@ -774,17 +804,20 @@ mod test {
     #[tokio::test]
     async fn can_clone_user() -> Result<(), Error> {
         let alice_key = SecretKey::new();
+        let alice_signer = signer::BoxedSigner::from(signer::SomeSigner {
```
FintanH, line 807: A quality of life change might be to have `impl From<SecretKey> for BoxedSigner`

proxy/coco/src/announcement.rs
FintanH: Nvm

FintanH (approved):

proxy/coco/src/announcement.rs
xla: For others, this is not part of this change as we still need to land some improvements in link. Until then we at least migrate the majority of the codebase to use signer.

proxy/coco/src/peer.rs
xla: Rusts foreign rules would make that impossible to do in upstream. Do you think this should land in link?

FintanH: Ya, that's what I meant :) If we notice annoyances in proxy then we should make changes in link if they make sense.

rudolfs (approved):

FintanH (approved):

PR #856, opened 2020-09-01
feat(proxy): fetch project updates

### LINKED ISSUE #845: [fetching] Implement Fetch
We want to be able to fetch changes for a repo via [`fetch_repo`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/storage.rs#L715).

This should be similar to the `clone` functionality.

FintanH:
Closes #845 

Adding fetch to the set of methods on Api. This allows us to contact a
peer and fetch any changes. We test this by doing exactly that exchange,
where one peer commits and the other follows up with a fecth.

### DIFF
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -772,6 +772,42 @@ mod test {
     }
 
     #[tokio::test]
+    async fn can_clone_user() -> Result<(), Error> {
+        let alice_key = SecretKey::new();
+        let bob_key = SecretKey::new();
+
+        let alice_tmp_dir = tempfile::tempdir().expect("failed to create temdir");
+        let config = config::default(alice_key.clone(), alice_tmp_dir.path())?;
+        let alice_peer = Api::new(config).await?;
+
+        let bob_tmp_dir = tempfile::tempdir().expect("failed to create temdir");
+        let config = config::default(bob_key.clone(), bob_tmp_dir.path())?;
+        let bob_peer = Api::new(config).await?;
+
+        let alice = alice_peer.init_user(&alice_key, "alice")?;
+        let bobby = bob_peer.clone();
+        let user_urn = tokio::task::spawn_blocking(move || {
+            bobby.clone_user(
+                alice.urn().into_rad_url(alice_peer.peer_id()),
+                vec![alice_peer.listen_addr()].into_iter(),
+            )
+        })
+        .await
+        .expect("failed to join thread")?;
+
+        assert_eq!(
+            bob_peer
+                .list_users()?
+                .into_iter()
+                .map(|user| user.urn())
+                .collect::<Vec<_>>(),
+            vec![user_urn]
+        );
+
+        Ok(())
+    }
+
+    #[tokio::test]
     async fn can_fetch_project_changes() -> Result<(), Error> {
         let alice_key = SecretKey::new();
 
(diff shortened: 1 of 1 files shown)


proxy/coco/src/peer.rs
```diff
@@ -113,6 +113,20 @@ impl Api {
...
+    /// Check the storage to see if we have the given commit for project at `urn`.
+    ///
+    /// # Errors
+    ///
+    ///   * Checking the storage for the commit fails.
+    ///
+    /// # Panics
+    ///
+    ///   * Unable to acquire the lock.
+    pub fn has_commit(&self, urn: &RadUrn, oid: impl Into<git2::Oid>) -> Result<bool, Error> {
```
xla, line 125: We have `coco::oid` now and probably should use that in the public api of this crate. It supports `into` to `git2::Oid`.

proxy/coco/src/peer.rs
```diff
@@ -250,6 +264,24 @@ impl Api {
         Ok(storage.metadata_of(urn, peer)?)
     }
 
+    /// Fetch any updates at the given `RadUrl`, providing address hints if we have them.
+    ///
+    /// **N.B.** This needs to be run with `tokio::spawn_blocking`.
+    ///
+    /// # Errors
```
xla, line 271:
```suggestion
    /// # Errors
    ///
```

proxy/coco/src/peer.rs
```diff
@@ -702,6 +736,102 @@ mod test {
...
+            let mut rad = repo.find_remote(config::RAD_REMOTE)?;
+            rad.push(&[&format!("refs/heads/{}", project.default_branch())], None)?;
+            commit_id
+        };
+
+        let bobby = bob_peer.clone();
+        let url = project_urn.into_rad_url(alice_peer_id.clone());
+        tokio::task::spawn_blocking(move || bobby.fetch(url, vec![alice_addr]))
+            .await
+            .expect("failed to join thread")?;
```
xla, line 817:
```suggestion
            .expect("failed to fetch")?;
```

xla (changes requested):

proxy/coco/src/peer.rs
FintanH: Aw yis, saw that change but forgot. Thanks :+1:

proxy/coco/src/peer.rs
FintanH: Wellll actually, the error I'm expecting on is `JoinThread`. The `?` is the failure to fetch.

proxy/coco/src/peer.rs
FintanH: It's kind of weird that I would take `coco::Oid` just call `into` and then in my test case do `coco::Oid::from`, it's just a no-op. So I think leaving this as `Into<git2::Oid>` is fine, because then it can take a `coco::Oid` _or_ a `git2::Oid`.

xla: DIdn't see the `Into` part. Alles gut.

proxy/coco/src/peer.rs
xla: Sneaky.

xla (approved):
😇 🕯 🐥 🔆

rudolfs (approved):

Issue #754, opened 2020-08-05
ui: Navigation/Routing refactor


xla:
# Navigation Refactor

We have accumulated a plethora of issues which stem from our current routing
strategy and/or the libraries we use. To address this a strategy has been
devised to build up a simple scheme to support our needs and play into the
abstractions we have been established.

## Tracking issues

* #148
* #542
* #648
* #650

## TODOS

* [ ] Implement View based routing (#622)
* [ ] Implement new Overlay
* [ ] Implemetn new global dropdowns
* [ ] Document architecture and usage

#### constraints

* built around native Svelte components
* nested navigations (i.e. Screen 1:N Fragment)
* optionally prefetching before navigating to/rendering
* optionally pass of typed props
* easy to extend for helper code to build reoccurring nav elements.

At the core is the concept of a `View` which represents a parameterised and
prepared component. A map of views and their typed keys as the datastructure
will make it possible to construct a `Navigation` that holds the history of
views and allows - when constructed as a singleton - to be easily advanced from
any component. Such a `Navigation` will be passed to a `Router` component which
owns the rendering of the latest `View` and can be used in Svelte natively.

``` typescript
// src/view.ts

export Component extends SvelteComponent {
  readonly key: K;
  readonly prefetch()?: Promise<void>;
}

export type Props = Record<string, 0 | string>;
export type Map<Key extends string> = Required<Record<Key, Component>>;

export interface View<K extends string> {
  readonly component: Component;
  readonly key: K,
  readonly props?: Props;
  readonly prefetch()?: Promise<void>;
}

export interface Navigation<K extends string> {
  readonly current: Readable<View<K>>;
  back(): void;
  set(key: Key, props?: Props): void;
}

export const create = <K extends string, C extends typeof SvelteComponent>(
  componentMap: Map<K, C>,
  initial: K
): Navigation<K> => {
  const hist = history.create<View<K>>({
    component: componentMap[initial],
    key: initial,
  });

  return {
    current: hist.current,
    back: (): void => {
      hist.pop();
    },
    set: (key: K, props?: Props): void => {
      const component = componentMap[key];

      if (component.prefetch) {
        await component.prefetch();
      }

      hist.push({
        component: component,
        key,
        props,
      });
    },
  };
};
```

``` svelte
<!-- View/Router.svelte --->

<script lang="ts">
  import * as view from "../src/view";

  export let nav: view.Navigation<string>;
  const current = nav.current;
</script>

{#if $current.props}
[...]

PR #858, opened 2020-09-01
refactor(ui): project creation modal

### LINKED ISSUE #826: Keyboard shortcuts don't work as expected after navigating with subrouter
To reproduce:
1) Go to a `Project`
2) Click to Issues or Revisions tab
3) Try any keyboard shortcut (`modifierKey` + `,`; `modifierKey` + `p`, etc)

Modal will not appear

4) Now try clicking anywhere in the window and a keyboard shortcut - modal will appear

### LINKED ISSUE #835: Overhaul new project modal
- [x] use new floating style modal
- [x] change wording
- [x] add emoji
- [x] move repository selectors to the top
- [ ] Update the disabled field globally to include new background color
- [x] add tooltip to project name field when starting with an existing repo

MeBrei:
Closes #826 
Closes #835


rudolfs (commented):
Looks good!

Some things I noticed:

- [ ] circular deps
```
(!) Circular dependencies
ui/src/modal.ts -> ui/Screen/ProjectCreation.svelte -> ui/DesignSystem/Component/index.js -> ui/DesignSystem/Component/ModalOverlay.svelte -> ui/src/modal.ts
ui/src/modal.ts -> ui/Screen/ProjectCreation.svelte -> ui/DesignSystem/Component/index.js -> ui/DesignSystem/Component/Sidebar.svelte -> ui/src/modal.ts
```
- [ ] lock screen during create request broken
- [ ] keyboard shortcuts don't work when opening the modal via the "new project" button (might be related to https://github.com/radicle-dev/radicle-upstream/issues/826)

xla:
> ```
> (!) Circular dependencies
> ui/src/modal.ts -> ui/Screen/ProjectCreation.svelte -> ui/DesignSystem/Component/index.js -> ui/DesignSystem/Component/ModalOverlay.svelte -> ui/src/modal.ts
> ui/src/modal.ts -> ui/Screen/ProjectCreation.svelte -> ui/DesignSystem/Component/index.js -> ui/DesignSystem/Component/Sidebar.svelte -> ui/src/modal.ts
```

I ran into that as well when working on the navigation refactor. It might be good practice to separate the library code providing the machinery and the functionality mapping to actual components.

File: ui/Screen/Profile/ProjectsMenu.svelte [hunk 1]
```diff
@@ -1,15 +1,16 @@
 <script>
-  import { push } from "svelte-spa-router";
-
   import * as path from "../../src/path.ts";
+  import * as modal from "../../src/modal.ts";
```
xla, line 3:
```suggestion
  import * as modal from "../../src/modal.ts";
  import * as path from "../../src/path.ts";
```

xla (changes requested):
I failed to bring this up during the original introduction of the new modal, but I don't see a reason why we have to hold onto the `path` concept even so we don't use a router anymore. An enum for our differrent modals could look like this:

``` ts
// app.ts

import { SvelteComponent } from "svelte";

import Help from "../Modal/Help.svelte";
import ProjectCreation from "../Modal/ProjectCreation.svelte";
import Shortcuts from "../Modal/Shortcuts.svelte";

export interface View<K extends string> {
  readonly component: unknown;
  readonly key: K,
  readonly props?: Props;
}

export type Props = Record<string, 0 | string>;
export type Map<Key extends string, C extends typeof SvelteComponent> = Required<
  Record<Key, C>
>;

export enum Modal {
  Help = "Help",
  ProjectCreation = "ProjectCreation",
  Shortcuts = "Shortcuts",
}

export const modalMap: Map<Modal, typeof SvelteComponent> = {
  [Modal.Help]: Help,
  [Modal.ProjectCreation]: ProjectCreation,
  [Modal.Shortcuts]: Shortcuts,
};
```

MeBrei:
Closes #826 

3ec3d9a should also fix the issue above

MeBrei:
> I failed to bring this up during the original introduction of the new modal, but I don't see a reason why we have to hold onto the `path` concept even so we don't use a router anymore. An enum for our differrent modals could look like this:
> ...

I refactored the setup of the modals in 511155f36c5cbfd56b57632611fa8b9e11b1bd3a
I kept the routes for now, this way, I like that everything is set up the same (ie routes map in `App.svelte` and the routes being accessible via `path.ts`). Having the mapping of the routes in `App.svelte` also avoids the circular dependencies. I also moved the modals into `Screen` as some of them have advanced functionality (eg API calls) that don't fit into `Component` and the modals basically form a screen in themselves.

xla:
> I kept the routes for now, this way, I like that everything is set up the same (ie routes map in `App.svelte` and the routes being accessible via `path.ts`). Having the mapping of the routes in `App.svelte` also avoids the circular dependencies. I also moved the modals into `Screen` as some of them have advanced functionality (eg API calls) that don't fit into `Component` and the modals basically form a screen in themselves.

The consistency argument is a good one and I think we can drop the route concept once we migrate away from router based navigation altogether in #754.

I also believe that modals should be a first class concept and be treated like screens in the hierarchy and in their allowed capabilities/access. It feels strange to move them into the screen directory, I'd put them `ui/Modal`. So we can have:

```
./
../
ui/
ui/Modal/MyModal.svelte
ui/Modal/MyModal/Fragment1.svelte
ui/Modal/MyModal/Fragment2.svelte
ui/Screen/MyScreen.svelte
ui/Screen/MyScreen/Fragment1.svelte
ui/Screen/MyScreen/Fragment2.svelte
```

xla (approved):
🌙 🎁 🕧 🎸

rudolfs:
Found some edge cases, that don't make sense after the reshuffling of the inputs in the form. Feel free to create a follow up issue if it's too much hassle to address in this PR.

- [ ] when choosing "create new repo", after selecting the directory path it should focus on the name field instead of immediately showing a validation error
Screenshot 2020-09-04 at 10 50 28

- [ ] when there's only one character in the name field, and the validations don't pass, the "create project" button is already enabled, it should stay disabled until all validations pass
Screenshot 2020-09-04 at 10 50 11

MeBrei:
> Found some edge cases, that don't make sense after the reshuffling of the inputs in the form. Feel free to create a follow up issue if it's too much hassle to address in this PR.
> 

I think it would make sense to refactor the validation set up for this modal before addressing the edge cases, so I created a new issue #869

rudolfs (approved):

Issue #549, opened 2020-06-23
Clean up API playground


MeBrei:
- [ ] some input examples are out of date, making test requests fail (eg register project - wrong `maybeCocoId`)
- [ ] some return examples are out of date (eg outdated coco id representation)
- [ ] coco ids that sent via the playground (eg find project by ID) have encoded special characters which can't be parsed (returns a `LibradParseUrn InvalidNID` error)

PR #859, opened 2020-09-01
refactor(proxy): drop warp documentation

### LINKED ISSUE #846: Drop warp documentation
Originally we wanted to have a self-documenting API, we came a long way, but the overhead it causes to keep the docs/playground up-to-date is significant. Especially now as things are moving a lot, it's not giving us enough in return. This also means we can depend on warp mainline again and don't have to maintain a fork.

xla:
Closes #846

### DIFF
--- a/proxy/Cargo.lock
+++ b/proxy/Cargo.lock
@@ -962,7 +962,6 @@ checksum = "86b45e59b16c76b11bf9738fd5d38879d3bd28ad292d7b313608becb17ae2df9"
 dependencies = [
  "autocfg",
  "hashbrown",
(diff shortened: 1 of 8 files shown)


NunoAlexandre (approved):
:broom: 

It's a sensible thing to do given where we are right now.

FintanH (approved):
I never did like this trait business :+1:

rudolfs:
Closes: https://github.com/radicle-dev/radicle-upstream/issues/549.

rudolfs (approved):
Byeee 👋

PR #860, opened 2020-09-01
refactor(proxy): move caching out of coco crate

### LINKED ISSUE #857: Take a trait as announcement store
https://github.com/radicle-dev/radicle-upstream/pull/838#discussion_r481135351

xla:
Make the caching of announcements a session concern.

Closes #857

### DIFF
--- a/proxy/Cargo.lock
+++ b/proxy/Cargo.lock
@@ -309,7 +309,6 @@ name = "coco"
 version = "0.1.0"
 dependencies = [
  "futures 0.3.5",
- "kv",
  "lazy_static",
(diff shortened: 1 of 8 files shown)


rudolfs (approved):

FintanH (approved):
Ah right, because we even called `load` and `save` only from `main`! It still might be nice to have `trait`s to model behaviours in the long run but I'm happy with this.

xla:
> Ah right, because we even called `load` and `save` only from `main`! It still might be nice to have `trait`s to model behaviours in the long run but I'm happy with this.

We can do that if we want to inject the actual implementation, for now we can group store/on-disk concerns under the session.

PR #862, opened 2020-09-01
chore(proxy): update crate metadata

### LINKED ISSUE #844: Update crate metadata consistently
Our [BDFL](https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/coco/Cargo.toml#L4) should be replaced with the monadic/radicle details :)

xla:
Closes #844


rudolfs (approved):

rudolfs (approved):

proxy/api/Cargo.toml
```diff
@@ -4,10 +4,10 @@ version = "0.1.0"
 license = "GPL-3.0-only"
 authors = ["radicle team <dev@radicle.xyz>"]
 edition = "2018"
-description = "Proxy to serve a specialised API to the Upstream UI."
+description = "Proxy to serve Upstream domain."
```
sarahscott, line 7: This is definitely an improvement; I'd probably say "Proxy to serve Upstream application" or  "Proxy to serve Upstream desktop application"

FintanH (approved):
Looks good :+1: I like @sarahscott's tweak to the description

proxy/api/Cargo.toml
xla: The reason I worded it this way is that the Upstream application consists of a native electron container, the ui and the proxy. So it's miss-leading to say the proxy serves the application. Furthermore you can run the proxy in isolation and interact with it and access what we understand as the domain surface of Upstream the product. /cc @FintanH

FintanH: Fair enough :+1:

PR #863, opened 2020-09-02
fix(ui): update and fix rollup

geigerzaehler:
While trying to integrate the `@walletconnect/client` I ran into some issues with rollup that this PR adresses

* We change the package name of some rollup plugins to get the latest versions
* We switch the order of the `commonjs` and `typescript` plugins. Now the `browser` field in NPM packages is respected and we’re using the browser builds for these packages where node builds may fail.


xla (changes requested):
The documentation of the node-resolve plugin indicates that there needs to be a certain order when used with **commonjs**: https://github.com/rollup/plugins/tree/master/packages/node-resolve#using-with-rollupplugin-commonjs

geigerzaehler:
> The documentation of the node-resolve plugin indicates that there needs to be a certain order when used with **commonjs**: https://github.com/rollup/plugins/tree/master/packages/node-resolve#using-with-rollupplugin-commonjs

Yes. It seem that it was only necessary to flip the order of `commonjs` and `typescript`. Now `commonjs` still comes after `nodeResolve`.

xla (approved):
🐟 🈹 ⚙ 🔝

rollup.app.js
```diff
@@ -28,20 +28,16 @@ export default {
...
-    // https://github.com/rollup/rollup-plugin-commonjs
     resolve({
       browser: true,
       dedupe: importee =>
         importee === "svelte" || importee.startsWith("svelte/"),
     }),
+
     commonjs(),
 
+    typescript(),
```
rudolfs, line 39: Should we maybe document here that the order is important?

rudolfs (approved):

Issue #864, opened 2020-09-02
Reintroduce notifications of meaningful events in the proxy


xla:
We used to have notification machinery in the proxy which could be use to inform the UI of meaningful events based on a push model. Given that we start to introduce more and more activity, esp. network related, in the proxy it would be helpful to surface these signals to the UI for at least debugging purposes. Ultimately we can enable more advanced use-cases like #725.

rudolfs:
@xla is this done?

xla:
There is still the UI part to be done, have a some infrastructure to set up listeners and show the events in ways that make sense, i.e.: notifications when project requests make advances.

rudolfs:
@xla now?

PR #865, opened 2020-09-02
feat(proxy): reintroduce notifications

xla:
Introduces notifications machinery in the proxy, which entails managing subscriptions, broadcasting new events and consume them via an SSE endpoint.

First part of #864


proxy/api/src/notification.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// Generator of unqiue keys for subscriptions.
+    next_id: Arc<AtomicUsize>,
+    /// Active subscribers.
+    subs: Arc<RwLock<HashMap<usize, mpsc::UnboundedSender<Notification>>>>,
+}
+
+impl Subscriptions {
+    /// Broadcast [`Notification`] to all active subscriptions.
+    pub async fn broadcast(&self, notification: Notification) {
+        // We use retain to remain all closed subscriptions.
```
cloudhead, line 31:
```suggestion
        // We use retain to discard all closed subscriptions.
```
Nice trick

proxy/api/src/notification.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use serde::Serialize;
+use tokio::sync::mpsc;
+use tokio::sync::RwLock;
+
+/// Significant events happening during proxy runtime.
+#[derive(Clone, Debug, Serialize)]
+pub enum Notification {
+    /// Our local peer started listening on a local socket.
+    LocalPeerListening(SocketAddr),
```
cloudhead, line 16: I find the "local peer" terminology kind of confusing in this context.. Maybe just `Listening`? :man_shrugging:

xla: The reason I prefixed it overly verbose is to account for potentiall other things we start to listen on a socket, we already listen for http requests on another addr, so `Listening` can be qutie ambiguous.

proxy/api/src/http/notification.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /// Maps the [`crate::notification::Notification`] from a subscription to an
+    /// [`sse::ServerSideEvent`].
+    fn map(
+        subscriber: mpsc::UnboundedReceiver<Notification>,
+    ) -> impl Stream<Item = Result<impl sse::ServerSentEvent, Infallible>> {
+        subscriber.map(|notification| match notification {
+            Notification::LocalPeerListening(addr) => {
+                Ok((sse::event("LOCAL_PEER_LISTENING"), sse::json(addr)))
```
cloudhead, line 47: To keep things in one place, could we have a method on `Notification` that either returns an `sse::event` or a `&'static str`?

proxy/api/src/notification.rs
cloudhead: What is this event supposed to mean to the frontend? Could something like `NetworkReady` work?

proxy/api/src/http/notification.rs
xla: Was about to do that, but there are a couple of oddities. Notifications shouldn't know about its usage as sse event. Was thinking about impling `Display`, but it would be a bit whack if it just renders the enum name. What do you suggest?

cloudhead: Coud you implement an `Into<sse::Event> for Notification` inside the `http/notifications` module?

cloudhead: Wouldn't "keep things in one place" but at least there would be only one way of constructing an event from it.

proxy/api/src/notification.rs
xla: To any api consumer it will signal that the peer we run locally as part of the proxy has physically been bound to a socket on the local machine and is listening on that. Examples of other notifications are: `PeerConnected`, `PeerDisconnected` - which refer to remote peers connecting to our local peer. Those are at least the terms I'm thinking of.

I see where you going with `NetworkReady` as it is trying to convey things on a higher level. But could also mean different things, maybe we consider readiness once we have connected to a couple of peers and have synced up. Not necessarily what I'm advocated for, my point is we would need to define what readiness meanss

proxy/api/src/http/notification.rs
xla: It would keep "things in the right place" as it would keep the concerns of how to construct the sse events in the context where they are used. Dunno if the `Into` is possible as `sse::SeverSentEvent` (what `sse::event` returns) is a trait.

proxy/api/src/notification.rs
FintanH: How is this event generated? Doesn't our app fail if the `Peer` doesn't `accept` to get the `PeerApi`?

xla: It's currently not generated yet. We will hook this up wherever we end up making sense of protocol events, currently:

https://github.com/radicle-dev/radicle-upstream/blob/9f8981de6fc9b00f2d564b322a48f0050ab5543b/proxy/coco/src/peer.rs#L63-L67

This will be one of the follow-ups to surface these to consumers. One of the use cases I have in mind at the moment, is a debug/developer view where you can see the log of all events from the proxy in the UI, to help understand what is going on and/or you don't have access to the logs.

FintanH: Would it not make more sense to do:
```rust
pub enum Notification {
  Protocol(ProtocolEvent),
}
```
using [`ProtocolEvent`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/net/protocol.rs#L56-L61) as the variant?

xla: I don't think it's wise to just pass through the `librad` protocol events and not make an active effort to curate what notifications we are emitting with what semantics.

FintanH: The curation can still be done by `match`ing on `Protocol`. On the flip side, not using `ProtocolEvent` means we have to maintain a mapping.

xla: My point is we want to maintain that mapping and `notification.rs` is the place to do that as this is where we curate the list of notifications the proxy can emit and will surface to clients. Let's take the `ProtocolEvent::Gossip` variant, a client needs to have deep knowledge of the protocol library in order to make it palatable for a user. Instead the notifications should be self-contained and ready to be presented. A use-case I'm thinking along is a`tail` of notifications showing us peer connection events, announcements and requests of heads to help understand network activity.

FintanH: Ok, but to expand what I mean by "The curation can still be done by `match`ing" is that we could have something like:
```rust
enum Message {
  Interesting { event_name: String, value: Json }
  Uninteresting,
}

impl Message {
  fn fire(&self, subscription: Subscription) {
    match self {
      Uninteresting => { /* do nothing */ }
      Interesting { event_name, value } => /* fire message off to Subscription */
    }
  }
}

enum Notification {
  Protocol(ProtocolEvent)
}

impl Notifiction {
  fn curate(&self) -> Message {
    match self {
      Protocol(protocol) => if protocol.is_gossip() { Uninteresting } else { make_interesting(protocol) }
    }
  }
}
```

A bit pseudo-codey but hopefully it gets my point across :)

xla: Where would you expect this to live, i.e. at which layer? Don't really see why we would need this extra machinery if we can already determine interesting events with the definition of the `Notification` variants.

FintanH: I could see both of these living here. The idea being that notification is rich with information, and this gets crushed down to something that's easy to communicate, i.e. this `Message` thing. Then we have the ability to transform or ignore the rich information into the easy to communicate information.

The advantage being that once we realise we're interested in having gossip notifications, we just change the handling of that case and turning into a message. Rather than adding a new case and adding handling logic for each place that case pops up.

Another advantage being that we don't need to come up with new names, which prompted this thought in the first place :smile:

FintanH (approved):

Issue #851, opened 2020-09-01
[track project] differentiate tracked projects and my projects


FintanH:
There is a difference between a tracked project and a project that we have created or contributed to. They are extremely similar, in that they're both replicated into our Monorepo. The difference, however, is that a project that we have created or contributed to is one that contains `signed_refs`. 

This means that we should have a way of differentiating them in our code. I would suggest a newtype over `Project`, i.e. `TrackedProject`, and also suggest to change `list_projects` to filter based on the above criteria.

PR #866, opened 2020-09-02
feat(proxy): differentiate tracked projects from my projects

sarahscott:
Screen Shot 2020-09-02 at 5 16 29 PM

A user will have two tabs in their own profile: one for **My Projects**, and another for **Following/Tracking**. A working definition for a "contributed" project (i.e. one that would show up in "My Projects") is a project that a user has contributed to, which should therefore contain [`signed_refs`](https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/identity_resolution.md#namespacing). A "tracked" (or followed?) project is almost exactly the same as an owned one, but the user hasn't contributed to it yet, so it shouldn't have `signed_refs`. 

- [x] Replace `projects/list` endpoint with `projects/contributed` and `projects/tracked`
- [x] Add a `newtype` for `Tracked` projects
- [x] Add a supertype `Projects` that contains both `tracked` and `contributed` `Project`'s
- [x] Separate endpoint for `projects/peer/<id>`

Closes https://github.com/radicle-dev/radicle-upstream/issues/851


proxy/api/src/http/project.rs
```diff
@@ -216,11 +218,30 @@ pub struct MetadataInput {
...
     user: Option<coco::Urn>,
+    /// Only include projects of the given `Status`
+    ///
+    /// **N.B.** This only works for _your_ projects; if you provide a `user`,
+    /// it will be ignored
+    ///
+    /// TODO(sos): If this becomes Option<Status>, I get a MissingOwner error 🤔
+    /// I think the error happens in `with_qs_opt` at the `match
+    /// serde_qs::from_str(&query)`
+    status: Option<String>,
```
FintanH, line 244: If `Status` is `Deserialize` we can just use the type directly. But now I'm reading your TODO comment :sweat_smile: `MissingOwner` means you haven't set the default owner for this session.

proxy/coco/src/peer.rs
```diff
@@ -259,6 +259,29 @@ impl Api {
...
+    ///
+    /// # Errors
+    ///
+    /// * if opening the storage fails
+    pub fn list_my_projects(&self) -> Result<Vec<librad_project::Project<entity::Draft>>, Error> {
+        let projects = self
+            .list_projects()?
+            .into_iter()
+            .filter(|project| {
+                // TODO(sos): either update list_project_refs() to return an Option
```
FintanH, line 272: `Refs` is an iterator, so if there is none the iterator will be empty, i.e. `refs.next().is_none()`

FintanH (commented):
Good start, just dropping down some thoughts for now

proxy/api/src/http/project.rs
xla: Adding this as a query parameter only makes sense if we ever want the unfiltered list . It also begs the question what is the behaviour in combination with the `user` query param. Adding this increases the combinations we have to think about for this endpoint.

With all that said, we will be better served with distinct endpoints. One for projects the parameterised user (defaults to current session idenitty when param is not set) has contributed to and one the parameterised user is tracking. It would avoid a lot of complexity imho.

proxy/api/src/project.rs
```diff
@@ -105,6 +108,26 @@ pub fn list_projects(api: &coco::Api) -> Result<Vec<Project>, error::Error> {
         .collect()
 }
 
+/// List all projects you have contributed to.
+///
+/// # Errors
+///
+///   * We couldn't get a project list
+///   * We couldn't get project stats
+pub fn list_my_projects(api: &coco::Api) -> Result<Vec<Project>, error::Error> {
```
xla, line 117: As mentioned in the context of the http changes, could this be generalised to a method which, given a user, return the list of projects that user has contributed to (contribution here as per definition in the PR description).

xla (changes requested):
Dope! There is some reshuffling which would get rid off of some of the complexity.

proxy/api/src/http/project.rs
FintanH: I was advocating for adding the parameter. My thinking was these are all projects in the user's view the only difference being this check if they're "mine" or "tracked". 
To get a feel, what's you criteria for when to add a parameter? So like, in this case, we have similar behaviour but there's a fork where return one type of that data vs another. So you would advocate for separate endpoints in these kinds of scenarios?

> It also begs the question what is the behaviour in combination with the user query param. 

Indeed, it's a bit weird. We _could_ differentiate user projects as owned and tracked too (via their signed_refs).

>  One for projects the parameterised user (defaults to current session idenitty when param is not set)

That actually doesn't work because behind the scenes we're checking the `remotes` using this `user: coco::Urn`. So passing your own user will give you back an empty list :upside_down_face:

proxy/api/src/project.rs
FintanH: Thinking about this, and I swear I had this idea before but maybe I miscommunicated or left it out, but I feel like a lot of this could be cleared up by creating a variant of [`all_metadata`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/storage.rs#L231-L241), called `all_metadata_of` that filters the list based on `remotes/<peer_id>`. 

That way you pass a `PeerId` when you want to look at specific profile.

proxy/api/src/http/project.rs
xla: > To get a feel, what's you criteria for when to add a parameter? So like, in this case, we have similar behaviour but there's a fork where return one type of that data vs another. So you would advocate for separate endpoints in these kinds of scenarios?

Generally whenever whenever we have a mild cardinality explosion and becomes to reason about all combinations. Furthermore it's often an indicator of watering down the purpose of a single endpoint and trying to put too much functionality in one place. Think about explaining to a client accessing the endpoint how to parameterise it to what outcome.

> That actually doesn't work because behind the scenes we're checking the `remotes` using this `user: coco::Urn`. So passing your own user will give you back an empty list upside_down_face

Why not, that's the current behaviour. If you don't provide a urn it will use the current user. Or am I miss-understanding your point?

FintanH: > Think about explaining to a client accessing the endpoint how to parameterise it to what outcome. If that exercise is hard, it's probably time to revise the design of the proposed design.

Ok, so in this case _I feel_ like the explanation is straight forward. The `/projects` endpoint gives you back projects, where `mine` filters by projects you've contributed to, and `tracked` filters by projects you are merely tracking.
Now, that's subjective and maybe I'm missing the complication :)

> Why not, that's the current behaviour. If you don't provide a urn it will use the current user. Or am I miss-understanding your point?

So if this parameter is `Some(urn)`:
1. We get the default user's projects
2. For each project get its tracked list
3. Filter the tracked list by the user pointed to by the urn

If the parameter is `None`:
1. List the projects where `rad/self` == `default_rad_self` <-- this filter condition is the confusing part because of session state and monorepo state being out of sync.

xla: > Ok, so in this case _I feel_ like the explanation is straight forward. The `/projects` endpoint gives you back projects, where `mine` filters by projects you've contributed to, and `tracked` filters by projects you are merely tracking.
> Now, that's subjective and maybe I'm missing the complication :)

The complication comes from the fact that the endpoint supports two query params, neither of them is required. Which gives us three 2 states for the user one and 3 for the status one. Building up that matrix of the combinations is what I'm referring to when explaining what to expect from the response.

Let's run with it for now. Isn't the list of contributed projects just a sub-set of tracked project for a user? If that's the case I think we can simplify by having boolean query param `Contributed`, which if not passed or set to false will return all tracked projects for the user.

> 1. List the projects where `rad/self` == `default_rad_self` <-- this filter condition is the confusing part because of session state and monorepo state being out of sync.

What are the scenarios where the session state and monorepo state are out-of-sync?

proxy/api/src/project.rs
xla: That makes a lot of sense to me! 

> That way you pass a `PeerId` when you want to look at specific profile.

Wouldn't we need to multiple calls for all `PeerId`s we know for the user?

xla: radicle-dev/radicle-link#306

FintanH: > Wouldn't we need to multiple calls for all PeerIds we know for the user?

Potentially, but I think it might be safe to assume we're looking at one user for now. I think that simplification is assumed in a few places at this stage.

proxy/api/src/http/project.rs
FintanH: > The complication comes from the fact that the endpoint supports two query params, neither of them is required. Which gives us three 2 states for the user one and 3 for the status one. Building up that matrix of the combinations is what I'm referring to when explaining what to expect from the response.

Ah I wholeheartedly agree with you there. If we're adding `Status` then it should be mandatory. And I would say splitting off the user one into a separate endpoint.

> Let's run with it for now. Isn't the list of contributed projects just a sub-set of tracked project for a user? If that's the case I think we can simplify by having boolean query param Contributed, which if not passed or set to false will return all tracked projects for the user.

But this is also true, so I think that might be a better way forward :raised_hands: 

> What are the scenarios where the session state and monorepo state are out-of-sync?

1. Open app create an identity
2. Create a project (note the above user gets set as the `rad/self` for this project)
3. Clear the session (note the monorepo is still persisted)
4. Create the new identity

At this stage we have two identities and we're in a weird state where we need to filter out the project we created before. My feeling is that we should ask the monorepo what our default self was instead of creating a new identity. Then if we want to handle multiple users logging in then we actually give that some thought and design. We seem to be in a half-way house of horror :scream: What do you think?

xla: > Ah I wholeheartedly agree with you there. If we're adding `Status` then it should be mandatory. And I would say splitting off the user one into a separate endpoint.

Dope, unfortunately query params are optional by design. Which means we accept that and non-presence has a sensible default meaning OR we encode as fragment in the path to make it required.

> At this stage we have two identities and we're in a weird state where we need to filter out the project we created before. My feeling is that we should ask the monorepo what our default self was instead of creating a new identity. Then if we want to handle multiple users logging in then we actually give that some thought and design. We seem to be in a half-way house of horror scream What do you think?

Wouldn't that be remedied by not allowing a new identity to be created until we destroy the old state/monorepo? To me it sounds like an impossible state and comes from the fact we haven't scrutinized the flows around "logging out"/session termination properly.

FintanH: > Dope, unfortunately query params are optional by design. Which means we accept that and non-presence has a sensible default meaning OR we encode as fragment in the path to make it required.

I ended up putting it in the path for this reason exactly :smile: _changes pending_

> Wouldn't that be remedied by not allowing a new identity to be created until we destroy the old state/monorepo? To me it sounds like an impossible state and comes from the fact we haven't scrutinized the flows around "logging out"/session termination properly.

Well, I wouldn't destroy the monorepo :smile: I'd advocate for archiving it and notifying the user. Wouldn't be fun if you ended up destroying all your monorepo state by accident.
So, I mean is it "impossible" in that we don't expect users to clear their session? Because this seems to be something we've been doing a lot but maybe that's just for testing?

proxy/api/src/project.rs
```diff
@@ -71,6 +75,76 @@ where
...
+    ///     a. Created themselves using the application.
+    ///     b. Replicated (see tracked above), checked out a working copy, and push changes to
+    ///     references.
+    ///
+    /// The conditions imply that a project is "mine" if I am the maintainer or I have contributed
+    /// to the project.
+    pub mine: Vec<Project>,
+}
+
+impl Projects {
```
xla, line 108: Introducing this type might impede on the ergonomics, as we in all places expect a `Vec` which can easily be manipulated like calling `first`. Additionally it might not communicate as well the actual shape of the data. Would it be feasible to approach this differently with lists of:

``` rust
enum ProjectRelation {
    Contributed(Project),
    Tracked(Project),
}

fn list(api: &coco::Api) -> Vec<ProjectRelation> {
    todo!()
}
```

With the help of that the caller can normally operate on the list by filtering, mapping and whatnot.

proxy/api/src/project.rs
```diff
@@ -71,6 +75,76 @@ where
...
+            } else {
+                projects.mine.push(project)
+            }
+        }
+
+        Ok(projects)
+    }
+
+    /// Give back an iterator over all the [`Project`]s.
+    pub fn into_iter(self) -> impl Iterator<Item = Project> {
```
xla, line 141: Shouldn't we impl [`IntoIterator`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) if we want this kind of API?

proxy/api/src/http/project.rs
```diff
@@ -216,11 +201,36 @@ pub struct MetadataInput {
...
+    /// The project exists in your monorepo and you're interested in it, but you've
+    /// not yet contributed to it.
+    Tracked,
+    /// You've contributed to this project and it exists in your monorepo.
+    Mine,
+}
+
+impl Default for Status {
+    fn default() -> Self {
+        Status::Mine
```
xla, line 216:
```suggestion
        Self::Mine
```

proxy/api/src/http/project.rs
```diff
@@ -216,11 +201,36 @@ pub struct MetadataInput {
...
 }
 
+/// The status of a project in your monorepo.
+#[derive(Deserialize, Serialize, Debug)]
+pub enum Status {
+    /// The project exists in your monorepo and you're interested in it, but you've
+    /// not yet contributed to it.
+    Tracked,
+    /// You've contributed to this project and it exists in your monorepo.
+    Mine,
```
xla, line 211:
```suggestion
    Contributed,
```

xla (changes requested):

proxy/api/src/project.rs
FintanH: The caller could do `into_iter().first()`. But if they wanted to get the first of a tracked or contributed project it seems easier to do `projects.tracked.first()` over something that needs to do a `filter` followed by `first`. 

Since the idea is these things are generally separated, I'd like to try to keep it in this form. Unless you see any other strong reasons why it would be better to have the above?

sarahscott:
I do feel a little weird about the `Projects { tracked, contributed }` struct; @FintanH can you remind me why we want tracked projects to have a `newtype` as opposed to the `Project` struct having a field like `contributed: boolean`?

Also, given the `newtype`, would it make sense to replace the `Projects` struct with filtering functions like `get_tracked_projects(...) -> Result<Vec<Tracked>>` & `get_contributed_projects(...) -> Result<Vec<Project>>`?

FintanH:
> I do feel a little weird about the Projects { tracked, contributed } struct; @FintanH can you remind me why we want tracked projects to have a newtype as opposed to the Project struct having a field like contributed: boolean?

Can you expand on what "feels weird" possibly?

If you use a newtype you're signalling to the compiler that these two things are different but have the same underlying structure. Which is the case for a `Project`. On top of that, it means that if we start using `Tracked` in function parameters, then we can't mix them up with contributed projects.

If we used a bool, we'd then have to start checking that `bool` everywhere. These kinds of things are examples of ["parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

> Also, given the newtype, would it make sense to replace the Projects struct with filtering functions like get_tracked_projects(...) -> Result<Vec<Tracked>> & get_contributed_projects(...) -> Result<Vec<Project>>?

We don't need filters, the fields are `pub`. If you want contributed projects then do `projects.contributed` and for tracked projects do `projects.tracked` :grin: No `Result` needed :tada:

proxy/api/src/http/project.rs
```diff
@@ -575,21 +578,48 @@ mod test {
         let owner = ctx.peer_api.init_owner(&ctx.signer, "cloudhead")?;
 
         coco::control::setup_fixtures(&ctx.peer_api, &ctx.signer, &owner)?;
-        let project = &project::list_projects(&ctx.peer_api)?[0];
+
+        let projects = project::Projects::list(&ctx.peer_api)?;
+        let project = projects.into_iter().next().unwrap();
```
FintanH, line 583: I don't think this is quite going to work how you expect it to. Because, since our peer created the project we have `signed_refs` in it.

To test this we need to create a second peer we clone the project from and then list our tracked projects, which should consist of the project we cloned.

proxy/api/src/project.rs
```diff
@@ -69,6 +74,86 @@ where
...
+            } else {
+                projects.contributed.push(project)
+            }
+        }
+
+        Ok(projects)
+    }
+}
+
+impl IntoIterator for Projects {
```
FintanH, line 142: We should probably add `impl Iterator` too while we're at it.

proxy/api/src/project.rs
```diff
@@ -113,17 +177,22 @@ pub fn list_projects(api: &coco::Api) -> Result<Vec<Project>, error::Error> {
 pub fn list_projects_for_user(
     api: &coco::Api,
     user: &coco::Urn,
-) -> Result<Vec<Project>, error::Error> {
-    let all_projects = list_projects(api)?;
+) -> Result<Vec<Tracked>, error::Error> {
```
FintanH, line 180: I wouldn't use `Tracked` here because we're not ensuring that these projects have `signed_refs`

proxy/api/src/http/project.rs
```diff
@@ -150,12 +162,21 @@ mod handler {
         let projects = if let Some(user) = query.user {
             project::list_projects_for_user(&ctx.peer_api, &user)?
```
FintanH, line 163: This should be a separate endpoint, as discussed

xla:
Merged latest master and fixed an issue where some log lines of the proxy wouldnt show. When call in the **contribtued** endpoint we see this error:

```
`Coco(Storage(Refsig(InvalidSignature(Refs { heads: {"refs/heads/dev": Oid(27acd68c7504755aa11023300890bb85bbd69d45), "refs/heads/master": Oid(223aaf87d6ea62eef0014857640fd7c8dd0f80b5)}, remotes: Remotes({}) }))))])
```

FintanH:
This is to do with setting up fake peers. Now that we have cloning and what not working properly, I'm going to fix this up.

FintanH:
So this test failure is due to our stop-gap solution finally leaking. After discussing with @xla we should opt to disable this particular test and create a follow-up ticket.

What we really need to do is come up with a proper test harness for peer-to-peer proxy's interacting with each other and testing the relevant components there. That is to say, if we can test the endpoints where they're defined then that's great, but as soon as we need to have peers represented in the monorepo then the test needs to live in this future integration suite.

Apologies for the frustration cause here @sarahscott :bow: 

Also, correct me if I'm wrong on any of the details above @xla

xla:
What @FintanH said. Let's capture an issue which outlines a proxy testnet (ideally based on exisiting radicle-link tooling) and enumerate what cases we want to test.

BTW we already have that scoped to just the coco crate. If we can assert the behaviour for now in the context of the `coco/tests/replication.rs` it would do us a massive service.

ui/src/project.ts
```diff
@@ -93,12 +95,21 @@ interface Fetch extends event.Event<Kind> {
   id: string;
 }
 
-interface FetchList extends event.Event<Kind> {
-  kind: Kind.FetchList;
+interface FetchContributed extends event.Event<Kind> {
+  kind: Kind.FetchContributed;
   urn?: string;
```
FintanH, line 100: This can be deleted now, I believe

FintanH (changes requested):
Almost there :ok_hand:

proxy/api/src/http/project.rs
```diff
@@ -49,13 +50,36 @@ fn get_filter(ctx: context::Ctx) -> impl Filter<Extract = impl Reply, Error = Re
...
+    ctx: context::Ctx,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    path("contributed")
+        .and(warp::get())
+        .and(http::with_context(ctx))
+        .and(path::end())
+        .and_then(handler::list_contributed)
+}
+
+/// `GET /peer/<id>
```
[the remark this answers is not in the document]
sarahscott: Would `projects/user/<id>` make more sense?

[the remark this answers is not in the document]
FintanH: `user` definitely makes more sense to me. I suppose `<id>` does too in that context since a user entity's id is a `RadUrn`

ui/src/project.ts
```diff
@@ -81,6 +81,7 @@ enum Kind {
   Create = "CREATE",
   Fetch = "FETCH",
   FetchList = "FETCH_LIST",
+  FetchUser = "FETCH_User",
```
FintanH, line 84:
```suggestion
  FetchUser = "FETCH_USER",
```

proxy/api/src/http/project.rs
```diff
@@ -137,22 +161,32 @@ mod handler {
         Ok(reply::json(&project::get(&state, &urn)?))
     }
 
-    /// List all known projects.
-    ///
-    /// If [`super::ListUser::user`] is given we only return projects that this user tracks.
-    pub async fn list(
+    /// List all projects tracked by the current.
```
FintanH, line 164:
```suggestion
    /// List all projects tracked by the current user.
```

FintanH (approved):

NunoAlexandre:
I'm having a look this morning :watch: :eyes:

proxy/api/src/http/project.rs
```diff
@@ -137,22 +161,32 @@ mod handler {
...
+    pub async fn list_contributed(ctx: context::Ctx) -> Result<impl Reply, Rejection> {
+        let ctx = ctx.read().await;
+        let state = ctx.state.lock().await;
+        let projects = project::Projects::list(&state)?;
+
+        Ok(reply::json(&projects.contributed))
+    }
+
+    /// List all known projects tracked by user of given `user_id`
+    pub async fn list_for_user(
```
NunoAlexandre, line 183: I find this confusing. The fn is called `list_for_user` which doesn't provide enough info, and reading the docs I see it lists projects tracked by the given `user_id`. So two suggestions

1. Could we rename it to `list_tracked_by_user` or just `list_tracked_by`
   Not only makes it easier to understand but it's also consistent with the above `fn list_tracked`.

2. Rename the underlying used function, `list_projects_for_user`, the same way

proxy/api/src/http/project.rs
```diff
@@ -563,59 +572,66 @@ mod test {
     }
 
     #[tokio::test]
-    async fn list() -> Result<(), error::Error> {
+    async fn list_for_user() -> Result<(), error::Error> {
```
NunoAlexandre, line 575:
```suggestion
    async fn list_tracked_by_user() -> Result<(), error::Error> {
```

proxy/api/src/project.rs
```diff
@@ -1,13 +1,13 @@
...
 
-use std::collections::HashSet;
+use std::{collections::HashSet, ops::Deref};
 
 use serde::{Deserialize, Serialize};
 
 use crate::error;
 
-/// Object the API returns for project metadata.
+/// Object encapsualating project metadata.
```
NunoAlexandre, line 10:
```suggestion
/// Object encapsulating project metadata.
```

proxy/api/src/project.rs
NunoAlexandre: I follow your point, @FintanH, and agree with the value it provides :+1:

proxy/api/src/project.rs
```diff
@@ -69,6 +73,139 @@ where
...
+    /// made any changes to. A project is still considered tracked if they checked out a working
+    /// copy but have not performed any commits to the references.
+    pub tracked: Vec<Tracked>,
+    /// A project that has been *contributed* to is one that the user has either:
+    ///     a. Created themselves using the application.
+    ///     b. Has replicated (see tracked above), checked out a working copy, and pushed changes
+    /// to     references.
+    ///
+    /// The conditions imply that a project is "contributed" if I am the maintainer or I have
+    /// contributed to the project.
```
NunoAlexandre, line 104:
```suggestion

    /// A project that has been *contributed* to is one that the user has either:
    ///     a. Created themselves using the application.
    ///     b. Has replicated (see tracked above), checked out a working copy, and pushed changes
    ///     to references.
    ///
    /// The conditions imply that a project is "contributed" if I am the maintainer or I have
    /// contributed to the project.
```

xla (approved):
🌉 🍸 🚸 🐕

proxy/api/src/project.rs
NunoAlexandre, line 104: neat docs :ok_hand:

proxy/api/src/project.rs
```diff
@@ -69,6 +73,139 @@ where
...
+    fn next(&mut self) -> Option<Self::Item> {
+        self.contributed
+            .next()
+            .or_else(|| match self.tracked.next() {
+                Some(tracked) => Some(tracked.0),
+                None => None,
+            })
+    }
+}
+
```
NunoAlexandre, line 208: hm, am I getting it right that when we iterate on a `Projects` that it will return an iterator on the `contributed` projects implicitly?

proxy/api/src/http/project.rs
```diff
@@ -49,13 +50,36 @@ fn get_filter(ctx: context::Ctx) -> impl Filter<Extract = impl Reply, Error = Re
...
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    path("contributed")
+        .and(warp::get())
+        .and(http::with_context(ctx))
+        .and(path::end())
+        .and_then(handler::list_contributed)
+}
+
+/// `GET /user/<id>
+fn peer_filter(ctx: context::Ctx) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
```
NunoAlexandre, line 76:
```suggestion
fn tracked_by_user_filter(ctx: context::Ctx) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
```

NunoAlexandre (changes requested):
@sarahscott kudus for this Rust adventure :)

I left some comments, mostly regarding how the REST endpoints are designed at the moment.

cypress/integration/project_source_browsing.spec.js
```diff
@@ -1,10 +1,9 @@
 before(() => {
   cy.nukeAllState();
   cy.onboardUser("cloudhead");
-  cy.createProjectWithFixture("platinum", "Best project ever.", "master", [
-    "ele",
-    "abbey",
-  ]);
+
+  // TODO(sos): add fake peers again when we have a peer testnet
```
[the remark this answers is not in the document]
FintanH: There is one https://github.com/radicle-dev/radicle-upstream/issues/908

proxy/api/src/http/project.rs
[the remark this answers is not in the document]
FintanH: So right now we're not differentiating `tracked` vs `contributed` for other users. The creation of this endpoint was due to splitting it out from the "current user" endpoints. 

I'll create a ticket to follow up with your thoughts, but I don't think this work should be done in this PR.

cypress/integration/project_source_browsing.spec.js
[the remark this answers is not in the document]
NunoAlexandre: cool!

proxy/api/src/http/project.rs
FintanH: Again, these aren't necessarily "tracked" projects as per our definition of tracked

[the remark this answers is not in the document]
NunoAlexandre: > So right now we're not differentiating tracked vs contributed for other users

Do you mean that we only differentiate it for the local user at the moment?

In any case, we are differentiating it for other users since this endpoint now filters specifically tracked users for a given user.

Note that my thought only involves namespace this endpoint with its semantics, just like we do for the local user, so I don't think it should be postponed.

proxy/api/src/project.rs
FintanH: It returns an iterator over _all_ the projects whether tracked or contributed

NunoAlexandre: Ah I follow now. So we first iterate over the `contributed` and when that's done we iterate over the `tracked`. Thanks!

proxy/api/src/http/project.rs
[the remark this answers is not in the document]
FintanH: > Do you mean that we only differentiate it for the local user at the moment?

Yes, that's the current change in this PR.

> In any case, we are differentiating it for other users since this endpoint now filters specifically tracked users for a given user.

Hmmm so you're talking about the _user_ being tracked and not the tracked user's _projects_?

If so, there's currently no such thing as non-tracked user in our local view. We have no way of knowing about users that are not in our monorepo :)

proxy/api/src/project.rs
FintanH: Exactly :white_check_mark:

proxy/api/src/http/project.rs
NunoAlexandre: hm that's not what the docs of `pub async fn list_for_user` nor `pub fn list_projects_for_user` suggest, neither the implementation of `pub fn list_projects_for_user`. 

is this a matter of language? :de: :netherlands: :england:

FintanH: Ya, I'm looking at them now and the docs are wrong :smile: 

I'm going to update them and the function name now :+1: Thanks for calling it out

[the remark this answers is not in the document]
NunoAlexandre: Ah, yes! That's where my confusion steemed. All clear, thanks for bearing with me :)

NunoAlexandre: Ah great :ok_hand:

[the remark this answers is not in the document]
NunoAlexandre: Ok, solved in https://github.com/radicle-dev/radicle-upstream/pull/866#discussion_r490071679 :100:

package.json
```diff
@@ -112,7 +112,7 @@
     "start:test": "run-p --race svelte:watch proxy:start:test electron:start",
     "test": "TZ='UTC' yarn test:unit && TZ='UTC' yarn test:integration",
     "test:integration": "TZ='UTC' run-p --race proxy:start:test wait:test",
-    "test:integration:debug": "RUST_LOG=proxy=debug TZ='UTC' run-p --race svelte:watch proxy:start:test wait:debug",
+    "test:integration:debug": "TZ='UTC' run-p --race svelte:watch proxy:start:test wait:debug",
```
rudolfs, line 115: Why do we want to remove this?

xla: It's incorrect syntax and hid relevant log lines from proxy, i.e. log level `error`.

NunoAlexandre (approved):
nice :surfer:

package.json
rudolfs: Is there a way to enable all log levels for `test:integration:debug`?

xla: We could do `RUST_LOG=debug` but that trickles down to all the dependencies and will be extremely noisy. One sane default could be to enable it only for our local crates with `RUST_LOG=api=debug,coco=debug`.

xla (approved):
🛣 💆 🆔 ⬜️

PR #867, opened 2020-09-02
chore(ui): convert ProjectList and children to ts

sarahscott:
✨ 

Pretty self-explanatory with some notes in-line.


ui/DesignSystem/Component/List.svelte
```diff
@@ -1,11 +1,12 @@
...
   import { createEventDispatcher } from "svelte";
 
   const dispatch = createEventDispatcher();
 
-  export let dataCy = null;
-  export let items = null;
-  export let style = null;
+  export let items: any[];
+
+  export let dataCy: string | undefined = undefined;
```
sarahscott, line 8: Not the prettiest syntax; an alternative could be 

```typescript
export let dataCy = "";
```

wherein the `string` type would be inferred, and we would check truthiness like `dataCy.length > 0`.

ui/DesignSystem/Component/ProjectCard.svelte
```diff
@@ -50,7 +52,7 @@
         </div>
       {/if}
       {#if badge}
-        <Badge style="margin-left: 0.5rem" variant={badge} />
+        <Badge style="margin-left: 0.5rem" variant={BadgeType.Maintainer} />
```
sarahscott, line 55: There's only one `BadgeType` right now; is there a reason to maintain the `variant` prop?

ui/DesignSystem/Component/ProjectList.svelte
```diff
@@ -1,19 +1,22 @@
-<script>
-  import { BadgeType } from "../../src/badge.ts";
+<script lang="ts">
+  import { BadgeType } from "../../src/badge";
+  import * as project from "../../src/project";
```
sarahscott, line 3: It's not straightfoward to use `interface`s as named imports with rollup right now - see https://github.com/rollup/plugins/issues/481

I personally don't mind sticking with this ESM syntax for now:
```typescript
import * as project from "../../src/project";
```

but I wanted to mention this in case someone gets an error like:
```
[!] Error: 'Project' is not exported by ui/src/project.ts, imported by ui/DesignSystem/Component/ProjectList.svelte
```

when importing this way:
```typescript
import { Project } from "../../src/project";
```

ui/DesignSystem/Component/List.svelte
FintanH: _Maybe_, we could define a type for `dataCy` in general:
```ts
interface NoData {
  kind: "no-data";
}

interface Data {
  kind: "data";
  value: string;
}

type DataCy = Data | NoData
```

In general, does anybody know of any `Maybe`/`Optional` type in typescript? Or even one step further, is there any good collection of functional-esque resources as ts library?

Having a `Maybe` type would help with all the other `undefined` cases as well. For example, we would have:
```ts
let badge: Maybe<BadgeType> = Nothing;
```

ui/DesignSystem/Component/ProjectList.svelte
```diff
@@ -1,19 +1,22 @@
...
+  export let projects: project.Project[];
+  export let urn: string;
 
-  const projectCardProps = project => ({
+  const projectCardProps = (project: project.Project) => ({
     title: project.metadata.name,
     description: project.metadata.description,
-    showRegisteredBadge: project.registration,
-    badge: project.metadata.maintainers.includes(urn) && BadgeType.Maintainer,
+    showRegisteredBadge: !!project.registration,
```
FintanH, line 16: What's this `!!`? :eyes:

ui/DesignSystem/Component/List.svelte
xla:
```suggestion
  export let dataCy = "";
```

There is [prelude-ts](https://github.com/emmanueltouzery/prelude-ts) which offers FP shenanigans and among them an `Option`. Maybe - no pun intended - we can start putting together a little prelude with some FP sprinkles.

It seems if a prop is optional it needs a default value, which would avoid presence tests alltogether.

ui/DesignSystem/Primitive/Flex.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
-  export let dataCy: string;
-  export let style: string;
   export let align: "left" | "right" | undefined = undefined;
+  export let dataCy: string | undefined = undefined;
```
xla, line 3:
```suggestion
  export let dataCy = "";
```

ui/DesignSystem/Primitive/Icon/IconWrapper.svelte
```diff
@@ -1,6 +1,6 @@
-<script>
-  export let dataCy = null;
-  export let style = null;
+<script lang="ts">
+  export let dataCy: string | undefined = undefined;
```
xla, line 2:
```suggestion
  export let dataCy = "";
```

ui/DesignSystem/Primitive/Icon/RegisteredSmall.svelte
```diff
@@ -1,8 +1,8 @@
-<script>
+<script lang="ts">
   import IconWrapper from "./IconWrapper.svelte";
 
-  export let dataCy = null;
-  export let style = null;
+  export let dataCy: string | undefined = undefined;
```
xla, line 4:
```suggestion
  export let dataCy = "";
```

xla (changes requested):
Dope!

ui/DesignSystem/Component/List.svelte
FintanH: These also seem like good options:
* https://gigobyte.github.io/purify/
* https://gcanti.github.io/fp-ts/

`fp-ts` looks pretty complete :ok_hand: Might become a front end engineer if we use it because it has higher kinded types and the whole shebang of classes

xla: We could give it a spin if we have actual use-cases. In thise specific instance I believe a default solves it.

FintanH: Sure, but there were other cases where I saw `| undefined` was being used. If we get some breathing room I'd love to see someone spearhead a `radicle-decision` on a sanctified lib for it :)

But agreed, the default is fine here. `dataCy` is just magic to me anyway. Basically a macro in my eyes

ui/DesignSystem/Component/ProjectList.svelte
sarahscott: Makes it a `boolean` 🔮

sarahscott: More specifically, the double bang evaluates the truthiness of `registration` because it is not a `boolean` value.

ui/DesignSystem/Component/ProjectCard.svelte
```diff
@@ -1,11 +1,15 @@
...
 
-  export let title = null;
-  export let description = null;
-  export let showRegisteredBadge = false;
-  export let badge = null;
+  export let title: string;
+
+  export let description = "";
+
+  /* eslint-disable @typescript-eslint/no-inferrable-types */
```
sarahscott, line 10: This is annoying, but `svelte-check` needs the explicit type annotation for `boolean` props with default values.

ui/DesignSystem/Component/List.svelte
```diff
@@ -1,11 +1,12 @@
-<script>
+<script lang="ts">
   import { createEventDispatcher } from "svelte";
 
   const dispatch = createEventDispatcher();
 
-  export let dataCy = null;
-  export let items = null;
-  export let style = null;
+  export let items: any[];
```
xla, line 6: Is this not of type `SvelteComponent | HTMLElement`?

ui/DesignSystem/Component/ProjectCard.svelte
xla: We should move your comment to an inline comment so other collaborators know why we have to leave it in.

xla (dismissed):
Dope!

🀄️ 🌜 💍 🌼

cloudhead (approved):

ui/DesignSystem/Component/List.svelte
sarahscott: In this context, `items` are passed as props to the `slot` as an array of `Project`s. I think `any` is fine here because it's a flexible component that we use for several different use cases.

xla (approved):
⏏ 🐎 🏹 🍾

cloudhead (approved):

PR #870, opened 2020-09-04
fix(ui): Rollup watch mode work again

geigerzaehler:
Before, updates to TypeScript source files would not be picked up by rollup in watch mode. (Even though rollup would recompile the bundle).

We fix this by disabling forcing the poll mode on chokidar. I don’t know why this works though.


xla (approved):
This workout is from over a year ago and always seemed like a stop-gap solution, so this is probably using the proper filesystem events underneath again now that polling is not enforced.

🛎 🎯 😅 👭

PR #875, opened 2020-09-07
chore(proxy): integrate ask urn providers

### LINKED ISSUE #600: Ask for a URN
The integration work for https://github.com/radicle-dev/radicle-link/issues/221

NunoAlexandre:
Closes #600


proxy/coco/src/peer.rs
```diff
@@ -725,6 +748,30 @@ mod test {
...
+            id: Hash::hash(b"project0"),
+            proto: uri::Protocol::Git,
+            path: "user/imperative-language"
+                .parse::<uri::Path>()
+                .map_err(Error::from)?,
+        };
+
+        let mut peers = api.providers(unkown_urn).await;
+        assert!(
+            peers.next().await.is_none(),
```
kim, line 768: I would be surprised if this ever returns. Perhaps the `upstream` wrapper should take a timeout or deadline parameter, and give up polling the stream once that elapses?

NunoAlexandre:
@xla @kim my sincere apologies for the noise, I meant to open this as a draft. If I don't request reviews myself please don't feel pushed to do so in any case.

proxy/coco/src/peer.rs
```diff
@@ -260,6 +262,25 @@ impl Api {
...
+    /// attempting a clone, even if it isn't a direct response to our query.
+    ///
+    /// Note that there is no guarantee that a peer who claims to provide the
+    /// [`RadUrn`] actually has it, nor that it is reachable using any of
+    /// the addresses contained in [`PeerInfo`]. The implementation may~F
+    /// change in the future to answer the query from a local cache first.
+    pub fn providers(
+        &self,
+        urn: RadUrn,
+    ) -> impl Future<Output = impl futures::Stream<Item = PeerInfo<IpAddr>>> {
```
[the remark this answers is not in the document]
NunoAlexandre: So I changed it to this. Is this what you mean?

``` rust
pub async fn providers(&self, urn: RadUrn) -> impl futures::Stream<Item = PeerInfo<IpAddr>> {
     let api = self.peer_api.lock().expect("unable to acquire lock");
     api.providers(urn).await
}
```

proxy/coco/src/peer.rs
NunoAlexandre: What triggered that suspicion for you? Cause I had just realized that this assert always succeeds. I thought it might have been related to your point above on `await` but that change didn't make a difference here.

proxy/coco/src/peer.rs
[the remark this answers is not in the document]
FintanH: That's right :+1: 

The `Future` return type would have eventually caused the compiler complain that `Api` is not `Send`/`Sync` across `await` points.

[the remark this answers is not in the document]
NunoAlexandre: @FintanH hm that's the exact issue I am having _now_ that I made this change, it didn't complain about that before :thinking:

proxy/coco/src/peer.rs
NunoAlexandre: I'm talking nonsense. You are right, it never returns, that's why I couldn't make it fail. Gotta fix https://github.com/radicle-dev/radicle-upstream/pull/875#discussion_r484291916.

kim: It's that the returned `Stream` will just be in pending state forever when there is nobody to answer. Of course it could also be the caller's responsibility to ensure it doesn't wait indefinitely, but it might also be the right place to _force_ them to think about this.

proxy/coco/src/peer.rs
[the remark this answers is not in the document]
xla: I ran into that issue when implementing `with_protocol`. Maybe it helps as inspiration:

https://github.com/radicle-dev/radicle-upstream/blob/bd9aed435db2efa9d5680501c3c4dbd496266250/proxy/coco/src/peer.rs#L376-L392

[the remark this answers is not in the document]
kim: Ah nevermind, the `PeerApi` is behind a _sync_ mutex, so that can't be held past the `.await`.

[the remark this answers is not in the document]
NunoAlexandre: @xla I am note sure I understand why all that machinery is necessary though? :thought_balloon: 

I tried to change it to:

``` rust
pub fn protocol(&self) -> Protocol<PeerStorage<keys::SecretKey>, Gossip> {
  let api = self.peer_api.lock().expect("unable to acquire lock");
  api.protocol().clone()
}
```

and adjusted the test using `with_protocol` and it all seems to work.

NunoAlexandre:
My knowledge of Rust and Rusty Future's, in particular, is lacking to face some of the challenges here.
What we are missing, to start with, are tests covering the `providers` feature on `link`, which is currently untested.

Once we have the guarantee from `link` that `providers` is working, then we can eliminate potential bugs there and nitpick all of these technicalities around sync mutexes, boxes, pins, streams, etc, imho.

FintanH:
I'd be happy to see it tested in radicle-link. If you'd like to attempt
that first that'd be awesome. As I've mentioned the testing harness is good
there and more future manageable.

For others: I've tried making the testnet code generic in TestPeer because
it holds a Peer instead of a Mutex over the PeerApi. It didn't really go
anywhere. It would be nice to be able to use it Upstream though.

proxy/coco/src/peer.rs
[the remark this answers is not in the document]
xla: Because I don't know how to `clone`. :> Removed it in a later iteration. Glad you found a cleaner way.

xla:
Currently working on splitting on integration tests in the `coco` trait, would very much appreciate using the testnet code there. Happy to help move further in that direction.

proxy/coco/src/peer.rs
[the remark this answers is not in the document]
NunoAlexandre: > Because I don't know how to clone. :>

I can relate so well to this :laughing:

proxy/coco/Cargo.toml
```diff
@@ -21,8 +21,7 @@ thiserror = "1.0"
 tokio = { version = "0.2", features = [ "dns", "macros", "time" ] }
 
 [dependencies.librad]
-git = "https://github.com/radicle-dev/radicle-link.git"
-rev = "dabbec104273eda48e943c1ff9fae8eedd8ba917"
+path = "/home/nuno/link/librad"
```
xla, line 24: Local first.

NunoAlexandre: in draft :hand:

NunoAlexandre:
@kim @xla @FintanH, regarding https://github.com/radicle-dev/radicle-upstream/pull/875#discussion_r484340666 and the whole discussion around timeout:

My opinion is that we should not take care of that in here. The reason being that we need to `timeout` on `next()`, not on the top-level future. That means that if we were to receive a timeout duration as a param, we'd have to timeout on `next()` ourselves, which would result in `providers` not providing a `stream` anymore but a future of `Option<Peer`> (pseudo-code), or to receive a closure that consumes it.

If we pass on that responsibility to the caller, we don't introduce unnecessary complexity here and we can easily then control what happens in the ["waiting area"](https://github.com/radicle-dev/radicle-upstream/issues/803): having the stream, wait for `next()` with a certain timeout (which we might want to adjust at each failed attempt) and once a provider is found, try and verify that said provider actually holds that URN and either call `next()` on failure or finish.

xla:
@NunoAlexandre As long as the upstream PR is not merged, this PR shouldn't go out of draft state.

For your comment about timeouts, I think it's safe to push it to the caller for now until we see the right pattern emerge and handle it inline. I also foresse some changes coming to where we have the providers live and handled once we integrate the proper urn request handling.

NunoAlexandre:
> @NunoAlexandre As long as the upstream PR is not merged, this PR shouldn't go out of draft state.

Did you mean the `link` PR? If so, do you say that from a perspective of good practices (not merge a PR using something that is untested in the used revision)? Just trying to understand :v: 

> For your comment about timeouts, I think it's safe to push it to the caller for now until we see the right pattern emerge and handle it inline. I also foresse some changes coming to where we have the providers live and handled once we integrate the proper urn request handling.

Yap :+1:

xla:
> Did you mean the `link` PR? If so, do you say that from a perspective of good practices (not merge a PR using something that is untested in the used revision)? Just trying to understand v

Good pratice and potentially merging changesets into master which point to a commit in a feature branch which could go away by ways of force pushing and therefore breaking master.

NunoAlexandre:
> > Did you mean the `link` PR? If so, do you say that from a perspective of good practices (not merge a PR using something that is untested in the used revision)? Just trying to understand v
> 
> Good pratice and potentially merging changesets into master which point to a commit in a feature branch which could go away by ways of force pushing and therefore breaking master.

Note that the revision of librad we point to is the same, which is on master, for the sake of clarity.

proxy/coco/src/peer.rs
```diff
@@ -261,6 +274,24 @@ impl Api {
         Ok(project_meta)
     }
 
+    /// Query the network for providers of the given [`RadUrn`].
+    ///
+    /// Notes:
```
xla, line 279:
```suggestion
    /// NB(nuno):
```

Let's treat these nota benes like TODOs, FIXMED, etc.

proxy/coco/src/peer.rs
```diff
@@ -261,6 +274,24 @@ impl Api {
...
+    ///     nor that it is reachable using any of the addresses contained in [`PeerInfo`].
+    ///
+    ///   * there is no guarantee that any peer ever replies to the query, meaning that awaiting for
+    ///     a reply might never terminate, requiring callers of this function to timeout on
+    ///     `next()`.
+    #[warn(clippy::future_not_send)]
+    pub async fn providers(
+        &self,
+        urn: RadUrn,
+    ) -> BoxFuture<'static, impl futures::Stream<Item = PeerInfo<IpAddr>>> {
```
xla, line 290:
```suggestion
    ) -> impl Futue<Output = futures::Stream<Item = PeerInfo<IpAddr>>> {
```

What's the reason that we can't return an `impl` here?

proxy/coco/src/peer.rs
```diff
@@ -518,10 +549,16 @@ impl entity::Resolver<user::User<entity::Draft>> for FakeUserResolver {
...
 #[allow(clippy::panic)]
 mod test {
+
+    use futures::stream::StreamExt;
+    use std::convert::TryInto;
     use std::env;
     use std::path::PathBuf;
     use std::process::Command;
+    use std::time::Duration;
+    use tokio::time::timeout;
```
xla, line 559:
```suggestion
    use std::convert::TryInto;
    use std::env;
    use std::path::PathBuf;
    use std::process::Command;
    use std::time::Duration;
    
    use futures::stream::StreamExt;
    use tokio::time::timeout;
```

proxy/coco/src/peer.rs
```diff
@@ -261,6 +274,24 @@ impl Api {
...
+    /// Query the network for providers of the given [`RadUrn`].
+    ///
+    /// Notes:
+    ///   * there is no guarantee that a peer who claims to provide the [`RadUrn`] actually has it,
+    ///     nor that it is reachable using any of the addresses contained in [`PeerInfo`].
+    ///
+    ///   * there is no guarantee that any peer ever replies to the query, meaning that awaiting for
+    ///     a reply might never terminate, requiring callers of this function to timeout on
+    ///     `next()`.
+    #[warn(clippy::future_not_send)]
```
xla, line 286: This is a red flag, if we see this warning that means we need to refactor until it is `Send`. If not when used/applied anywhere it will lead to compilation errors.

proxy/coco/src/peer.rs
```diff
@@ -99,6 +102,16 @@ impl Api {
         api.paths().git_dir().join("")
     }
 
+    /// Get the [Seed] representation for this [Api] instance.
```
xla, line 105:
```suggestion
    /// Get the [`Seed`] representation for this [`Api`] instance.
```

proxy/coco/src/peer.rs
```diff
@@ -99,6 +102,16 @@ impl Api {
...
+    #[must_use]
+    pub fn seed(&self) -> Seed {
```
xla, line 107: An impl for `From` would be the more elegant solution for this.

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
...
+            id: Hash::hash(b"project0"),
+            proto: uri::Protocol::Git,
+            path: "user/imperative-language"
+                .parse::<uri::Path>()
+                .map_err(Error::from)?,
+        };
+
+        let res = timeout(
+            Duration::from_secs(5),
+            api.providers(unkown_urn).await.await.next(),
```
xla, line 772: Why is there a double await here?

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
...
+            "Should have timed out without obtaining any provider",
+        );
+
+        Ok(())
+    }
+
+    /// Verify that asking the network for a URN owned by a seed peer returns said peer.
+    #[tokio::test]
+    #[allow(clippy::unwrap_used)]
+    async fn get_urn_providers_works() -> Result<(), Error> {
```
xla, line 787:
```suggestion
    async fn get_urn_providers() -> Result<(), Error> {
```

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
...
+        let target_urn = tokio::task::spawn_blocking(move || {
+            let project = radicle_project(repo_path.clone());
+            let user = alice.init_owner(&signer, "cloudhead").unwrap();
+            let created_project = alice.init_project(&signer, &user, &project).unwrap();
+            created_project.urn()
+        })
+        .await
+        .unwrap();
+
+        // Have peer 2 ask the network for providers for `target_urn`
```
xla, line 822:
```suggestion
```

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
...
+        let key = SecretKey::new();
+        let config = config::configure(
+            config::Paths::FromRoot(tmp_dir.path().to_path_buf()).try_into()?,
+            key,
+            *config::LOCALHOST_ANY,
+            vec![alice.seed()],
+        );
+        let bob = Api::new(config).await?;
+
+        // Create the targeted project in peer 1
```
xla, line 812:
```suggestion
```

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
...
+        let repo_path = tmp_dir.path().join("radicle");
+        let key = SecretKey::new();
+        let signer = signer::BoxedSigner::from(signer::SomeSigner {
+            signer: key.clone(),
+        });
+        let config = config::default(key.clone(), tmp_dir.path())?;
+        let alice = Api::new(config).await?;
+        let alice_peer_id = alice.peer_id();
+
+        // Peer #2
```
xla, line 800:
```suggestion
```

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
...
+        );
+
+        Ok(())
+    }
+
+    /// Verify that asking the network for a URN owned by a seed peer returns said peer.
+    #[tokio::test]
+    #[allow(clippy::unwrap_used)]
+    async fn get_urn_providers_works() -> Result<(), Error> {
+        // Peer #1
```
xla, line 788:
```suggestion
```

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
...
+
+        assert!(
+            res.is_err(),
+            "Should have timed out without obtaining any provider",
+        );
+
+        Ok(())
+    }
+
+    /// Verify that asking the network for a URN owned by a seed peer returns said peer.
```
xla, line 784:
```suggestion
```

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
         Ok(())
     }
 
+    /// Verify that asking the network for an unkown urn returns no providers.
```
xla, line 755:
```suggestion
```

xla (changes requested):
Really good work, thanks for putting it together. Inlined some things.

proxy/coco/src/peer.rs
FintanH: The `From` was a bit of a pain because `PeerApi` is parameterised over an `S` which needs to be a `Signer` et al. But I think I still agree, `From` would be good :+1: Just follow the trait bounds in `librad` on `PeerApi` to get the incantations right.

proxy/coco/src/peer.rs
FintanH: I'd disagree that these are TODOs. The below are semantics that a user of this function should be aware of. So they're acting as documentation that the user of this function should be aware of.

kim: Is that the same thing? He's documenting what callers should be aware of.

proxy/coco/src/peer.rs
```diff
@@ -715,6 +752,96 @@ mod test {
...
+    #[tokio::test]
+    async fn get_urn_providers_none() -> Result<(), Error> {
```
FintanH, line 757:
```suggestion
    async fn get_urn_providers_is_none() -> Result<(), Error> {
```

proxy/coco/src/peer.rs
NunoAlexandre: Could you elaborate on why? I find such small doc comments to be super useful if you haven't looked into a piece in a while, while not having it only saves some bytes. 

Same for the suggested deletions above. The test gets pretty big, adding those small annotations is pretty helpful in my experience.

proxy/coco/src/peer.rs
FintanH: Or something like `can_get_provider_for_urn`

proxy/coco/src/peer.rs
kim: Don't get it either -- the desugared return type is 

```rust
impl Future<Output = Pin<Box<dyn Future<Output = Stream<Item = PeerInfo<IpAddr>> + 'a + Send>>>
``` 
Double future!

proxy/coco/src/peer.rs
xla: > I'd disagree that these are TODOs. The below are semantics that a user of this function should be aware of. So they're acting as documentation that the user of this function should be aware of.

I did not say that these are TODOs. My point is that we should keep the same notation style, where `<TOKEN>(<author>): <Comment>`.

proxy/coco/src/peer.rs
xla: :rainbow: :rainbow:

proxy/coco/src/peer.rs
xla: As mentioned in other places, if the code doesn't communicate what these small comments allude to, it needs to change. Here the test function should give indication enough that a comment is not necessary.

proxy/coco/src/peer.rs
FintanH: Ah, I see. I got confused by your wording. I find the author being in documentation strange.

If you wanted to track them down just `git blame` :sweat_smile:

NunoAlexandre: exactly, these are documentation notes, not notes for me.

xla: Having the author in there helps contextualising and potentially knowing who to ask when advise is needed. Git blame can fall apart esp. when moving code and on large refactors.

proxy/coco/src/peer.rs
NunoAlexandre: I am confused :upside_down_face: 

From this comment (https://github.com/radicle-dev/radicle-upstream/pull/875#discussion_r484346892), as well as https://github.com/radicle-dev/radicle-upstream/pull/875#discussion_r484342112, and given the warnings we were getting, I understood that the way to address the issue was to wrap the stream in a `BoxFuture`.

> What's the reason that we can't return an impl here?

Because [`Stream` is not `Send`](https://buildkite.com/monadic/radicle-upstream/builds/4515#e234d536-6616-4279-b94d-efc23e8094ea/626-634), if I don't misunderstand the issue.

proxy/coco/src/peer.rs
NunoAlexandre: I used to be an extreme believer in that principle until my pursue for that principle got in the way of other developers and their experience.

I'd welcome suggestions to make the code so that readers don't benefit from these small annotations while making sense from a trade-off pov.

proxy/coco/src/peer.rs
NunoAlexandre: This is my bad, I had set it to see if there would be any further warnings just as an experiment and I forgot to drop it. Done now. Sorry!

proxy/coco/src/peer.rs
NunoAlexandre: Sorry, I can't see the point. This is documenting the caveats of the function, there is no need to know whom to contact. I sense that if I had inlined these notes that this wouldn't be a discussion anymore?

proxy/coco/src/peer.rs
NunoAlexandre: `From` would be more elegant, but as pointed out it's not trivial and offers no further benefit compared to what's already in. Could leave this for last but don't see the point all things considered.

proxy/coco/src/peer.rs
NunoAlexandre: Having https://github.com/radicle-dev/radicle-upstream/pull/875#discussion_r485619418 in mind, the double await comes from first unboxing the future and then operate on it. Happy to reconsider that approach :hand:

proxy/coco/src/peer.rs
kim: The thing is that `async fn` desugars to something which is a `Future` and captures `self` (I don't know exactly how, but it's good enough to think about it like this). Minimally, `self` has to be `Send`, so it can be moved to the executor threadpool. Most of the time, however, it also has to be `Sync`, which `PeerApi` isn't.

The trick employed in `PeerApi` is that the returned `Future` does **not** capture `self`, but a clone of `Protocol`, which _is_ `Sync`. Thus, if you just return the `Future` returned from `PeerApi::providers`, all should be good.

It would also be possible to `.await` that future and return just the `Stream`, but for this to work, `self.peer_api` would need to be mutexed by an _async aware_ mutex (e.g. `futures::lock::Mutex`).

NunoAlexandre: @kim, thanks a ton for such an insightful explanation. I got it, pushed :)

proxy/coco/src/peer.rs
NunoAlexandre: > Here the test function should give indication enough that a comment is not necessary.

In all the best spirit, I see a substantial inconsistency between this point :point_up: and requiring _every_ `struct` or `function` to be documented, where said documentation is absolutely redundant and adds no useful information nor value. Just a few examples are [this](https://github.com/radicle-dev/radicle-upstream/blob/3377508a3d01f8638a921d23244256b6c17580c9/proxy/api/src/avatar.rs#L90) and [this](https://github.com/radicle-dev/radicle-upstream/blob/3377508a3d01f8638a921d23244256b6c17580c9/proxy/api/src/avatar.rs#L203) (nothing against `avatar.rs` :laughing:)

xla: When I read the code I didn't benefit from the comments and therefore suggested the changes. The code as is was clear enough to follow. It seems more like a liability in this instance. Then again I'm familiar with the code-base and most of it's darker corners, so take it with a grain of salt. Wouldn't be fuzzed if you don't accept the suggestions, :dog:

proxy/coco/src/peer.rs
xla: Would something like this not be sufficient:

``` rust

impl From<&Api> for Seed {
    fn from(api: &Api) -> Self {
        Self {
            addr: api.listen_addr(),
            peer_id: api.peer_id(),
        }
    }
}
```

proxy/coco/src/peer.rs
xla: Not planning to die on this hill. :japanese_ogre:

kim: I just made the decision for you, so feel free to replace the commentary with "RTFM": radicle-dev/radicle-link#314

proxy/coco/src/peer.rs
NunoAlexandre: > Wouldn't be fuzzed if you don't accept the suggestions, :dog: 

Ok :) in that case, I appreciate the sensitivity in accepting other experiences here.

proxy/coco/src/peer.rs
NunoAlexandre: You are right, @xla! When Fintan and I were pairing on this we were operating on the underlying `PeerApi`, which is parameterised, but we didn't think of implementing it for `Api` instead. Done!

xla (dismissed):
💌 🕙 📇 🗣

proxy/coco/src/peer.rs
FintanH: Well, I did. But I just didn't want to interact with the `lock` in a `From` impl :sweat_smile:

proxy/coco/src/peer.rs
```diff
@@ -715,6 +744,89 @@ mod test {
...
+        .unwrap();
+
+        // Have peer 2 ask the network for providers for `target_urn`
+        let res = bob
+            .providers(target_urn, Duration::from_secs(5))
+            .await
+            .next()
+            .await;
+
+        if let Some(peer_info) = res {
```
FintanH, line 818:
```suggestion
        assert_eq!(res.map(|info| info.peer_id), Some(alice_peer_id))
```

NunoAlexandre: better :ok_hand:

xla (approved):
📩 🅾️ 🔩 🔷

FintanH (approved):
Thanks, Nuno :clap:

NunoAlexandre:
> Thanks, Nuno

No, and I don't mean it in a cute way, thank you and @xla! :yellow_heart:

Issue #852, opened 2020-09-01
sync on startup


FintanH:
When our peer is offline then this implies we miss announcements that might have happened during this period. To ensure that our Monorepo is up to date we want to do a round of fetches on the `RadUrn`s + `PeerId`s we have in our Monorepo. We should only do this upon startup of the peer.

xla:
As we looked into it a bit more in depth in a call, we see the possibility that the syncing is kicked off by gossiping a set of `Want`s for the projects the peer is tracking. Leaving this here for completeness sake.

FintanH:
Wait, where/when did we say it was sending off `Want`s?

xla:
In the replication call yesterday. The flow we hinted at was for the peer on startup to send wants for all its tracked projects. At least that's the part I remember, as we concluded it's so closely related to reacting to announements of `Have`s that we should tackle these together.

FintanH:
I think that was said in passing. I need to figure out if that's possible/what we want (no pun intended).

xla:
# Coco::Sync

To keep the local state as up-to-date as possible we want to introduce
routine(s) for syncs. Most notably when starting up, but potentially after long
periods of being offline. Syncing is synonymous with performing fetch
operations from connected peers for repos the local peer is tracking. For that
a new coco module is proposed offering functionality to:

* build the list of repos we want to sync
* run the fetch against a conencted peer
* machinery to react to peer connections

## Flow

We have two potential ways to address syncing, which are described below.

### Shotgun fetch

Here we fetch indiscriminately from a number of freshly connected peers.

* build up list of repos we want to sync
* listen to `N` amount of `ProtocolEvent::Connected` events and for each
  * iterate over the list of to-be-synced repos and for each
  * do a fetch from the newly connected peer
  * ignore error indicating the peer doesn't have the repo

#### Pros

* latency as we sync as soon as peers are available
* due to its naïve approach relatively simple to implement

#### Cons

* wasteful as we do a lot of blind operations
* might miss the peers which actually have updates for us

### One-time gossip

Here we announce a set of `Want`s for the list of projects we track and wait
for peers in the network to reply with `Have`s.

* build up list of repos we want to sync
* wait for `N` amount of `ProtocolEvent::Connected` events and for each
  * announce a `Want`
  * wait for the correspondening `Have`s to return
  * fetch the repo from the peer which the `Have` came from

#### Pros

* we use the gossip layer
[...]

PR #878, opened 2020-09-07
refactor(proxy): prep sync work

xla:
Preparations for #852 

* bump link deps
* push librad into coco crate


proxy/coco/src/keystore.rs
```diff
@@ -10,46 +10,45 @@ use radicle_keystore::{
     file, FileStorage, Keystore, SecretKeyExt,
 };
 
-/// File path to librad key
-const LIBRAD_KEY: &str = "librad.key";
+/// Last component of the file path to the key.
```
FintanH, line 13:
```suggestion
/// File name component of the file path to the key.
```

proxy/coco/src/peer.rs
```diff
@@ -80,7 +80,11 @@ impl Api {
             run_loop.await;
         });
 
-        // Register the rad:// transport protocol
+        // Turns out we don't need to call the git commands from the command line.
+        // All we need is to register the git transport and this is the same
+        // thing. This means we can use git2 to execute actions. So instead of
+        // implementing the checkout via git clone as a Command, we use
+        // [`git2::RepoBuilder`].
```
FintanH, line 87: This comment seems out of place. I get that it's trying to explain the history of a misconception we had. I think it would make more sense to explain what it buys us, for example:

This registers the `rad://` transport protocol so that we can perform git actions between the Monorepo and a working copy. This means that we can easily use the `git2` library to perform these actions and be sure that the `signer` is used to sign any git artifacts.

FintanH (commented):
Looks good, minor suggestions

FintanH (approved):

PR #879, opened 2020-09-08
refactor(proxy): Simplify with_protocol

NunoAlexandre:
The function `with_protocol` in `coco` can just be a simple delegated
getter to the underlying coco peer api. No need to receive a function as an
argument and operate on BoxFuture.

Triggering context: https://github.com/radicle-dev/radicle-upstream/pull/875#discussion_r484342112


xla (approved):
Dope!

🤓 ⌛️ 📳 📓

MeBrei (approved):

PR #880, opened 2020-09-08
refactor(ui): new project modal validations

### LINKED ISSUE #869: Refactor project creation validations
- [ ] extract the validations using `validation.ts`

- [ ] add new functionality to cover edge cases mentioned [here](https://github.com/radicle-dev/radicle-upstream/pull/858#issuecomment-687021693)

MeBrei:
Closes #869 

ToDo

- [x] focus on name input after new repository selection
- [x] reset validateName after new selection
- [x] remove current Selection validation message
- [x] do not show `Product name is required` validation error

### DIFF
--- a/ui/DesignSystem/Primitive/Input/Directory.svelte
+++ b/ui/DesignSystem/Primitive/Input/Directory.svelte
@@ -1,4 +1,5 @@
 <script>
+  import { createEventDispatcher } from "svelte";
   import { getDirectoryPath } from "../../../../native/ipc.js";
 
   import Button from "../Button.svelte";
@@ -10,8 +11,11 @@
   export let validation = null;
   export let buttonVariant = "primary";
 
+  const dispatch = createEventDispatcher();
+
   const openFileDialog = async () => {
     path = await getDirectoryPath();
+    if (path) dispatch("chosen");
   };
 </script>
 
--- a/ui/DesignSystem/Primitive/Input/Directory.svelte
+++ b/ui/DesignSystem/Primitive/Input/Directory.svelte
@@ -15,7 +15,7 @@
 
   const openFileDialog = async () => {
     path = await getDirectoryPath();
-    if (path) dispatch("chosen");
+    if (path) dispatch("selected");
   };
 </script>
 
--- a/ui/Modal/ProjectCreation.svelte
+++ b/ui/Modal/ProjectCreation.svelte
@@ -46,7 +46,6 @@
   let newRepositoryPath = "";
   let existingRepositoryPath = "";
 
-  let validatingName = false;
   let nameValidation = nameValidationStore();
 
   let loading = false;
@@ -54,7 +53,6 @@
   const setSelection = type => {
     currentSelection = type;
     // Reset the name validation on selection switch
-    validatingName = false;
     nameValidation = nameValidationStore();
   };
 
@@ -103,10 +101,7 @@
   $: pathValidation = repositoryPathValidationStore(isNew);
 
   $: {
-    if (name.length > 0) {
-      validatingName = true;
-    }
-    if (validatingName) nameValidation.validate(name);
+    if (name.length > 0) nameValidation.validate(name);
   }
 
   $: repositoryPath = isNew ? newRepositoryPath : existingRepositoryPath;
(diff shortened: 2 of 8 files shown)


sarahscott:
Looking good so far! I think it would be easier to reason about this functionality with some unit tests, specifically around the validation logic and interdependent stores. It'd be nice to have a test for how these three change with respect to one another:
```typescript
export const localState = writable<LocalState | string>("");
export const localStateError = writable<string>("");
export const defaultBranch = writable<string>(DEFAULT_BRANCH_FOR_NEW_PROJECTS);
```

e.g.
```typescript
describe("branch fetching", () => {
  it("reverts to defaults whenever the path changes", () => {
      // ...
  });
  
  it("sets default branch if found in local state", () => {
    // ...
  });
});
```

ui/Modal/ProjectCreation.svelte
```diff
@@ -28,131 +33,30 @@
...
-    validations = validatejs(
-      {
-        name: name,
-        currentSelection: currentSelection,
-        newRepositoryPath: newRepositoryPath,
-        existingRepositoryPath: existingRepositoryPath,
-      },
-      constraints
-    );
+  const setSelection = type => {
```
sarahscott, line 53: `setSelection` is a little vague; something like `setRepoType` would be more clear.

ui/Modal/ProjectCreation.svelte
```diff
@@ -28,131 +33,30 @@
   } from "../DesignSystem/Component";
 
   let currentSelection;
+  let nameInput;
   export let content;
 
-  const projectNameMatch = "^[a-z0-9][a-z0-9._-]+$";
   const dispatch = createEventDispatcher();
 
   $: isNew = currentSelection === RepoType.New;
```
sarahscott, line 41: I don't love having these two reactive variables - isn't a project always either `New` or `Existing`? I think this would make more sense with just one `isNew` variable:

```
 isNew ? { do something because it's new } : { do something else because it exists already }
```

It's confusing to have certain ui elements depend on `isExisting` and others on `isNew`.

MeBrei: `currentSelection` is `null` at the beginning of the modal so that neither `isNew` and `isExisting` are `true`.
For some elements with the current set up, we need `isExisting`, like the tooltip and disabled-ness of the project name input field should only happen if the `currentSelection` is set and it is set to  `RepoType.Existing`. I guess it could be replaced by `currentSelection && currentSelection === RepoType.Existing`, but I find `isExisting` easier to read.

ui/DesignSystem/Primitive/Input/Directory.svelte
```diff
@@ -10,8 +11,11 @@
   export let validation = null;
   export let buttonVariant = "primary";
 
+  const dispatch = createEventDispatcher();
+
   const openFileDialog = async () => {
     path = await getDirectoryPath();
+    if (path) dispatch("chosen");
```
xla, line 18: Sounds very epic! :D Maybe `selected` is better.

ui/src/project.ts
```diff
@@ -198,3 +203,92 @@ export const fetchList = event.create<Kind, Msg>(Kind.FetchList, update);
 
 // Fetch initial list when the store has been subcribed to for the first time.
 projectsStore.start(fetchList);
+
+// NEW PROJECT
+
+export const localState = writable<LocalState | string>("");
```
xla, line 209: We usually don't export the writables. Rather we attempt to derive a readable and have mutations done through functions/events.

xla (changes requested):

ui/Modal/ProjectCreation.svelte
MeBrei: Renamed it to `setCurrentSelection`

ui/src/project.ts
MeBrei: We set up the `readable` for remote stores. These stores are more simple and we export simple `writables` throughout the app. Could you explain why we would want to avoid exporting them?

xla: Mostly for semantics and trace of write patterns as in who are the writers to this state. We don't need to go all out on this for this at the moment.

xla (approved):
🕓 🤓 🎬 ©️

PR #881, opened 2020-09-08
refactor(proxy): introduce coco::peer

### LINKED ISSUE #887: CoCo::peer
# CoCo::peer

With the advent of a range of ongoing network tasks (#600, #602, #786, #803, #852)
the coco crate is in need of an entity to manage these, while advancing the
network protocol - to observe and react to signifcant events from peers.
Addtionally striving for the ability to have our the algorithms for the
mentioned isues easy to understand and iterate on, given that we for most of
them are in exploration phase.

So far our organisation of network related code was haphazard at best, with
setting up some subscribers in the `coco::Api` and starting a single
background task in our main. Neither does that give clear contracts to test, nor
does it help collaborators contribute safely when tasked with the introduction
of a new subroutines.

To remedy these points, the design proposed here introduces a dedicated
`coco::Peer` to manage all network responsibilities to uphold:

* separate state access from network machinery
* manage network subroutines concurrently
* surface "peer protocol"
* expose significant peer events so UX features can be satisfied
* give contributors simple tools to add/iterate on subroutines
* enable integration test suite

## Interface

To illustrate the peer API we use a trait, might not be needed in the final
implementation.

``` rust
trait Peer {
    /// Consume the `RunLoop` we obtain when calling accept on a librad peer. We hold onto it
    /// until `run` is explicitely scheduled.
    fn new(run_loop: RunLoop, state: State, store: kv::Store) -> Self;

    /// Long-running task to be scheduled that consumes the `Peer` and keeps the machinery running.
    /// Responsible to keep the `RunLoop` alive and of the peers background tasks.
    async fn run(self) -> Result<(), Error>;

    /// Subscribe to peer events emitted while running the "peer protocol".
    ///
    /// NB(xla): Must be called before `run` is started, as that consumes the peer.
    fn subscribe(&self) -> tokio::sync::broadcast::Receiver<Event>;
}
```

Zooming into the run loop of the peer this is how the central, concurrent way
of advancing its protocol could look like:

``` rust
/// Start up the internal machinery to advance the underlying protocol, react to significant
/// events and keep auxiliary tasks running.
///
/// # Errors
///
/// * if one of the handlers of the select loop fails
pub async fn run(self) -> Result<(), Error> {
    // Start announcement timer.
    let mut announce_timer = todo!();

    // Subscribe to API events.
    let api_subscriber = todo!();

    // Subscribe to protocol events.
    let protocol_subscriber = todo!();

    // Subscrbe to project urn requests.
    let request_projects_subscriber = todo!();

    // Advance the librad protocol.
    tokio::spawn(self.run_loop);

    loop {
        let res = tokio::select! {
            _ = announce_timer.tick() => {
                let updates = todo!();
                Ok(Event::Announced(updates))
            },
            Some(event) = api_subscriber.next() => {
                // React to api events.
                Ok(Event::API(event))
            },
            Some(event) = protocol_subscriber.next() => {
                // React to protocol events like fetching.
                // Issues feteches for the first N connected peers on startup.
                Ok(Event::Protocol(event))
            },
            Some(request) = request_projects_subscriber.next() => {
                // Issue a Want.
                // Update request.
                Ok(Event::Requested(request))
            },
            else => break,
        };

        match res {
            // Propagate if one of the select handlers failed.
            Err(err) => return Err(err),
            Ok(event) => {
                // Send will fail, if there are no active receivers.
                if self.subscriber.receiver_count() > 0 {
                    if let Err(err) = self.subscriber.send(event).map_err(Error::Broadcast) {
                        return Err(err);
                    }
                }
            }
        }
    }

    Ok(())
}
```

Events emitted by the peer could be shaped like this:

``` rust
/// Significant events that occur during a [`Peer`]s lifetime.
#[allow(clippy::large_enum_variant)]
#[derive(Clone)]
pub enum Event {
    /// Gossiped a list of updates of new heads in our [`crate::state::State`]`.
    Announced(announcement::Updates),
    /// Received a low-level api event.
    API(peer::PeerEvent),
    /// Received a low-level protocol event.
    Protocol(protocol::ProtocolEvent<Gossip>),
    /// Requested a project from the network
    Requested(Request),
}
```

_The current design does not explicitely propose a "peer protocol", this is
left for future iterations._

xla:
High-level design is in #887 

Closes #887
Part of #852

***
* separate peer and state
* make announcements a peer responsibility
* setup peer event broadcasting
* introduce coco integration tests


proxy/coco/src/lib.rs
```diff
@@ -55,3 +57,34 @@ pub use source::{
...
+{
+    let peer = config.try_into_peer().await?;
+    let (api, run_loop) = peer.accept()?;
+
+    let state = State::new(api, signer);
+    let state = state::Lock::from(state);
+    let peer = Peer::new(run_loop, state.clone(), store);
+
+    Ok((peer, state))
+}
```
cloudhead, line 90: This API is a bit weird to me, maybe it could work more like this:
```
let peer = Peer::from_config(config, signer, store)?; // this is basically implemented as the above function
let state = peer.state.clone();
```
?

cloudhead: Or even better, could this be `Peer::new`?

proxy/coco/src/peer.rs
```diff
@@ -1,1111 +1,153 @@
...
-            } else {
-                let _repo = storage.create_repo(&user)?;
+            match res {
+                // Propagate if one of the select failed.
+                Err(err) => return Err(err),
+                Ok(event) => {
+                    log::info!("{:?}", event);
+
+                    // Send will fail, if there are no active receivers.
+                    if self.subscriber.receiver_count() > 0 {
```
cloudhead, line 127: Do we need this check?

proxy/coco/src/peer.rs
```diff
@@ -1,1111 +1,153 @@
...
+        // Start announcement timer.
+        let mut announce_timer = tokio::time::interval(std::time::Duration::from_secs(1));
+
+        // Advance the librad protocol.
+        tokio::spawn(self.run_loop);
+
+        loop {
+            let res = tokio::select! {
+                _ = announce_timer.tick() => {
+                    let updates = Self::announce(self.state.clone(), &self.store).await?;
```
cloudhead, line 111: You sure an error here should exit the run loop?

kim (commented):
Okay, so, I don’t know how to say this in a nicer way, but: this change does not make any sense to me.

a. `PeerApi` was introduced after a conversation where we concluded that it would be good for librad to provide a single entry-point and high-level API, driven by the patterns emerging from application development. Instead of contributing to that, you’re creating a “high-level wrapper”. Cool, feedback taken, `PeerApi` will be removed at the next occasion.

b. What made you conclude that it will improve your concurrency story (and performance) by guarding resource acquisition behind a mutex **as well as** acquiring a new libgit handle virtually every time you need to perform a git operation? You’re paying twice the price for half the benefit.

What you want is to move out the `Protocol` — which doesn’t require external synchronisation — and maintain thread-local copies of the storage. Afaics, the `thread_local` crate would be a better choice than the standard library. Alternatively, you could just roll your own by indexing into an array by thread id (mod the length of the array).

xla:
> Okay, so, I don’t know how to say this in a nicer way, but: this change does not make any sense to me.

No need to sugar code, happy to discuss your concerns.

> a. `PeerApi` was introduced after a conversation where we concluded that it would be good for librad to provide a single entry-point and high-level API, driven by the patterns emerging from application development. Instead of contributing to that, you’re creating a “high-level wrapper”. Cool, feedback taken, `PeerApi` will be removed at the next occasion.

The "higher-level" wrapper is to unify the application concerns, if any of these:

* announcing diffs of heads
* syncing on startup
* managing requests for project urns

can be pushed down to the `PeerApi` then I'm happy to move that work to radicle-link. While it may be true that `PeerApi` is high-level, it is from the protocol perspective. And I think two things are true: a) the protocol hasn't stabilized yet, which is fair and expected b) we are in an ongoing negotiation of where to draw the lines between protocol and application. Neither is this change an attempt to reverse our design decisions, nor an attempt to avoid contributing to the protocol. Intention here was informed by my understanding what is application specific, my understanding is likely limited and am eager to get a better intuition for where to put things.

Which parts would you conclude are in the realm of the protocol? We can break it apart and contribute those back to `PeerApi`.

> b. What made you conclude that it will improve your concurrency story (and performance) by guarding resource acquisition behind a mutex **as well as** acquiring a new libgit handle virtually every time you need to perform a git operation? You’re paying twice the price for half the benefit.

It was never a goal to improve performance. The aim was to make the handful of concurrent tasks/streams manageable and for additional ones to have a place, so we can iterate/experiment on how these work (sync, requests). For the resource acquisition, we always needed some form of lock around the `PeerApi` as we had to hand it into our HTTP async soup which wants it all `Send` and `Sync`. To the libgit handle I can't speak to much as mostly @FintanH drove this part of the code, maybe he can provide more context on that.

> What you want is to move out the `Protocol` — which doesn’t require external synchronisation — and maintain thread-local copies of the storage. Afaics, the `thread_local` crate would be a better choice than the standard library. Alternatively, you could just roll your own by indexing into an array by thread id (mod the length of the array).

I'm starting to share this understanding, esp. the protocol part and plan to explore a more sane resource acquisition strategy. This has been mostly limited by me acquiring deep enough understanding of the protocol internals and having enough time.

kim:
> Which parts would you conclude are in the realm of the protocol? 

Ya I mean, most methods in `State` are rather shallow wrappers, and that's what I imagined would go into `PeerApi`. But it's fine, we can just return a "dumb" struct from `Peer::accept`, which doesn't contain any `Storage` (and is therefore `Sync`), and from which you can move out fields as you wish.

> To the libgit handle I can't speak to much

Alright, let's solve this once and for all, making anyone even _think_ about this seems, uhm, unergonomic.

I think thread-local storage was basically the right idea, but the total number of `Storage` values needs to be bounded. So, what we need is a good ol' pool. There are a couple available on crates.io, [deadpool](https://crates.io/crates/deadpool) seems quite nice (you can also tell it to not require a manager mwwaha).

I don't mind whether this is set up here, or the return value of `Peer::accept` contains some sort of `StorageFactory`, which is by default initialised with a capacity == the number of CPUs.

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+        {
+            let storage = self.api.storage().reopen()?;
+
+            if storage.has_urn(&urn)? {
+                return Err(Error::EntityExists(urn));
+            } else {
+                let _repo = storage.create_repo(&user)?;
+            }
+        }
+
```
NunoAlexandre, line 404:
```suggestion
        // Initialise the user in the storage
        if self.api.storage().reopen()?.has_urn(&urn)? {
            return Err(Error::EntityExists(urn));
        } else {
            let _repo = storage.create_repo(&user)?;
        }

```

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+        signer: &signer::BoxedSigner,
+        handle: &str,
+    ) -> Result<user::User<entity::Draft>, Error> {
+        // Create the project meta
+        let mut user =
+            user::User::<entity::Draft>::create(handle.to_string(), signer.public_key().into())?;
+        user.sign_owned(signer)?;
+        let urn = user.urn();
+
+        // Initialising user in the storage.
```
NunoAlexandre, line 394: We had a discussion in another PR precisely about such comments. For me, _after_ reading the code below it's clear what it does, _however_ I'd still benefit from having this comment.

This to ask, where do you stand on this in the end? :hamster:

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+    }
+
+    /// Query the network for providers of the given [`RadUrn`] within a given `timeout`.
+    pub fn providers(
+        &self,
+        urn: RadUrn,
+        timeout: Duration,
+    ) -> impl Future<Output = impl futures::Stream<Item = PeerInfo<IpAddr>>> {
+        self.api.providers(urn, timeout)
+    }
```
NunoAlexandre, line 415: Beautiful :star_struck:

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+
+    /// Wrapper around the storage track.
```
NunoAlexandre, line 417: We know this is a wrapper but it would be more useful to document what it means.

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// Get the [`user::User`]s that are tracking this project, including their [`PeerId`].
+    ///
+    /// # Errors
+    ///
+    /// * If we could not acquire the lock
+    /// * If we could not open the storage
+    /// * If did not have the `urn` in storage
+    /// * If we could not fetch the tracked peers
+    /// * If we could not get the `rad/self` of the peer
+    pub fn tracked(&self, urn: &RadUrn) -> Result<Vec<(PeerId, user::User<entity::Draft>)>, Error> {
```
NunoAlexandre, line 435: `nit`: "tracked" suggests a boolean check, thus I suggest renaming it to:

```suggestion
    pub fn trackers(&self, urn: &RadUrn) -> Result<Vec<(PeerId, user::User<entity::Draft>)>, Error> {
```

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+            state.init_project(&signer, &kalt, &fakie_project(repo_path).into_existing())?;
+
+        // Attempt to initialise a browser to ensure we can look at branches in the project
+        let _stats = state.with_browser(&fakie.urn(), |browser| Ok(browser.get_stats()?))?;
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn create_with_existing_remote() -> Result<(), Error> {
```
NunoAlexandre, line 700: I would highly benefit from having tests such as this one with a simple comment briefly explaining what it's testing. I'd need to read the whole thing to understand what it is trying to create and what it means to create _it_with an existing remote.

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+        let (api, _run_loop) = config.try_into_peer().await?.accept()?;
+        let state = State::new(api, signer.clone());
+
+        let kalt = state.init_owner(&signer, "kalt")?;
+
+        let fakie = state.init_project(&signer, &kalt, &fakie_project(repo_path.clone()))?;
+
+        let fake_fakie = repo_path.join("fake-fakie");
+
+        let copy = Command::new("cp")
```
NunoAlexandre, line 715:
```suggestion
        let kalt = state.init_owner(&signer, "kalt")?;
        let fakie = state.init_project(&signer, &kalt, &fakie_project(repo_path.clone()))?;
        let fake_fakie = repo_path.join("fake-fakie");
        let copy = Command::new("cp")
```

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+        let signer = signer::BoxedSigner::from(key.clone());
+        let config = config::default(key, tmp_dir.path())?;
+        let (api, _run_loop) = config.try_into_peer().await?.accept()?;
+        let state = State::new(api, signer.clone());
+
+        let kalt = state.init_owner(&signer, "kalt")?;
+
+        let fakie = state.init_project(&signer, &kalt, &fakie_project(repo_path.clone()))?;
+
+        let fake_fakie = repo_path.join("fake-fakie");
```
NunoAlexandre, line 713: Though I enjoy some humour, I got no idea what `fake_fakie` means to represent without digesting the whole block.

proxy/coco/tests/gossip.rs
```diff
@@ -0,0 +1,189 @@
...
+use coco::{Hash, Urn};
+
+mod common;
+use common::{
+    build_peer, build_peer_with_seeds, init_logging, radicle_project, shia_le_pathbuf,
+    wait_connected,
+};
+
+#[tokio::test]
+async fn announce_solo() -> Result<(), Box<dyn std::error::Error>> {
```
NunoAlexandre, line 19: Again, a test I'd have no idea what it's doing without reading dozens of lines of code. This should be ignored lightly, it creates unnecessary strain in the reader that is not as familiar with the whole domain.

proxy/coco/tests/gossip.rs
```diff
@@ -0,0 +1,189 @@
...
+        })
+        .map(|_| ());
+    tokio::pin!(announced);
+    timeout(Duration::from_secs(1), announced.next()).await?;
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn providers_is_none() -> Result<(), Box<dyn std::error::Error>> {
```
NunoAlexandre, line 116: My comment on this test was dropped when moving this test to here...

proxy/coco/tests/gossip.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    let res = state
+        .lock()
+        .await
+        .providers(unkown_urn, Duration::from_secs(5))
+        .await
+        .next()
+        .await;
+
+    assert!(res.is_none(), "expected no results",);
```
NunoAlexandre, line 138:
```suggestion
    assert!(res.is_none(), "didn't expected to obtain any providers");
```

proxy/coco/tests/gossip.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn providers() -> Result<(), Box<dyn std::error::Error>> {
```
NunoAlexandre, line 144: Again, my comment on this test was dropped when moving this test to here...

proxy/api/src/http/identity.rs
```diff
@@ -168,18 +175,24 @@ mod test {
...
-        let user = ctx.peer_api.init_user(&ctx.signer, "cloudhead")?;
-        let urn = user.urn();
-        let handle = user.name().to_string();
-        let peer_id = ctx.peer_api.peer_id();
-        let shareable_entity_identifier = (peer_id.clone(), user).into();
+
+        let user = {
+            let state = ctx.state.lock().await;
+            state.init_user(&ctx.signer, "cloudhead")?
+        };
```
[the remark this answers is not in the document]
FintanH: Note that the semantics are different in your suggestion. You're holding onto the lock so no other thread can get ahold of it. Whereas in the original case the lock is dropped and a thread that's waiting could progress itself.

xla:
> Ya I mean, most methods in `State` are rather shallow wrappers, and that's what I imagined would go into `PeerApi`. But it's fine, we can just return a "dumb" struct from `Peer::accept`, which doesn't contain any `Storage` (and is therefore `Sync`), and from which you can move out fields as you wish.

@kim Ha! I see now where you pointing at. While the `State` is indeed a shallow wrapper we always had it in upstream since we started integrating librad. Seems to be the diff view of github that makes us beleive it was introduced in this change-set:

https://github.com/radicle-dev/radicle-upstream/blob/ccfa9a4842b676d42f6dd76b7dbbaa6104a7b9f4/proxy/coco/src/peer.rs#L56-L508

Nonetheless we should audit it and move the parts into librads `PeerApi` that make sense as high-level protocol offering. Created #890 for that.

> I don't mind whether this is set up here, or the return value of `Peer::accept` contains some sort of `StorageFactory`, which is by default initialised with a capacity == the number of CPUs.

The latter sounds somewhat more desirable given our experience so far. We also see that a large range of our functionality doesn't depend on a running peer/protocol and exclusively does modifications to the storage. When we need the run_loop, we also need the protocol. Maybe accept could return something like: `(GenericStorageFactoryManager, Protocol, RunLoop)`. In any case we should track this somewhere, would be best to start an issue in radicle-link for this?

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// Create a new [`State`] given a [`PeerApi`].
+    #[must_use]
+    pub fn new(api: PeerApi<keys::SecretKey>, signer: signer::BoxedSigner) -> Self {
+        let paths = api.paths();
+
+        // Turns out we don't need to call the git commands from the command line.
+        // All we need is to register the git transport and this is the same
+        // thing. This means we can use git2 to execute actions. So instead of
+        // implementing the checkout via git clone as a Command, we use
+        // [`git2::RepoBuilder`].
```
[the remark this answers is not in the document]
FintanH: You're right, it's weird that there's context alluded to here.

Does it make any more sense if we said:
"Register the transport so we can use git2 to execute actions such as checkouts, fetch, and push. The transport will then handle the interaction with the monorepo for us."

proxy/api/src/http/identity.rs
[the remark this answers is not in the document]
NunoAlexandre: That's right, thanks for pointing that out

xla:
> I understand why you felt motivated towards this refactor, although I find it difficult to visualise how well the problems - somewhat generic - stated in the original issue are addressed here.
> 
> On another note, I don't see (and this might just as well be my bad and not your demerit) such an added value by introducing this change, especially in a time where we are focusing on releasing. Now that's there, however, it's surely worthy of appreciation, of course :)

@NunoAlexandre Thanks for the thorough review. Ths is not a refactor for style reason or beautification of the codebase. It's a neceesry prerequisite to full-fill our planned release features. To illustrate, so far we started to advance the protocol network incidently in:

https://github.com/radicle-dev/radicle-upstream/blob/ccfa9a4842b676d42f6dd76b7dbbaa6104a7b9f4/proxy/coco/src/peer.rs#L75-L82

This is the result of evolving the code-collaboration and tackle one thing at a time, for the longest time we disregarged the run_loop there until we wanted to assert that we can clone from another peer. What that meant is that when we had to introduce a continuous background task that full-fills a specific feature #602 - it ended up here:

https://github.com/radicle-dev/radicle-upstream/blob/ccfa9a4842b676d42f6dd76b7dbbaa6104a7b9f4/proxy/api/src/main.rs#L110-L137

As it was the only place which did task management. It didn't feel great but landed the feature, but was pretty much impossible to integration test, which we achieve with relative ease in this change, which for me is a measure of success. Furthermore it means we have structure in place to handle concurrent tasks concerning peer/network operations. For me that means being focused on the release. :>

> Amongst the comments I left, a special note to the comments I had left in some tests I introduced in the last days that were discretely deleted when moving the tests around. In a pursue of trust and solid collaboration, I highly appreciate my non-unsensible (however subjective) development experience to be considered, especially when it adds objective value for the "average" contributor of this project.

I understand that this is off-putting and looks like your view was disregarded. Will introduce your comments where preservable and take care of the places where current names don't signal enough.

NunoAlexandre:
> > I understand why you felt motivated towards this refactor, although I find it difficult to visualise how well the problems - somewhat generic - stated in the original issue are addressed here.
> > On another note, I don't see (and this might just as well be my bad and not your demerit) such an added value by introducing this change, especially in a time where we are focusing on releasing. Now that's there, however, it's surely worthy of appreciation, of course :)
> 
> @NunoAlexandre Thanks for the thorough review. Ths is not a refactor for style reason or beautification of the codebase. It's a neceesry prerequisite to full-fill our planned release features. To illustrate, so far we started to advance the protocol network incidently in:
> 
> https://github.com/radicle-dev/radicle-upstream/blob/ccfa9a4842b676d42f6dd76b7dbbaa6104a7b9f4/proxy/coco/src/peer.rs#L75-L82
> 
> This is the result of evolving the code-collaboration and tackle one thing at a time, for the longest time we disregarged the run_loop there until we wanted to assert that we can clone from another peer. What that meant is that when we had to introduce a continuous background task that full-fills a specific feature #602 - it ended up here:
> 
> https://github.com/radicle-dev/radicle-upstream/blob/ccfa9a4842b676d42f6dd76b7dbbaa6104a7b9f4/proxy/api/src/main.rs#L110-L137
> 
> As it was the only place which did task management. It didn't feel great but landed the feature, but was pretty much impossible to integration test, which we achieve with relative ease in this change, which for me is a measure of success. Furthermore it means we have structure in place to handle concurrent tasks concerning peer/network operations. For me that means being focused on the release. :>
> 


Thanks for the thorough reply :)

For clarity, I didn't mean to say I saw this as purely organizational refactoring, but rather that I found it difficult to see what value it offers _in practice_ for the release. I appreciate your effort in explaining as you made it much clearer for me :ok_hand:

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+
+        Ok(project_meta)
+    }
+
+    /// Retrieves the [`librad::git::refs::Refs`] for the state owner.
+    ///
+    /// # Errors
+    ///
+    /// * if opening the storage fails
+    pub fn list_owner_project_refs(&self, urn: &RadUrn) -> Result<Refs, Error> {
```
FintanH, line 229: Can we get a patch for this @xla /cc @sarahscott 

It's causing Sarah issues on her work :)

xla: #891

proxy/coco/src/peer/announcement.rs
```diff
@@ -0,0 +1,283 @@
...
+/// # Errors
+///
+/// * if listing of the projects fails
+/// * if listing of the Refs for a project fails
+pub async fn build(state: Lock) -> Result<Updates, Error> {
+    let state = state.lock().await;
+    let mut list: Updates = HashSet::new();
+
+    match state.list_projects() {
+        // TODO(xla): We need to avoid the case where there is no owner yet for the peer api, there
```
FintanH, line 73: Is there an issue for this, I have an idea :ghost:

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+use crate::user::{verify as verify_user, User};
+
+/// Thread-safe wrapper for [`State`].
+pub type Lock = Arc<Mutex<State>>;
+
+impl From<State> for Lock {
+    fn from(state: State) -> Self {
+        Arc::new(Mutex::new(state))
+    }
+}
```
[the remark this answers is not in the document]
FintanH: I wouldn't say it's "locking the state". It's just packing the state in a mutex, which can _later_ be locked.

If you think about it this and your proposal are the exact same. There's a function that turns `State` into `Lock`. The difference just being the name.

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+    #[must_use]
+    pub fn new(api: PeerApi<keys::SecretKey>, signer: signer::BoxedSigner) -> Self {
+        let paths = api.paths();
+
+        // Turns out we don't need to call the git commands from the command line.
+        // All we need is to register the git transport and this is the same
+        // thing. This means we can use git2 to execute actions. So instead of
+        // implementing the checkout via git clone as a Command, we use
+        // [`git2::RepoBuilder`].
+        transport::register(transport::Settings {
```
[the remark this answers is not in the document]
FintanH: I suppose these things are affected by how much you're familiar with the surrounding imports. For example, I know well that this `transport` comes from `librad` and where it's located.

With that in mind, I think I prefer the original way because if you need to know where it comes from you can grep `transport` and go up to the imports. But that's because I have a personal preference for 1 qualified path without ambiguity and 2 if we need to disambiguate. 

Again, that's personal preference and knowledge of the codebases in question :) What do you think with that in mind?

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// Set the default owner for this `PeerApi`.
+    ///
+    /// # Errors
+    ///
+    ///   * Fails to set the default `rad/self` for this `PeerApi`.
+    pub fn set_default_owner(&self, user: User) -> Result<(), Error> {
+        Ok(self.api.storage().set_default_rad_self(user)?)
+    }
+
+    /// Initialise a [`User`] and make them the default owner of this `PeerApi`.
```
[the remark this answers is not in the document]
FintanH: What do you mean by redundancy?

There can be only one owner, and setting the default owner lets us know which user entity that is -- because there will be multiple users in the monorepo which are other peer's users.

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+        log::debug!("Created project with Urn '{}'", urn);
+
+        let repo = project.setup_repo(LocalUrl::from_urn(urn, self.api.peer_id().clone()))?;
+        log::debug!("Setup repository at path '{}'", repo.path().display());
+
+        Ok(meta)
+    }
+
+    /// Create a [`user::User`] with the provided `handle`. This assumes that you are creating a
+    /// user that uses the secret key the `PeerApi` was configured with.
```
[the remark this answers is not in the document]
FintanH: If another key is used with the user, for example to create a project, the operation will fail because of key mismatches.

proxy/coco/src/peer/announcement.rs
xla: @FintanH You reckon it's a link issue, or something we should track in upstream?

proxy/coco/src/state.rs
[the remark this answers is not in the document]
xla: As @FintanH said, it's not locking the state. Just a value-to-value conversion. Which is the main use-ase for `From` and gives extra utility over the alternative you propose.

proxy/coco/tests/common.rs
```diff
@@ -0,0 +1,97 @@
+use std::{path::PathBuf, time::Duration};
```
FintanH, line 1: `fixtures` might be a good name for this module to be inline with `librad`. Just a suggestion though :)

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+/// High-level interface to the coco monorepo and gossip layer.
+pub struct State {
+    /// Internal handle on [`PeerApi`].
+    pub(crate) api: PeerApi<keys::SecretKey>,
+}
+
+impl State {
+    /// Create a new [`State`] given a [`PeerApi`].
+    #[must_use]
+    pub fn new(api: PeerApi<keys::SecretKey>, signer: signer::BoxedSigner) -> Self {
```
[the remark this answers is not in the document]
xla: I prefer the former, because the qualified use gives you context where it originates from. Esp. in a codebase like this where a lot of the sub-domains reappear and are reused.

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+            signer,
+        });
+
+        Self { api }
+    }
+
+    /// Returns the [`PathBuf`] to the underlying monorepo.
+    #[must_use]
+    pub fn monorepo(&self) -> PathBuf {
+        self.api.paths().git_dir().join("")
```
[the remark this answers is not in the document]
xla: Will be looked at in the context of #890

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+    }
+
+    /// Returns the underlying [`paths::Paths`].
```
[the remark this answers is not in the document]
xla: This is link concenr and another todo while working on #890

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+    ///
+    ///   * Unable to acquire the lock.
+    pub fn has_commit(&self, urn: &RadUrn, oid: impl Into<git2::Oid>) -> Result<bool, Error> {
+        Ok(self.api.storage().has_commit(urn, oid.into())?)
+    }
+
+    /// Our current peers [`PeerId`].
+    #[must_use]
+    pub fn peer_id(&self) -> PeerId {
+        self.api.peer_id().clone()
```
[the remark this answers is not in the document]
xla: #890

proxy/coco/src/peer/announcement.rs
FintanH: Hrmmm good question. My idea is that we could use the ol' `PhantomData` trick to make sure we get a `Store` that has its default rad self is set. I'll create an issue in link and can discuss there.

proxy/coco/src/state.rs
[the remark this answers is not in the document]
xla: @FintanH Can probably answer that.

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+        let user = verify_user(user)?;
+
+        self.set_default_owner(user.clone())?;
+
+        Ok(user)
+    }
+
+    /// Given some hints as to where you might find it, get the urn of the project found at `url`.
+    ///
+    /// **N.B.** This needs to be run with `tokio::spawn_blocking`.
```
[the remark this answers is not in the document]
xla: TODO for #890

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+            }
+        }
+
+        Ok(entities)
+    }
+
+    /// Given some hints as to where you might find it, get the urn of the user found at `url`.
+    ///
+    /// **N.B.** This needs to be run with `tokio::spawn_blocking`.
+    ///
```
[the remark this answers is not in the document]
xla: #890

proxy/coco/src/peer/announcement.rs
FintanH: https://github.com/radicle-dev/radicle-link/issues/316

proxy/coco/src/state.rs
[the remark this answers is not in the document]
xla: The signing will fail.

proxy/coco/src/state.rs
xla: Convinced these comments are a liability and just add noise. If the handful of lines below it do not communicate clearly what is going on we failed in way more severe ways.

proxy/coco/src/state.rs
xla: In your proposal `storage` won't be available in the else branch.

proxy/coco/src/state.rs
xla: #890

proxy/coco/src/state.rs
xla: #890

proxy/coco/tests/gossip.rs
xla: What would you propose as an alternative?

proxy/coco/src/state.rs
xla: What would make it easier for you to understand?

proxy/coco/src/state.rs
xla: @FintanH knows the fakies.

proxy/coco/src/peer/announcement.rs
xla: Looks like another thing to look at in #890

proxy/coco/src/state.rs
FintanH: Do you have a suggestion here?

We know that it's a `PathBuf` because of `repo_path.join` and it's used as a target directory for `cp`. The whole context is the test itself.

`copied_fakie` maybe? :shrug:

proxy/coco/tests/common.rs
xla: It's not only fixtures tho, also a bunch of utility.

proxy/coco/src/state.rs
```diff
@@ -0,0 +1,733 @@
...
+impl From<State> for Lock {
+    fn from(state: State) -> Self {
+        Arc::new(Mutex::new(state))
+    }
+}
+
+/// High-level interface to the coco monorepo and gossip layer.
+pub struct State {
+    /// Internal handle on [`PeerApi`].
+    pub(crate) api: PeerApi<keys::SecretKey>,
```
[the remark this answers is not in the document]
xla: #890 again

proxy/coco/src/lib.rs
xla: #890

proxy/coco/tests/gossip.rs
FintanH: `can_announce_new_project` maybe?

proxy/coco/tests/common.rs
FintanH: :+1:

proxy/coco/src/state.rs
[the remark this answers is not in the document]
FintanH: :+1:

FintanH (dismissed):

NunoAlexandre (approved):
:surfer: :1st_place_medal: :mount_fuji:

proxy/coco/src/state.rs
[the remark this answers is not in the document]
kim: > the peer_id is not a stateful matter

It is, actually, because everything below `refs/<namespace>/rad/*` is expected to be signed by the `PeerId` which was advertised during the handshake. Iow, one cannot simply change the local key without rewriting the repository, or else bad things will happen (although we haven't yet enumerated all of them).

proxy/coco/src/state.rs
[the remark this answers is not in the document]
NunoAlexandre: thanks for the clarification, I had misunderstood the issue as being a typed effect :+1:

kim (commented):
Do you maybe want to remove the `storage.reopen()` calls?

xla:
@kim Removed all of them, looks like no libgit handles were harmed in the process.

NunoAlexandre (approved):
:space_invader:

FintanH (approved):

PR #883, opened 2020-09-09
chore(proxy): rustfmt merge_imports

NunoAlexandre:
Enable Rust cargo fmt `merge_imports`.

**Why**:

- This is enabled on `link`, therefore enabling it here as well reduces the gap between the two projects
- I have come to appreciate this on `link`, as it reduces the amount of duplicated `import <crate>::` lines
- Neater imports

### DIFF
--- a/proxy/.rustfmt.toml
+++ b/proxy/.rustfmt.toml
@@ -5,6 +5,7 @@ comment_width = 100
 wrap_comments = true
 
 # imports
+merge_imports = true
 reorder_imports = true
 
 # strings
--- a/proxy/api/src/bin/nuke.rs
+++ b/proxy/api/src/bin/nuke.rs
@@ -1,13 +1,10 @@
-use std::fs::remove_dir_all;
-use std::io::ErrorKind;
-use std::process::exit;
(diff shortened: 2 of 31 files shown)


cloudhead (dismissed):

xla (changes requested):
Strongly opposed to this change, it makes it way harder to edit the imports and trace them quickly.

MeBrei (dismissed):

FintanH:
>  it makes it way harder to edit the imports and trace them quickly.

Could you expand on these points?

From my point of view, editing doesn't seem hard. You add the new `sub-crate`, `function`, or `type` into the list.
And for tracing, they're still alphabetical and grouped under their common structure.
That's just my opinion on those two points and would appreciate to get an insight on what annoyances you see :)

sarahscott (approved):
i'm a fan of consistency; if this is how it's done in link i see no reason for it to be different in proxy.

but i am neither 'strongly opposed' nor 'strongly in favor', fwiw.

xla (approved):

Issue #592, opened 2020-06-25
discovery: wire up track button


FintanH:
When a user clicks the track button on the discovery page, they should start tracking that project and its updates.

rudolfs:
While discussing the "Track Button" feature it was also not clear where the state of that something is being tracked is stored? Does the proxy store this state or does it live somewhere in the mono-repo/protocol?

PR #886, opened 2020-09-09
refactor(ui): project page

### LINKED ISSUE #834: Overhaul project page
- [x] extract user selector into the header
- [x] move branches and contributor counters into the header (and they shouldn't be links anymore, just icon+text)
- [x] rename and move the tracking button into the header 
- [x] move and rework the navigation bar (tabs: source & commits, no additional drop down, no support button, add checkout button)
- [x] remove space character between `Read more` link and `.`

MeBrei:
Closes #834 

**ToDo**
smaller items
- [x] style everything
- [x] scrolling + fixed header on all project tabs
- [x] fix issue where the peer dropdown sometimes doesn't close after selection (see failing test)

bigger issues
- [x] refactor how params/variables are passed? The main project page will function now more than before as the central place for fetching and passing data to the tabs (passing around the selected peer (`PeerSelector`) and the current `revision` (`RevisionSelector` from `Source`)). My suggestion would be to move the data out of the `path` params and into `Context` or stores.
- [x] refactor the use of `revisions`, `revision`, `currentRevision` as it sometimes refers to the content of the `revisionStore` and sometimes to a construct from the `RevisionSelector`
- [x] on peer select, update the source page (keep `currentObjectPath` etc) instead of going to the README + highlight Source tab
- [x] Move `Commits` into tab
  - [x] Does the commit count depend on the selected peer?
  - [x] What happens if the peer is selected while looking at the commit tab?

Follow up/outstanding issues
- show actual commit count of the branch instead of project stats #924
- collapse the header on scroll #923
- rename & style follow/unfollow button #592


rudolfs (commented):
Massive feat!! 🚀 🚀 🚀 


- [ ] clicking on the "Source" tab doesn't redirect to project root if we're already in source browsing

- [ ] these files are not used anymore and should be deleted:
```
/Project/SourceMenu.svelte
/Project/IssuesMenu.svelte
/Project/RevisionsMenu.svelte
```

- [ ] revision selector text jumps around (can be done in a styling follow up)
  browser-jump

- [ ] revision selector doesn't truncate selected branch (can be done in a styling follow up)
  Screenshot 2020-09-17 at 16 49 09


- are we testing that the selected peer is highlighted when dropdown is open (same for revision selector)?

- what happens if we switch peers and the new peer doesn't have the branch we had selected for the previous peer?

cypress/integration/project_source_browsing.spec.js
```diff
@@ -225,51 +225,30 @@ context("source code browsing", () => {
         cy.contains(".i-too-am-hidden").should("exist");
       });
 
-      it("allows selecting different peers", () => {
+      it("updates when selecting different peers", () => {
```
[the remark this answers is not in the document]
sarahscott: Agree, especially because I imagine peer-related functionality will be changing a lot in the near future.

ui/DesignSystem/Component/Header/Large.svelte
```diff
@@ -4,7 +4,7 @@
 
   export let style = null;
   export let entity = null;
-  export let variant = null; // profile | org
+  export let variant = null; // profile | project | org
```
sarahscott, line 7: This component is a little easier to reason about now - after you merge `master` you can just specify the `name`, `urn`, etc.

ui/src/source.ts
```diff
@@ -127,6 +127,15 @@ export const object = objectStore.readable;
 const revisionsStore = remote.createStore<Revisions>();
 export const revisions = revisionsStore.readable;
 
+export const objectType = writable(ObjectType.Tree);
```
sarahscott, line 130: I think these actions would make sense as events in the event loop, so we can keep things consistent. If that's a lot for this PR, we should make a followup issue. Having both approaches (some updates use the event structure, others don't) adds some unnecessary complexity, so I feel we should go with one or the other.

sarahscott:
Looking lovely so far ✨! Lmk if you'd rather address the `source.ts` feedback in a subsequent issue or PR; this one is already a lot of really great work and I don't want it to languish forever unmerged 😄

ui/src/source.ts
xla: On point observation.

xla (commented):
:clap: :clap: :clap: - Great work, the move of data access into the screen and out of components is a net positive!

ui/src/source.ts
MeBrei: > I think these actions would make sense as events in the event loop, so we can keep things consistent. If that's a lot for this PR, we should make a followup issue. Having both approaches (some updates use the event structure, others don't) adds some unnecessary complexity, so I feel we should go with one or the other.

I am not sure what you mean with `event in the event loop`. We set up an infrastructure for remote stores, ie stores where we have a loading, error etc. state. The stores added here are simpler, so setting and reading them is actually enough. For me using them as remote stores would add more complexity.

ui/DesignSystem/Component/Header/Large.svelte
MeBrei: I added more variables now (description, stats) and depended showing the avatar on if an avatar fallback is provided

ui/src/source.ts
rudolfs: Have to agree with @MeBrei here, I don't see what wrapping everything in the event code would add here besides another layer of indirection.

MeBrei:
> Massive feat!! 🚀 🚀 🚀
> 
> *  clicking on the "Source" tab doesn't redirect to project root if we're already in source browsing

Since the source path does not include any other params than the project id, a reload of the page can't be triggered via `push(path.projectSource(project.id))`, since svelte does not reload without changes. For it to jump back to the readme, the stores need to be reset on tab click (on page load they are reseted). That would mean adding a custom click handler to the Horizontal Menu for the Source tab. 
That seemed a bit like a hack to me. Happy to implement something if there are other ideas how to solve it, but I also talked to @juliendonck, who said it would be fine for now.

> * are we testing that the selected peer is highlighted when dropdown is open (same for revision selector)?

added a test in 350da18, was green before merge. On master tests connected to revision/peer selector are skipped.

> * what happens if we switch peers and the new peer doesn't have the branch we had selected for the previous peer?

After master merge, faking peers is no longer possible. I think this needs a bit more work and a better test set up, also finding all the edge cases, but for now I observed that if the peer doesn't have that branch, it jumps to the default branch, and the file that was opened stays opened, but is updated to that branch's state. But if the peer does not have that file on their master branch there is an `Invalid input` error notification.

rudolfs (dismissed):
🎩 ab!

rudolfs (approved):

ui/DesignSystem/Component/Header/Large.svelte
```diff
@@ -1,34 +1,29 @@
...
-  import type { Identity } from "../../../src/identity";
-  import type { Org } from "../../../src/org";
+  import type { Stats } from "../../../src/project";
 
   import { Avatar, Icon } from "../../Primitive";
   import Urn from "../Urn.svelte";
 
   export let name: string;
   export let urn: string;
+  export let description: string | undefined = undefined;
```
sarahscott, line 10:
```suggestion
  export let description = "";
```

sarahscott: bc it's an optional string prop ✨

MeBrei: For me `optional` means, the `description` would be `string | undefined`. But if I understand correctly, with the above notation it is `string(set by user) | string(empty)`.
That means if I don't provide a description it is set to `""` and code relying on if description was provided has to do `{#if description.length > 0}` instead of `{#if description}`

sarahscott: Correct, we discussed this [here](https://github.com/radicle-dev/radicle-upstream/pull/867#discussion_r482532377) and went with the empty string as a default value, which also ensures we handle the empty string state in UI (i.e. if the parent passes an empty string, we probably shouldn't render empty `div`s).

MeBrei: 9f0bdbd

rudolfs (approved):

ui/src/source.ts
sarahscott: Cool, we can discuss elsewhere

sarahscott (approved):

PR #891, opened 2020-09-11
fix(proxy): signed refs retrieval for projects

xla:
Splits up the functionality of getting signed refs for a project, to
make mis-use hard we have two methods now. One to obtain the `Refs` for
the owner of the monorepo and one to get the them for a peer.

***
Backport from #881 and sparked by https://github.com/radicle-dev/radicle-upstream/pull/881#discussion_r487044797.


FintanH (approved):
Thanks :+1:

sarahscott (approved):

Issue #855, opened 2020-09-01
Make announcement period configurable


xla:
Currently we hard-code a ten second interval for the announcement of updated project heads. This should be coming from some form of configuration/settings.

xla:
Done in #896 - how/if to tune these knobs by an end-user is for future iterations.

PR #896, opened 2020-09-14
feat(proxy): sync with peers when coming online

### LINKED ISSUE #852: sync on startup

xla:
Closes #852


xla:
This is ready for review, sans an integration test that is currently underway. Hope to get some eyes on it now.

NunoAlexandre:
@xla before reviewing, were there any special technical decisions or trade-offs you made along the way that you find would be useful to mention? #justcurious

xla:
> @xla before reviewing, were there any special technical decisions or trade-offs you made along the way that you find would be useful to mention? #justcurious

@NunoAlexandre The state machine in `run_state` is not implemented with the pattern that emerges in our codebases. It's outlined in [this post](https://hoverbear.org/blog/rust-state-machine-pattern/) and used in varying degrees (like type witnessing). Which makes the transitioning code a big ole match blob, likely to become unwieldy if extended further. Streamlining it should be left to a follow-up exercise.

Beyond that there is a lot of CSP style channel communication used for the plumbing of data flowing through the peer. To the extend that errors are just fed in as failure variants of events. This is done to not crash the run loop, but these errors should be audited to see if there are any non-recoverable states.

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+}
+
+/// Internal state representation the current mode of operation for the local peer.
+#[derive(Debug)]
+enum Status {
+    /// Nothing is setup, not even a socket to listen on.
+    Stopped(Instant),
+    /// Local peer is listening on a socket, but no connected peers yet.
+    Started(Instant),
+    /// The loccal peer lost connections to all peers.
```
FintanH, line 79:
```suggestion
    /// The local peer lost its connections to all its peers.
```

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+    Offline(Instant),
+    /// Phase where the local peer tries get up-to-date.
+    Syncing(Instant, usize),
+    /// Normal operation with other other peers being connected.
```
FintanH, line 83:
```suggestion
    /// The local peer is operational and is able to interact with the peers it has connected to.
```

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+    Online(Instant),
+}
+
+/// Set of knobs to change the behaviour of the [`RunState`].
+#[derive(Default)]
+pub struct Config {
```
FintanH, line 89: The configuration parameters that have TODOs might fit in here, where the `Default` impl uses the `const` values.

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+    /// Enables the syncing stage when coming online.
+    pub sync_on_startup: bool,
+}
+
+/// State kept for a running local peer.
+pub struct RunState {
+    /// Confiugration to change how input [`Event`]s are interpreted.
```
FintanH, line 96: How does it change the interpretation?

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+    #[cfg(test)]
+    const fn new(config: Config, connected_peers: HashSet<PeerId>, status: Status) -> Self {
+        Self {
+            config,
+            connected_peers,
+            status,
+        }
+    }
+
+    /// Applies the `input` and based on the current state transforms to the new state and in some
```
FintanH, line 125:
```suggestion
    /// Applies the `event` and based on the current state transforms to the new state and in some
```

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+    /// cases produes commands which should be executed in the appropriate sub-routines.
+    pub fn transition(&mut self, event: Event) -> Vec<Command> {
+        match (&self.status, event) {
+            // Go from [`Input::Stopped`] to [`Input::Offline`] once we are listening.
+            (Status::Stopped(_since), Event::Protocol(ProtocolEvent::Listening(_addr))) => {
```
FintanH, line 130: Could be nice to log these values at least in `trace`

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+                    ]
+                } else {
+                    self.status = Status::Online(Instant::now());
+
+                    vec![]
+                }
+            },
+            // Sync until configured maximum of peers is reached.
+            (Status::Syncing(since, syncs), Event::Protocol(ProtocolEvent::Connected(peer_id)))
+                if *syncs < SYNC_MAX_PEERS =>
```
FintanH, line 157: `SYNC_MAX_PEERS` definitely could be defined in `Config`, I would say

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+                } else {
+                    self.status = Status::Syncing(*since, syncs + 1);
+                }
+
+                vec![Command::SyncPeer(peer_id)]
+            }
+            // TODO(xla): Also issue sync if we come online after a certain period of being
+            // disconnected from any peer.
+            // Issue more syncs if we connect to new peers while syncing.
+            (
```
FintanH, line 171: Is it possible for this match to be hit? :thinking:

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+                if self.connected_peers.len() == 1 =>
+            {
+                self.connected_peers.remove(&peer_id);
+                self.status = Status::Offline(Instant::now());
+
+                vec![]
+            }
+            // Remove peer that just disconnected.
+            (_, Event::Protocol(ProtocolEvent::Disconnecting(peer_id))) => {
+                self.connected_peers.remove(&peer_id);
```
FintanH, line 196: I think we could merge these two branches:
```rust
self.connect_peers.remove(&peer_id)
if self.connected_peers.is_empty() {
  self.status = Status::Offline(Instant::now())
}
vec![]
```

proxy/coco/src/state.rs
```diff
@@ -228,8 +228,10 @@ impl State {
     ///
     /// * if opening the storage fails
     pub fn list_owner_project_refs(&self, urn: &RadUrn) -> Result<Refs, Error> {
-        let storage = self.api.storage();
-        storage.rad_signed_refs(urn).map_err(Error::from)
+        let storage = self.api.storage().reopen()?;
```
FintanH, line 231: Do we know if we have to `reopen` here? :thinking:

proxy/coco/src/state.rs
```diff
@@ -228,8 +228,10 @@ impl State {
...
+        let refs = storage.rad_signed_refs(urn).map_err(Error::from)?;
```
FintanH, line 232:
```suggestion
        let refs = storage.rad_signed_refs(urn)?;
```

I don't think we need to do the `map_err` here, but maybe I'm wrong?

proxy/coco/tests/common.rs
```diff
@@ -11,17 +11,48 @@ use librad::{keys::SecretKey, net::protocol::ProtocolEvent, peer::PeerId, signer
...
+    }};
+    ( $receiver:expr , $pattern:pat if $cond:expr ) => {{
+        $crate::await_event!($receiver, |res| match res.unwrap() {
+            $pattern if $cond => future::ready(Some(())),
+            _ => future::ready(None),
+        })
+    }};
+}
+
+#[allow(dead_code)]
```
FintanH, line 41: Eh? :thinking:

proxy/coco/tests/replication.rs
```diff
@@ -215,3 +247,75 @@ async fn can_fetch_project_changes() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn can_sync_on_startup() -> Result<(), Box<dyn std::error::Error>> {
```
FintanH, line 252: How do you feel about using the `testnet` code here instead?

proxy/coco/tests/replication.rs
```diff
@@ -215,3 +247,75 @@ async fn can_fetch_project_changes() -> Result<(), Box<dyn std::error::Error>> {
...
+            alice_store,
+            RunConfig {
+                sync_on_startup: true,
+            },
+        ));
+        tokio::task::spawn(bob_peer.run(bob_state.clone(), bob_store, RunConfig::default()));
+        connected(bob_events, &alice_peer_id).await?;
+    };
+
+    assert_event!(
```
FintanH, line 315: I also expected an assertion that bob has the updates in his `Storage` after syncing up with alice

proxy/coco/src/peer.rs
```diff
@@ -87,60 +77,96 @@ impl Peer {
...
+            self.subscriber.send(event.clone()).ok();
+            log::debug!("{:?}", event);
+
+            for cmd in run_state.transition(event) {
+                match cmd {
+                    Command::Announce => {
+                        Self::announce(state.clone(), store.clone(), announce_sender.clone());
+                    },
+                    Command::SyncPeer(peer_id) => {
+                        Self::sync(state.clone(), peer_sync_sender.clone(), peer_id.clone()).await;
```
FintanH, line 130: Does this mean we're constantly syncing while the application is live?

FintanH (commented):
Lot's of good work :grin: 

Just a few suggestions and questions :ok_hand:

proxy/coco/tests/replication.rs
xla: I didn't go through the length of setting up a project both peers track and that has new changes since last being online. That's mostly because I dunno how to disconnect and reconnect a peer.

proxy/coco/tests/replication.rs
xla: Let's address that in #908 - I fear it might introduce too much noise in this PR which already is not that lightweight.

proxy/coco/tests/replication.rs
FintanH: Hmmm, I wonder if we create and drop the peer in a block, then create it again outside of that block would it be equivalent?

proxy/coco/src/state.rs
xla: Leftovers.

proxy/coco/tests/replication.rs
FintanH: Well what I mean is do the same style as: https://github.com/radicle-dev/radicle-upstream/blob/fintan/testnetz/proxy/coco/tests/replication.rs#L14-L74

And I'm just suggesting this test in particular since it's being introduced fresh anyway. That way we don't need to go back and change it later.

Feel free to resolve this if you'd still rather not.

proxy/coco/src/peer/run_state.rs
xla: It is and it's test in `issue_sync_command_until_max_peers`.

proxy/coco/src/state.rs
FintanH: Thought as much :grin:

proxy/coco/src/peer/run_state.rs
xla: For sure.

proxy/coco/src/peer/run_state.rs
xla: For example it doesn't sync on startup if the config field for it is set to false.

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,361 @@
...
+
+/// Scheduled timeouts which can occur.
+#[derive(Clone, Debug)]
+pub enum TimeoutEvent {
+    /// Grace period is over signaling that we should go offline, no matter how many syncs have
+    /// succeeded.
+    SyncPeriod,
+}
+
+/// Internal state representation the current mode of operation for the local peer.
```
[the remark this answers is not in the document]
xla: It's representing current operational status of the local peer. E.g. offline when no peers are connected or syncing when it came online and catches up with the network. What could be rephrased here to make that more clear?

proxy/coco/src/peer/run_state.rs
FintanH: Ya, I think this became clearer as I read the function below

proxy/coco/src/peer.rs
xla: Nope, this command is only issued when coming online and `sync_on_startup` is set. Also the run_state factors in how many peers to sync with at max and will stop syncing alltogether after a certain period has expired.

proxy/coco/src/peer.rs
```diff
@@ -87,60 +77,96 @@ impl Peer {
...
+            let state = state.lock().await;
             let protocol = state.api.protocol();
             protocol.subscribe().await
         };
         tokio::pin!(protocol_subscriber);
 
         // Start announcement timer.
-        let mut announce_timer = tokio::time::interval(std::time::Duration::from_secs(1));
+        // TODO(xla): Find a more structured approach to manage timings.
+        let mut announce_timer = interval(std::time::Duration::from_secs(1));
```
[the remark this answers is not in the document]
xla: Will be addressed in #855.

proxy/coco/tests/replication.rs
xla: For that we have to have a way to hold onto it's monorepo/start and assoicated key/signer.

proxy/coco/src/peer.rs
FintanH: Nice :ok_hand: So if I'm understanding correctly, that startup logic is isolated in the `run_state` module?

proxy/coco/src/peer/run_state.rs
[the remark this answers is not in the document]
FintanH: Well, the sentence seems off I think because of this portion "representation the current mode", I'm guessing a typo or a missed word.

But also "mode of operation" seems a bit vague. Like the variants describe states of syncing and connectivity which doesn't seem like a mode of operation to me.

Don't wanna bikeshed on this so I'm just gonna suggest to fix the sentence and you can do what you'd like with my above misunderstanding :smile:

proxy/coco/src/peer/run_state.rs
FintanH: So I deleted this case locally and the test still passed :shrug:

proxy/coco/tests/replication.rs
FintanH: We could create the key outside of the block and initialise the storage inside and outside the block. So in pseudo-code:
```rust
let key = SecretKey::new();

{
  let bob = init_peer(key.clone());
  let repo = bob.clone_repo(urn);
}

let commit = alice.commit();
let bob = init_peer(key.clone());
wait_for_sync(bob);
assert!(bob.has_commit(comit))
```

proxy/coco/src/peer.rs
xla: That's correct, this loop just advances the `run_state` and executes any commands if any.

FintanH: I dig it :+1:

proxy/coco/src/peer/run_state.rs
xla: You sir, are correct! Who allowed this `#[dead_code]`?? Wonder why we don't get an `unreachable` on this one.

FintanH: Probably because of the catch-all case at the bottom of the match arms :sweat_smile:

proxy/coco/src/peer/run_state.rs
xla: True, tried to be heaps fancy with these guards.

FintanH: And now I'm here to check your work ;)

proxy/coco/src/peer.rs
```diff
@@ -87,60 +72,98 @@ impl Peer {
...
-        let old = announcement::load(store)?;
-        let new = announcement::build(state.clone()).await?;
-        let updates = announcement::diff(&old, &new);
-
-        announcement::announce(state, updates.iter()).await;
+    fn announce(state: Lock, store: kv::Store, mut sender: mpsc::Sender<AnnounceEvent>) {
+        tokio::spawn(async move {
+            match announcement::run(state, &store).await {
+                Ok(updates) => sender.send(AnnounceEvent::Succeeded(updates)).await.ok(),
+                Err(_err) => sender.send(AnnounceEvent::Failed).await.ok(),
```
FintanH, line 142: I feel like it would be good to log errors initially

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,383 @@
...
+                Event::Announce(AnnounceEvent::Tick),
+            ) => vec![Command::Announce],
+            _ => vec![],
+        }
+    }
+}
+
+#[allow(clippy::needless_update, clippy::panic, clippy::unwrap_used)]
+#[cfg(test)]
+mod test {
```
FintanH, line 221: Are there any tests for going to the `Offline` state? Or in general, do you think there are missing tests for state transitions?

Also, tests for events that come in but we're not in the happy path state?

I've used `cargo-tarpaulin` in the past to see which code paths have I tested for getting code coverage.

FintanH (commented):
A couple more suggestions

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,383 @@
...
+    pub sync_on_startup: bool,
+    /// Duration until the local peer goes online regardless if and how many syncs have succeeded.
+    pub sync_period: Duration,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            sync_max_peers: DEFAULT_SYNC_MAX_PEERS,
+            sync_on_startup: false,
```
NunoAlexandre, line 108: I might answer it throughout the review, but what is the idea behind defaulting to no sync on startup? I could imagine that since it requires more infrastructure to be in place that it might get in the way when running tests and what not, so just leaving a note for now.

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,383 @@
...
+    Offline(Instant),
+    /// Phase where the local peer tries get up-to-date.
+    Syncing(Instant, usize),
+    /// The local peer is operational and is able to interact with the peers it has connected to.
+    Online(Instant),
+}
+
+/// Set of knobs to change the behaviour of the [`RunState`].
+pub struct Config {
+    /// Number of peers a full sync is attempting with up on startup.
```
NunoAlexandre, line 96: I _think_ you meant:

```suggestion
    /// Number of peers that a full sync is attempted with upon startup.
```

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,383 @@
...
+    pub sync_max_peers: usize,
+    /// Enables the syncing stage when coming online.
+    pub sync_on_startup: bool,
+    /// Duration until the local peer goes online regardless if and how many syncs have succeeded.
+    pub sync_period: Duration,
```
NunoAlexandre, line 101: The docs refer to this field as "duration" and the time is also `Duration`, so I'd suggest:

 ```suggestion
     pub sync_duration: Duration,
```

File: proxy/coco/src/peer/run_state.rs [hunk 23]
```diff
@@ -0,0 +1,383 @@
...
+    /// The local peer lost its connections to all its peers.
+    Offline(Instant),
+    /// Phase where the local peer tries get up-to-date.
+    Syncing(Instant, usize),
+    /// The local peer is operational and is able to interact with the peers it has connected to.
+    Online(Instant),
+}
+
+/// Set of knobs to change the behaviour of the [`RunState`].
+pub struct Config {
```
NunoAlexandre, line 95: `nit` As far as I see, the only use case for `Config` is to support the syncing stage, thus I'd suggest:

```suggestion
pub struct SyncConfig {
```

+ and drop "sync_" as prefix from all fields. leave it up to you :ping_pong: 
+ move it to `sync.rs`

Up to you :raised_hands:

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+
+    /// Applies the `event` and based on the current state transforms to the new state and in some
+    /// cases produes commands which should be executed in the appropriate sub-routines.
+    pub fn transition(&mut self, event: Event) -> Vec<Command> {
+        log::trace!("run_state: ({:?}, {:?})", self.status, event);
+
+        match (&self.status, event) {
+            // Go from [`Input::Stopped`] to [`Input::Offline`] once we are listening.
```
NunoAlexandre, line 151: Not sure if I get this comment. Would this be correct?

```suggestion
            // Go from [`Status::Stopped`] to [`Status::Started`] once we are listening.
```

If yes, not sure the comment is necessary. but I might not be understanding things here

proxy/coco/src/peer/run_state.rs
```diff
@@ -0,0 +1,383 @@
...
+                if syncs + 1 == self.config.sync_max_peers {
+                    self.status = Status::Online(Instant::now());
+                } else {
+                    self.status = Status::Syncing(*since, syncs + 1);
+                }
+
+                vec![Command::SyncPeer(peer_id)]
+            }
+            // Go online if we exceed the sync period.
+            (Status::Syncing(_since, _syncs), Event::Timeout(TimeoutEvent::SyncPeriod)) => {
```
NunoAlexandre, line 193: neat :ok_hand:

proxy/coco/src/peer/sync.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use librad::{peer::PeerId, uri::RadUrl};
+
+use crate::state::Lock;
+
+use super::Error;
+
+/// Initiaites a fetch for all locally tracked projects from the given [`PeerId`].
+pub async fn sync(state: Lock, peer_id: PeerId) -> Result<(), Error> {
+    log::debug!("Starting sync for {}", peer_id);
```
NunoAlexandre, line 11:
```suggestion
    log::debug!("Starting sync from/with {}", peer_id);
```

proxy/coco/src/state.rs
```diff
@@ -353,7 +355,7 @@ impl State {
         owner: &User,
         project: &project::Create<P>,
     ) -> Result<librad_project::Project<entity::Draft>, Error> {
-        let storage = self.api.storage();
+        let storage = self.api.storage().reopen()?;
```
NunoAlexandre, line 358: what triggered this change?

proxy/coco/tests/common.rs
```diff
@@ -11,31 +11,59 @@ use librad::{keys::SecretKey, net::protocol::ProtocolEvent, peer::PeerId, signer
...
+    }};
+    ( $receiver:expr , $pattern:pat if $cond:expr ) => {{
+        $crate::await_event!($receiver, |res| match res.unwrap() {
+            $pattern if $cond => future::ready(Some(())),
+            _ => future::ready(None),
+        })
+    }};
+}
+
+pub async fn connected(
```
NunoAlexandre, line 41: A doc here wouldn't hurt

NunoAlexandre (commented):
Great work, @xla! :ok_hand:

proxy/coco/src/peer/run_state.rs
xla: Yeah not a bad suggestion, just that this struct will be used to house all run loop configuration. Good candidate is the announcement tick duration and other knobs in the future. What could be done is to have separate struct bundling all sync concerns that is a field of `Config`.

proxy/coco/src/state.rs
xla: Residue, will remove.

proxy/coco/src/peer/run_state.rs
xla: It's quite a noisy and bandwidth intense behaviour and should be done with care. It's a stop-gap solution until more effective dissemination strategies have settled in the protocol. Where this application change will help to inform. I'd err on the side of caution here, even so if we make it the default behaviour for upstream instances for the foreseeable future.

It does help for the tests as well as it is a happening on startup, in either case it's good to have a knob. Happy to revise the default if it is a bigger surprise to see it disabled.

proxy/coco/src/peer/run_state.rs
xla: I'd like to avoid that this looks like including the type in the variable name. Personally the phonetics are more pleasing and I like this definition:

> the interval of time between successive occurrences of the same state in an oscillatory or cyclic phenomenon, such as a mechanical vibration, an alternating current, a variable star, or an electromagnetic wave.

proxy/coco/src/peer/run_state.rs
xla: Good catch, trusty mate tarpaulin indeed revealed that this case wasn't covered. For future generations #939.

proxy/coco/tests/replication.rs
xla: Yeah, as much as I agree with the approach -- it will divert from this change-set. Already got feedback that this PR is largish and not easy to review. We should feel safe with #908 being there to keep us honest.

proxy/coco/src/peer/run_state.rs
NunoAlexandre: Thanks for the insight! Yeah, this seems like the right thing to do then :)

proxy/coco/src/peer/run_state.rs
NunoAlexandre: the type in the same would only be coincidental but this is good as is :+1:

NunoAlexandre (approved):
This work deserves a great v60 cup :100: :coffee:

proxy/coco/tests/replication.rs
FintanH: :+1:

proxy/coco/tests/replication.rs
FintanH: fyi: I'd really like to see this test. Are we going to cover it in another PR or are you going to cover it here? It's what I'm holding out on before I give a :+1: :)

xla: It's under way, but quite involved. Our test suite is not at a point where these kind of scenarios can be expressed concisely, as a result the test code feels more like a liability than a confidence improvement.

FintanH: Ok, maybe we should tackle it as a next step then?

proxy/api/src/main.rs
```diff
@@ -79,14 +79,22 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        peer.run().await.expect("peer run loop crashed");
+        peer.run(
+            state,
+            store,
+            RunConfig {
+                sync_on_startup: true,
+            },
+        )
+        .await
+        .expect("peer run loop crashed");
```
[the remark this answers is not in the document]
kim: Does `run` actually ever return the `Error` it says it may return in the type signature?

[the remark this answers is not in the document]
xla: Good catch, it used. But in the latest iteration it never crashes just return once all channels and streams are closed.

proxy/coco/tests/replication.rs
xla: https://github.com/radicle-dev/radicle-upstream/issues/908#issuecomment-696020816

FintanH (approved):

NunoAlexandre (approved):
:cactus:

PR #901, opened 2020-09-14
feat(ui): show app version

rudolfs:
As specified in: https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=4147%3A7518.

Screenshot 2020-09-14 at 18 08 34

### DIFF
--- a/native/ipc.js
+++ b/native/ipc.js
@@ -1,5 +1,6 @@
-export const DIALOG_SHOWOPENDIALOG = "IPC_DIALOG_SHOWOPENDIALOG";
 export const CLIPBOARD_WRITETEXT = "IPC_CLIPBOARD_WRITETEXT";
+export const DIALOG_SHOWOPENDIALOG = "IPC_DIALOG_SHOWOPENDIALOG";
+export const GET_VERSION = "GET_VERSION";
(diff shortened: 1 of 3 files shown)


native/ipc.js
```diff
@@ -9,6 +10,8 @@ export const OPEN_PATH = "IPC_OPEN_PATH";
 //
 // The workaround is to use the electron native open dialog. As a bonus we
 // can configure it to allow users to create new directories.
+export const getVersion = () => window.electron.ipcRenderer.invoke(GET_VERSION);
```
xla, line 13: It looks like this function got in between the comment above that belongs to the function below.

xla (changes requested):
Dope! Just one mix-up.

native/ipc.js
rudolfs: Good catch!

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/901/commits/e85f82566712684a9e0f32e5e16c531a48c11192

xla (approved):
🍬 🔍 🛡 🌀

PR #903, opened 2020-09-15
feat(proxy): waiting room requests

FintanH:
:construction: Under Construction :construction: 

- [x] Use real `PeerId`
- [x] Use the real `Repo`
- [x] Test `next` under different strategies
- [x] Consider pushing the intermediate `Requested` states, i.e. `Found`, `Cloning`, and `Cloned` into a type parameter as well ala this [pattern](https://hoverbear.org/blog/rust-state-machine-pattern/)
- [x] De/serialise implementations
- [x] Docs

Part of ##803


proxy/coco/src/request.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+}
+
+impl<S, T> Request<S, T> {
+    fn coerce<R>(self, timestamp: T) -> Request<R, T> {
+        Request {
+            urn: self.urn,
+            state: PhantomData,
+            found_peers: self.found_peers,
```
xla, line 103: This answers my question above, we should start think about how we track which peer we cloned from as well and mark them when cloning failed.

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,303 @@
...
+            |request| match request {
+                SomeRequest::Requested(request) => Some(request),
+                _ => None,
+            },
+            |prev| Ok(prev.cancel(timestamp)),
+            urn,
+        )
+    }
+
+    pub fn timed_out(&mut self, urn: &RadUrn, timestamp: T) -> Result<Request<TimedOut, T>, Error>
```
xla, line 200: Doesn't seem right that someone is calling this, the waiting room should know what requests are timed out based on the heuristics of `MAX_QUERIES`/`MAX_CLONES`.

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,303 @@
...
+            |request| match request {
+                SomeRequest::Created(request) => Some(request),
+                _ => None,
+            },
+            |previous| Ok(previous.request(timestamp)),
+            urn,
+        )
+    }
+
+    pub fn attempted(
```
xla, line 111: As we refer to attempts both for queries and clones, this name might be bettah.

```suggestion
    pub fn queried(
```

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,303 @@
...
+            |prev| Ok(prev.timed_out(timestamp)),
+            urn,
+        )
+    }
+
+    pub fn list(&self) -> impl Iterator<Item = &RadUrn> {
+        self.requests.keys()
+    }
+
+    pub fn next<R: Rng>(&self, strategy: Strategy<R>) -> Option<&Request<Created, T>>
```
xla, line 218: We probably want this kind of method for candidates to be cloned as well.

xla (changes requested):
Dope! Had a first pass.

It seems there is still some gap when it comes to transition a request during clone attemtps, where the order would be: `Found -> Cloning -> CloningFailed/TimedOut -> Cloning` or `Found -> Cloning -> Cloned`. We want to make explicit when a request is being cloned to avoid scheduling it for anohter clone. Furthermore we should get our timeouts looked at, as we have the timeouts for a single operation and the terminal state of the request alltogether having timed out.

The dependency direction seems skewed with the waiting room depending on its super while being under the request module hierarchy. Maybe both request and waiting_room can be next to each other.

In its current form are we supporting repo requests for any repo not only projects? Or did I miss-understand the current imimplementation?

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,240 @@
...
+use std::convert::TryFrom;
+use std::marker::PhantomData;
+use std::time::Duration;
+
+use librad::net::peer::types::Gossip;
+use librad::uri::RadUrn;
+
+pub mod waiting_room;
+
+type PeerId = String;
```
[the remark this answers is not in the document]
FintanH: Right, I'll stick that as a TODO. I just wanted ease of making the POC

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,240 @@
...
+
+#[derive(Clone, Debug, Hash, PartialEq)]
+pub struct Repo {
+    urn: RadUrn,
+    name: String,
```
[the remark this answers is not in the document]
FintanH: I'll use the real `Repo` to :smile:

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,240 @@
...
+const PERIOD: Duration = Duration::from_secs(1); // Not for the whole request but for re-request
+
+pub fn exponential_backoff(attempts: Attempts, interval: Duration) -> Duration {
+    let exp = u32::try_from(attempts.queries + attempts.clones).unwrap_or(u32::MAX);
+    Duration::from_millis(u64::pow(2, exp)) + interval
+}
+
+#[derive(Clone, Debug, Hash, PartialEq)]
+pub enum State {
+    Found(PeerId),
```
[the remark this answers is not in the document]
FintanH: They're tracked in the `Request` remember?

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,240 @@
...
+    urn: RadUrn,
+    name: String,
+}
+
+const MAX_QUERIES: usize = 1;
+const MAX_CLONES: usize = 1;
+const PERIOD: Duration = Duration::from_secs(1); // Not for the whole request but for re-request
+
+pub fn exponential_backoff(attempts: Attempts, interval: Duration) -> Duration {
+    let exp = u32::try_from(attempts.queries + attempts.clones).unwrap_or(u32::MAX);
```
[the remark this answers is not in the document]
FintanH: > current in process operation

What do you mean by this?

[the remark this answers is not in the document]
xla: We either in the process of querying or cloning, but we calculate the `exp` for the sum of these. My argument is that cloning shouldn't be backing off further only becaue the querying went through more attempts. It should be separate.

proxy/coco/src/request.rs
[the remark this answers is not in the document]
xla: Do we need this `PeerId` in the variant then? Or differently what is it used for?

NunoAlexandre:
Would love to review this once ready :100:

proxy/coco/src/request.rs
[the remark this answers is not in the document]
FintanH: Got ya :+1:

xla (commented):
:eyes:

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,198 @@
...
+        S: TimeOut,
+    {
+        if self.attempts.queries > max_queries {
+            Either::Right(Request {
+                urn: self.urn,
+                attempts: self.attempts,
+                timestamp,
+                state: self.state.time_out(Kind::Query),
+            })
+        } else if self.attempts.queries > max_clones {
```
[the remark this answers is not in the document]
FintanH: Might make newtypes while I'm here :joy:

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,198 @@
...
+    {
+        Request {
+            urn: self.urn,
+            attempts: self.attempts,
+            timestamp,
+            state: self.state.cancel(),
+        }
+    }
+
+    pub fn found_peer(mut self, peer_id: PeerId, timestamp: T) -> Request<S, T>
```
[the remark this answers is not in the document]
FintanH: I wouldn't think so. The request is in a terminal state in all those states, so really there should be no more operations happening.

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,198 @@
...
+    {
+        self.state
+            .peers()
+            .entry(peer_id)
+            .or_insert(Status::Available);
+        self.timestamp = timestamp;
+        self
+    }
+
+    pub fn timed_out(
```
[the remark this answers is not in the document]
FintanH: Ya, `timed_out` isn't in its TRUE FORM yet. I'm still thinking about what would be the nicest semantics with your above recommendation in mind :)

proxy/coco/src/request/waiting_room.rs
```diff
@@ -119,16 +119,42 @@ impl<T> WaitingRoom<T> {
...
                 _ => None,
             },
-            |previous| previous.queried(timestamp),
+            |previous| previous.queried(max_queries, max_clones, timestamp),
+            urn,
+        )
+    }
+
+    pub fn queried_found(
+        &mut self,
```
xla, line 143: This one and the method above seem to assume knowledge on the caller side that implies state being kept per request. Rather whenever a `Have` is observed the caller should call `found`, if there is no request for it -- it's a noop otherwise the waiting room advances the request.

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,296 @@
...
+        Request {
+            urn: self.urn,
+            attempts: self.attempts,
+            timestamp,
+            state: Found { peers },
+        }
+    }
+}
+
+// TODO(finto): I think we need a state to transition back to `IsRequested` if there's no peers
```
FintanH, line 221: I think this is all that's left todo.

proxy/coco/src/request/existential.rs
```diff
@@ -0,0 +1,177 @@
+// I reserve the right to not match all the arms when picking out particular cases, thank you very
```
rudolfs, line 1: Why is this module called "existential"?

FintanH: The request (temporarily) forgets what state it's in and becomes `SomeRequest`

xla: > **Existentialism** (/ˌɛɡzɪˈstɛnʃəlɪzəm/ or /ˌɛksəˈstɛntʃəˌlɪzəm/) is a form of philosophical enquiry that explores the nature of existence by emphasizing experience of the human subject—not merely the thinking subject, but the acting, feeling, living human individual.

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,311 @@
...
+                timestamp,
+                state: self.state.time_out(TimedOut::Clone),
+            })
+        } else {
+            self.timestamp = timestamp;
+            Either::Left(self)
+        }
+    }
+
+    /// When a `Request` is queried it we increment the `queries` count -- tracked via the
```
sarahscott, line 154:
```suggestion
    /// When a `Request` is queried, we increment the `queries` count -- tracked via the
```

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,311 @@
...
+    /// `attempts` of the `Request`. If incrementing this count makes it exceed the maximum then
+    /// the `Request` transitions into the `TimedOut` out state.
```
sarahscott, line 156:
```suggestion
    /// the `Request` transitions into the `TimedOut` state.
```

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+}
+
+impl<T> Request<Cloning, T> {
+    /// Transition from the `Cloning` state back to the `Found` state.
+    ///
+    /// This signifies that the `peer` failed to clone the identity and we mark it as failed.
+    pub fn failed(self, peer: PeerId, timestamp: T) -> Request<Found, T> {
+        let mut peers = self.state.peers;
+        // TODO(finto): It's weird if it didn't exist but buh
```
sarahscott, line 276: 😂  How does one resolve this `TODO`?

proxy/coco/src/request/existential.rs
```diff
@@ -0,0 +1,177 @@
...
+#![allow(clippy::wildcard_enum_match_arm)]
+
+use serde::{Deserialize, Serialize};
+
+use super::{
+    Cloned, Clones, Cloning, Either, Found, IsCanceled, IsCreated, IsRequested, Queries, RadUrn,
+    Request, RequestKind, TimedOut,
+};
+
+/// Since a `Request` is parameterised over its state, it makes it difficult to talk a `Request` in
```
sarahscott, line 12:
```suggestion
/// Since a `Request` is parameterised over its state, it makes it difficult to talk about a `Request` in
```

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+pub struct WaitingRoom<T> {
+    /// The set of requests keyed by their `RadUrn`. This helps us keep only unique requests in the
+    /// waiting room.
+    requests: HashMap<RadUrn, SomeRequest<T>>,
+
+    /// The configuration of the waiting room.
+    config: Config,
+}
+
+/// The `Conig` for the waiting room tells it what are the maximum number of query and clone
```
sarahscott, line 75:
```suggestion
/// The `Config` for the waiting room tells it what are the maximum number of query and clone
```

sarahscott (commented):
Epic, thorough, meticulously-typed Waiting Room, bravo!

proxy/coco/src/request/states.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+impl HasPeers for Cloning {
+    fn peers(&mut self) -> &mut HashMap<PeerId, Status> {
+        &mut self.peers
+    }
+}
+
+/// If a state type implements this trait it means that their is a valid transition from that state
```
rudolfs, line 270: Did you mean "this"?
```suggestion
/// If a state type implements this trait it means that this is a valid transition from that state
```

proxy/coco/src/request/states.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+impl Cancel for IsCreated {}
+impl Cancel for IsRequested {}
+impl Cancel for Found {}
+impl Cancel for Cloning {}
+impl Cancel for IsCanceled {}
+
+/// If a state type implements this trait it means that their is a valid transition from that state
```
rudolfs, line 291: Also here, what did you mean by "their"?

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+    /// When looking up a `RadUrn` in the [`WaitingRoom`] it was missing.
+    #[error("the URN '{0}' was not found in the waiting room")]
+    MissingUrn(RadUrn),
+
+    /// When performing an operation on the a [`Request`] in the [`WaitingRoom`] it was found to be
+    /// in the wrong state for the desired operation.
+    ///
+    /// For example, if we tried to call `cloning` on a request that has only been created then
+    /// this would be an invalid transition.
+    #[error("the state fetched '{0}' from the waiting room was not one of the expected state")]
```
rudolfs, line 30:
```suggestion
    #[error("the state fetched '{0}' from the waiting room was not one of the expected states")]
```

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+    /// Initialise a new `WaitingRoom` with the supplied `config`.
+    #[must_use]
+    pub fn new(config: Config) -> Self {
+        Self {
+            requests: HashMap::new(),
+            config,
+        }
+    }
+
+    /// Get the underlying [`SomeRequest`] for the given `urn.
```
rudolfs, line 144:
```suggestion
    /// Get the underlying [`SomeRequest`] for the given `urn`.
```

rudolfs (approved):
Nice state machine!

proxy/coco/src/request.rs
FintanH: Probably with `dd` in vim :shushing_face:

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+    #[must_use]
+    pub fn get(&self, urn: &RadUrn) -> Option<&SomeRequest<T>> {
+        self.requests.get(urn)
+    }
+
+    /// Create a fresh [`Request`] with the given `urn`.
+    ///
+    /// If the `Request` already existed in the `WaitingRoom` then we get back the original request
+    /// as `Some`. This means we did nothing with the `urn` or `timestamp`.
+    pub fn create(&mut self, urn: RadUrn, timestamp: T) -> Option<SomeRequest<T>>
```
xla, line 156:
```suggestion
    pub fn request(&mut self, urn: RadUrn, timestamp: T) -> Option<SomeRequest<T>>
```

Communicates more clearly: waiting_room request urn

proxy/coco/src/request/states.rs
xla:
```suggestion
/// If a state type implements this trait it means that there is a valid transition from that state
```

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+            },
+            urn,
+        )
+    }
+
+    /// Tell the `WaitingRoom` that we failed the attempt to clone from the `peer` for the given
+    /// `urn`.
+    ///
+    /// If the underlying `Request` was in the `Found` state then it will transition to the
+    /// `Cloning` state.
```
MeBrei, line 305:
```suggestion
    /// If the underlying `Request` was in the `Cloning` state then it will transition to the
    /// `IsRequested ` state.
```

proxy/coco/src/request/states.rs
FintanH: I meant "there". How embarrassing :flushed:

proxy/coco/src/request/states.rs
FintanH: copy-pasta

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+    ///
+    /// # Errors
+    ///
+    ///   * If the `urn` was not in the `WaitingRoom`.
+    ///   * If the underlying `Request` was not in the expected state.
+    pub fn failed(&mut self, peer: PeerId, urn: &RadUrn, timestamp: T) -> Result<(), Error>
```
xla, line 311:
```suggestion
    pub fn failed(&mut self, peer: PeerId, urn: &RadUrn, timestamp: T) -> Result<(), Error>
```

Might be a bit ambiguous otherwise.

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+    #[test]
+    fn cannot_create_twice() {
+        let mut waiting_room: WaitingRoom<()> = WaitingRoom::new(Config::default());
+        let urn: RadUrn = "rad:git:hwd1yre85ddm5ruz4kgqppdtdgqgqr4wjy3fmskgebhpzwcxshei7d4ouwe"
+            .parse()
+            .expect("failed to parse the urn");
+        waiting_room.create(urn.clone(), ());
+        let request = waiting_room.create(urn.clone(), ());
+
+        assert_eq!(request, Some(SomeRequest::Created(Request::new(urn, ()))));
```
MeBrei, line 497: Doesn't this only prove that `request` was "successful" in creating the request?

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+                SomeRequest::Created(request) => Some(request),
+                _ => None,
+            },
+            strategy,
+        )
+    }
+
+    /// Return a `Request` that is in the `Cloned` state, if any, based off of the supplied
+    /// `strategy`.
+    pub fn ready<R: Rng>(&self, strategy: Strategy<R>) -> Option<&Request<Cloned, T>>
```
xla, line 403: What's the reason to have `ready` and `next`?

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,700 @@
...
+    ///   * If the underlying `Request` was not in the expected state.
+    pub fn canceled(&mut self, urn: &RadUrn, timestamp: T) -> Result<(), Error>
+    where
+        T: Clone,
+    {
+        self.transition(|request| request.cancel(timestamp).right(), Ok, urn)
+    }
+
+    /// Return the list of all `RadUrn`s in the `WaitingRoom`.
+    pub fn list(&self) -> impl Iterator<Item = &RadUrn> {
```
xla, line 365: Why not list the entire `Request`?

xla (changes requested):
First pass, focusing on the waiting room semantics.

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,311 @@
...
+pub mod waiting_room;
+
+/// Private trait for sealing the traits we use here.
+mod sealed;
+
+/// The maximum number of query attempts that can be made for a single request.
+const MAX_QUERIES: Queries = Queries::new(1);
+
+/// The maximum number of clone attempts that can be made for a single request.
+const MAX_CLONES: Clones = Clones::new(1);
```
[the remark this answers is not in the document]
FintanH: Ya, actually thought that too! Slipped my mind when it came to it

proxy/coco/src/request/waiting_room.rs
FintanH: I'm playing spot the difference here and I'm losing

proxy/coco/src/request/waiting_room.rs
FintanH: Can you elaborate on the question, please :)

proxy/coco/src/request/waiting_room.rs
FintanH: Fair :+1:

proxy/coco/src/request/waiting_room.rs
xla: To expand, probably two streams (in the general sense) are need, one that provides the next candidate to query and another which provides the next candidate to clone.

proxy/coco/src/request/waiting_room.rs
FintanH: Well... Does it? It hasn't been _requested_ it's just created and is _ready_ to be requested.

proxy/coco/src/request/waiting_room.rs
FintanH: Ah, haha thanks for the edit :+1:

proxy/coco/src/request/waiting_room.rs
FintanH: Noticed that myself :+1: Already resolved :)

proxy/coco/src/request/waiting_room.rs
FintanH: The returned request _is_ the original request. If it was a fresh request then it would have been `None`. I can add to the test though to move it into the next state and show we get back that request

proxy/coco/src/request/waiting_room.rs
```diff
@@ -0,0 +1,748 @@
...
+    /// `HasPeers::all_failed` method.
+    ///
+    /// If the underlying `Request` was in the `Found` state then it will transition to the
+    /// `IsRequested` state.
+    ///
+    /// # Errors
+    ///
+    ///   * If the `urn` was not in the `WaitingRoom`.
+    ///   * If the underlying `Request` was not in the expected state.
+    pub fn found_failed(&mut self, urn: &RadUrn, timestamp: T) -> Result<(), Error>
```
xla, line 293: This should be collapsable into `cloning_failed` if that would take the `PeerId`. Otherwise the question would rise how does the caller keep track of that all found peers have been exhausted.

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+}
+
+impl<T> Request<Cloning, T> {
+    /// Transition from the `Cloning` state back to the `Found` state.
+    ///
+    /// This signifies that the `peer` failed to clone the identity and we mark it as failed.
+    pub fn failed(self, peer: PeerId, timestamp: T) -> Request<Found, T> {
+        let mut peers = self.state.peers;
+        // TODO(finto): It's weird if it didn't exist but buh
```
xla, line 275:
```suggestion
        // FEELS(finto): It's weird if it didn't exist but buh
```

proxy/coco/src/request/waiting_room.rs
xla: The caller doesn't know if it hasn't been created yet, the signature encodes that as well, returning `None` if the urn is already associated with an request. Which could be read as `Noop`.

proxy/coco/src/request/waiting_room.rs
xla: Scratch that, totally miss-understood. My real question is, when is that method used?

proxy/coco/src/request.rs
FintanH: lmao

proxy/coco/src/request/waiting_room.rs
MeBrei: That would make it clearer!

proxy/coco/src/request/waiting_room.rs
FintanH: Ya, ok. It's a bit of grey area semantically for me because of the other state `IsRequested` but I can see your point. And since the next call on that request should be `queried` I'm happy to change this :+1:

xla: Ref. https://github.com/radicle-dev/radicle-upstream/pull/903#discussion_r493358673 to back the point.

proxy/coco/src/request/waiting_room.rs
FintanH: After all the effort I put into docs :(

```rust
    /// This should mean that the set of peers in the `Found` state have been exhausted and are all
    /// `Status::Failed`. It left up to the caller to determine this by using the
    /// `HasPeers::all_failed` method.
```

In other words, if we're in the `Found` state and have exhausted all our options then we need a way to fail back to `IsRequested` so that the caller can say `queried` again.

In terms of our actual use case, the `providers` stream has been emptied and none of the clone attempts worked so we need to call `providers` again to make another attempt.

FintanH: We discussed on a call and came to the conclusion that if the `Cloning` `failed` _and_ all the peers have failed then we can go back to `IsRequested`.

FintanH: As well as that, `IsRequested` should hold a `HashMap` of peers so that it doesn't forget what peers it has tried.

proxy/coco/src/request/states.rs
```diff
@@ -0,0 +1,344 @@
...
+
+use librad::{peer::PeerId, uri::RadUrn};
+
+use super::sealed;
+
+impl sealed::Sealed for IsCreated {}
+impl sealed::Sealed for Requested {}
+impl sealed::Sealed for Found {}
+impl sealed::Sealed for Cloning {}
+impl sealed::Sealed for IsCanceled {}
```
xla, line 20: For consistency should all `Is` prefixes be removed?

xla (dismissed):
♌️ 🎁 🕸 🏌

rudolfs (dismissed):

xla (approved):
(=´∀｀)

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,336 @@
...
+            origin: None,
+        }
+    }
+}
+
+impl<S, T> Request<S, T> {
+    /// Get the [`RadUrn`] that this `Request` is searching for.
+    pub const fn urn(&self) -> &RadUrn {
+        &self.urn
+    }
```
NunoAlexandre, line 92: TIL on `pub const`

Out of curiosity, why having this getter instead of making `urn` public?

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,336 @@
...
+    {
+        timestamp - self.timestamp.clone()
+    }
+
+    /// Transition this `Request` into an `Cancelled` state. We can only transition a particular
+    /// subset of the states which are: `{Created, Requested, Found, Cloning, Cancelled}`.
+    ///
+    /// That is, attempting to cancel a `Cloned` `Request` is not permitted and will complain at
+    /// compile time.
+    pub fn cancel(self, timestamp: T) -> Request<Cancelled, T>
```
NunoAlexandre, line 110: I am not understanding why we need to pass a `timestamp` when we cancel a request. Is it so to capture when the request was cancelled?

proxy/coco/src/request.rs
```diff
@@ -0,0 +1,336 @@
...
+    {
+        self.state.peers().entry(peer).or_insert(Status::Available);
+        self.timestamp = timestamp;
+        self
+    }
+
+    /// A `Request` transitions into a timed out state if it exceeds the maximum number of queries
+    /// or maximum number of clones. Otherwise, the `Request` proceeds as normal.
+    ///
+    /// The subset of states that can transition to the `TimedOut` out state consist of
```
NunoAlexandre, line 136:
```suggestion
    /// The subset of states that can transition to the `TimedOut` state consist of
```

proxy/coco/src/request/existential.rs
```diff
@@ -0,0 +1,199 @@
...
+/// Since a `Request` is parameterised over its state, it makes it difficult to talk about a
+/// `Request` in general without the compiler complaining at us. For example, we cannot have
+/// something like `vec![created, requested, cloning, timedout]` since they all have distinct types
+/// where they differ in states.
+///
+/// To allow us to do this we unify all the states into `SomeRequest` where each state is a variant
+/// in the enumeration.
+///
+/// When we pattern match we get back the request parameterised over the specific state and can
+/// work in a type safe manner with this request.
```
NunoAlexandre, line 26: nice docs :ok_hand:

proxy/coco/src/request/existential.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+impl<T> From<Request<TimedOut, T>> for SomeRequest<T> {
+    fn from(request: Request<TimedOut, T>) -> Self {
+        Self::TimedOut(request)
+    }
+}
+
+impl<T, L: Into<SomeRequest<T>>, R: Into<SomeRequest<T>>> From<Either<L, R>> for SomeRequest<T> {
```
NunoAlexandre, line 108: :mag: :book: :notebook:

proxy/coco/src/request/states.rs
```diff
@@ -0,0 +1,340 @@
...
+impl sealed::Sealed for Found {}
+impl sealed::Sealed for Cloning {}
+impl sealed::Sealed for Cancelled {}
+
+// State Types
+
+/// An enumeration of the different states a `Request` can be in. This is useful if we want to
+/// convey the state information without any of the other state data.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum RequestKind {
```
NunoAlexandre, line 26:
```suggestion
pub enum RequestState {
``` 

would be truer to its meaning and purpose.

proxy/coco/src/request/states.rs
```diff
@@ -0,0 +1,340 @@
...
+impl fmt::Display for RequestKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+/// The initial state for a `Request`. It has simply been created.
+#[derive(Clone, Debug, Hash, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// The initial state for a `Request`. It has simply been created.
```
NunoAlexandre, line 58:
```suggestion
```

proxy/coco/src/request/states.rs
```diff
@@ -0,0 +1,340 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct Requested {
+    /// A set of found peers and the lifecycle of clone attempts made on those peers.
+    pub(super) peers: HashMap<PeerId, Status>,
+}
+
+/// `Status` represents the lifecycle of a clone attempt, when paired with a `PeerId`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum Status {
```
NunoAlexandre, line 72:
```suggestion
pub enum CloneStatus {
```

The namespace doesn't specify that it's related to "Clone"ing so I'd prefer to see it in the type.

proxy/coco/src/request/states.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn add(self, other: usize) -> Self::Output {
+        Self(self.0 + other)
+    }
+}
+
+impl AddAssign<usize> for Queries {
+    fn add_assign(&mut self, other: usize) {
+        *self = Self(self.0 + other)
+    }
+}
```
NunoAlexandre, line 194: cool

NunoAlexandre (approved):
Excellent work, @FintanH :clap: 

left some comments but wouldn't mind if you skip them.

rudolfs (approved):

proxy/coco/src/request.rs
FintanH: I think it'd be disastrous if a user could change the `RadUrn` of the request in the `WaitingRoom`. #eNcApSuLaTiOn

proxy/coco/src/request.rs
FintanH: You actually understood correctly :) The timestamp is to record the last action. Which plays nicely with `filter` for getting items that may have elapsed some `delta`

proxy/coco/src/request.rs
FintanH: Argghhh!!! Thank you :blush:

proxy/coco/src/request/existential.rs
FintanH: Merci :flag-fr:

proxy/coco/src/request/existential.rs
FintanH: Would you like some explanation? :)

proxy/coco/src/request/states.rs
FintanH: Ya, I'm in agreement with this :+1:

proxy/coco/src/request/states.rs
FintanH: It's also not necessarily specific to cloning. This data could represent any kind of process that has three states. I suppose the only weird thing is that it doesn't have a `Done` since that's handled by a Request's state transition.

That is to say, I understand where the suggestion is coming from but disagree that it should be done.

proxy/coco/src/request/states.rs
FintanH: :sunglasses:

proxy/coco/src/request.rs
NunoAlexandre: would that require it to be `mut`?

proxy/coco/src/request/states.rs
NunoAlexandre: > It's also not necessarily specific to cloning

Hm, the docs suggest otherwise:

> `Status` represents the lifecycle of a clone attempt

proxy/coco/src/request.rs
FintanH: Ya, it would I believe.

proxy/coco/src/request/states.rs
FintanH: touché! I shall change the docs

NunoAlexandre (approved):
:fire:

xla (approved):
🕓 ⚔ 🍻 🌙

rudolfs (approved):

Issue #900, opened 2020-09-14
App opens in onboarding on every app start


rudolfs:
Steps to reproduce:

1. start app
2. create identity and complete onboarding
3. close the app
4. start the app again

Expected outcome: the app opens in the profile view without going through onboarding again.

The identity still exists, but the session endpoint always returns an empty identity on app start:
```json
{
  "identity": null,
  "settings": {
    "appearance": {
      "theme": "light",
      "hints": {
        "showRemoteHelper": true
      }
    },
    "coco": {
      "seeds": []
    }
  }
}
```

Seems like this regression was introduced in: https://github.com/radicle-dev/radicle-upstream/pull/881. Other settings (like theme) aren't persisted either.

xla:
Tracked down the issue being introduced in this commit: https://github.com/zshipko/rust-kv/commit/9d208d4be60b99a7ed9f505b36d325c58ce74dca

Going to prepare a patch to integrate on a working commit and look into how to report this meaningfully to the upstrea project.

rudolfs:
Fixed in #907.

xla:
I left this open as the underlying issue wasn't addressed and we only landed a stop-gap.

juliendonck:
@xla check this out and clooooose please

geigerzaehler:
@xla: I’m closing this. Feel free to reopen or create a new issue with the underlying issue details if this is still relevant.

PR #907, opened 2020-09-15
fix(proxy): session not loaded on restart

xla:
A regression was introduced in our upstream dependency `kv` which breaks
the loading of our session from disk. Until we land a patch, this is the
stopgap solution.

Stop-gap for #900


proxy/api/Cargo.toml
```diff
@@ -15,7 +15,7 @@ default-run = "api"
 anyhow = "1.0"
 directories = "2.0"
 futures = { version = "0.3", features = [ "compat" ] }
-kv = { version = "0.22", features = [ "json-value" ] }
+# kv = { version = "0.21", features = [ "json-value" ] }
```
rudolfs, line 18: Should this be left here?

proxy/coco/Cargo.toml
```diff
@@ -12,7 +12,7 @@ repository = "git@github.com:radicle-dev/radicle-upstream.git"
 
 [dependencies]
 futures = { version = "0.3", features = [ "compat" ] }
-kv = { version = "0.22", features = [ "json-value" ] }
+# kv = { version = "0.21", features = [ "json-value" ] }
```
rudolfs, line 15: Ditto.

FintanH (approved):

NunoAlexandre (approved):
:paperclips:

rudolfs (approved):

PR #909, opened 2020-09-15
chore(ui): types for key components

sarahscott:
if the components & modules we use a lot (e.g. `svelte-spa-router`, `Remote`) are typed, it'll be a lot easier to write new components in typescript

- [x] `svelte-spa-router` (will consider fleshing the types out more if we indeed want to keep using this library)
- [x] `Remote.svelte`
- [x] `Profile.svelte` & children
- [x] All icons


ui/DesignSystem/Component/Remote.svelte
```diff
@@ -1,35 +1,46 @@
...
+  import * as remote from "../../src/remote";
 
   import WithContext from "./WithContext.svelte";
 
-  export let store = null;
-  export let context = null;
+  export let store: remote.Store<any>;
+  export let context: string | undefined = undefined;
+
+  // Shorthand for casting these states
```
sarahscott, line 11: Not the prettiest syntax, but a more elegant solution will require a refactor of `remote.ts`, which was built before we could use typescript in components.

xla: Can we capture a follow-up issue for that?

xla (approved):
Herculean feat! Very extra dope.

🕌 ⏏ 🈁 🔌

ui/DesignSystem/Primitive/Avatar.svelte
```diff
@@ -119,11 +122,7 @@
 
 <div data-cy={dataCy} class={`container ${size}`} {style}>
   {#if imageUrl}
-    <img
-      class={`image ${avatarClass}`}
-      src={imageUrl}
-      alt="user-avatar"
-      onerror="this.style.display='none'" />
```
rudolfs, line 126: Why aren't we keeping the `onerror`?

rudolfs (commented):
Looks great, learned a lot about typing stuff!

Added some questions clarification.
I think aside the stray console log and the org stuff in `Large.svelte` this should be mergeable.

ui/DesignSystem/Primitive/Emoji.svelte
```diff
@@ -1,10 +1,12 @@
...
+<script lang="ts">
   import twemoji from "twemoji";
 
-  export let emoji = null;
+  export let emoji: string;
 
-  export let size = "regular";
-  export let style = null;
+  type EmojiSize = "small" | "regular" | "medium" | "big" | "huge";
+  export let size: EmojiSize = "regular";
```
NunoAlexandre, line 7: does `size` here stand for default size?

NunoAlexandre (approved):
Excited to work more with types in the UI :tada: 

I second @rudolfs' comments, didn't find anything to add further than that.

ui/Screen/Profile.svelte
```diff
@@ -47,10 +47,11 @@
   ];
 
   const session = getContext("session");
+  console.log(session);
```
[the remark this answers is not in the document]
sarahscott: i think this is from an old commit 😉

ui/DesignSystem/Primitive/Avatar.svelte
sarahscott: Because it's [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#Browser_compatibility).

ui/DesignSystem/Component/Urn.svelte
```diff
@@ -1,13 +1,15 @@
...
   import { Icon } from "../Primitive";
 
+  export let urn: string;
+
   export let notificationText = "Copied to your clipboard";
+
   export let showOnHover = false;
-  export let showCopyOnlyOnHover = false;
+  export let showCopyOnlyOnHover: boolean = false;
   export let truncate = false;
```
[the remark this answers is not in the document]
sarahscott: Good catch! They're both optional boolean props so `truncate` definitely also needs the explicit type

ui/DesignSystem/Component/Header/Large.svelte
```diff
@@ -97,7 +92,7 @@
               style="display: flex; align-items: center;">
               {name}
             </h1>
-            {#if variant === 'org' || entity.registered}
+            {#if !isIdentity}
```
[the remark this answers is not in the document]
sarahscott: I parameterized more of the functionality so it can still be used for orgs as they used to be (RIP) but also for users 😊

ui/DesignSystem/Primitive/Emoji.svelte
sarahscott: The emoji has a default `size` of `regular` if no `size` prop is provided, but it can be changed to suit the spec like:
```typescript
    <Emoji emoji={'🎷'} size="huge" style="margin-bottom: 1.5rem;" />
```

ui/DesignSystem/Component/Remote.svelte
sarahscott: https://github.com/radicle-dev/radicle-upstream/issues/926

NunoAlexandre (approved):
:100:

xla (approved):
➿ 🍁 🐈 🔹

PR #913, opened 2020-09-16
feat(ui): introduce experimental flag

### LINKED ISSUE #888: Decouple the feature flag from the env
When starting the app with `yarn start` etc. it starts electron with `NODE_ENV=development`. The feature flag `isDev()` depends on this environment. 

To flexibly develop the app and see what features the production user will see, it should be possible to easily and independently set the feature flag and environment.

NunoAlexandre:
Closes #888 

This is a feature flag that is consumed by the app as `isExperimental`,
signalling whether we are running the app with the experimental mode on,
which toggles experimental/under construction features. This flag is
only allowed in development mode.
A new yarn command is added, `yarn start:public`, which starts the app
in public mode, i.e, hiding all the experimental features. Such run
simulates a production preview while running in development mode.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -101,6 +101,7 @@ Here is a list of the most commonly used ones:
 
 ```sh
 yarn start                  # Start Upstream in development mode
+yarn start:public           # Start Upstream in the public mode, hidding experimental features.
 
 yarn test                   # Run all ui tests
 yarn test:integration       # Run only integration tests
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -68,6 +68,20 @@ On Linux:
 3. Start Upstream in development mode: `yarn start`.
 
 
+### Feature flagging
+
+UI features that are experimental or under construction that find their way into the main
+branch must be placed behind the feature flag, to make them inaccessible for the general public.
+
+We do that by using `native > ipc.js > isExperimental` as a feature flag to enable or
+disable said features accordingly to the mode in which we are running the app.
+
+See the [scripts](#scripts) section below to learn which commands to use to toggle this flag
+accordingly to your current workflow.
+
+The feature flag is only available in development mode. It is always disabled in production.
+
+
 ### Running tests
 
 Before running UI end-to-end tests locally you'll need to check out the latest
@@ -101,7 +115,7 @@ Here is a list of the most commonly used ones:
 
 ```sh
 yarn start                  # Start Upstream in development mode
-yarn start:experimental           # Start Upstream in experimental mode, showing unfinished features
+yarn start:experimental     # Start Upstream in experimental mode, showing unfinished features
 
 yarn test                   # Run all ui tests
 yarn test:integration       # Run only integration tests
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -101,7 +101,7 @@ Here is a list of the most commonly used ones:
 
 ```sh
 yarn start                  # Start Upstream in development mode
-yarn start:public           # Start Upstream in the public mode, hidding experimental features.
+yarn start:experimental           # Start Upstream in experimental mode, showing unfinished features
 
 yarn test                   # Run all ui tests
 yarn test:integration       # Run only integration tests
--- a/native/ipc.js
+++ b/native/ipc.js
@@ -28,7 +28,7 @@ export const isDev = () => {
 
 // Informs whether it's running in experimental mode, where
 // features under construction are enabled and can thus be used.
-// This option can only be enaled iff `isDev()` as we should only
+// This option can only be enabled iff `isDev()` as we should only
 // want to toggle it while in development mode.
 export const isExperimental = () => {
   return isDev() && window.electron.isExperimental;
--- a/native/ipc.js
+++ b/native/ipc.js
@@ -21,6 +21,15 @@ export const copyToClipboard = text =>
 export const openPath = path =>
   window.electron.ipcRenderer.invoke(OPEN_PATH, path);
 
+// Informs whether it's running in a development environment.
 export const isDev = () => {
   return window.electron.isDev;
+};
+
+// Informs whether it's running in experimental mode, where
+// features under construction are enabled and can thus be used.
+// This option can only be enaled iff `isDev()` as we should only
+// want to toggle it while in development mode.
+export const isExperimental = () => {
+  return isDev() && window.electron.isExperimental;
 };
--- a/package.json
+++ b/package.json
@@ -106,7 +106,9 @@
     "wait-on": "^5.1.0"
   },
   "scripts": {
-    "start": "run-p --race svelte:watch proxy:start electron:start",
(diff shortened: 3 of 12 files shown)


native/ipc.js
```diff
@@ -21,6 +21,15 @@ export const copyToClipboard = text =>
...
   return window.electron.isDev;
 };
+
+// Informs whether it's running in experimental mode, where
+// features under construction are enabled and can thus be used.
+// This option can only be enaled iff `isDev()` as we should only
+// want to toggle it while in development mode.
+export const isExperimental = () => {
+  return isDev() && window.electron.isExperimental;
+};
```
NunoAlexandre, line 35: Happy to reconsider if others find this strictness unhelpful.

package.json
```diff
@@ -115,7 +117,7 @@
     "test:unit:watch": "jest --watchAll",
     "wait:test": "wait-on tcp:8080 && yarn svelte:build && yarn cypress:run",
     "wait:debug": "wait-on tcp:8080 && yarn cypress:open",
-    "dist": "rm -rf ./dist && mkdir ./dist && yarn proxy:clean && yarn svelte:clean && yarn svelte:build && yarn proxy:build:release && cp proxy/target/release/api dist/proxy && cp proxy/target/release/git-remote-rad dist && electron-builder --publish never",
+    "dist": "rm -rf ./dist && mkdir ./dist && yarn proxy:clean && yarn svelte:clean && RADICLE_UPSTREAM_EXPERIMENTAL=false NODE_ENV=production yarn svelte:build && yarn proxy:build:release && cp proxy/target/release/api dist/proxy && cp proxy/target/release/git-remote-rad dist && electron-builder --publish never",
```
rudolfs, line 120: This command is used to build a binary package, so these env variables don't make much sense here.
A user will later take the generated package and install it on their system, after that they would run it from the GUI (by double clicking), without passing any env variables by default.

When running the app from a package electron automatically sets the `NODE_ENV` to `production`. This is how `isDev` works.

package.json
```diff
@@ -106,7 +106,9 @@
     "wait-on": "^5.1.0"
   },
   "scripts": {
-    "start": "run-p --race svelte:watch proxy:start electron:start",
+    "start": "RADICLE_UPSTREAM_EXPERIMENTAL=true start:app",
+    "start:public": "RADICLE_UPSTREAM_EXPERIMENTAL=false start:app",
```
rudolfs, line 110:
```suggestion
    "start": "RADICLE_UPSTREAM_EXPERIMENTAL=false start:app",
    "start:experimental": "RADICLE_UPSTREAM_EXPERIMENTAL=true start:app",
```
As a dev I'm used to run `yarn start` when working with the app. I think the initial idea was that by default we devs should see what a normal user would see. If we're working on new features we should be forced to use env variables to enable these unfinished features.

Maybe @cloudhead can confirm if I understood his reasoning correctly?

DEVELOPMENT.md
```diff
@@ -101,6 +101,7 @@ Here is a list of the most commonly used ones:
 
 ```sh
 yarn start                  # Start Upstream in development mode
+yarn start:public           # Start Upstream in the public mode, hidding experimental features.
```
rudolfs, line 104: See comment in `package.json`...

```suggestion
yarn start:experimental
```

package.json
NunoAlexandre: Right. I have worked in other projects that were bundled and took the relevant environment variables into account, I thought the same would happen here.

 > When running the app from a package electron automatically sets the NODE_ENV to production. This is how isDev works.

I see, so `NODE_ENV` is a `electron`-reserved variable. I thought it was ours and was going to suggest namespace it better.

package.json
NunoAlexandre: I don't think it makes much sense to have the default developer workflow not being focus on active development. The use case raised by Merle is that she wanted to run the app and see what the general public will see without going on changing unrelated environment variables, which `yarn start:public` is dedicated for. So the default dev workflow is `yarn start` which shows it all as it makes sense to be in development, and `yarn start:public` to have a public preview of the app.

xla (dismissed):
It still uses `isDev` for the checks, does that not go against the premise set in the issue?

NunoAlexandre:
> It still uses `isDev` for the checks, does that not go against the premise set in the issue?

Which checks are you referring to? AFAIK the relevant checks for feature flagging are now using `isExperimental`.

package.json
NunoAlexandre: Fixed, thanks.

xla:
> Which checks are you referring to? AFAIK the relevant checks for feature flagging are now using `isExperimental`.

Which in turn uses `isDev` if I read the code correctly.

NunoAlexandre:
> > Which checks are you referring to? AFAIK the relevant checks for feature flagging are now using `isExperimental`.
> 
> Which in turn uses `isDev` if I read the code correctly.

That's not telling the whole picture. The original issue here is that we want to be able to have a preview of the app as it will look like to the general public (i.e., without the flagged experimental features) without changing the `NODE_ENV` whose purpose is not to feature flag. 

The code you read says that we run in experimental mode `if` we are in development mode `&&` the experimental flag is enabled. That's just a safety measure to not have the experimental flag turned on in production. So, as for the use cases that this feature concerns with, we are covered. As I noted, I'm happy to reconsider this strict check tho.

package.json
NunoAlexandre: > I think it makes sense to be working with what the user will see by default, because it makes it harder to miss things that we end up shipping.

Cool, I actually really like that :) I had misuderstood the poit initially. Thank you both, will adust!

native/ipc.js
```diff
@@ -21,6 +21,15 @@ export const copyToClipboard = text =>
...
   window.electron.ipcRenderer.invoke(OPEN_PATH, path);
 
+// Informs whether it's running in a development environment.
 export const isDev = () => {
   return window.electron.isDev;
 };
+
+// Informs whether it's running in experimental mode, where
+// features under construction are enabled and can thus be used.
+// This option can only be enaled iff `isDev()` as we should only
```
sarahscott, line 31:
```suggestion
// This option can only be enabled iff `isDev()` as we should only
```

sarahscott:
Just so I know how to use these going forward, in which instances would I use `isDev` instead of `isExperimental`?

native/ipc.js
NunoAlexandre: sounds like inhale, what was I thinking about :mushroom: #iamamsterdam

NunoAlexandre:
> Just so I know how to use these going forward, in which instances would I use `isDev` instead of `isExperimental`?

Does the docs added here help with that question? https://github.com/radicle-dev/radicle-upstream/pull/913/files#diff-5a0360ca6ab64541cfe0d4d1131a943a

If not, they need to be improved, or we need to further document this stuff in `DEVELOPMENT.md` I'd say. 

Good to have your input :)

rudolfs (approved):

MeBrei (approved):

PR #914, opened 2020-09-16
feat(ui): add input field hint

### LINKED ISSUE #898: Add ⌘+V box in the search box
Oh and the symbol should update based on the OS, like in the keyboard shortcuts modal.

From the design: 
image

juliendonck:
closes #898 

Onboarding


ui/DesignSystem/Primitive/Input/Text.svelte
```diff
@@ -156,6 +173,8 @@
       <div class="validation-row">
         <p>{validation.message}</p>
       </div>
+    {:else if hint !== null && (value === '' || value === null)}
+      <p class="typo-text hint">{hint}</p>
```
juliendonck, line 177: This duplicated what I've done on [line 156](https://github.com/radicle-dev/radicle-upstream/pull/914/files#diff-10dae02f6dfcb8b7213057c861dde888R156). I didn't know how else to match that it's not being validated, without duplicating what's already on line 160 - 175. Open to suggestions

sarahscott: If we type this component, the truthiness could be easier to check, e.g.
```typescript
  export let hint = "";
  export let value = "";

  ...

  $: showHint = hint && !value.length && !validation
```

rudolfs:
- [ ] we don't require the `⌘`  modifier key here, `Enter` is enough.
Screenshot 2020-09-17 at 13 45 38

- [ ] regression: input field has red validation outline even though there are no validation errors
Screenshot 2020-09-17 at 13 46 35

- [ ] we should add the `Enter` hint to the password inputs as well
Screenshot 2020-09-17 at 13 47 43

rudolfs:
Also wondering whether we should rename `Press enter` to `↵`. Otherwise it's a bit confusing that the placeholder says `Enter a name` and the hint `Press enter`.

ui/DesignSystem/Primitive/Input/Text.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/914/commits/719f4ac0b9e8a9e1b1886073bf1c520b76fa5dc4

juliendonck:
> Also wondering whether we should rename `Press enter` to `↵`. Otherwise it's a bit confusing that the placeholder says `Enter a name` and the hint `Press enter`.

@rudolfs added the `↵`, had to do some magic. Don't think this is the most elegant solution. But don't know how else to do it.

ui/DesignSystem/Component/KeyHint.svelte
```diff
@@ -0,0 +1,33 @@
+<script>
```
sarahscott, line 1:
```suggestion
<script lang="ts">
```



get in comrade, we have types now
bday

ui/DesignSystem/Component/KeyHint.svelte
```diff
@@ -0,0 +1,33 @@
...
+  import { isMac } from "./../../src/settings.ts";
```
sarahscott, line 2:
```suggestion
  import { isMac } from "./../../src/settings";
```

ui/DesignSystem/Component/KeyHint.svelte
```diff
@@ -0,0 +1,33 @@
...
+
+  export let hint = null;
+  export let noModifier = false;
+  export let style = null;
```
sarahscott, line 6:
```suggestion
  export let hint = "";
  export let noModifier: boolean = false;
  export let style = "";
```

sarahscott: In case you're wondering why we have to have the `boolean` type annotation here and not for the `string`'s, it's an idiosyncrasy of svelte at this time and the compiler needs to know that it's a `boolean` instead of `false | undefined`. 

tldr; if it's an optional prop, we should specify a default value. If it's an optional prop that happens to be a `boolean`, we should also let the compiler know it's a `boolean`. If it's a required prop, e.g. something that the component should be passed every single time, we can just specify the type, à la:
```typescript
    export let alwaysProvideThisOrCompilerShouldComplain: string;
```

ui/DesignSystem/Component/KeyHint.svelte
```diff
@@ -0,0 +1,33 @@
...
+
+  const modifierKey = isMac ? "⌘" : "ctrl";
```
sarahscott, line 8: We don't have to do this now, but it might be nice in the future as we add more keyboard functionality: https://github.com/nfriend/ts-key-enum/blob/master/Key.enum.d.ts

ui/DesignSystem/Component/KeyHint.svelte
```diff
@@ -0,0 +1,33 @@
...
+    border-radius: 0.25rem;
+  }
+
+  .hint:first-child {
+    margin-right: 0.25rem;
+  }
+</style>
+
+<div class="hint-container" {style}>
+  {#if hint.length === 1 && !noModifier && hint !== '↵'}
```
sarahscott, line 29: I would probably have this validation in a variable:
```typescript
    const addModifier = hint.length === 1 && !noModifier && hint !== '↵`;
```

sarahscott (dismissed):
Super slick UX improvement ✨ 

There are keys with names of `length > 1` and different OS compatibilities to consider, so I've created a followup ticket here for dealing with keyboard-related complexity: https://github.com/radicle-dev/radicle-upstream/issues/927

That's for a new PR though - this one has my approval once the new component is typed.

rudolfs (dismissed):
Looks good.

ui/DesignSystem/Component/KeyHint.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/914/commits/4da650c03cb5846cdd55d435f473cb6dde0749fd

ui/DesignSystem/Component/KeyHint.svelte
juliendonck: yup, next iteration.

ui/DesignSystem/Component/KeyHint.svelte
juliendonck: Ok, that makes sense. I'm sure I'll have to go through the motions a couple of time to get it completely. Thanks for taking the time to explain it :)

ui/DesignSystem/Component/KeyHint.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/914/commits/4da650c03cb5846cdd55d435f473cb6dde0749fd

ui/DesignSystem/Component/KeyHint.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/914/commits/4da650c03cb5846cdd55d435f473cb6dde0749fd

ui/DesignSystem/Component/KeyHint.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/914/commits/4da650c03cb5846cdd55d435f473cb6dde0749fd

xla (approved):
🍣 🖊 🗂 🍰

rudolfs (approved):

Issue #917, opened 2020-09-17
Support link to irc channel is not working


rudolfs:
When clicking the link we get this error message in the console:
```
User tried opening URL with invalid URI scheme: irc://freenode:1/radicle
```

Refs: https://github.com/radicle-dev/radicle-upstream/issues/912.

<img width="586" alt="Screenshot 2020-09-17 at 11 26 39" src="https://user-images.githubusercontent.com/158411/93452425-b1554000-f8d8-11ea-8fce-a86e27cb336d.png">

PR #921, opened 2020-09-17
fix(ui): allow irc:// as external protocol

rudolfs:
If anyone wonders why we passlist: https://security.stackexchange.com/questions/225799/dangers-of-electrons-shell-openexternal-on-untrusted-content.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/917.


xla:
We might not need this at all as we are going to move away from IRC for our community. @abbey-titcomb Can shed some light.

abbey-titcomb:
Yeah, but we don't have the replacement spun up yet, so I don't see why we should keep thee broken link around. Can sub for Matrix chat once we have it figured out.

xla:
> Yeah, but we don't have the replacement spun up yet, so I don't see why we should keep thee broken link around. Can sub for Matrix chat once we have it figured out.

Because we have to remember to remove the exceptiosn added. Can we assume matrix lands before launch?

sarahscott (approved):
I think this makes sense for as long as we're using irc.

xla (approved):
🖖 ☃ 🕕 ✅

PR #922, opened 2020-09-17
docs: document release QA procedure

rudolfs:
This formalises our successful trial QA procedure run https://github.com/radicle-dev/radicle-upstream/issues/912.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/842.
Closes: https://github.com/radicle-dev/radicle-upstream/issues/910.

PR #941, opened 2020-09-21
docs: add additional explanations to QA checklist

rudolfs:
Follow-ups for: https://github.com/radicle-dev/radicle-upstream/pull/922.

### DIFF
--- a/QA.md
+++ b/QA.md
@@ -54,9 +54,13 @@
     - is at least 2 characters long
     - can contain the letters `a`-`z`, `A`-`Z`, digits `0`-`9` and special
       characters `-`, `_`
+    - all of the following examples should show a validation error:
+      `_john`, `x`, `linu$$$`, `-moira`
   - passphrase
     - is at least 4 characters long
     - the repeated passphrase is equal to the first one
+    - all of the following example pairs should show a validation error:
+      `1`/`1`, `12`/`12`, `123`/`123`, `supersecret`/`supersceret`
 - [ ] After passphrase input the identity is created and a copyable URN is
       provided
 - [ ] After completion we land on our profile page which contains a placeholder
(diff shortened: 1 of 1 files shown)


QA.md
```diff
@@ -49,6 +49,14 @@
 - [ ] It's possible to go to the next screen by pressing <kbd>enter</kbd> if
       the input field validations allow it
 - [ ] Handle and passphrase validations work
+  - handle
+    - starts with a letter or digit
+    - is at least 2 characters long
+    - can contain the letters `a`-`z`, `A`-`Z`, digits `0`-`9` and special
+      characters `-`, `_`
```
FintanH, line 56: While positive cases are good, I think negative cases should be here too. Even just examples the person could copy and paste into the field.

rudolfs: In my experience the best bugs pop up when people try to use software without any instructions, that's why I thought it might be valuable to let it open to interpretation.

rudolfs: But can also add a list of positive and negative examples if you still think it's valuable.

FintanH:
```
- [ ] Go wild!
```

:smile: I agree but we should also have consistent examples that we expect to fail/pass for quality assurance purposes :)

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/941/commits/e8e7097d8ad3e93bc99ce0b1df955427b736ff12

FintanH: And maybe for projects as well? :grimacing:

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/941/commits/b3669a283c3dc657f73a96920d626b7cfb9ed484

xla (approved):
🚋 🕗 🍌 📫

FintanH (approved):
Amazing :clap:

PR #944, opened 2020-09-21
refactor(proxy): move to coco pooled storage, cancellable async everywhere

kim:


proxy/coco/src/state.rs
```diff
@@ -97,8 +76,12 @@ impl State {
...
     ///
     ///   * Unable to acquire the lock.
-    pub fn has_commit(&self, urn: &RadUrn, oid: impl Into<git2::Oid>) -> Result<bool, Error> {
-        Ok(self.api.storage().has_commit(urn, oid.into())?)
+    pub async fn has_commit(&self, urn: RadUrn, oid: impl Into<git2::Oid>) -> Result<bool, Error> {
+        let oid = oid.into();
+        Ok(self
+            .api
+            .with_storage(move |storage| storage.has_commit(&urn, oid))
+            .await??)
```
kim, line 84: Before you ask: `??` is pronounced `The wat? operator`, was stabilised in Rust'98, and is part of the Rust Dawallah Level II certification programme.

kim:
This is unfortunately very invasive, as it changes virtually everything to `async`.

Note that resetting the on-disk state is removed entirely, as that was done in a way very likely to cause unexplicable bugs down the road -- both on-disk state and network stack must be managed in lockstep.

proxy/coco/src/state.rs
FintanH: I'm looking forward to the interrobang RFC. `?!` will be the `The WAT?! operator` and will indicate to code reviewers that they should move along

xla:
> This is unfortunately very invasive, as it changes virtually everything to `async`.

We don't accept any other non-invasive PRs.

> Note that resetting the on-disk state is removed entirely, as that was done in a way very likely to cause unexplicable bugs down the road -- both on-disk state and network stack must be managed in lockstep.

What do you propose in order to enable the UI integration tests to reset the states inbetween test runs?

kim:
> What do you propose in order to enable the UI integration tests to reset
> the states inbetween test runs?

I don't know, actually. A "safe" API which would take care of polling the
network future would probably require `async_closure`.

Is it absolutely necessary that the entire state is wiped for those tests? Like,
do they depend on the keypair being used? Or could they just randomise the
projects being created?

xla:
>  Is it absolutely necessary that the entire state is wiped for those tests? Like, do they depend on the keypair being used? Or could they just randomise the projects being created?

A lot of the tests depend on the fact that the monorepo is without an userr/owner -- mainly to test onboarding. To take a step back, what we want is to have state isolation between test runs to avoid spurious test failures from residue of other runs, maybe there is a saner way than resetting the entire state carried in the context.

xla:
To expand on this, the granularity between these tests varies but is mostly on the level of user flows closely modeled after figma user flows.

FintanH:
Do the frontend tests need to use the same proxy process or can they change between test runs? What I'm thinking is that a cypress test could consist of `proxy --root .tmp/some/dir`

xla:
> Do the frontend tests need to use the same proxy process or can they change between test runs? What I'm thinking is that a cypress test could consist of `proxy --root .tmp/some/dir`

AFAIK cypress doesn't control the proxy, maybe @rudolfs can shed some light.

xla:
It looks like `Context` needs to own the peer handling -- then whenever we reset all the things we start a new peer backed by the new monorepo. Unfortunately [`JoinHandle`](https://github.com/tokio-rs/tokio/issues/1830) is not canceling on drop. To remedy that I can see two strategies:

* wrap the `Peer::run` in an [`Abortable`](https://docs.rs/futures/0.3.5/futures/future/struct.Abortable.html) and the context holds onto the `AbortHandle`
* evaluate exposing cancellation in links `Peer`/`PeerApi` with a channel or other construct

@kim Is there any precedence for `RunLopp` cancellation in link?

rudolfs:
>> Do the frontend tests need to use the same proxy process or can they change between test runs? What I'm thinking is that a cypress test could consist of proxy --root .tmp/some/dir

>AFAIK cypress doesn't control the proxy, maybe @rudolfs can shed some light.

At the moment we start proxy first via a yarn task that also runs the tests. I think technically it would be possible to control the proxy lifecycle from within tests, but I wonder whether it would not slow down the test suite considerably because the proxy takes a while to start up?

kim:
Should be good now. Will add some more prose later.

proxy/coco/tests/gossip.rs
```diff
@@ -12,30 +12,33 @@ use common::{
...
+                interval: Duration::from_millis(100),
+            },
+            ..RunConfig::default()
+        },
+    )
+    .await?;
     let alice_events = alice_peer.subscribe();
 
-    tokio::task::spawn(alice_peer.run(alice_state.clone(), alice_store, RunConfig::default()));
+    let _alice_runs = alice_peer.into_running();
```
kim, line 33: Nb funny semantics: we don't actually need to `spawn` this, just prevent it from being dropped. It also doesn't _hurt_ to spawn it, though, so whatever people think is preferable in tests

proxy/api/src/main.rs
```diff
@@ -20,46 +32,121 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 
-    let paths_config = if args.test {
-        std::env::set_var("RAD_HOME", temp_dir.path());
-        coco::config::Paths::FromRoot(temp_dir.path().to_path_buf())
-    } else {
-        coco::config::Paths::default()
-    };
-    let paths = coco::Paths::try_from(paths_config)?;
+            Some(()) = sighup.recv() => {
+                log::info!("SIGHUP received, reloading...");
```
kim, line 59: To the SIG HUP you drop gitit?

Because with `--test`, everything is in a temp dir, this is equivalent to a reset. Without `--test`, it's equivalent to, well, restarting.

proxy/api/src/main.rs
```diff
@@ -70,67 +157,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         });
         let config = coco::config::configure(paths, key, *coco::config::LOCALHOST_ANY, seeds);
 
-        coco::into_peer_state(config, signer.clone()).await?
+        coco::into_peer_state(config, signer.clone(), store.clone(), RunConfig::default()).await?
     };
 
-    if args.test {
-        let state = state.lock().await;
-        // TODO(xla): Given that we have proper ownership and user handling in coco, we should
```
kim, line 78: It did not make any difference whether or not those were set up, so I removed it.

proxy/coco/src/peer.rs
```diff
@@ -71,100 +100,235 @@ impl Peer {
...
     }
 
-    /// Sync timeout subroutine.
-    fn start_sync_timeout(sync_period: Duration, sender: mpsc::Sender<TimeoutEvent>) {
-        tokio::spawn(async move {
-            tokio::time::delay_for(sync_period).await;
-            sender.clone().send(TimeoutEvent::SyncPeriod).await.ok();
-        });
-    }
+    announcer_handle.abort();
```
kim, line 298: Not entirely sure about this, actually. Could be we need to wake the tasks when the parent gets aborted.

proxy/coco/src/state.rs
```diff
@@ -328,12 +347,12 @@ impl State {
     /// * If no project for the `urn` was found.
     /// * If the [`git::Browser`] fails.
     /// * If the passed `callback` errors.
-    pub fn with_browser<F, T>(&self, urn: &RadUrn, callback: F) -> Result<T, Error>
+    pub async fn with_browser<F, T>(&self, urn: RadUrn, callback: F) -> Result<T, Error>
```
kim, line 350: Might be worth considering to maintain a pool of `Browser`s here. Also, `spawn_blocking` the callback seems advisable.

xla: #973

xla (approved):
:clap: :clap: :clap: This is extra dope! Kill all the locks. :clap: :clap: :clap: 

🌔 ✋ 💅 🚓

xla (approved):
👩 📧 🌬 🐣

FintanH (approved):
Thank you :bow: It's good to see your hand over in the proxy :heart:

proxy/api/src/main.rs
```diff
@@ -20,46 +32,121 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let mut args = pico_args::Arguments::from_env();
     let args = Args {
-        reset_state: args.contains("--reset-state"),
```
rudolfs, line 23: Now that this is removed, we can close: https://github.com/radicle-dev/radicle-upstream/issues/960.
However we also need to update the QA docs: https://github.com/radicle-dev/radicle-upstream/blob/master/QA.md#prerequisites.

@kim Would you prefer to do that in this issue or should we open another one?

rudolfs (approved):
Looks solid!
We should update the QA docs, but I won't block on it.

proxy/api/src/main.rs
NunoAlexandre: If the PR affects those bits we should address it in here imo

Issue #908, opened 2020-09-15
testnet for p2p networking


sarahscott:
As discussed in #866, we need a way to test interactions between peers in both `proxy` and `ui` integration tests.

FintanH:
To shed some more light on this issue:

We have certain tests with the proxy that involve having peer interactions. Some examples of these are:
* Revision selector with multiple peers to select from
* Having tracked projects in your monorepo
* Coming Soon: asking the network for a peer

These types of integration tests are best simulated with two proxy's talking to each other since this will be the actuality of how they will interact.

xla:
> * Coming Soon: asking the network for a peer

Do you mean asking for a project?

FintanH:
Correct, thanks for catching :+1:

xla:
TODOs to be tackeld in this context:

* https://github.com/radicle-dev/radicle-upstream/pull/896#discussion_r490850297
* https://github.com/radicle-dev/radicle-upstream/pull/896#discussion_r490851360

rudolfs:
Superseded by https://github.com/radicle-dev/radicle-upstream/issues/1475.

PR #946, opened 2020-09-22
feat(proxy): configure include file during checkout

### LINKED ISSUE #894: Create include file on checkout
When we checkout we should generate include file using `radicle-link` [utilities](https://github.com/radicle-dev/radicle-link/pull/300). 

So when we are [creating](https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/coco/src/project/checkout.rs#L52) the working copy we should also be creating an include file for that working copy.

This means we need to place the include file in a canonical location. This means that we should use a data directory for placing the set of include files. I believe upstream already has it's own data directory, so this should be passed to the `Checkout` struct.

We should end up with something like `$XDG_DATA_HOME/includes/<project.urn.id>.inc` and the working copy should create a config entry to point to this include path.

There's an integration test called [working_copy.rs](https://github.com/radicle-dev/radicle-link/blob/master/librad/tests/working_copy.rs) to see an example use of creating the working copy with this include file.

### LINKED ISSUE #930: Keep include files fresh and clean
Once #894 has landed we should tie into significant actions affecting the tracking graph of a project to update the corresponding include file, so that local working copies can access to the most up-to-date information.

xla:
Closes #894 
Closes #930

### DIFF
--- a/proxy/api/src/http/project.rs
+++ b/proxy/api/src/http/project.rs
@@ -139,21 +139,11 @@ mod handler {
         super::CheckoutInput { path, peer_id }: super::CheckoutInput,
     ) -> Result<impl Reply, Rejection> {
         let ctx = ctx.read().await;
-        let project = ctx
+        let path = ctx
             .state
             .lock()
             .await
-            .get_project(&urn, peer_id)
-            .map_err(Error::from)?;
-
-        let include = ctx
-            .state
-            .lock()
-            .await
-            .prepare_include(&project)
-            .map_err(Error::from)?;
-        let path = coco::project::Checkout::new(project, path, include)
-            .run(ctx.state.lock().await.peer_id())
+            .checkout(&urn, peer_id, path)
             .map_err(Error::from)?;
 
         Ok(reply::with_status(reply::json(&path), StatusCode::CREATED))
@@ -285,7 +275,7 @@ mod test {
         };
         let res = request()
             .method("POST")
-            .path(&format!("/{}/checkout", urn))
+            .path(&format!("/{}/checkout", urn.clone()))
             .json(&input)
             .reply(&api)
             .await;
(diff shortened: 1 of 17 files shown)


proxy/coco/src/project.rs
```diff
@@ -12,6 +13,18 @@ pub use create::{Create, Repo};
...
+fn set_include_path<P>(
+    repo: &git2::Repository,
+    include: &include::Include<P>,
+) -> Result<(), git2::Error>
+where
+    P: AsRef<std::path::Path>,
+{
+    let mut config = repo.config()?;
+
+    config.set_str("include.path", &include.file_path().display().to_string())
```
xla, line 25: @FintanH Can we get rid of the magic string here? Maybe `librad;:git::include` could export a const?

proxy/coco/src/state.rs
```diff
@@ -443,6 +444,36 @@ impl State {
...
+        P: Into<Option<PeerId>>,
+    {
+        let proj = self.get_project(&urn, peer_id).map_err(Error::from)?;
+        let include = self.prepare_include(&proj).map_err(Error::from)?;
+        let checkout = project::Checkout::new(proj, destination, include);
+
+        checkout.run(self.peer_id()).map_err(Error::from)
+    }
+
+    pub fn prepare_include(
```
xla, line 464:
```suggestion
    pub fn update_include(
```

We are going to call this for many project actions as per #930

xla:
@FintanH This is early and only does the setup during checkout in this iteration. I'm wondering what a valid strategy is to test this functionality proper.

proxy/coco/src/project.rs
FintanH: You didn't add this after all the forrays into link? :scream:

proxy/coco/src/state.rs
```diff
@@ -443,6 +445,48 @@ impl State {
             })
             .collect()
     }
+
+    /// Creates a workign copy for the project of the given `urn`.
```
cloudhead, line 449:
```suggestion
    /// Creates a working copy for the project of the given `urn`.
```

proxy/coco/src/project.rs
xla: :facepalm: forgot my own todo...PR forthcoming

xla: radicle-dev/radicle-link#335

proxy/api/src/http/project.rs
```diff
@@ -308,13 +304,30 @@ mod test {
...
+            .get_entry(coco::include::GIT_CONFIG_PATH_KEY)
+            .map_err(coco::Error::from)?
+            .value()
+            .unwrap()
+            .to_string();
+        assert_eq!(
+            include_path,
+            format!(
+                "{}/git-includes/{}.inc",
+                tmp_dir.path().display().to_string(),
```
FintanH, line 326: Could make this more resilient using the `Paths` to fill in for `git-includes`

FintanH (approved):

rudolfs (approved):
Looks great!

A follow-up issue for some tests for this functionality would be nice. Perhaps could be just a note in https://github.com/radicle-dev/radicle-upstream/issues/908.

PR #951, opened 2020-09-23
chore(ui): typing project creation

sarahscott:
Opened a can of worms with this one:
- [x] `Dropdown.svelte` & `Option.svelte`
- [x] `Tooltip.svelte`
- [x] `ProjectCreation.svelte` & smaller children components
- [x] Fix bug in `ProjectCreation` wherein `localState` store persists after cancel - to reproduce on `master` or here:
            - Click "New Project"
            - Choose an existing project with multiple branches
            - Select a branch that isn't `master`
            - Click "Cancel" or close the modal another way
            - Click "New Project" & existing project again - the branches from the first project will still be there even though 
                nothing is selected
- [x] Tests pass


ui/DesignSystem/Component/Dropdown/Option.svelte
```diff
@@ -42,18 +53,17 @@
...
 
 <div class="option" on:click={clickHandler} class:selected {style}>
-  {#if variant === 'avatar'}
+  {#if avatarProps}
     <Avatar
       size="small"
       style="overflow:hidden; text-overflow: ellipsis; margin: 0 42px 0 8px;
       --title-color: var(--color-foreground-level-6);"
-      {...avatarProps}
-      {disabled} />
```
rudolfs, line 51: Where did `{disabled}` go?

ui/Modal/ProjectCreation.svelte
```diff
@@ -232,14 +248,14 @@
...
-      {#if $settings && $settings.appearance.hints.showRemoteHelper}
+      {#if showRemoteHelper}
         <RemoteHelperHint on:hide={dismissRemoteHelperHint} />
       {/if}
     </div>
 
     <Tooltip
-      value={isExisting && 'The project name is taken from the repository you selected'}
-      position="top">
+      value={isExisting ? 'The project name is taken from the the repository you selected' : ''}
```
rudolfs, line 257:
```suggestion
      value={isExisting ? 'The project name is taken from the repository you selected' : ''}
```

ui/src/project.ts
```diff
@@ -104,8 +113,20 @@ interface FetchUser extends event.Event<Kind> {
...
+
+type Msg =
+  | ClearLocalState
+  | Create
+  | Fetch
+  | FetchList
+  | FetchLocalState
+  | FetchUser;
 
+// / REQUEST INPUTS
```
rudolfs, line 129: Is this a typo or some special type of comment in ts? Couldn't find anything on it...

```suggestion
// REQUEST INPUTS
```

tsconfig.json
```diff
@@ -42,7 +42,7 @@
     /* Strict Type-Checking Options */
     "strict": true /* Enable all strict type-checking options. */,
     /* Raise error on expressions and declarations with an implied 'any' type. */
-    // "noImplicitAny": true,
+    "noImplicitAny": false,
```
rudolfs, line 45: Was this only used for debugging, or do we want to disable it entirely?

ui/src/style.ts
```diff
@@ -0,0 +1,16 @@
+// / Shared types and functionality related to styling
```
rudolfs, line 1:
```suggestion
// Shared types and functionality related to styling
```

ui/Modal/ProjectCreation.svelte
```diff
@@ -16,14 +18,16 @@
     extractName,
     repositoryPathValidationStore,
     RepoType,
-  } from "../src/project.ts";
-  import { ValidationStatus } from "../src/validation.ts";
-  import * as screen from "../src/screen.ts";
+  } from "../src/project";
+  import * as testing from "../src/project";
```
rudolfs, line 22:
```suggestion
```

ui/Modal/ProjectCreation.svelte
```diff
@@ -75,7 +79,7 @@
 
       // Re-fetch session so we have the right permissions to enable the
       // project registration button rithout a page-reload.
-      await fetchSession();
+      fetchSession();
```
rudolfs, line 82: Lol, not the first time I tripped over the ts event thing not returning a promise. 🤦 

I wonder if we need this at all, as the registry is gone...

rudolfs (commented):
Great work!

Noted some very minor things, should be good to merge otherwise.

xla (changes requested):
Good stuff!!

ui/DesignSystem/Component/Dropdown.svelte
```diff
@@ -1,22 +1,38 @@
...
-  export let style = null;
-  export let optionStyle = null;
-  export let valid = true;
-  export let validationMessage = null;
-  export let validationPending = false;
+  export let dataCy = "";
+  export let style = "";
+  export let optionStyle = "";
+
+  export let valid: boolean = true;
```
FintanH, line 27: Could we convert these booleans into variants? For example,
```
enum Input {
  Raw(string),
  Valid(string),
}
```

Although really we'd like a newtype `Valid` that has a **smart constructor** `string -> Result<Valid, ValidationError>`

ui/src/project.ts
```diff
@@ -232,21 +271,19 @@ export const extractName = (repoPath: string) =>
...
-  localState.set("");
+  fetchLocalState({ path });
+
   localStateError.set("");
   defaultBranch.set(DEFAULT_BRANCH_FOR_NEW_PROJECTS);
 
   // This is just a safe guard. Since the validations on the constraints are
   // executed first, an empty path should not make it this far.
-  if (path === "") {
+  if (!path.length) {
```
FintanH, line 281: Is there no `empty` function or something? :thinking:

FintanH (commented):
Just having a peek

ui/DesignSystem/Component/Dropdown/Option.svelte
sarahscott: The `Avatar` component has no `disabled` prop! It never existed in the first place 👻

sarahscott: Or if it used to, it no longer does 🤷‍♀️

ui/src/project.ts
sarahscott: Just a typo 🤡

tsconfig.json
sarahscott: I was running into an error that I can't reproduce 🤔   if it surfaces again, I'll make a more informative issue. Will revert this change.

ui/Modal/ProjectCreation.svelte
sarahscott: I think you're right - it doesn't make sense to re-fetch until we actually need to. Will remove it!

ui/DesignSystem/Component/Dropdown.svelte
```diff
@@ -25,11 +41,12 @@
 
     expanded = !expanded;
   };
+
   const hideMenu = () => {
     expanded = false;
   };
 
-  const optionSelectedHandler = event => {
+  const optionSelectedHandler = (event: CustomEvent<{ value: string }>) => {
```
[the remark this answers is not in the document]
sarahscott: Yes - it's a `CustomEvent` dispatched by `Option.svelte`: https://github.com/radicle-dev/radicle-upstream/blob/master/ui/DesignSystem/Component/Dropdown/Option.svelte#L20

This could be exported from a lib file:
```ts
type OptionSelect = CustomEvent<{ value: string }>
```

But I was thinking that would be overkill, and might be better served by a general approach to typing dispatched events. What do you think?

ui/DesignSystem/Component/Dropdown.svelte
```diff
@@ -1,22 +1,38 @@
...
-  export let placeholder = null;
+  type Option =
+    | { value: string; title: string }
+    | {
+        value: string;
+        title: string;
+        avatarProps: {
+          avatarFallback: EmojiAvatar;
+          title: string;
+          variant?: "circle" | "square";
```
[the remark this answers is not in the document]
sarahscott: They could! But then the parent components would have to import the type from a `.ts` file instead of just passing `circle` or `square` as a prop.

ui/DesignSystem/Component/Dropdown.svelte
sarahscott: Interesting topic - we actually have typed states for validation: https://github.com/radicle-dev/radicle-upstream/blob/master/ui/src/validation.ts#L11

I'll make an issue for adding this to `Dropdown.svelte` because this pattern already exists for the other inputs.

ui/src/project.ts
sarahscott: Not in JS 😜

tsconfig.json
rudolfs: Got it. In your revert this line got commented out, but with the new value. To save another round-trip of reviewing & fixing I did a follow up which reverts the change proper: https://github.com/radicle-dev/radicle-upstream/pull/951/commits/5942685d054c07a5bccd10f190338938f24738bf.

rudolfs (approved):

ui/DesignSystem/Component/Dropdown.svelte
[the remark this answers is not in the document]
FintanH: What's wrong with that? :eyes:

ui/src/project.ts
FintanH: OOf

ui/DesignSystem/Component/Dropdown.svelte
FintanH: Sounds good to me :+1:

juliendonck (commented):
Generally looks good to me. Little nitpick, I find the enum name `CSSPosition` a bit clunky. It's not really a CSSPosition. It's the position of the tooltip.

xla (approved):
☢ 🌘 😌 🌃

MeBrei (approved):

sarahscott:
@juliendonck the `CSSPosition` type is just a reference to css's position-setting properties - `top`, `bottom`, `left`, `right` - for any component to use, not just the Tooltip. Maybe the Tooltip warrants its own type though 🤔

PR #953, opened 2020-09-23
fix(ui): improve identity and project name validation

### LINKED ISSUE #949: Make user handle and project name validation messages friendlier
[While doing QA](https://github.com/radicle-dev/radicle-upstream/issues/912#issuecomment-693499113.) we discovered that the display name and project name validation messages can be confusing to users. Particularly it's not instantly obvious why a single character is not valid (names have to be at least 2 characters).

Currently we only show the regex that the input is validated against.

Screenshot 2020-09-22 at 15 38 01
Screenshot 2020-09-22 at 15 38 14

My proposal is to split up the validations in parts and write a separate message for each of them, like so:

- **display name:**
  - Must start with a letter or digit
  - Must be at least 2 characters long
  - Only letters `a`-`z`, `A`-`Z`, digits `0`-`9` and special characters `-`, `_` are allowed

- **project name:**
  - Must start with a letter or digit
  - Must be at least 2 characters long
  - Only letters `a`-`z`, `A`-`Z`, digits `0`-`9` and special characters `-`, `_`, `.` are allowed

We would still want to show only one validation at a time to save space.

MeBrei:
Closes #949

### DIFF
--- a/ui/src/validation.ts
+++ b/ui/src/validation.ts
@@ -39,6 +39,10 @@ interface RemoteValidation {
   validationMessage: string;
 }
 
+interface firstHandleCharOptions {
+  valueName: string;
+}
+
 interface FormatConstraints {
   presence?: {
     message: string;
@@ -48,7 +52,21 @@ interface FormatConstraints {
     pattern: RegExp;
     message: string;
   };
+  firstHandleChar?: firstHandleCharOptions;
 }
+
+validatejs.validators.firstHandleChar = (
+  value: string,
+  options: firstHandleCharOptions,
+  key: any,
+  attributes: any
+) => {
+  if (!value.match(new RegExp("^[a-z0-9]", "i"))) {
+    return (
+      `Your ${  options.valueName  } should start with a letter or a number.`
+    );
+  }
+};
 
 export const createValidationStore = (
   constraints: FormatConstraints,
--- a/ui/src/validation.ts
+++ b/ui/src/validation.ts
@@ -58,13 +58,11 @@ interface FormatConstraints {
 validatejs.validators.firstHandleChar = (
   value: string,
   options: firstHandleCharOptions,
-  key: any,
-  attributes: any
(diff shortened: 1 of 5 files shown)


ui/src/validation.ts
```diff
@@ -48,8 +52,22 @@ interface FormatConstraints {
...
+validatejs.validators.firstHandleChar = (
+  value: string,
+  options: firstHandleCharOptions,
+  key: any,
+  attributes: any
+) => {
+  if (!value.match(new RegExp("^[a-z0-9]", "i"))) {
+    return (
+      `Your ${  options.valueName  } should start with a letter or a number.`
+    );
```
rudolfs, line 67:
```suggestion
    return `Your ${options.valueName} should start with a letter or a number.`;
```

rudolfs (approved):

rudolfs (dismissed):

ui/src/validation.ts
```diff
@@ -39,6 +39,10 @@ interface RemoteValidation {
   validationMessage: string;
 }
 
+interface firstHandleCharOptions {
```
sarahscott, line 42:
```suggestion
interface FirstHandleCharOptions {
```

ui/src/validation.ts
```diff
@@ -48,8 +52,20 @@ interface FormatConstraints {
     pattern: RegExp;
     message: string;
   };
+  firstHandleChar?: firstHandleCharOptions;
```
sarahscott, line 55: Alternatively, you could skip making a separate `interface` and just do:
```suggestion
  firstHandleChar?: {
      valueName: string
  };
```

sarahscott (changes requested):
Looking good apart from the `interface` thing - should be capitalized or built into the existing `FormatConstraints` ⭐

rudolfs (approved):

sarahscott (approved):

PR #962, opened 2020-09-24
fix: show commit count on project page

### LINKED ISSUE #924: Show commit count based on revision
The commit tab currently shows the number of commits from the project stats. Instead the number of commits on the selected revision should be shown.

MeBrei:
* extends the commit endpoint with stats
* shows the correct commit count of the selected peer and branch
* cleans up the variables project page components

Note: Changing the selected peer can't be tested right now.

Closes #924 

**TODO**
- [x] fix source browsing test
- [x] fix jumpy horizontal header (jump between no-count & count)

### DIFF
--- a/cypress/integration/project_source_browsing.spec.js
+++ b/cypress/integration/project_source_browsing.spec.js
@@ -22,6 +22,8 @@ context("repository stats", () => {
 context("commit browsing", () => {
   context("commit history", () => {
     it("shows the commit history for the default branch", () => {
+      // Wait for the commit tab to be updated
+      cy.pick("horizontal-menu", "Commits", "counter").contains("14");
       cy.pick("horizontal-menu", "Commits").click();
       cy.pick("commits-page").should("exist");
       cy.pick("commit-teaser")
--- a/cypress/integration/project_source_browsing.spec.js
+++ b/cypress/integration/project_source_browsing.spec.js
@@ -9,15 +9,15 @@ before(() => {
 beforeEach(() => {
   cy.visit("./public/index.html#/profile/projects");
   cy.contains("platinum").click();
-  // cy.contains("Source").click();
+  // Make sure there is enough time to load the commits which refreshes the horizontal menu
+  cy.wait(500);
+  cy.contains("Source").click();
 });
 
 context("repository stats", () => {
   it("shows the correct numbers", () => {
     cy.pick("header", "project-stats").contains("2 Branches");
     cy.pick("header", "project-stats").contains("4 Contributors");
-    // Make sure there is enough time to load the commits
-    cy.wait(500);
     cy.pick("horizontal-menu", "Commits").contains("Commits 14");
   });
 });
--- a/cypress/integration/project_source_browsing.spec.js
+++ b/cypress/integration/project_source_browsing.spec.js
@@ -9,16 +9,13 @@ before(() => {
 beforeEach(() => {
   cy.visit("./public/index.html#/profile/projects");
   cy.contains("platinum").click();
-  // Make sure there is enough time to load the commits which refreshes the horizontal menu
-  cy.wait(500);
-  cy.contains("Source").click();
 });
 
 context("repository stats", () => {
   it("shows the correct numbers", () => {
     cy.pick("header", "project-stats").contains("2 Branches");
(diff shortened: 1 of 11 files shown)


rudolfs:
Found a bug, where switching between projects doesn't reload the revision selector properly.

branch

proxy/api/src/http/source.rs
```diff
@@ -729,10 +730,11 @@ mod test {
         })?;
 
         http::test::assert_response(&res, StatusCode::OK, |have| {
-            assert_eq!(have, json!(want));
-            assert_eq!(have.as_array().unwrap().len(), 14);
+            assert_eq!(have["headers"], json!(want.headers));
+            assert_eq!(have["headers"].as_array().unwrap().len(), 14);
```
FintanH, line 734: Is this check not made redundant by the equality check above?

proxy/api/src/http/source.rs
```diff
@@ -729,10 +730,11 @@ mod test {
         })?;
 
         http::test::assert_response(&res, StatusCode::OK, |have| {
-            assert_eq!(have, json!(want));
-            assert_eq!(have.as_array().unwrap().len(), 14);
+            assert_eq!(have["headers"], json!(want.headers));
```
FintanH, line 733: I feel like this test should simply be `assert_eq!(have, json!(want))`, is there a reason for the extra cases?

FintanH (changes requested):

rudolfs:
It seems like the `README.md` is loading twice, causing some flickering.

double-load

cypress/integration/project_source_browsing.spec.js
```diff
@@ -9,14 +9,16 @@ before(() => {
 beforeEach(() => {
   cy.visit("./public/index.html#/profile/projects");
   cy.contains("platinum").click();
+  // Make sure there is enough time to load the commits which refreshes the horizontal menu
+  cy.wait(500);
```
rudolfs, line 13: If possible we should avoid fixed `wait`s, because it makes the tests brittle.
A better approach is to check for the presence of an element on screen. This way cypress will wait only as long as necessary and is guaranteed to succeed.

In this case I'd add a `data-cy` to the commit counter and check that the content equals to the number of commits.

Something like: `cy.pick("commit-count").contains("14")`.

cypress/integration/project_source_browsing.spec.js
```diff
@@ -40,6 +42,8 @@ context("commit browsing", () => {
     it("shows the commit history for another branch", () => {
       cy.pick("revision-selector").click();
       cy.get('[data-branch="dev"]').click();
+      // Make sure there is enough time to load the commits
+      cy.wait(500);
```
rudolfs, line 46: Ditto.

proxy/coco/src/source.rs
```diff
@@ -185,6 +185,26 @@ impl Serialize for CommitHeader {
     }
 }
 
+/// A selection of commit headers and their statistics.
+pub struct Commits {
```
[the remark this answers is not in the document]
MeBrei: `Serialize` seems enough here. Would it be good to add the others anyway?

proxy/api/src/http/source.rs
MeBrei: You are right! I just extended the tests without re-thinking the existing ones

proxy/coco/src/source.rs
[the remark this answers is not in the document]
FintanH: Ya, the others are useful regardless of whether they're needed right now. They're generally useful for debugging tests and what not.

proxy/api/src/http/source.rs
```diff
@@ -691,6 +691,7 @@ mod test {
     }
 
     #[tokio::test]
+    #[allow(clippy::indexing_slicing)]
```
FintanH, line 694:
```suggestion
```

I think this isn't needed now that we cleaned up the testing situation

FintanH (changes requested):

MeBrei:
> It seems like the `README.md` is loading twice, causing some flickering.
> 

@rudolfs I created an issue for this bug #970

FintanH (approved):
Looks good :+1:

rudolfs (approved):

PR #967, opened 2020-09-25
feat(proxy): waiting room subroutine

### LINKED ISSUE #955: Waiting room subroutine
Once https://github.com/radicle-dev/radicle-upstream/pull/903 lands we want to link it up with the async subroutine handlers so that we can issue `Want`s for projects, and transition the `WaitingRoom` through its states.

xla:
Closes #955

### DIFF
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -24,8 +24,8 @@ mod request;
 
 mod run_state;
 pub use run_state::{
-    AnnounceConfig, AnnounceEvent, Config as RunConfig, Event, RequestCommand, RequestConfig,
-    RequestEvent, SyncConfig, SyncEvent, TimeoutEvent,
+    AnnounceConfig, AnnounceEvent, Config as RunConfig, Event, RequestCommand, RequestEvent,
+    SyncConfig, SyncEvent, TimeoutEvent,
 };
 use run_state::{Command, RunState};
 
@@ -103,15 +103,9 @@ impl Peer {
         let (peer_sync_sender, mut peer_syncs) = mpsc::channel::<SyncEvent>(RECEIVER_CAPACITY);
         let (timeout_sender, mut timeouts) = mpsc::channel::<TimeoutEvent>(RECEIVER_CAPACITY);
 
-        let request_queries = waiting_room::stream::Queries::new(
-            waiting_room.clone().value,
-            run_config.request.query_delay,
-        );
+        let request_queries = waiting_room::stream::Queries::new(waiting_room.clone().value);
         tokio::pin!(request_queries);
-        let request_clones = waiting_room::stream::Clones::new(
-            waiting_room.clone().value,
-            run_config.request.clone_delay,
-        );
+        let request_clones = waiting_room::stream::Clones::new(waiting_room.clone().value);
         tokio::pin!(request_clones);
 
         // Advance the librad protocol.
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -9,10 +9,7 @@ use tokio::{
     time::interval,
 };
 
-use librad::{
-    net::peer::RunLoop,
-    peer::PeerId,
-};
+use librad::{net::peer::RunLoop, peer::PeerId};
 
 use crate::{
     request::waiting_room::{self, WaitingRoom},
@@ -137,7 +134,9 @@ impl Peer {
                     Command::Announce => {
                         Self::announce(state.clone(), store.clone(), announce_sender.clone());
                     },
-                    Command::Request(command) => command.run(state.clone(), waiting_room.clone()).await,
+                    Command::Request(command) => {
+                        command.run(state.clone(), waiting_room.clone()).await
+                    },
                     Command::SyncPeer(peer_id) => {
                         Self::sync(state.clone(), peer_sync_sender.clone(), peer_id.clone()).await;
                     },
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -38,6 +38,9 @@ mod sync;
 /// Upper bound of messages stored in receiver channels.
 const RECEIVER_CAPACITY: usize = 128;
 
+/// The period at which we ping the `select!` loop.
+const PING_PERIOD: Duration = Duration::from_millis(500);
+
 /// Peer operation errors.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
@@ -108,6 +111,10 @@ impl Peer {
         // TODO(xla): Find a more structured approach to manage timings.
         let mut announce_timer = interval(run_config.announce.interval);
 
+        // TODO(finto): We're not sure why pinging the select loop helps drive the state forward.
+        // This should be investigated.
+        let mut ping = interval(PING_PERIOD);
+
         let (announce_sender, mut announcements) =
             mpsc::channel::<AnnounceEvent>(RECEIVER_CAPACITY);
         let (peer_sync_sender, mut peer_syncs) = mpsc::channel::<SyncEvent>(RECEIVER_CAPACITY);
@@ -133,6 +140,7 @@ impl Peer {
                 Some(urn) = request_queries.next() => Event::Request(RequestEvent::Query(urn)),
                 Some(url) = request_clones.next() => Event::Request(RequestEvent::Clone(url)),
                 Some(request_event) = requests.next() => Event::Request(request_event),
+                _ = ping.tick() => Event::Ping,
                 else => {
                     break
                 },
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -106,7 +106,7 @@ impl Peer {
             mpsc::channel::<AnnounceEvent>(RECEIVER_CAPACITY);
         let (peer_sync_sender, mut peer_syncs) = mpsc::channel::<SyncEvent>(RECEIVER_CAPACITY);
         let (timeout_sender, mut timeouts) = mpsc::channel::<TimeoutEvent>(RECEIVER_CAPACITY);
-        let (cloned_sender, mut requests) = mpsc::channel::<RequestEvent>(RECEIVER_CAPACITY);
+        let (request_sender, mut requests) = mpsc::channel::<RequestEvent>(RECEIVER_CAPACITY);
 
         let request_queries = waiting_room::stream::Queries::new(waiting_room.clone().value);
         tokio::pin!(request_queries);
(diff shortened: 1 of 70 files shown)


proxy/coco/src/peer/run_state.rs
```diff
@@ -29,12 +34,108 @@ const DEFAULT_SYNC_PERIOD: Duration = Duration::from_secs(5);
...
+
+#[derive(Debug, PartialEq)]
+pub enum RequestCommand {
+    Clone(RadUrl),
+    Found(RadUrl),
+    Query(RadUrn),
+}
+
+impl RequestCommand {
+    pub async fn run<W>(self, state: Lock, waiting_room: W)
```
xla, line 58: I see where you going with this. Yet it feels out of place to have this in the run_state module where we so far have been exclusively be concerned with the state transition of the peer event machine. It should follow the organisation of the other module like announcement and sync.

proxy/coco/src/peer/run_state.rs
```diff
@@ -430,4 +565,160 @@ mod test {
 
         assert!(cmds.is_empty(), "expected no command");
     }
+    // TODO(finto): Test that we only issue Query command if Online or Syncing
+    // TODO(finto): Test that we only issue Clone command if Online or Syncing
```
xla, line 569: The cases where once we come online we start draining requests, either read from persisted state or created while we were offline, need to be addressed as well.

proxy/coco/src/request.rs
```diff
@@ -316,12 +316,12 @@ impl<T> Request<Cloning, T> {
     /// This signifies that the clone was successful and that the whole request was successful,
     /// congratulations.
     #[allow(clippy::use_self, clippy::missing_const_for_fn)]
-    pub fn cloned(self, repo: RadUrn, timestamp: T) -> Request<Cloned, T> {
+    pub fn cloned(self, timestamp: T) -> Request<Cloned, T> {
```
xla, line 319: It might interesting to record the url we cloned from.

proxy/coco/src/shared.rs
```diff
@@ -0,0 +1,29 @@
+use std::{ops::Deref, sync::Arc};
+
+use tokio::sync::RwLock;
+
+pub struct Shared<T> {
```
xla, line 5: We've come full circle.

xla (commented):
Dope!

proxy/coco/src/peer/run_state.rs
FintanH: Not sure I understand

proxy/coco/src/request.rs
FintanH: _smh_ thought we'd decided otherwise? :sweat_smile:

proxy/coco/src/shared.rs
FintanH: Hahaha ya, but this time I know better tricks

proxy/coco/src/peer/run_state.rs
xla: Say we restore a waiting_room from disk and it will take a while to get online, will we be able to start querying requests that have been stored in the persisted waiting room?

proxy/coco/src/request.rs
xla: Can't remember that we ever had the url as an input. It might be of interest to see what peer satisfied our request.

proxy/coco/src/shared.rs
xla: Hubris!

proxy/coco/src/request.rs
FintanH: Ah the url, sorry was thinking urn. `RadUrl` makes sense :+1:

proxy/coco/src/shared.rs
FintanH: :smiling_imp:

proxy/coco/src/peer/run_state.rs
FintanH: Ah, ok I'm with you. So the test case would look like:
1. Have a waiting room with created/requested requests inside it
2. Initialise the run state to online
3. See the commands for every one of the requests from step 1.

Right?

xla: Sounds like a viable approach. What we want to ensure is that the pending requests are not accidentally advanced to timed out because the peer was offline.

proxy/coco/src/peer/run_state.rs
```diff
@@ -24,6 +24,9 @@ const DEFAULT_ANNOUNCE_INTERVAL: Duration = std::time::Duration::from_secs(60);
 /// TODO(xla): Revise number.
 const DEFAULT_SYNC_MAX_PEERS: usize = 5;
 
+/// How much a stream should delay before polling the waiting room again.
+const DEFAULT_REQUEST_DELAY: Duration = Duration::from_millis(500);
```
xla, line 28: Why do we have such low poll rate? This will result in 2Hz.

proxy/coco/src/peer/run_state.rs
```diff
@@ -192,6 +197,23 @@ impl Default for SyncConfig {
     }
 }
 
+/// Set of knobs to alter request behaviour.
+pub struct RequestConfig {
```
xla, line 201: These poll rates don't seem like a thing that is frequently configured or changed from the default. They are in fact internals and this seem to leak them.

proxy/coco/src/request/waiting_room/stream.rs
```diff
@@ -85,9 +90,13 @@ impl Clones {
     /// exists then the stream does nothing and waits until the next poll. We note that the stream
     /// is infinite with this respect and will never return `None`.
     #[must_use]
-    pub fn new(waiting_room: Arc<RwLock<WaitingRoom<Instant, Duration>>>) -> Self {
+    pub fn new(
+        waiting_room: Arc<RwLock<WaitingRoom<Instant, Duration>>>,
+        duration: Duration,
```
xla, line 95: As mentioned above, it seems odd that this is configurable. It is internal to the proper implementation of the stream as it ensures that the waker is called and we make progress on this future.

proxy/coco/tests/gossip.rs
```diff
@@ -236,3 +240,93 @@ async fn providers() -> Result<(), Box<dyn std::error::Error>> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn can_ask_and_clone_project() -> Result<(), Box<dyn std::error::Error>> {
```
xla, line 245: :black_heart:

proxy/coco/src/peer.rs
```diff
@@ -167,4 +196,68 @@ impl Peer {
             sender.clone().send(TimeoutEvent::SyncPeriod).await.ok();
         });
     }
+
+    /// Request subroutine.
+    async fn request(
```
xla, line 201: To keep the core peer impl light, this could follow the precedence of `announcement` and `sync` and live in `coco::peer::request`.

xla (commented):
More dopeness.

proxy/coco/src/request/waiting_room/stream.rs
FintanH: It's configurable because otherwise our tests would be stuck with the "reasonable" value which may be too long for them making our test suite slow.

xla: I'd argue the reasonable value is off. We should be able to poll at 100Hz without a sweat.

FintanH: I don't think I'm understanding. I picked a rate at random tbh. If you have a better default then I'm all ears :ear: However, I'm not sure why making it configurable is not ok

xla: Because it's the poll rate of the `impl Stream`, just an internal number to make sure the stream advances -- exclusively in the sense of feeding the async runtime. So it does not mean anything to the domain of the waitng_room/requests. It's not altering any domain behaviour. More so it provides a footgun, when missconfigured to potentially delay the progress of requests unnecessarily.

FintanH: But does not define how long the stream will wait before it tries to look for the next request? Like say the delay was set to 5 seconds, and our test missed the boat on checking the waiting room, then it has to wait for the next window, no? Adding a whole 5 seconds to our test suite :(

xla: There seems to be a miss-conception here. The delay should be in the order of max 10 millis. It's really a cheap solution to avoid a busy loop hogging the cpu while still making progress. We need this because we don't have a more effective way of signaling the waker passed in via context to `poll` that there is a `Ready` value. If we find a more effective, as in evented way, to signal the async runtime that we have a new value, this delay will vanish alltogether. It's a very deep implementation detail and should not be surfaced.

FintanH: Ya, I was completely misunderstanding :+1: Thanks for the patience :innocent:

proxy/coco/src/peer/run_state.rs
FintanH: I'm not exactly sure how I can test this scenario come to think of it :thinking: fwiw I don't think we can run into this scenario because we only look at query requests if we're online.

FintanH:
Well, shit... `can_ask_and_clone_project` seems flakey :thinking:

proxy/coco/src/request/waiting_room/stream.rs
```diff
@@ -0,0 +1,264 @@
...
+        self.delay = delay_for(REQUEST_DELAY)
+    }
+}
+
+impl Stream for Queries {
+    type Item = RadUrn;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let waiting_room = self.waiting_room.clone();
+        let waiting_room = futures::executor::block_on(waiting_room.read());
```
kim, line 60: Uh-oh, blocking is verboten. You need to store the untypeable future returned from `RwLock::read()` in `self`, and poll that. If it returns you the guard, proceed, else return `Poll::Pending`.

At least, I don't know a better way to desugar `async/await` here.

FintanH (approved):
Looks great, amazing work, you're amazing, everything is amazing. No flakes here...

proxy/coco/src/request/waiting_room/stream.rs
FintanH: Thanks, that might actually be the cause of our previous bug :thinking: Created a follow up issue: https://github.com/radicle-dev/radicle-upstream/issues/981

rudolfs (approved):

PR #968, opened 2020-09-25
fix(proxy): move delta to config

FintanH:
The delta should only be a concern for `Requested` states rather than `Created` or `Found` states. This is because `Requested` states can live long and need to be retried. Whereas `Created` and `Found` states are once off and should always be returned immediately.

### DIFF
--- a/proxy/coco/src/request/waiting_room.rs
+++ b/proxy/coco/src/request/waiting_room.rs
@@ -69,13 +69,13 @@ impl<T> From<Request<TimedOut, T>> for Error {
 /// request on the outside.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct WaitingRoom<T> {
+pub struct WaitingRoom<T, D> {
     /// The set of requests keyed by their `RadUrn`. This helps us keep only unique requests in the
     /// waiting room.
     requests: HashMap<RadUrn, SomeRequest<T>>,
 
     /// The configuration of the waiting room.
-    config: Config,
+    config: Config<D>,
 }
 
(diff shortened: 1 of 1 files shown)


NunoAlexandre:
@FintanH would appreciate your help here :)

1. What problem more specifically is this solving?
2. Do we really need delta to be a parameterized type? Don't see more use cases for it besides `usize` so now it only seems to add complexity.

FintanH:
> What problem more specifically is this solving?

So, this has some insider knowledge based off of https://github.com/radicle-dev/radicle-upstream/pull/967/files. The crux is that we want to ask the `WaitingRoom` for certain `Request` so we can fire off actions, so essentially the `WaitingRoom` becomes a `Stream`.
In the initial PR I misunderstood a requirement from @xla and we both didn't catch it initially. This misunderstanding was that the `delta` is only useful for the case of the state being `Requested`. This change shows that in the `next_query` function.

 > Do we really need delta to be a parameterized type? Don't see more use cases for it besides usize so now it only seems to add complexity.

Yes, the reason the tests use `usize` is because they're tests. Using actual `Duration`s and `Instant`s is going to make testing hell. Any time there's "time" involved, I'm of the opinion that you should parameterise it out so that you just plug in the time type you want later. The complexity in the single type signature is worth the gains of not worrying about time in tests.

NunoAlexandre (approved):
Thanks for the clarification ✌️

FintanH:
No problemo, lemme know if there's anything else that's unclear :)

rudolfs (approved):

xla (approved):
🐕 🎒 🐝 🏁

PR #969, opened 2020-09-29
feat(ui): use search bar to kick off project search

sarahscott:
🚧  depends on work in fintolex's branch 🚧


FintanH:
So as I mentioned `Instant` has no `Serialize` impl, however this lib provides one https://docs.rs/serde_millis/0.1.1/serde_millis/

Maybe what we _could_ do is try:
```rust
impl Serialize for WaitingRoom<Instant, Duration>
```

writing it manually and for the timestamp field we just call https://docs.rs/serde_millis/0.1.1/serde_millis/fn.serialize.html. I believe that should work. So you can disregard my comment about making a new `WaitingRoom` newtype around `Instant` + `Duration`.

ui/src/search.ts
```diff
@@ -1,7 +1,9 @@
 import { writable } from "svelte/store";
 
+import * as api from "./api";
 import * as event from "./event";
-
+import * as remote from "./remote";
+// import * as waitingRoom from "./waitingRoom";
```
rudolfs, line 6:
```suggestion
```

FintanH: I wanna leave it there so we reintroduce it when we _actually_ respond with waiting room data, if that's cool with you?

ui/Screen/Profile/Tracking.svelte
```diff
@@ -4,55 +4,39 @@
...
-        description: "Lemuel is a villager from Butcher Creek.",
-        maintainers: [],
-      },
-      registration: false,
-      shareableEntityIdentifier: "lemuel@38590438594",
-      maintainers: [],
-    },
-  ];
+  fetchTracked();
+  $: console.log($tracked);
```
rudolfs, line 16: What about this one?

xla (approved):
🏖 🎤 🔋 🍌

FintanH (dismissed):

xla (approved):

FintanH (approved):

proxy/coco/src/request.rs
```diff
@@ -71,6 +71,53 @@ pub struct Request<S, T> {
...
+    T: Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut state = serializer.serialize_struct("Request", 4)?;
+        state.serialize_field("urn", &self.urn)?;
+        state.serialize_field("attempts", &self.attempts)?;
+        state.serialize_field("timestamp", &WithMillis(self.timestamp))?;
```
FintanH, line 98: @sarahscott I think we could just do `serde_millis::serialize(self.timestamp)` here and remove  `WithMillis`

PR #978, opened 2020-09-30
docs(qa): advise to use temporary user accounts

kim:
For those unverbesserliche, we also provide a bash script to remove state and config data.

### DIFF
--- a/QA.md
+++ b/QA.md
@@ -4,6 +4,30 @@
 
 ### Prerequisites
 
+- [ ] make sure you are installing into an environment with no old configuration
+      and user data
+
+      **Safe method**: Use a temporary user account on your computer
+
+      - on macOS:
+        - If you are **not** using FileVault, switch to the "Guest User". You
+          may have to enable this in "System Preferences -> Users & Groups".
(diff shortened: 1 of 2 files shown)


NunoAlexandre (dismissed):
Sick :ok_hand:

QA.md
```diff
@@ -4,8 +4,32 @@
...
+
+     **Dangerous method**: Remove all directories manually
+
+     You can use [this script](./scripts/reset-state.sh). Make sure you have a
+     backup of your data, or are using this in combination with the safe method
+     (i.e. while logged in with a temporary user account).
+
 - [ ] download the binary package from the location provided above
-- [ ] remove any old configuration and user data
+- [ ] install the binary package
```
kim, line 32: Is this dance on macOS still necessary?

rudolfs: Yeah, we should leave it in as it's not very obvious. Much macOS, such secure!

rudolfs (approved):
Very nice, I like! 🚀

NunoAlexandre (approved):

NunoAlexandre (approved):

PR #980, opened 2020-09-30
fix(ui): show search input hint on empty input

### LINKED ISSUE #976: Keyboard hint appears *after* user types
The keyboard hint should be visible *before* the user types anything into the search box.

Since we already have the keyboard hint in there, should we also remove the "Paste it here..."?

paste

MeBrei:
Closes #976

### DIFF
--- a/ui/DesignSystem/Primitive/Input/Text.svelte
+++ b/ui/DesignSystem/Primitive/Input/Text.svelte
@@ -36,7 +36,7 @@
     }
   };
 
-  $: showHint = hint.length > 0 && value.length > 0;
+  $: showHint = hint.length > 0 && value.length === 0;
 </script>
 
 <style>
--- a/ui/Modal/SearchModal.svelte
+++ b/ui/Modal/SearchModal.svelte
@@ -86,7 +86,7 @@
     <Input.Text
       autofocus
       bind:value
-      placeholder="Have a Radicle project ID? Paste it here..."
+      placeholder="Have a Radicle project ID?"
       showLeftItem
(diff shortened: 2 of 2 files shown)


rudolfs (approved):

NunoAlexandre (approved):

sarahscott:
Looks good! I want to make sure this is the desired behavior - is the hint supposed to disappear as soon as the user types? Wouldn't we want the hint to be there when they finish typing, so they can use it?

sarahscott (approved):

rudolfs:
> Looks good! I want to make sure this is the desired behavior - is the hint supposed to disappear as soon as the user types? Wouldn't we want the hint to be there when they finish typing, so they can use it?

AFAIK, that's how @juliendonck implemented it first.

I think I'd also find it more intuitive if it was always there, but I wonder whether it would not conflict with the icons of the *waiting* and *validated* input states.

rudolfs:
A good example where the show/hide logic doesn't sense to me is the password input fields:
pw

juliendonck:
> Looks good! I want to make sure this is the desired behavior - is the hint supposed to disappear as soon as the user types? Wouldn't we want the hint to be there when they finish typing, so they can use it?

You make a good point and I had tried to do it as you suggested, but it would be a much more involved solution. And I felt is wasn't warranted the time and effort. 

The reason it's so much more involved is that it would need to work with the validation on the one hand but we'd also have to resize the input text field depending on the size of the hint.

So even though it's not perfect, I think it does the job for now.

PR #986, opened 2020-10-01
fix(ui): show cursor wait on commits loading

### LINKED ISSUE #920: Navigating to "commits" tab should have visual feedback
On larger projects with many commits there is a substantial delay between clicking on the "commits" tab and actually navigating there. We should at least indicate that there is some activity to the user via a waiting cursor.

Refs: https://github.com/radicle-dev/radicle-upstream/issues/912.

commits

MeBrei:
The `wait` cursor is shown when the commit tab is active, but the commits are still loading. If the cursor hovers over other clickable items than the Commit tab, the cursor changes according to those components (eg revision selector or Source tab) and they are still clickable, ie the screen is not locked.

Closes #920 (the beta-1 version)


rudolfs:
For me the cursor only sometimes switches to the `waiting` state, I can get it to work only if I move the mouse around (but also not always).

no-wait

MeBrei:
> For me the cursor only sometimes switches to the `waiting` state, I can get it to work only if I move the mouse around (but also not always).
> 
It's a bit hard to see what happens first in your gif. But the loading cursor only shows up, if the commit tab is `active` while the commits are loading. In the gif it looks like you only click on it once it is already loaded

rudolfs:
> It's a bit hard to see what happens first in your gif. But the loading cursor only shows up, if the commit tab is active while the commits are loading. In the gif it looks like you only click on it once it is already loaded

Let me try again.

try2

In this gif I:

1) go to the project page → wait for it to load
2) click on the *Commits* tab → wait for it to load (here I would have expected to see a loading cursor)
3) click on the page to make sure it's focussed → hit <kbd>⌘</kbd>+<kbd>r</kbd> → wait for it to load (here I would also expect to see a loading cursor)
4) click on the *Source* tab to go back to the main screen

MeBrei:
> In this gif I:
> 
> 1. go to the project page → wait for it to load
> 2. click on the _Commits_ tab → wait for it to load (here I would have expected to see a loading cursor)

This is expected behavior. The commits are loaded in the main project page. As an indicator, once you see the number of commits next to the `Commit` tab the store has successfully loaded. So the cursor does not change. If you click the Commit tab before you see the count, you should see the wait cursor.

> 3. click on the page to make sure it's focussed → hit ⌘+r → wait for it to load (here I would also expect to see a loading cursor)

It seems like in this scenario, the cursor only changes if you move the mouse ever so slightly. 🤷  I will look into it.

> 4. click on the _Source_ tab to go back to the main screen

rudolfs:
> This is expected behavior. The commits are loaded in the main project page. As an indicator, once you see the number of commits next to the Commit tab the store has successfully loaded. So the cursor does not change. If you click the Commit tab before you see the count, you should see the wait cursor.

If the commits are loaded in the main project page, why is there a substantial pause between me clicking on the tab and the UI actually showing it? I would expect it to be instant.

MeBrei:
There is a difference between the commits `store` loading and the commits page it self. I understood that the issue was just about the store, but realised now that it is about the wait until the page itself is loaded. 

I am closing this PR as we need to tackle this problem in a different way, eg connect the cursor change to the `Remote` component being in loading state, but also address the problem above where the cursor only changes on move. It might make sense to deal with this in a more general way as suggested in #920

Issue #836, opened 2020-08-28
Adjust the wording according to the Glossary


MeBrei:
Go through all copy of the app and check if it is up to date with the outcome of the [Glossary](https://www.notion.so/monadic/Glossary-5026cc02dca14629bb9d4058c7fa527a). Changes should be listed here to be worked on in PRs.

abbey-titcomb:
I've added comments with outstanding copy changes & questions to the [Figma file](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=4147%3A7246). Will wait to resolve these before moving ahead with changes.

Main points of conflict are between use of RadicleID throughout App and the distinction between Tracking & Following.

abbey-titcomb:
Some copy updates being addressed here -> https://github.com/radicle-dev/radicle-upstream/issues/877

rudolfs:
Another small wording change addressed here -> #919

abbey-titcomb:
Reviewed latest build (https://buildkite.com/monadic/radicle-upstream/builds/5170) and came up with some outstanding TODOs & comments. Capturing here:

- [ ] Replace with correct links once website is up (will create an issue for this)

- [ ] Add "Be sure to put this in a safe place" on second part of onboarding (@MeBrei )

- [ ] Replace Radice ID and Project ID with user handle and project handle respectively *after* #840 is resolved. Needs to be replaced in onboarding and search bar

**Some bugs**

- [ ] Does Project Description have a max? Seems that there is no way to expand the description when it gets to long...
<img width="1067" alt="image" src="https://user-images.githubusercontent.com/18275452/94798417-df12ac80-03e1-11eb-8c2b-0dad45492647.png">

- [ ] After something is pasted the <kbd>⌘</kbd><kbd>V</kbd> should go away right?
<img width="445" alt="image" src="https://user-images.githubusercontent.com/18275452/94799759-ec309b00-03e3-11eb-96bf-b346445137c7.png">

- [ ] Project URN does not appear in text box
<img width="469" alt="image" src="https://user-images.githubusercontent.com/18275452/94799793-fd79a780-03e3-11eb-8ee8-ef043608eda7.png">

@MeBrei @rudolfs - should I make these into issues?

rudolfs:
@abbey-titcomb 

> Does Project Description have a max? Seems that there is no way to expand the description when it gets to long...
[...]

PR #990, opened 2020-10-02
fix(ui): extend password text

MeBrei:
Contributes to resolving #836 

Adds `Be sure to put this in a safe place!` to the description on the onboarding password screen

### DIFF
--- a/ui/Screen/Onboarding/EnterPassphrase.svelte
+++ b/ui/Screen/Onboarding/EnterPassphrase.svelte
@@ -126,7 +126,7 @@
       computer’s password. <span class="typo-text-bold">
         You can’t recover your account with it,
       </span> but it prevents someone from accessing your account if this computer
-      is stolen or hacked.
+      is stolen or hacked. Be sure to put this in a safe place!
     </p>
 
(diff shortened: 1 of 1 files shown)


rudolfs (approved):

xla (changes requested):
This is a strange advise, are we implying to use a password manager?

juliendonck:
> This is a strange advise, are we implying to use a password manager?

@xla Are we? Do you consider a password manager a safe place? If not, then we're not implying that. I don't think it's our job to teach cyber security to our users.

juliendonck (approved):

xla:
> @xla Are we? Do you consider a password manager a safe place? If not, then we're not implying that. I don't think it's our job to teach cyber security to our users.

That would be a bit much in the context of this field. Yet the wording might help more if we remind the user that they make sure they won't loose "access" to the password.

abbey-titcomb:
This change is just to keep the app copy consistent with what we have in the Figma file.

I don't see how it's strange but am open to alternate suggestions on wording if have some @xla

PR #992, opened 2020-10-03
fix(proxy): prevent halting peer subroutines

### LINKED ISSUE #981: Don't block Stream
As pointed out here: https://github.com/radicle-dev/radicle-upstream/pull/967#discussion_r497503768

xla:
The polling of subroutines can starve if the channels are not completely
drained. To remedy that this refactor restructures the internals of the
peer, most notably by have the RunState drive the WaitingRoom
transitions.

* move WaitingRoom into RunState
* use SelectAll for subroutines
* introduce handle for peer control

Fixes #981
Fixes #1009


xla:
Sorry all, this escalated a bit. The refactor addresses a bunch of concerns, first and foremost the halting subroutines as captured in #981. As the change-set is quite large, it's best to check [subroutine.rs](https://github.com/radicle-dev/radicle-upstream/blob/xla/tame-the-locks/proxy/coco/src/peer/subroutines.rs) for the move to `SelectAll` and [runstate.rs](https://github.com/radicle-dev/radicle-upstream/blob/xla/tame-the-locks/proxy/coco/src/peer/run_state.rs) for the move of the `WaitingRoom`.

@FintanH @geigerzaehler @kim Thanks for all the input and guidance along the way.

@rudolfs The newly introduced [control.rs](https://github.com/radicle-dev/radicle-upstream/blob/xla/tame-the-locks/proxy/coco/src/peer/control.rs) is the way I imagined of how to inspect the state of a running peer.

proxy/coco/src/peer/run_state.rs
```diff
@@ -278,45 +381,87 @@ impl RunState {
...
-        cmds
+                if let Some(urn) = self.waiting_room.next_query(Instant::now()) {
+                    cmds.push(Command::Request(RequestCommand::Query(urn)));
+                }
+                if let Some(url) = self.waiting_room.next_clone() {
+                    cmds.push(Command::Request(RequestCommand::Clone(url)));
+                }
+                cmds
+            },
+            // FIXME(xla): Come up with a strategy for the results returned by the waiting room.
```
xla, line 424: @FintanH What's ur take on the Results here and how to preserve semantics?

proxy/coco/src/peer/control.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use std::time::Instant;
+
+use tokio::sync::{mpsc, oneshot};
+
+use librad::uri::RadUrn;
+
+use crate::request;
+
+/// Reqests sent to the peer.
```
NunoAlexandre, line 11:
```suggestion
/// Requests sent to the peer.
```

proxy/coco/src/peer/control.rs
```diff
@@ -0,0 +1,61 @@
...
+        RadUrn,
+        Instant,
+        oneshot::Sender<Option<request::SomeRequest<Instant>>>,
+    ),
+}
+
+/// Returned responses from the peer.
+#[derive(Debug)]
+pub enum Response {
+    /// Response to a urn request.
```
NunoAlexandre, line 25:
```suggestion
    /// Response to a `Request::Urn` request.
```

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
+//! Management of subroutines.
```
NunoAlexandre, line 1: Could you expand this doc? What problems does it address and what should/should NOT go in here would be nice to add

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
...
+        RequestInput, RunState, SyncInput, TimeoutInput,
+    },
+    sync, RECEIVER_CAPACITY,
+};
+
+/// Management of "subroutine" tasks.
+#[allow(clippy::missing_docs_in_private_items)]
+pub struct Subroutines {
+    pending_tasks: FuturesUnordered<SpawnAbortable<()>>,
+    inputs: SelectAll<BoxStream<'static, Input>>,
```
NunoAlexandre, line 41: Could you document these fields? What is `inputs` and input for? for `pending_tasks`? I'd say no but don't know

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
...
+    state: State,
+    store: kv::Store,
+    run_state: RunState,
+
+    input_sender: mpsc::Sender<Input>,
+    subscriber: broadcast::Sender<Event>,
+}
+
+impl Subroutines {
+    /// Constructor.
```
NunoAlexandre, line 52:
```suggestion
```

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
...
+        state: State,
+        store: kv::Store,
+        run_config: RunConfig,
+        protocol_events: BoxStream<'static, ProtocolEvent<Gossip>>,
+        subscriber: broadcast::Sender<Event>,
+        control_receiver: mpsc::Receiver<control::Request>,
+    ) -> Self {
+        let announce_timer = interval(run_config.announce.interval);
+        let waiting_room_timer = interval(run_config.waiting_room.interval);
+
```
NunoAlexandre, line 63:
```suggestion
```

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
...
+        let (input_sender, inputs) = mpsc::channel::<Input>(RECEIVER_CAPACITY);
+
```
NunoAlexandre, line 65:
```suggestion
```

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
...
+        let run_state = RunState::from(run_config);
+
```
NunoAlexandre, line 67:
```suggestion
```

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
...
+                    return Poll::Ready(Err(e));
+                },
+                Poll::Ready(Some(Ok(()))) => continue,
+                // Either pending, or FuturesUnordered thinks it's done, but
+                // we'll enqueue new tasks below
+                Poll::Ready(None) | Poll::Pending => break,
+            }
+        }
+
+        loop {
```
NunoAlexandre, line 135: Could you document this loop?

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
...
+            log::error!("announce error: {:?}", err);
+            sender
+                .send(Input::Announce(AnnounceInput::Failed))
+                .await
+                .ok();
+        },
+    }
+}
+
+/// Control response subroutine.
```
NunoAlexandre, line 202: These docs don't add much since we already have the module and the fn name. as a reader I'd better like to be informed of what these do and when they are useful.

proxy/coco/tests/common.rs
```diff
@@ -146,6 +138,7 @@ pub fn radicle_project(path: PathBuf) -> project::Create<PathBuf> {
     }
 }
 
+#[allow(dead_code)]
```
NunoAlexandre, line 141: delete?

NunoAlexandre (commented):
I like the overall technical decisions :v:

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,280 @@
...
+                // Either pending, or FuturesUnordered thinks it's done, but
+                // we'll enqueue new tasks below
+                Poll::Ready(None) | Poll::Pending => break,
+            }
+        }
+
+        loop {
+            match self.inputs.poll_next_unpin(cx) {
+                Poll::Ready(Some(input)) => {
+                    log::debug!("handling subroutine input: {:?}", input);
```
geigerzaehler, line 136: For understanding this method and keeping it focused on the concurrency logic it might be beneficial to extract this block into a separate function.

proxy/coco/tests/common.rs
xla: It's used in the integration tests. Haven't found a way for the compiler not to claim it's dead code.

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,278 @@
...
+    run_state::{
+        AnnounceInput, Command, Config as RunConfig, ControlCommand, Event, Input, RequestCommand,
+        RequestInput, RunState, SyncInput, TimeoutInput,
+    },
+    sync, RECEIVER_CAPACITY,
+};
+
+/// Management of "subroutine" tasks.
+#[allow(clippy::missing_docs_in_private_items)]
+pub struct Subroutines {
```
geigerzaehler, line 39: If I understand this structure correctly, it does two things: It handles input events and the business logic around them and it manages tasks and concurrency. Maybe we could separate these two things more clearly. More concretely we could have one logical thread that takes input events and builds futures that are then pushed on a worker to be executed as tasks. The worker would be responsible for spawning tasks and terminating when one of the tasks errors. It would also ensure that all tasks are aborted on an error

proxy/coco/src/peer/control.rs
```diff
@@ -0,0 +1,61 @@
...
+    Urn(
```
kim, line 26: Took me a moment to figure out -- nice!

proxy/coco/src/request/waiting_room.rs
```diff
@@ -150,14 +148,14 @@ impl<T, D> WaitingRoom<T, D> {
     ///
     /// If there is no such `urn` then it create a fresh `Request` using the `urn` and `timestamp`
     /// and it will return `None`.
-    pub fn request(&mut self, urn: RadUrn, timestamp: T) -> Option<SomeRequest<T>>
+    pub fn request(&mut self, urn: &RadUrn, timestamp: T) -> Option<SomeRequest<T>>
```
FintanH, line 151: Out of curiosity, why change this?

proxy/coco/src/peer/run_state.rs
```diff
@@ -104,11 +156,21 @@ pub enum RequestEvent {
         /// The reason the clone failed.
         reason: String,
     },
+    /// Query the network for the `RadUrn`.
+    Queried(RadUrn),
+    /// A urn was requested.
+    Requested(
+        RadUrn,
+        Instant,
+        Option<oneshot::Sender<Option<SomeRequest<Instant>>>>,
```
kim, line 165: When can this be `None`?

kim (commented):
Nice work!

proxy/coco/src/bin/generate-key.rs
```diff
@@ -0,0 +1,15 @@
+use std::{fs::File, io::prelude::*};
+
+use librad::{keys::SecretKey, peer::PeerId};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
xla: Accidental, deleted it in a follow-up comment.

proxy/coco/src/peer/subroutines.rs
xla: To clarify are you suggesting to introduce something like `handle_input` on `Subroutines` to reduce the size of this method body?

proxy/coco/src/peer/subroutines.rs
xla: > It handles input events and the business logic around them and it manages tasks and concurrency

The business logic is encapsulated in the `RunState`.

> The worker would be responsible for spawning tasks and terminating when one of the tasks errors. It would also ensure that all tasks are aborted on an error

What you describe is the `Subroutines` responsibility already, I don't see what an extra worker abstraction would add to this.

kim: If by "business logic" you mean the morphism `Command -> SpawnAbortable`, then yeah, I also thought about how this could be abstracted out. It's not so obvious how to capture the parameters though (namely `state` and `store`), sans partial application. (Did someone say `TaskFactoryImpl`?)

proxy/coco/src/peer/run_state.rs
```diff
@@ -42,44 +57,83 @@ pub enum Command {
...
+    PeerSynced(PeerId),
+    /// Request fullfilled with a successful clone.
+    RequestCloned(RadUrl),
+    /// Requested urn was queried on the network.
+    RequestQueried(RadUrn),
+    /// Waiting room interval ticked.
+    RequestTick,
+}
+
+impl MaybeFrom<&Input> for Event {
```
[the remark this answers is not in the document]
FintanH: Are we using `MaybeFrom` in any other places? _Maybe_ we just need a function :shrug:

proxy/coco/src/peer/run_state.rs
xla: We call it a lot when we don't care about the answer, namely in the tests.

proxy/coco/src/peer/run_state.rs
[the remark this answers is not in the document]
xla: Went down that path, but it feels an abuse. It's not the same semantics. I can drop the `trait` if we feel strongly about it.

proxy/coco/src/peer/run_state.rs
FintanH: A very important `Error` result here is the `TimedOut` variant. We will likely want to communicate that back. I'm almost thinking that we could just communicate the `Result` back on the channel so that the endpoint can also communicate that error.

What do you think?

xla: Could you expand which of the calls below returns `TimedOut`? Sounds almost more important to react to the returns inline here and change the state machine semantics. Currently the only input that carries a channel is the `Requested` variant.

proxy/coco/src/peer/subroutines.rs
xla: Could we expand for what we are optimising for? Do we want to be able to test this code block in isolation? Are we keen to find static guarantees for the `Command -> SpawnAbortable` mapping?

There is already quite a lot of infrastructure involved to run the peer machinery which tips the noise-to-signal ratio unfavourably, so any more of it should give us sick gains.

proxy/coco/src/peer/run_state.rs
FintanH: > Could you expand which of the calls below returns TimedOut?

`queired` and `cloning` do -- check out their docs

> Sounds almost more important to react to the returns inline here and change the state machine semantics. Currently the only input that carries a channel is the Requested variant.

Ah right, wasn't thinking through fully. It's more like that we'll want to report the `TimedOut` errors when we hook up notifications -- as well as the successful cases of "found peer" and "cloned repo".

On top of that, logging warnings could be useful to know when we're requesting things with the wrong state transitions.

proxy/coco/src/request/waiting_room.rs
xla: It was inconsistent with the other methods. Was there a semantic significance to expect a value instead of a reference?

FintanH: I can't remember :sweat_smile: Thanks for making it more consistent :+1:

proxy/coco/src/peer/subroutines.rs
kim: Extensibility. If you want to add a new command, you need to change code in at least three places, one of which is `poll` -- which should not at all change, but the reviewer will have to look at the diff to convince herself of this being the case.

kim: I guess moving the pattern match to an inherent method would be good enough for that

xla: a087b84

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,278 @@
...
+                    return Poll::Ready(Err(e));
+                },
+                Poll::Ready(Some(Ok(()))) => continue,
+                // Either pending, or FuturesUnordered thinks it's done, but
+                // we'll enqueue new tasks below
+                Poll::Ready(None) | Poll::Pending => break,
+            }
+        }
+
+        loop {
```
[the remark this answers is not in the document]
xla: I have not forgotten https://github.com/radicle-dev/radicle-upstream/pull/1009#discussion_r501277948 - can you think of a way to reliably trigger that case, so we can assert the hand-wired `wake_by_ref` operates as we like?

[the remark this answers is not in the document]
kim: How do you mean that? I'm not sure there is a way to build a test case around this. But actually we can ask `FuturesUnordered::is_empty`, so all we need is to conditionally wake on that in the `Poll::Pending` branch.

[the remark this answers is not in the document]
xla: fce40f5


dunno what I'm doing

proxy/coco/src/peer/subroutines.rs
geigerzaehler: > If by "business logic" you mean the morphism `Command -> SpawnAbortable`, then yeah, I also thought about how this could be abstracted out.

Yes, this is (almost) what I meant. I was talking about `Command -> impl Future`.

> Could we expand for what we are optimising for?

Separation of concerns. The concurrency/worker logic is not easy to understand follow, so keeping it separate aids that. Moreover we could write tests for the worker more easily. Don’t let this be a blocker, though.

proxy/coco/src/peer/run_state.rs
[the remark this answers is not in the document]
geigerzaehler: I’d strongly favor removing this abstraction but will leave it up to you.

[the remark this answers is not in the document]
FintanH: I'd second that because it's only used in one place. If you wanted to abstract we could do our favourite impl trick!
```
impl From<&Input> for Option<Event>`
```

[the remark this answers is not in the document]
FintanH: Oh wait, no you can't

proxy/coco/src/peer/gossip.rs
```diff
@@ -0,0 +1,36 @@
...
+            urn: urn.clone(),
+            rev: Some(Rev::Git((*rev).into())),
+            origin: None,
+        })
+        .await;
+}
+
+/// Emit a [`Gossip`] request for the given `urn`.
+///
+/// # Errors
```
FintanH, line 24: There are no errors :eyes:

proxy/coco/src/peer/run_state.rs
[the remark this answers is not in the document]
xla: @FintanH I tried that, dunno how to call `from` correctly with that impl.

FintanH (commented):
One moooore thing

proxy/coco/src/peer/run_state.rs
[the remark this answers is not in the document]
FintanH: Ya, I was wrong anyway because we can't impl `From`/Into` for `Option`

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,299 @@
...
+                    for cmd in self.run_state.transition(input) {
+                        let task = self.spawn_command(cmd);
+
+                        self.pending_tasks.push(task);
+                    }
+                },
+                Poll::Ready(None) => return Poll::Ready(Ok(())),
+                Poll::Pending => {
+                    if !self.pending_tasks.is_empty() {
+                        cx.waker().wake_by_ref()
```
geigerzaehler, line 191: I believe this will make the executor spin and repeatedly poll this future. Let’s assume we have one task in `self.pending_tasks` that is blocked by IO and will return `Poll::Pending` for the next couple of seconds. Let’s also assume that there are now inputs arriving on `self.inputs`. If you follow the code in this method if will always break out of the first loop that calls `self.pending_tasks.poll_next_unpin(cx)` and land here. Since we are calling the waker, the task that runs this future will schedule the `poll()` here to be called again. This brings us back to the begining. And since nothing has changed (we’re still waiting on IO from the other task) we will be stuck in a loop.

The code wants to address the following issue

>  Once self.inputs returns Pending and we did enqueue new tasks, we should summon the waker before returning Pending.

This is handled by `FuturesUnordered`. We are passing the context to `FuturesUnordered::poll_next_unpin()`. If one of the tasks in `self.pending_tasks` finishes, `FuturesUnordered` will call the waker which ensures that this future is polled again.

proxy/coco/src/peer/subroutines.rs
```diff
@@ -0,0 +1,299 @@
...
+                        self.subscriber.send(event).ok();
+                    }
+
+                    for cmd in self.run_state.transition(input) {
+                        let task = self.spawn_command(cmd);
+
+                        self.pending_tasks.push(task);
+                    }
+                },
+                Poll::Ready(None) => return Poll::Ready(Ok(())),
```
geigerzaehler, line 188: Is it correct that we want to abort all running tasks once the input stream stops?

proxy/coco/src/peer/subroutines.rs
kim: I don't think your reasoning is correct:

a. There is no "make the executor spin" -- calling `wake_by_ref` puts us on the back of the poll queue, nothing more. Iow, its semantics is "yield".
b. Morally, we **want** to recurse here, because our `FuturesUnordered` was most likely empty before we entered the second `loop` -- ie. it returned `Ready` from being polled. However, since we **know** that we enqueued new tasks, we must ensure that we are polled again soonish, in order to poll the inputs queue.
c. If there are no new inputs **nor** have enqueued new tasks, we don't wake -- because there's nothing new we can expect when polled again. So, we only return `Pending`.

proxy/coco/src/peer/subroutines.rs
kim: I think so, because the only way the stream can signal `Ready` is because all the channels the `SelectAll` is closing over return `Ready`. This can only happen when all their senders have been dropped.

proxy/coco/src/peer/subroutines.rs
kim: Oh I see, we want to distinguish between `Pending` and `Ready` as the result of polling the `FuturesUnordered` -- it it was `Pending`, it wasn't empty before.

geigerzaehler: The second part of my analysis was wrong. If `pending_tasks.poll_next_unpin()` returns `Poll::Ready(None)` we will need to ensure that `Subroutine::poll` will be called again if we push an item to `pending_tasks`. This is because `FuturesUnordered` will not wake up any task when the pushed future finishes. I guess this was the point that you made, @kim.

> Oh I see, we want to distinguish between `Pending` and `Ready` as the result of polling the `FuturesUnordered` -- it it was `Pending`, it wasn't empty before.

Indeed. It is exactly this case that makes the executor spin. The executor will poll every future on its queue and these polls are inexpensive and don’t block. So even if we are at the back and there are a lot of other futures queued we will call `Subroutine::poll` again pretty quickly. And if nothing has changed (in contrast to `yield`) we will put the future on the queue again.

xla: eeacd3c

proxy/coco/src/peer/subroutines.rs
geigerzaehler: I’m just wondering whether we want to wait for all the tasks to be finished.

proxy/coco/src/peer/run_state.rs
```diff
@@ -199,29 +284,59 @@ impl From<Config> for RunState {
             config,
             connected_peers: HashSet::new(),
             status: Status::Stopped(Instant::now()),
+            waiting_room: WaitingRoom::new(waiting_room::Config::default()),
```
FintanH, line 287: The default here will set the delta to `0` because I believe that's the default for `Duration`. We should probably define `Duration` that makes more sense as a `const` to set it as, or pass it from the `Config`.

kim (dismissed):
image

proxy/coco/src/peer/run_state.rs
```diff
@@ -199,29 +284,59 @@ impl From<Config> for RunState {
...
 impl RunState {
     /// Constructs a new state.
     #[cfg(test)]
-    const fn new(config: Config, connected_peers: HashSet<PeerId>, status: Status) -> Self {
+    fn new(config: Config, connected_peers: HashSet<PeerId>, status: Status) -> Self {
         Self {
             config,
             connected_peers,
             status,
+            waiting_room: WaitingRoom::new(waiting_room::Config::default()),
```
FintanH, line 300: But then this one is safe I would say because it's useful to have no delta during testing.

FintanH (commented):
Sorry, should have spotted this one sooner

proxy/coco/src/peer/run_state.rs
xla: fe90e77

geigerzaehler (approved):

FintanH (approved):
:ship: it

kim:
> I’m just wondering whether we want to wait for all the tasks to be
> finished.

Initially, I wrote this to poll the task results queues only once, but that
risks missing events from expiring timers -- in that case, we may not have any
pending tasks, nor pushed any new ones. So, we'd just return `Pending`, and
there's no way to know for the runtime that it should wake _our_ poll when a new
event is available.

It is potentially that this "subroutine" loop is just not a good fit for async
Rrrrrrrust -- there isn't really a reason to yield back to the executor unless
all in-flight tasks are blocked on IO, or we hit the circumstance that there's
no input to react to (neither from outside triggers nor timers). In a sane
environment, one would implement this thing simply as an OS thread, and never
care about whether it blocks.

PR #993, opened 2020-10-05
refactor(proxy): module errors

### LINKED ISSUE #850: Refactor proxy errors
Currently we maintain general error modules, which violates our dependency boundaries as sub modules have to import the top-level mod. Instead we should have per module error types.

FintanH:
Fixes #850 

The aim of the game is to have `Error` enums located next to their module, splitting up the mega enum. Some advantages to this approach:
* This minimises the chance of cross-module interactions since we can't conflate errors of the same domain
* Where the error comes from is more specific, since we would get something like `State(Source(Git(...)))` vs `State(Git(...))`. We know the former is a git related error located in source browsing vs a git related error during state interactions.

### DIFF
--- a/proxy/api/src/error.rs
+++ b/proxy/api/src/error.rs
@@ -1,84 +1,23 @@
 //! Proxy library errors usable for caller control flow and additional context for API responses.
 
-use std::time::SystemTimeError;
-
-/// Project problems.
-#[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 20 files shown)


NunoAlexandre (approved):
sick :ok_hand:

rudolfs (approved):

xla (commented):
errors

🤖 🍁 🍷 🐻

PR #994, opened 2020-10-05
refactor(proxy): move De/Serialize

FintanH:
We can utilise serde's macro magic using the `with` and `bound` macros
to tell serde that we'll use the Milliseconds trait for serialisation.

Note that this enforces to use the time types defined in serde_millis
when we want serialise any of these types. Otherwise, we will have to
hand-roll implementations.

### DIFF
--- a/proxy/coco/src/request.rs
+++ b/proxy/coco/src/request.rs
@@ -10,11 +10,10 @@
 use std::{
     collections::HashMap,
     ops::{Deref, Sub},
-    time::Instant,
 };
 
 use either::Either;
-use serde::{ser::SerializeStruct, Deserialize, Serialize};
+use serde::{Deserialize, Serialize};
 
 use librad::{
     net::peer::types::Gossip,
@@ -22,10 +21,10 @@ use librad::{
     uri::{self, RadUrl, RadUrn},
 };
 
-pub mod states;
-pub use states::*;
 pub mod existential;
 pub use existential::SomeRequest;
+pub mod states;
+pub use states::*;
 pub mod waiting_room;
 
 /// Private trait for sealing the traits we use here.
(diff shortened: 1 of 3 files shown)


kim (dismissed):
lgtm. I suppose there's some anti-macro policy in effect?

FintanH:
`Instant` doesn't have impls so we have to go through the `serde_millis` crate. Yet another case of traits/typeclasses being unergonomic for composition across package boundaries 😭

kim:
> `Instant` doesn't have impls so we have to go through the `serde_millis`
> crate.

If that's the only reason, you can also just add `#[serde(with = "serde_millis")]` to the timestamp field.

FintanH:
Do you think that will work when we have the generic parameter for timestamp? 👀 I can try it today

kim:
> generic parameter for timestamp

Yes, but you'll have to tell serde about a trait bound, e.g.:

    #[serde(with = "serde_millis", bound = "T: serde_millis::Milliseconds")]

That trait is sealed, though, so if you need to instantiate `T` to something
else (and still need the serde impls), you'll need to make a wrapper around
`serde_millis` which accepts types convertible to `serde_millis::Milliseconds`
(via your own trait).

FintanH:
Ya, so this works for `Request` but then to derive `Serialize` on `SomeRequest` it wants `T` to be bound with `Milliseconds` which is annoying. I want to avoid having time types as the `T` in tests so that I don't have to worry about accounting for it.

I guess I'll have to implement the `De/``Serialize` for `SomeRequest` manually :thinking:

kim (commented):
You now have

```rust
impl Serialize for SomeRequest<Instant>
```

and the generated code should be

```rust
impl<T: serde_millis::Milliseconds> Serialize for SomeRequest<T>
```

So, the net effect should be the same -- if you use `SomeRequest<()>`, you just
don't have a `Serialize` impl in scope. Or no?

FintanH:
Ya, sorry in my local changes when I smack the `derive(Serialize, Deserialize)` on `SomeRequest` it wants me to add the bound to the `T` directly, i.e.

```rust
#[derive(Deserialize, Serialize)]
enum SomeRequest<T: serde_millis::Millisecond>
```

Which wouldn't allow me to do `SomeRequest<()>`.

I'm finding `serde` quite unintuitive when it comes to their API :(

FintanH:
I think I was being dense. I just needed to add `bound =` to the top of `SomeRequest` as well.

FintanH:
Thanks for the help here @kim :bow:

cloudhead (approved):
Much cleaner :+1:

proxy/coco/src/request.rs
```diff
@@ -59,65 +58,19 @@ mod sealed;
...
 /// API. Note that it makes it easy to test by just choosing `()` for the timestamp.
-#[derive(Clone, Debug, PartialEq, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Request<S, T> {
     /// The identifier of the identity on the network.
     urn: RadUrn,
     /// The number of attempts this request has made to complete its job.
     attempts: Attempts,
     /// The timestamp of the latest action to be taken on this request.
+    #[serde(with = "serde_millis", bound = "T: serde_millis::Milliseconds")]
```
sarahscott, line 68: 🎉

sarahscott (approved):

Issue #840, opened 2020-08-28
Unify shareable identifiers across app


rudolfs:
@cloudhead, @brandonhaslegs and I decided that the format of shareable identifiers for β1 in Upstream should be like this:

Shown in the UI: `@hyndb5gs95gwtsf37tncz4ag3wqrg4ejw3qqga6x1srw9jp8jw59d6.git`
Copied to clipboard: `@hyndb5gs95gwtsf37tncz4ag3wqrg4ejw3qqga6x1srw9jp8jw59d6.git#rudolfs`
Passed down as URN to proxy: `rad:git:hyndb5gs95gwtsf37tncz4ag3wqrg4ejw3qqga6x1srw9jp8jw59d6`
👇 
```json
{
  "displayName": "rudolfs",
  "urn": "rad:git:hyndb5gs95gwtsf37tncz4ag3wqrg4ejw3qqga6x1srw9jp8jw59d6"
}
```

Figma design is [here](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=3372%3A10405).

xla:
> Passed down as URN to proxy: `rad:hyndb5gs95gwtsf37tncz4ag3wqrg4ejw3qqga6x1srw9jp8jw59d6.git#rudolfs`

This is not a valid `RadUrl` afaik. Again I urge for us to devise an upstream url/urn scheme which can be translated to and from the underlying protocol. We are doing us a disservice by not approaching this structurally.

Also what's up with the random `@` prefixes?

cloudhead:
Passed down to proxy should be `rad:git:<id>` without the `#rudolfs`

cloudhead:
> Also what's up with the random @ prefixes?

This identifies a random string of alphanumerics as an identity.

rudolfs:
> Passed down to proxy should be rad:git:<id> without the #rudolfs

@cloudhead: But if we don't pass it down when searching, then we would not be able to easily tell what is queued up for tracking in this screen:

<img width="785" alt="Screenshot 2020-08-28 at 13 51 35" src="https://user-images.githubusercontent.com/158411/91558022-d12ac100-e935-11ea-9951-c4628d197890.png">

cloudhead:
We can pass the identifier separately from the URN, I don't think that should be a problem.

rudolfs:
[...]

PR #995, opened 2020-10-06
refactor(ui): source browsing

### LINKED ISSUE #970: Project readme flickers on branch switch
When switching branches via the revision selector while on the README file, it is loaded twice which causes flickering. 
Visuals can be found [here](https://github.com/radicle-dev/radicle-upstream/pull/962#issuecomment-698894915)

sarahscott:
**Why?**

When we created `Remote.svelte`, we didn't have the option to type-check data in components. To enable its usage as we migrate from JS -> TS, I [asserted](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions) an `any` type [here](https://github.com/radicle-dev/radicle-upstream/blob/master/ui/DesignSystem/Component/Remote.svelte#L10):
```ts
  export let store: Readable<remote.Data<any>>;
```

...which essentially destroys any notion of type safety in the `data` value (YOLO🤪). While this is convenient, it doesn't allow us to use the magic of the type system in the most critical places - i.e. when we're rendering data that we expect to be of a certain shape.

The `Remote.svelte` component does make it very easy to load `slot`s based on the remote store status (`Loading`/`Success`/`Error`), which I think is ergonomic and useful and can perhaps be its only function until we have [generics in svelte](https://github.com/sveltejs/rfcs/pull/38). With this in mind, I'd like to transition away from the `let:data={something}` pattern and towards one where we know exactly what the `T` in  `Data<T>` is before we render components.

To kick off this effort, I've typed the `Project` screen and its associated components, created a single source of truth for the project -> peer -> revisions -> commits data flow, and added a lot of comments to clarify what exactly these types are supposed to be.

Closes #970


MeBrei (changes requested):
Great effort tackling the refactoring of this code! When working on it, I also felt the need for it, but didn't quite know where to start.

I am a bit worried, changing this code while we are not able to test it properly. Tests related to the peer selector are currently skipped and we also can't test that in the app directly.

I also found the following regression:

- [x] when switching from one project to another the page of the second project does not load and errors with the following:

Screenshot 2020-10-06 at 10 46 46

MeBrei:
It seems like your changes also solved #970

ui/src/source.ts
```diff
@@ -112,15 +117,24 @@ export enum RevisionType {
...
 }
 
+// Client representation of a git tag
 export interface Tag {
   type: RevisionType.Tag;
   name: string;
+  // The id of the project this tag belongs to
+  peerId: string;
+  // The id of the project this tag belongs to
+  projectId: string;
```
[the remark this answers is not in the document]
sarahscott: I think they should stay separate for now because the `currentRevision` can change independently of the `currentPeerId` - e.g. if the user selects a revision that belongs to the same peer.

sarahscott:
@MeBrei fixed the regression! I'd say it's the perfect time to strongly type everything peer-related so it's easier to reason about when we have actual peers to test 😁

cypress/integration/project_source_browsing.spec.js
```diff
@@ -317,6 +317,9 @@ context("source code browsing", () => {
     });
 
     it("allows navigating the tree structure", () => {
+      // Ensure the file has loaded before navigating away.
```
sarahscott, line 320: @MeBrei @rudolfs good catch here, this is way better than `cy.wait()`!

ui/Screen/Project.svelte
```diff
@@ -127,52 +169,56 @@
...
-        <div slot="left">
-          <div style="display: flex">
-            <div class="revision-selector-wrapper">
-              {#if $currentPeerId}
-                <RevisionSelector
-                  currentPeerId={$currentPeerId}
-                  bind:currentRevision={$currentRevision}
-                  {revisions} />
+  <Remote {store} context="project">
+    {#if project}
```
MeBrei, line 173: Why do we need this check? I thought that the slot (without name) in `Remote` is only rendered on `Success`.

MeBrei (commented):
Looks good to me! Just added a comment for understanding the new way of using remote.

I think the tests are failing, because some `contains` have to be replaced with `pick` and using data-cy. The workaround for `wait` only works with `get` / `pick`. (One example is trying to do `cy.contains(".i-am-well-hidden").click();` in the source browsing specs.

ui/src/identity.ts
```diff
@@ -14,6 +14,8 @@ export interface Avatar {
 
 export interface Identity {
   id: string;
+  peerId: string;
+  urn: string;
```
FintanH, line 18: I'd be verrrry interested in seeing a `RadUrn` type at some point, but maybe not for now.

ui/src/source.ts
```diff
@@ -112,15 +117,24 @@ export enum RevisionType {
   Sha = "sha",
 }
 
+// Client representation of a git branch
 export interface Branch {
   type: RevisionType.Branch;
   name: string;
-  peerId?: string;
+  // The id of the project this branch belongs to
+  projectId: string;
```
FintanH, line 125:
```suggestion
  projectUrn: string;
```

ui/src/source.ts
FintanH, line 137:
```suggestion
  projectUrn: string;
```

ui/src/source.ts
```diff
@@ -140,24 +163,56 @@ export const commits = commitsStore.readable;
...
+
+    const peerId = defaultRevisions.identity.peerId;
+    updateCurrentPeerId({ peerId });
+
+    // Now that we have a peer, set the current revision to the default branch.
+    // If not found, use the first branch returned from proxy.
+    const defaultRevision =
+      defaultRevisions.branches.find(
+        branch => branch.name === defaultRevisions.defaultBranch
+      ) || defaultRevisions.branches[0];
```
FintanH, line 182: And in the event that there are no branches?

ui/Screen/Project/PeerSelector.svelte
```diff
@@ -45,11 +36,13 @@
...
   const dispatch = createEventDispatcher();
-  const selectPeer = peerId => {
+  const selectPeer = (peerId: string) => {
     hideDropdown();
-    currentPeerId = peerId;
     dispatch("select", { peerId });
   };
+
+  $: currentPeer =
+    availablePeers.find(peer => peer.id === currentPeerId) || availablePeers[0];
```
FintanH, line 45: Again, what happens if there are no `availablePeers`?

FintanH (commented):
Mostly observing :eyes:

ui/DesignSystem/Component/HorizontalMenu.svelte
```diff
@@ -12,7 +12,7 @@
     title: string;
     href: string;
     looseActiveStateMatching: boolean;
-    counter?: string;
+    counter?: string | number;
```
xla, line 15: Why is a string allowe here?

ui/DesignSystem/Component/HorizontalMenu/MenuItem.svelte
```diff
@@ -7,7 +7,7 @@
   export let icon: typeof SvelteComponent;
   export let title: string;
   export let active: boolean;
-  export let counter: string | undefined;
+  export let counter: string | number | undefined = undefined;
```
xla, line 10: Same as above.

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -1,48 +1,26 @@
-<script>
-  import { getContext } from "svelte";
+<script lang="ts">
+  import { createEventDispatcher } from "svelte";
 
   import { isExperimental } from "../../../../native/ipc.js";
-  import { RevisionType } from "../../../src/source.ts";
+
```
xla, line 5:
```suggestion
```

ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -1,48 +1,26 @@
...
-
   const hideDropdown = () => {
     expanded = false;
   };
 
-  const selectRevision = revision => {
-    currentRevision = revision;
+  const dispatch = createEventDispatcher();
+
+  const selectRevision = (revision: source.Branch | source.Tag) => {
```
xla, line 22: This should probably be a union type to make futue extensibilty simpler.

ui/Screen/Project.svelte
```diff
@@ -113,9 +117,47 @@
...
 
   fetch({ id: projectId });
-  fetchRevisions({ projectId });
-  $: if ($currentRevision)
-    fetchCommits({ projectId, revision: $currentRevision });
+
+  // Workaround for stores: https://github.com/sveltejs/language-tools/issues/493
+  // TODO(sos): should be removed if/when this is fixed.
+  let revisions: source.PeerRevisions[] | undefined,
+    project: Project | undefined,
```
xla, line 124: Not too excited about this multi-line let. Makes it hader to edit a line and doesn't seem to be used anywhere else.

ui/src/transaction.ts
```diff
@@ -417,6 +417,8 @@ export const unknownOrg: Org = {
 // Identity counter part of `unknownOrg`.
 export const unknownIdentity: Identity = {
   id: "user@unknown.git",
+  peerId: "",
```
xla, line 420: Shouldn't `transaction.ts` be gone by now? Since we dropped wallet related functionality from mainline.

xla (changes requested):
Dope! Currently we can't just `export let store: Readable<remote.Data<T>>;` the situation?

ui/Screen/Project.svelte
sarahscott: With the old patten (`let:data={project}`), that was the case, but now `project` is supplied by the store itself, which can be `Project | undefined`.

Also, `Success` doesn't necessarily mean there is data - it could be possible in some cases for endpoints to return empty arrays successfully (e.g. the `projects/` endpoint could return `[]` if the user has no projects).

ui/src/identity.ts
sarahscott: Soon

ui/src/transaction.ts
sarahscott: Is it time to delete it? I have no objections to that. It might also necessitate deleting or updating all of the UI components that use `transaction.ts` though.

sarahscott: cc @NunoAlexandre because he's working on funding

ui/DesignSystem/Component/HorizontalMenu.svelte
sarahscott: It shouldn't be

ui/Screen/Project/PeerSelector.svelte
sarahscott: The `PeerSelector` wouldn't render in that case - this would mean that proxy didn't return any `Revisions`, or that it returned `Revisions` without an `Identity`, which I don't think it's supposed to do 🤔 .

Alternatively we could throw an error, or display the disabled `PeerSelector` with some sort of 'No peers found" messaging.

ui/src/source.ts
sarahscott: In this case, the revision selector won't show up but the project will still render. Before, the `RevisionSelector` would create a default revision that may or may not actually exist (https://github.com/radicle-dev/radicle-upstream/pull/995/files#diff-ac6e05ed01605cc8ea4ba88d402e6541L30).

Should we throw an error instead? I don't really like the idea of creating a revision in ui that we don't actually receive from proxy.

ui/src/source.ts
sarahscott: This is worthy of its own pr because `Project` has an `id` in proxy - relevant to #840

sarahscott:
> Dope! Currently we can't just export let store: Readable<remote.Data<T>>; the situation?

@xla we can't export generics from svelte components 😢

ui/Screen/Project/PeerSelector.svelte
FintanH: > which I don't think it's supposed to do .

Don't assume external resources will _always_ do the thing they're supposed to. That's what I'm trying to get at here: we should think about failure cases and show appropriate errors when possible. Indexing arrays is a code smell in my eyes.

ui/src/source.ts
FintanH: Ya, I definitely agree that creating a made-up revision doesn't make sense. Could we have like an error window in the selector saying something like, "We couldn't find any revisions o.O"?

Again, some sort of error handling would be good to think about when we see partial functions like these :)

xla:
Took this on and got it up-to-speed. PTAL.

rudolfs:
Steps to repro:

1. add `radicle-upstream` as a project
2. click on a file in the file browser, for example `CHANGELOG.md`
3. navigate back to your profile by clicking the profile icon in the sidebar
4. go back to the project by clicking on `radicle-upstream` in the project list

Expected result: see the project source.
Actual result: exception in the console.

Screenshot 2020-10-19 at 09 19 06

ui/Screen/Project/PeerSelector.svelte
FintanH: @xla do you plan to do anything about this in this PR, out of interest?

ui/src/source.ts
```diff
@@ -112,25 +117,46 @@ export enum RevisionType {
...
+  projectId: string;
 }
 
 export interface Sha {
   type: RevisionType.Sha;
   sha: string;
 }
 
+// Currently supported revision types in UI:
+export type SupportedRevision = Branch | Tag;
```
FintanH, line 146: Is `Sha` not supported? :thinking: Or what's the issue here?

ui/src/source.ts
FintanH: Same question here about whether we'll address these here.

ui/src/source.ts
```diff
@@ -140,24 +166,56 @@ export const commits = commitsStore.readable;
...
 export const objectType = writable(ObjectType.Tree);
 export const resetObjectType = (): void => objectType.set(ObjectType.Tree);
 export const objectPath = writable(null);
 export const resetObjectPath = (): void => objectPath.set(null);
-export const currentRevision = writable(null);
-export const resetCurrentRevision = (): void => currentRevision.set(null);
-export const currentPeerId = writable(null);
-export const resetCurrentPeerId = (): void => currentPeerId.set(null);
+
+export const currentRevision = writable<Branch | Tag | undefined>(undefined);
```
FintanH, line 196:
```suggestion
export const currentRevision = writable<SupportedRevision | undefined>(undefined);
```

We could re-use the above `type` right?

ui/src/source.ts
```diff
@@ -170,12 +228,13 @@ interface FetchCommit extends event.Event<Kind> {
 interface FetchCommits extends event.Event<Kind> {
   kind: Kind.FetchCommits;
   projectId: string;
-  revision: Branch;
+  revision: Branch | Tag;
```
FintanH, line 231:
```suggestion
  revision: SupportedRevision;
```

ui/src/source.ts
```diff
@@ -187,6 +246,16 @@ interface FetchObject extends event.Event<Kind> {
...
 }
 
+interface UpdateCurrentPeerId extends event.Event<Kind> {
+  kind: Kind.UpdateCurrentPeerId;
+  peerId: string;
+}
+
+interface UpdateCurrentRevision extends event.Event<Kind> {
+  kind: Kind.UpdateCurrentRevision;
+  revision: Branch | Tag;
```
FintanH, line 256:
```suggestion
  revision: SupportedRevision;
```

Issue #984, opened 2020-10-01
Wire up waiting room with search


FintanH:
In our haste to wire up the search feature, we cut a few corners in https://github.com/radicle-dev/radicle-upstream/pull/969.

Tasks that come to mind that we need to do:
- [ ] Load the `WaitingRoom` on startup
- [ ] Save the `WaitingRoom` (but not sure where and when)
- [x] `request` endpoint should be `PUT` [#1026]
- [x] Check that URN exists in the monorepo before searching [#1030]
- [x] Return the request metadata to the frontend [#1103]
- [x] show list of requests [#1104]
- [x] update progress of requests [#1117]
- [x] emit completion notification [#1117]

xla:
Superseded by #1147 #1148

PR #1004, opened 2020-10-07
feat(proxy): waiting room persistence

sarahscott:
🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 🚧 

Towards #984 

- [x] Implements `Default` for `WaitingRoom`
- [x] Adds `waiting_room` to `session` store
- [x] Loads session `waiting_room` upon startup
- [x] `projects/request/<id>` is now a `POST` request
- [ ] Store the waiting room in the `session` at appropriate intervals
- [ ] Test the `request/` endpoint properly

### DIFF
--- a/proxy/api/src/http/project.rs
+++ b/proxy/api/src/http/project.rs
@@ -54,12 +54,12 @@ fn get_filter(
         .and_then(handler::get)
 }
 
-/// `GET /request/<id>`
+/// `POST /request/<id>`
 fn request_filter(
     ctx: context::Context,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     path("request")
-        .and(warp::get())
+        .and(warp::post())
         .and(http::with_context(ctx))
         .and(path::param::<coco::Urn>())
         .and(path::end())
@@ -167,14 +167,19 @@ mod handler {
         Ok(reply::json(&project::get(&ctx.state, urn).await?))
     }
 
-    /// Kick off a network request for the [`project::Project`] of the given `id`.
+    // TODO(sos): Should there be a `request` module that can take `POST`/`GET` requests for any
+    // `RadUrn` (i.e. `User` | `Project`)?
+    ///
+    ///  Kick off a network request for the [`project::Project`] of the given `id`.
     pub async fn request(ctx: context::Context, urn: coco::Urn) -> Result<impl Reply, Rejection> {
         // TODO(finto): Check the request exists in the monorepo
         let mut waiting_room = ctx.waiting_room.write().await;
-        let _request = waiting_room.request(urn, Instant::now());
+        let request = waiting_room.request(urn, Instant::now());
 
-        // TODO(finto): Serialise request and respond with that.
-        Ok(reply::json(&true))
+        // TODO(finto/sos): Serialise request and respond with that
+        // `waiting_room.request()` returns `None` for new requests by design;
+        // should it return something more substantial?
+        Ok(reply::json(&request))
     }
 
     /// List all projects tracked by the current user.
@@ -248,6 +253,8 @@ pub struct MetadataInput {
 #[allow(clippy::panic, clippy::unwrap_used)]
 #[cfg(test)]
 mod test {
+    use std::time::Instant;
+
     use pretty_assertions::assert_eq;
     use serde_json::{json, Value};
     use warp::{http::StatusCode, test::request};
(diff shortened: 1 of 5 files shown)


xla (changes requested):
Preliminary feedback. It's probably not a session concern to hold onto the waiting room. Persistence can be modeled like the announcements. This will keep the separation and doesn't bloat the session with unrelated information.

proxy/api/src/http/project.rs
```diff
@@ -167,14 +167,19 @@ mod handler {
...
+    ///  Kick off a network request for the [`project::Project`] of the given `id`.
     pub async fn request(ctx: context::Context, urn: coco::Urn) -> Result<impl Reply, Rejection> {
         // TODO(finto): Check the request exists in the monorepo
         let mut waiting_room = ctx.waiting_room.write().await;
-        let _request = waiting_room.request(urn, Instant::now());
+        let request = waiting_room.request(urn, Instant::now());
 
-        // TODO(finto): Serialise request and respond with that.
-        Ok(reply::json(&true))
+        // TODO(finto/sos): Serialise request and respond with that
```
FintanH, line 179: We could have it so that it returns `Either` where `Left` means it's a request that already existed and `Right` means that the request is fresh

proxy/api/src/http/project.rs
```diff
@@ -508,14 +516,34 @@ mod test {
...
+        //         },
+        //         "state": {},
+        //         "timestamp": 1602031207235,
+        //         "urn": "rad:git:hwd1yreyfmtg5ntmm97g8eazmb1uh75xpbta8mdthkcj3ee3yxcci3d8b5y"
+        //     }
+        // });
+
         http::test::assert_response(&res, StatusCode::OK, |have| {
-            assert_eq!(have, json!(true));
+            assert_eq!(have, request);
```
FintanH, line 546: I wonder if we could deserialize the timestamp and use it above with creating the expected request, i.e. something along the lines of:

```rust
        let timestamp = Instant::deserialize(have["timestamp"]).unwrap();
        let request = json!(request::SomeRequest::Created(request::Request::new(
            urn,
            timestamp
        )));
```

proxy/coco/src/request.rs
```diff
@@ -66,7 +66,7 @@ pub struct Request<S, T> {
     attempts: Attempts,
     /// The timestamp of the latest action to be taken on this request.
     #[serde(with = "serde_millis", bound = "T: serde_millis::Milliseconds")]
-    timestamp: T,
+    pub timestamp: T,
```
FintanH, line 69: I'd prefer that we have a `fn timestamp(&self) -> &T`. The reason being that `Request` should generally be a blackbox that we can only peek into.

proxy/coco/src/request/waiting_room.rs
```diff
@@ -83,6 +83,15 @@ pub struct WaitingRoom<T, D> {
     config: Config<D>,
 }
 
+impl<T, D> Default for WaitingRoom<T, D>
```
FintanH, line 86: I actually don't think we should have this because that sets the `Config`'s delta to `0` which we don't want in a production setting. So this is a big foot gun.

proxy/coco/src/request/waiting_room.rs
```diff
@@ -149,16 +158,16 @@ impl<T, D> WaitingRoom<T, D> {
...
     where
         T: Clone,
     {
         match self.get(&urn) {
             None => {
                 let request = SomeRequest::Created(Request::new(urn.clone(), timestamp));
-                self.requests.insert(urn.id, request);
-                None
+                self.requests.insert(urn.id, request.clone());
+                Some(request)
```
FintanH, line 170: Ok, so here I'd like to point out some things I'd like you to think about with regards to types and semantics.
1. If we _always_ return a request then there is no need for an `Option` because we always return `Some` at the end. Essentially the function signature becomes a lie. It says that it's a possibility that this function will fail but it never does.
2. The semantics of this function was that we want to have some signal of when a request was new vs when it already existed. You've now removed those semantics and we never know this information.

I think it's good practice to keep these kinds of things in mind: "What is the type signature saying and is it true?" and "What are the semantics of this function? Am I changing them?" Sometimes you do need to change the signature and semantics but in this case, I think we only need to change the signature to use `Either`, as mentioned in a previous comment.

FintanH (changes requested):
Thanks for making a start on this :+1: 

+1 to @xla's feedback on the storing. We don't want `Session` to become the kitchen sink of serialisation :)

PR #1005, opened 2020-10-07
feat(proxy): track endpoint

### LINKED ISSUE #1006: Track endpoint
Add a `track` endpoint that allows us to track a peer for a given project. That is, given a `RadUrn` and a `PeerId` we should track the `PeerId` for this `RadUrn` in the monorepo by calling `Storage::track`.

FintanH:
Fixes #1006

Adds a `POST /track` endpoint that takes a `RadUrn` and `PeerId` as input and tracks the `PeerId` for the given project at `RadUrn`. 

What this PR doesn't do is attempt to add any calls to the front end for this. It's worth waiting for the design of this feature to do that /cc @juliendonck.

### DIFF
--- a/proxy/api/src/http/project.rs
+++ b/proxy/api/src/http/project.rs
@@ -666,7 +666,11 @@ mod test {
 
         let res = request()
             .method("PUT")
-            .path(&format!("/{}/track/{}", project.id, coco::control::generate_peer_id()))
+            .path(&format!(
+                "/{}/track/{}",
+                project.id,
+                coco::control::generate_peer_id()
+            ))
             .reply(&api)
             .await;
 
--- a/proxy/api/src/http/project.rs
+++ b/proxy/api/src/http/project.rs
@@ -112,15 +112,16 @@ fn discover_filter(
         .and_then(handler::discover)
 }
 
-/// `POST /track/<peer_id>`
+/// `PUT /<urn>/track/<peer_id>`
 fn track_filter(
     ctx: context::Context,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
-    path("track")
-        .and(warp::post())
-        .and(http::with_context(ctx))
+    http::with_context(ctx)
+        .and(warp::put())
+        .and(path::param::<coco::Urn>())
+        .and(path("track"))
+        .and(path::param::<coco::PeerId>())
         .and(path::end())
-        .and(warp::body::json())
         .and_then(handler::track)
 }
 
@@ -227,7 +228,8 @@ mod handler {
     /// Track the peer for the provided project.
     pub async fn track(
         ctx: context::Context,
-        super::TrackInput { urn, peer_id }: super::TrackInput,
+        urn: coco::Urn,
+        peer_id: coco::PeerId,
     ) -> Result<impl Reply, Rejection> {
         ctx.state.track(urn, peer_id).await.map_err(Error::from)?;
         Ok(reply::json(&true))
@@ -265,16 +267,6 @@ pub struct MetadataInput {
     description: String,
     /// Configured default branch.
     default_branch: String,
-}
-
-/// User provided input for which peer to track for which project.
-#[derive(Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
-pub struct TrackInput {
-    /// The project identifier to track.
-    urn: coco::Urn,
-    /// The peer to track for the given project.
-    peer_id: coco::PeerId,
 }
 
 #[allow(clippy::panic, clippy::unwrap_used)]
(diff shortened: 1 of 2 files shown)


proxy/api/src/http/project.rs
```diff
@@ -111,6 +112,18 @@ fn discover_filter(
         .and_then(handler::discover)
 }
 
+/// `Post /tracke/<peer id>`
```
NunoAlexandre, line 115:
```suggestion
/// `POST /track/<peer_id>`
```

proxy/api/src/http/project.rs
```diff
@@ -210,6 +223,15 @@ mod handler {
 
         Ok(reply::json(&feed))
     }
+
+    /// Track the peer for the provided project.
+    pub async fn track(
+        ctx: context::Context,
+        super::TrackInput { urn, peer_id }: super::TrackInput,
```
NunoAlexandre, line 230: Just a suggestion: Import `TrackInput` from `super` and simplify to:

```suggestion
        TrackInput { urn, peer_id }: TrackInput,
```

TIL btw :smiley:

NunoAlexandre (commented):
Looks good! I wonder what happens if a mal-formed URN or peer id is provided from the frontend. In my experience with other endpoints, that will lead to annoying issues where the targeted endpoint can't be resolved (due to type-strictness that finds no matching endpoint), which, depending on how far we can validate those two values, can be more of a pain than a gain.

proxy/api/src/http/project.rs
FintanH: This suggestion goes against the way things are organised in the http handlers in general. I'm operating under the "When in Rome" principle and adhering to the style that currently exists :)

FintanH:
> Looks good! I wonder what happens if a mal-formed URN or peer id is provided from the frontend. 

A lot of our endpoints take a `RadUrn`/`PeerId` already, and I believe the response would be a rejection saying that the body was malformed.

> In my experience with other endpoints, that will lead to annoying issues where the targeted endpoint can't be resolved (due to type-strictness that finds no matching endpoint), which, depending on how far we can validate those two values, can be more of a pain than a gain.

Can you show me these annoyances? :thinking: I'm not quite sure I understand how they would come about. I'm also not sure what the solution would be here, these things are going to have to be type-checked into a `RadUrn` and `PeerId` at some point.

proxy/api/src/http/project.rs
```diff
@@ -111,6 +112,18 @@ fn discover_filter(
         .and_then(handler::discover)
 }
 
+/// `POST /track/<peer_id>`
```
xla, line 115:
```suggestion
/// `PUT /<id>/track/<peer_id>`
```

Somehow the endpoints in here became inconsistent. All endpoints regarding a single resource should be bundled under that resource, the id/urn should be a param as seen in the get filter. And the peer id as well, as the docs currently lie, indicating that the peer id is part of the path.

I'd also suggest this to be a put to indicate that it's idempotent.

xla (changes requested):
See comment about HTTP semantics.

proxy/api/src/http/project.rs
FintanH: Nice :+1: Good suggestions :)

proxy/api/src/http/project.rs
NunoAlexandre: macht Sinn!

NunoAlexandre:
> Can you show me these annoyances?  I'm not quite sure I understand how they would come about. I'm also not sure what the solution would be here, these things are going to have to be type-checked into a `RadUrn` and `PeerId` at some point.

If my memory doesn't fail me, an endpoint involving a transfer of funds would receive an address and an amount. If we had the address parameter strongly-typed, whenever a user would input an ill-formed address, the API call would fail to resolve the appropriate handler, causing a 500 if I am not mistaken. That's because it couldn't find any endpoint where the given request could puzzle in. so we had to have the address being a `String` instead and, in the handler, try to parse it as the appropriate, strong Address type. If it'd fail to, we'd return a 400 error back, which we otherwise couldn't.

So, if this endpoint will be called using data provided by users which can be ill-formed, as long as we can validate it on the UI side we should be good, which was not the case regarding the addresses in the example above.

FintanH:
So I just gave this a quick go, and it turns out if you pass malformed params then it returns a [`405`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405). This seems like makes sense to me.

NunoAlexandre:
> So I just gave this a quick go, and it turns out if you pass malformed params then it returns a [`405`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405). This seems like makes sense to me.

That's cool, though that's not what I had seen on my case. Maybe we have bumped warp? In any case, good to go then :ok_hand:

FintanH:
> That's cool, though that's not what I had seen on my case. Maybe we have bumped warp?

Would be good to write these things down when seen so that we can reproduce :)

xla (approved):
(￣∇￣)

NunoAlexandre (approved):
:black_flag: _Fintan, the tracker_

Issue #999, opened 2020-10-06
Unseal key store with the user provided passphrase


MeBrei:
We want to use the password send from the API to the proxy since #1120 to actually unseal the keystore. At the moment we are still using a hard-coded dummy password.

Issue #998, opened 2020-10-06
Add and use unseal endpoint on app start


MeBrei:
Depends on #997

In this iteration, the signer should not automatically be created on app start, but the user should be prompted to enter the password (for now still the hardcoded passphrase). 
The `unseal` endpoint should set the signer, create a secret, store it in `Context` and send it as a cookie in its response. The ui keeps the cookie in memory and all requests after a successful unseal have to send this information in its header. The signed filter guard checks if it matches the one in `Context`.
The cookie is also set on identity creation.

The design for the password prompt can be found in [figma](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=4911%3A43513)

PR #1008, opened 2020-10-07
feat(proxy): Add unsealed guard

### LINKED ISSUE #997: Add sealed keystore filter guard
This prepares the api endpoints availability depending on the future key store and if it is un-/sealed.

We want to add a guard (similar to `with_owner_guard`) that guards those endpoints only available to unsealed key stores. In the first iteration, the guard checks if the `signer` in `Context` is present, if not `401 Unauthorized` is returned. `signer` will be optional in `Context`

In this first iteration, the signer will always be set on app start, so that no password input is prompted.

MeBrei:
Closes #997 

Can currently only be tested when setting the signer to `None` in `rig()` of `main.rs`

**Open tasks/questions**
- [x] add testing (on http.rs level)
- [x] ~Should any requests of session/avatar/identity be sealed?~ I would not guard any of them for now. We need the session or at least the identity for the passphrase re-enter screen. The guards can also be set at a later state when we know how the ui will distinguish/use the different requests

### DIFF
--- a/proxy/api/src/http.rs
+++ b/proxy/api/src/http.rs
@@ -314,4 +314,34 @@ mod test {
             );
         });
     }
+
+    #[tokio::test]
+    async fn with_unsealed_guard() -> Result<(), Box<dyn std::error::Error>> {
+        let tmp_dir = tempfile::tempdir()?;
+        let mut ctx = context::Context::tmp(&tmp_dir).await?;
+        ctx.signer = None;
+        let api = with_qs_opt::<Query>()
+            .map(|opt_query| warp::reply::json(&opt_query))
+            .and(super::with_unsealed_guard(ctx))
+            .recover(super::error::recover)
+            .boxed();
+
+        let res = warp::test::request()
+            .method("GET")
+            .path("/?value=72")
+            .reply(&api)
+            .await;
+
+        assert_response(&res, StatusCode::FORBIDDEN, |have| {
+            assert_eq!(
+                have,
+                serde_json::json!({
+                    "message": "Keystore is sealed",
+                    "variant": "FORBIDDEN"
+                })
+            );
+        });
+
+        Ok(())
+    }
 }
--- a/proxy/api/src/http.rs
+++ b/proxy/api/src/http.rs
@@ -132,7 +132,7 @@ fn with_unsealed_guard(ctx: context::Context) -> BoxedFilter<((),)> {
     warp::any()
         .and(with_context(ctx))
         .and_then(|ctx: context::Context| async move {
-            if let Some(_) = ctx.signer {
+            if ctx.signer.is_some() {
                 Ok(())
             } else {
                 Err(Rejection::from(error::Routing::SealedKeystore))
--- a/proxy/api/src/http.rs
+++ b/proxy/api/src/http.rs
@@ -40,16 +40,28 @@ pub fn api(
     enable_fixture_creation: bool,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     let avatar_filter = path("avatars").and(avatar::get_filter());
-    let control_filter = path("control").and(control::filters(
-        ctx.clone(),
-        selfdestruct,
-        enable_fixture_creation,
-    ));
+    let control_filter = path("control")
+        .and(with_unsealed_guard(ctx.clone()))
+        .untuple_one()
+        .and(control::filters(
+            ctx.clone(),
+            selfdestruct,
+            enable_fixture_creation,
+        ));
     let identity_filter = path("identities").and(identity::filters(ctx.clone()));
-    let notification_filter = path("notifications").and(notification::filters(subscriptions));
-    let project_filter = path("projects").and(project::filters(ctx.clone()));
+    let notification_filter = path("notifications")
+        .and(with_unsealed_guard(ctx.clone()))
+        .untuple_one()
+        .and(notification::filters(subscriptions));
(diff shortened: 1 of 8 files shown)


proxy/api/src/http.rs
```diff
@@ -40,16 +40,28 @@ pub fn api(
...
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     let avatar_filter = path("avatars").and(avatar::get_filter());
-    let control_filter = path("control").and(control::filters(
-        ctx.clone(),
-        selfdestruct,
-        enable_fixture_creation,
-    ));
+    let control_filter = path("control")
+        .and(with_unsealed_guard(ctx.clone()))
+        .untuple_one()
```
xla, line 45: Can't we move the `untuple_one` into the body of `with_unsealed_guard`? It would spare us the reptition and help avoid confusion when used.

proxy/api/src/http/error.rs
```diff
@@ -14,6 +14,8 @@ use crate::error;
 pub enum Routing {
     /// The currently active [`coco::User`] is missing.
     MissingOwner,
+    /// The signer is missing.
```
xla, line 17: The error variant makes sense, but the docs don't reflect it. In general we need to find a representation of three cases: keystore not present, keystore sealed, keystore unsealed.

proxy/api/src/http/error.rs
```diff
@@ -81,7 +84,7 @@ pub async fn recover(err: Rejection) -> Result<impl Reply, Infallible> {
             )
         } else if let Some(err) = err.find::<Routing>() {
             match err {
-                Routing::MissingOwner => {
+                Routing::MissingOwner | Routing::SealedKeystore => {
```
xla, line 87: We might benefit from separating them. The sealed keystore could be a `403` and the missing owner a `401`.

xla (changes requested):
This looks like the right approach and matches to some extend what I had in mind. It's unfortunate that we now have these litered `ok_or_else` and `unwraps`.

Maybe not in this PR but as part of that story we can rework the `Context` to something that encodes these states and then handlers can have an explicit dependency on which context information they required:

``` rust
enum Context {
    Empty(session),
    Sealed(session),
    Unsealed(session, signer),
}
```

Rather pseudo code just to illustrate my idea.

proxy/api/src/http.rs
```diff
@@ -114,6 +123,22 @@ fn with_owner_guard(ctx: context::Context) -> BoxedFilter<(coco::user::User,)> {
         .boxed()
 }
 
+/// Asserts presence of the signer and rejects the request early if missing.
+#[must_use]
+fn with_unsealed_guard(ctx: context::Context) -> BoxedFilter<()> {
```
FintanH, line 128: I think it would make sense to return the `signer` here. This means that you wouldn't repeat this check inside the handlers below

xla: It would and it was done initially. The issue that comes from that is, as we want to use it as a guard on the top level `filters`, it would require every handle down chain to expect a `signer` as the first argument. Most don't need one, tho. See https://github.com/radicle-dev/radicle-upstream/pull/1008#pullrequestreview-503901412 for more context of how to address that issue more generally.

MeBrei: That was my initial thought as well, but passing it down to the filters in each module was not straightforward. It would also mean having to handle it as an input in all handlers with the guard wether they use it or not. So I went with this implementation as a first version.

FintanH: Why wouldn't it being only added to the filters that require a signer then? :eyes:

MeBrei: As far as I understand we want to guard almost all requests if the signer is not provided (later on this means no key store exists or it is sealed). Listing all projects for a user, for instance, does not need a signer, but should only be possible with an unsealed key store

FintanH: Mmmmmk, well I think it might be easy enough to express `Context` in terms of sealed and unsealed. It's everybody's favourite technique -- a type parameter :tada: 

```rust
pub struct Context<S> {
  signer: S,
}

pub type Sealed = PhantomData<!>;
pub type Unsealed = BoxedSigner;

impl Context<Sealed> {
  pub fn new(...) -> Self

  pub fn unseal(...) -> Context<Unsealed>
}
```

Then `with_unsealed_guard` would convert a `Context<Sealed>` into `Context<Unsealed>`. 

However, I think we'd end up closer to @xla's enum suggestion again since the `Context` needs to be mutated so that when we unseal it then it stays unsealed across endpoints. So we'd probably actually have `with_unsealed_guard: Either<Context<Sealed>, Context<Unsealed>> -> Context<Unsealed>`.

The good thing being that in tests we can always construct `Context<Unsealed>` and don't need to unwrap.

Do we think it's worth putting in this upfront work here and now? The advantage being that we don't have to go back and start removing unwraps et al. 

P.S. This is how `Storage` works in `radicle-link` if you want to take a look there.

MeBrei: I'd prefer to do it in a later PR. We won't need the `Empty` state until #999 and I think we will gain more insight into it once we start implementing #998.

FintanH: Okey dokey :+1:

FintanH (dismissed):

xla (approved):
👴 🍢 🍹 😆

FintanH (approved):

PR #1009, opened 2020-10-07
fix(proxy): drain peer channels

xla:
The polling of subroutines can starve if the channels are not completely
drained. async/await while being young already is a form of witchcraft
only the elder gods can control. Who knows what else lurks in the depth
of its abyss. Only the brave willl risk the decent into madness.

### DIFF
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -384,7 +384,8 @@ impl Future for Subroutines {
     type Output = Result<(), SpawnAbortableError>;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
-        // Drain the task queue
+        // Ensure the `JoinHandle`s of the tasks are consumed, droped to release associated
+        // resources.
         loop {
             match self.pending_tasks.poll_next_unpin(cx) {
                 Poll::Ready(Some(Err(e))) => {
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -209,7 +209,7 @@ impl Future for Running {
                     Ok(Ok(())) => Ok(()),
                 };
                 Poll::Ready(val)
-            }
+            },
             Poll::Pending => Poll::Pending,
         }
     }
@@ -282,7 +282,7 @@ impl<T> Future for SpawnAbortable<T> {
                     Ok(Ok(t)) => Ok(t),
                 };
                 Poll::Ready(val)
-            }
+            },
         }
     }
 }
@@ -390,7 +390,7 @@ impl Future for Subroutines {
                 Poll::Ready(Some(Err(e))) => {
                     log::warn!("error in spawned subroutine task: {:?}", e);
                     return Poll::Ready(Err(e));
-                }
+                },
                 Poll::Ready(Some(Ok(()))) => continue,
                 // Either pending, or FuturesUnordered thinks it's done, but
                 // we'll enqueue new tasks below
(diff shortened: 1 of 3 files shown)


kim (approved):
:knive: :rooster:

proxy/coco/src/peer.rs
```diff
@@ -401,39 +401,39 @@ impl Future for Subroutines {
         // Collect any task results, and enqueue new tasks if applicable
         let mut events = Vec::with_capacity(9);
         {
-            if let Poll::Ready(Some(protocol_event)) = self.protocol_events.poll_next_unpin(cx) {
+            while let Poll::Ready(Some(protocol_event)) = self.protocol_events.poll_next_unpin(cx) {
```
kim, line 404: Alright, wait. I’ve been poking around a bunch of impls to get a better understanding of how this waking business works — and I think this is now exactly what the `select!` macros try to avoid: if one of those channels is very busy, we’ll never poll the later ones.

So instead, if we want to keep multiple channels, we should `loop` over the entire thing. I.e. poll each channel once, enqueue new tasks, rinse repeat. 

If we don’t have any new events (ie. all channels are `Pending`) **and** we have enqueued at least one new task, we should put ourselves on the poll queue by calling `cx.waker().wake_by_ref()` — this is referred to as “yielding”, but what actually happens is that we tell the executor to poll us again asap instead of at its discretion (see the implementation of `tokio::task::yield_now()`).

Note that keeping track of the in-flight tasks via the `FuturesUnordered` is still a good idea I think, because we can abort all of them from one place and don’t have to make the tasks themselves select between a shutdown signal and their actual work. Also, we can bubble up panics or other irregularities.

kim (changes requested):

proxy/coco/src/peer.rs
xla: > Alright, wait. I’ve been poking around a bunch of impls to get a better understanding of how this waking business works — and I think this is now exactly what the `select!` macros try to avoid: if one of those channels is very busy, we’ll never poll the later ones.

So you did venture into the abyss, did you come back with your sanity in tact?

> So instead, if we want to keep multiple channels, we should `loop` over the entire thing. I.e. poll each channel once, enqueue new tasks, rinse repeat.

Will address that in this PR and rework it to not potentially loop indiefinitely over one channel. After the waiting room is tucked away in the `RunState` we should transition to `Select/SelectAll` to maybe not be too worried about these subtleties in our app code.

> If we don’t have any new events (ie. all channels are `Pending`) **and** we have enqueued at least one new task, we should put ourselves on the poll queue by calling `cx.waker().wake_by_ref()` — this is referred to as “yielding”, but what actually happens is that we tell the executor to poll us again asap instead of at its discretion (see the implementation of `tokio::task::yield_now()`).

You came back with powerful spells. In essence returning `Poll::Pending` is distinct from `cx.waker().wake_by_ref()`. The latter doesn't require another wake, so will definitely be polled. Fom the `yield_now` docs: `Any other pending tasks will be scheduled. No other waking is required for the task to continue.`

> Note that keeping track of the in-flight tasks via the `FuturesUnordered` is still a good idea I think, because we can abort all of them from one place and don’t have to make the tasks themselves select between a shutdown signal and their actual work. Also, we can bubble up panics or other irregularities.

:100:

kim: > You came back with powerful spells.

Observe how `yield_now` is implemented: https://docs.rs/tokio/0.2.22/src/tokio/task/yield_now.rs.html#16-37

Once `YieldNow` is polled once, it puts itself on the wake queue and returns `Pending`. Next time, it'll be `Ready`, so the code would proceed after the `.await` point.

Also, observe the magic spell inscribed here: https://docs.rs/futures-util/0.3.6/src/futures_util/stream/futures_unordered/mod.rs.html#439

I have zero clue how one would "yield the thread" from there, but I read that code as: `FuturesUnordered` is not sure if its got new tasks, so it tells "the task system" to poll soonish, so it can find out.

FintanH: The fact that we have to think about wakers is a big-ass abstraction leak :sob:

xla: > The fact that we have to think about wakers is a big-ass abstraction leak sob

captain obvious

PR #1010, opened 2020-10-08
fix(proxy): checkout peer

### LINKED ISSUE #849: [working copy] Ensure checkout comes from selected peer
We have the checkout functionality, however, we have mostly tested on local projects. We want to confirm (or fix) that it works by checking out the peer that I select. For example, if I have `rudolfs@hsgfj` as a peer and I select his view of the project, then when I click checkout his default branch is selected for creating the working copy.

FintanH:
Closes #849 

This PR ensures that we when we do a checkout of a project that it comes from the correct peer. There are two cases:
* We're checking out a copy of our own version of a project
* We're checkout out a copy of a peer's version of a project

In the former, it looks similar to creating a new project. We clone the project and set `rad` as our default remote.
In the latter, we need to set up the project based off of the peer's reference and then also create the rad remote for our default remote.

The test creates three peers. Bob clones from Alice, and Eve clones from Bob. Bob makes a commit to his working copy. Then Eve creates a working copy based off of Alice's view, and we ensure the Eve does not see Bob's commit.


FintanH:
Note that we get this in the log output:
> `Oct 08 13:04:42.243 ERROR librad::git::local::transport: Error waiting on child process: failed to create temporary file '/tmp/.tmpsM4se9/git/objects/tmp_object_git2_mqJGcZ': No such file or directory; class=Os (2)`

But this isn't an error that breaks the tests.

kim:
lol sorry

xla:
Spooky action at a distance.

FintanH:
@kim The error is still popping up :thinking:

cloudhead:
When cloning from a peer, do we setup an additional `upstream` remote from which the user can pull changes?

proxy/coco/src/project/checkout.rs
```diff
@@ -35,6 +37,22 @@ where
...
 
+/// We want to know whether we're checking out from one of our own copies, or if we're checking out
+/// based off of a remote's branch.
+pub enum Ownership {
+    /// We're checking out our own copy of the project.
+    Local(PeerId),
+    /// We're checking out a remote's version of the project.
+    Remote {
+        /// The handle of the remote.
+        handle: String,
```
cloudhead, line 48: Is this the handle as in `hash(id.revs[0])`?

FintanH: Nope, it's the handle of the peer. I'll update the docs :+1:

cloudhead: Ok I'm not sure what that means, waiting to see the doc update :)

cloudhead: Ah I see, it's the non-unique (display) name of the user.

FintanH: :100: good to know the docs helped :smile:

cloudhead:
Aside: how did you come up with the `<name>@<id>` syntax?

FintanH:
> Aside: how did you come up with the <name>@<id> syntax?

It was originally specified here: https://github.com/radicle-dev/radicle-link/issues/235

proxy/coco/src/project/checkout.rs
```diff
@@ -78,20 +92,44 @@ where
...
+                builder.remote_create(move |repo, _remote_name, url| {
+                    Remote::new(url, name.clone()).create(repo)
+                });
+                let remote_url = LocalUrl::from_urn(self.project.urn(), remote).to_string();
+                let repo =
+                    git2::build::RepoBuilder::clone(&mut builder, &remote_url, &project_path)?;
 
-        include::set_include_path(&repo, self.include_path)?;
+                // Create a rad remote and push the default branch so we can set it as the
+                // upstream.
```
xla, line 117: Why are we pushing to the remote for another peer by default?

xla (changes requested):
Smol changes, dope! It's unclear what the behaviour is we expect, there is some default pushing going on. Wouldn't the expectation be that a checkout of another peer is more like an upstream set we sometimes fetch from? Like how the origin is tracked on a forked repo to get mainline changes in.

proxy/coco/src/project/checkout.rs
FintanH: We're not :) We're pushing for our peer so that they have a copy of the default branch under their `rad` remote. Similar to having the default branch under `origin` when you clone from `GitHub`. This then becomes your version of the default branch.

xla: I see, I guess. It's hard to follow which end is which in this code and what the to expected state at the end is for the working copy.

FintanH: Cool, that's good to know. I'll make it clearer by maybe breaking it out or more documentation. Tbh, I could only reason through it by looking at the outcomes locally so I'm not surprised it's not clear :)

rudolfs (dismissed):
I also see:
```
WARN  librad::git::local::transport > child process still running, sending SIGKILL
```
in the logs when creating projects and doing checkouts when manually testing. The functionality seems to work though.

Other than that, looks good from my end & can't wait to test it with real peers.

rudolfs (approved):

proxy/coco/src/state.rs
```diff
@@ -577,11 +581,29 @@ impl State {
     where
         P: Into<Option<PeerId>> + Clone + Send + 'static,
     {
-        let proj = self.get_project(urn.clone(), peer_id).await?;
-        let include_path = self.update_include(urn).await?;
+        let proj = self.get_project(urn.clone(), None).await?;
```
xla, line 584: What's the significance of not passing the peer_id here anymore?

xla (dismissed):
The expansion of the docs helped immensely!

🐍 🌛 🚜 🚴

proxy/coco/src/state.rs
FintanH: We can _always_ get a project without a `PeerId` so I thought that would make sense to get the project metadata that way. But now that I'm thinking about it, it might be significant if the user's project data differs from the peers. If the default branch has changed it might mean that the user tries to checkout the wrong branch. 

So, I think I'll change it back to passing the `peer_id` :+1:

FintanH:
Converting back to a draft because it wasn't fully working when testing in the application itself. There's some weirdness to do with the local transport protocol that I'm not quite understanding.

xla (approved):
🏜 💥 🔥 🚚

rudolfs (approved):
🛒

Issue #1007, opened 2020-10-07
Track UI


FintanH:
Depends on #1006 

- Track a peer outside the projects' maintainer graph (seeing as there's only projects of 1 maintainer)

User story:
1. Get a `PeerId` out of band
2. Go to a projects' member tab (network tab)
3. Press "add peer to replicate"
4. Paste `PeerId` there
5. Now you start to listen for that peers' project changes

/cc @juliendonck - I just copy+pasted your card :innocent:

juliendonck:
![Uploading manage-remote.gif…]()

PR #1014, opened 2020-10-09
feat(ui): project remote management

juliendonck:
This integrates the "Manage Remotes" feature.

Closes: #1007 

demo


juliendonck:
@FintanH The scaffolding is here.

juliendonck:
@FintanH the UI is all here. I'm making some assumptions of data which i'm sure will be wrong. I'm happy to adapt the front end logic once we have an idea what the data will look like. For now this illustrates the goal imo.

ui/DesignSystem/Component/Illustration.svelte
```diff
@@ -546,5 +546,21 @@
       7.45457C15.2806 12.5048 19.5315 12.6427 19.5315 12.6427C19.5315 12.6427
       17.8043 11.7216 16.8524 9.76546V9.76546Z"
       fill="#292F33" />
+  {:else if variant === Variant.Computer}
```
rudolfs, line 549: I wonder how I didn't realise this earlier, but all the illustrations are just emojis from the twemoji set. Which we already have integrated. So we could potentially get rid of all these custom illustration SVGs.

```diff
diff --git a/ui/Modal/ManageRemotes.svelte b/ui/Modal/ManageRemotes.svelte
index a49c5fc..b001d44 100644
--- a/ui/Modal/ManageRemotes.svelte
+++ b/ui/Modal/ManageRemotes.svelte
@@ -1,12 +1,9 @@
 <script lang="ts">
-  import { Variant as IllustrationVariant } from "../src/illustration";
-
   import {
-    Illustration,
     List,
     RemoteListItem,
   } from "../DesignSystem/Component";
-  import { Button, Input } from "../DesignSystem/Primitive";
+  import { Button, Emoji, Input } from "../DesignSystem/Primitive";

   const remotes = [
     {
@@ -56,9 +53,8 @@
 </style>

 <div data-cy="remotes-modal" class="container">
-  <Illustration
-    style="margin-bottom: 1.5rem;"
-    variant={IllustrationVariant.Computer} />
+  <Emoji emoji={'💻'} size="huge" style="margin-bottom: 1.5rem;" />
+
   <h1>Manage remotes</h1>
   <div class="input">
     <Input.Text
```

Emoji
Screenshot 2020-10-12 at 15 27 32

vs

Illustration

Screenshot 2020-10-12 at 15 28 01

juliendonck: nice

juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/903c0e71db7234671546ea6ece63c1cebdb3dee9

juliendonck: Ended up using the <Illustration> component which is then in turn using the <Emoji> component. Made more sense like this.

FintanH (commented):
Looks cool! One thought below on data representation.

I'm also wondering if the "Follow" disabling is going to happen in the next iteration of this?

ui/Modal/ManageRemotes.svelte
```diff
@@ -0,0 +1,93 @@
...
+  } from "../DesignSystem/Component";
+  import { Button, Input } from "../DesignSystem/Primitive";
+  import { Variant } from "../src/illustration";
+
+  const remotes = [
+    {
+      handle: "cloudhead",
+      peerID: "hwd1yreg4khbjfa4gsyrio3f7ehluwkdhyregs4k",
+      maintainer: true,
+      synced: true,
```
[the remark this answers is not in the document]
xla: What's the meaning of being synced with a peer? Couldn't imagine how we determine that at the moment.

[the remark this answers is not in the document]
FintanH: The idea behind it is that we have tracked them but haven't got their changes in our repo yet. So it's determined by having them tracked for that project, but no remote exists for them under that project.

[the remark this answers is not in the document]
xla: Might be a matter of wording then. We syncing as a mechanism in another way already.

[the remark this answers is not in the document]
xla: > but maybe is it too early to introduce the TS stuff? I'm not sure what the dev flow of UI components is :)

The component is already TS so no harm in modelling the data.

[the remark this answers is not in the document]
FintanH: > Might be a matter of wording then. We syncing as a mechanism in another way already.

Ya maybe something better is `type RemoteStatus = Tracking | Searching`. And actually, to refine the type above, it would be better represented by:
```ts
interface Tracked {
  kind: "tracked";
  handle: string;
  peerId: string;
  maintainer: boolean;
}

interface Searching {
  kind: "searching";
  peerId: string;
}

type RemoteStatus = Tracked | Searching;
```

Since we know nothing of the `Searching` variant other than the `peerId`, which gets rid of these **nasty** `null`s

juliendonck:
> I'm also wondering if the "Follow" disabling is going to happen in the next iteration of this?

Good question, could we already do it in this iteration? Like check if the peerID is the maintainer's and user's peerID? 
It's hard for me to say since I don't know what data we have available.

FintanH:
> Good question, could we already do it in this iteration? Like check if the peerID is the maintainer's and user's peerID?

So to me, it looks like the data model you came up with makes sense (with a few tweaks I mentioned above ;)). So couldn't we use this maintainer field to disallow the button?

I guess one thing I should have questioned, why would we even show ourselves in the list here as a user? Like, it's obvious that we have the project because we wouldn't be able to click this button otherwise. That would simplify the above to: "we can't unfollow maintainers".

ui/DesignSystem/Component/UntrackedRemoteListItem.svelte
```diff
@@ -0,0 +1,31 @@
...
+  export let projectId = null;
+
+  const firstSix = peer.peerId.substring(0, 7);
+  const lastSix = peer.peerId.substring(
+    peer.peerId.length - 7,
+    peer.peerId.length
+  );
+
+  const cancelRemoteTracking = () => {
+    removeRemote(projectId, peer.peerId);
```
xla, line 18: Same as above.

ui/DesignSystem/Component/TrackedRemoteListItem.svelte
```diff
@@ -0,0 +1,60 @@
...
+  import { BadgeType } from "../../src/badge.ts";
+  import { CSSPosition } from "../../src/style";
+  import { PeerType, removeRemote, Role } from "../../src/project";
+
+  export let peer = null;
+  export let projectId = null;
+  export let projectName = null;
+
+  const unfollowRemote = () => {
+    removeRemote(projectId, peer.peerId);
```
xla, line 16: Should be a dispatch this component should not do any data manipulation, ought to be responsibility of the modal.

ui/DesignSystem/Component/TrackedRemoteListItem.svelte
```diff
@@ -0,0 +1,60 @@
+<script>
+  import { Avatar, Flex } from "../../DesignSystem/Primitive";
+  import Badge from "./Badge.svelte";
+  import Tooltip from "./Tooltip.svelte";
+
+  import TrackToggle from "./TrackToggle.svelte";
+  import { BadgeType } from "../../src/badge.ts";
+  import { CSSPosition } from "../../src/style";
+  import { PeerType, removeRemote, Role } from "../../src/project";
```
xla, line 9:
```suggestion
  import { BadgeType } from "../../src/badge.ts";
  import { PeerType, removeRemote, Role } from "../../src/project";
  import { CSSPosition } from "../../src/style";
  
  import { Avatar, Flex } from "../../DesignSystem/Primitive";
  
  import Badge from "./Badge.svelte";
  import Tooltip from "./Tooltip.svelte";
  import TrackToggle from "./TrackToggle.svelte";

```

ui/DesignSystem/Component/TrackedRemoteListItem.svelte
```diff
@@ -0,0 +1,60 @@
+<script>
```
xla, line 1:
```suggestion
<script lang="typescript">
```

ui/DesignSystem/Component/TrackedRemoteListItem.svelte
```diff
@@ -0,0 +1,60 @@
...
+  import Tooltip from "./Tooltip.svelte";
+
+  import TrackToggle from "./TrackToggle.svelte";
+  import { BadgeType } from "../../src/badge.ts";
+  import { CSSPosition } from "../../src/style";
+  import { PeerType, removeRemote, Role } from "../../src/project";
+
+  export let peer = null;
+  export let projectId = null;
+  export let projectName = null;
```
xla, line 13:
```suggestion
  export let peer: Peer = null;
  export let projectUrn: Urn = null;
  export let projectName: string = null;
```

ui/DesignSystem/Component/UntrackedRemoteListItem.svelte
```diff
@@ -0,0 +1,31 @@
+<script>
```
xla, line 1:
```suggestion
<script lang="typescript">
```

ui/DesignSystem/Component/UntrackedRemoteListItem.svelte
```diff
@@ -0,0 +1,31 @@
...
+  import { Flex } from "../../DesignSystem/Primitive";
+  import TrackToggle from "./TrackToggle.svelte";
+
+  import { removeRemote } from "../../src/project";
+
+  export let peer = null;
+  export let projectName = null;
+  export let projectId = null;
```
xla, line 9:
```suggestion
  export let peer: Peer = null;
  export let projectName: string = null;
  export let projectUrn: Urn = null;
```

ui/DesignSystem/Component/UntrackedRemoteListItem.svelte
```diff
@@ -0,0 +1,31 @@
+<script>
+  import { Flex } from "../../DesignSystem/Primitive";
+  import TrackToggle from "./TrackToggle.svelte";
+
+  import { removeRemote } from "../../src/project";
```
xla, line 5:
```suggestion
  import { removeRemote } from "../../src/project";
  
  import { Flex } from "../../DesignSystem/Primitive";
  import TrackToggle from "./TrackToggle.svelte";
```

xla (changes requested):
The main thing to change is that we start to use remotes as concept even so it's the management of peers in the context of the peer selector. It would streamline if we use peers consistently.

rudolfs:
> The main thing to change is that we start to use remotes as concept even so it's the management of peers in the context of the peer selector. It would streamline if we use peers consistently.

Just to make sure I understand this correctly, you're suggesting we use to `peer` in code and show `remote` in the UI?

xla:
> @xla: just to make sure I understand this correctly, you're suggesting we use `peer` in code and show `remote` in the UI?

I'm surprised that we show remote in the UI as well, while it is the technically correct in the sense of the git file tree layout, it's peers we are referring to, supported by the fact that the remotes are keyed by peer id and it's waht we provide when following new views of a project.

ui/DesignSystem/Component/UntrackedRemoteListItem.svelte
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/1385636c46b3c1812fd6233db5b8731b57143038

ui/DesignSystem/Component/TrackedRemoteListItem.svelte
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/1385636c46b3c1812fd6233db5b8731b57143038

cloudhead:
We're adding remotes to the repository (which happen to be keyed by PeerId), hence why we're calling it remotes.

xla:
> We're adding remotes to the repository (which happen to be keyed by PeerId), hence why we're calling it remotes. Adding "peers" doesn't really having a meaning to the user. What's a peer? What does it do? What does adding a peer mean?

I doubt adding remotes has more significance.

cloudhead (approved):
Awesome.

The only thing I'd change is the copy on the button: I think it could simply say `Add` or `Follow`.

xla (changes requested):
Can't wait for this to land. Had a first pass with some comment sinline.

ui/src/project.ts
```diff
@@ -321,6 +493,56 @@ export const repositoryPathValidationStore = (
   }
 };
 
+const VALID_PEER_MATCH = /^[\w\d]{54}$/;
```
xla, line 496: Let's make sure this regex ony matches valid `base32`.

ui/DesignSystem/Component/FollowButton.svelte
```diff
@@ -1,31 +1,31 @@
 <script lang="typescript">
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/db394c30a83968aa320cf2dbf5887a62290bcada

ui/src/track.ts
```diff
@@ -1,4 +0,0 @@
-export enum Event {
```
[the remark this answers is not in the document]
rudolfs: Yeah, we were only using this type in `FollowToggle` and we couldn't use the types at the call-site, because of special svelte syntax for the handlers like `on:eventname`. `on:Event.Track` doesn't compile...

So I deleted this file:

Screenshot 2020-10-27 at 15 24 31

ui/src/project.ts
FintanH: Nooooo base58 hahaha https://github.com/radicle-dev/radicle-link/blob/master/spec/radicle-link.md#cryptographic-identities

ui/Modal/ManageRemotes.svelte
```diff
@@ -0,0 +1,145 @@
...
+
+  .peer-entry-field {
+    display: flex;
+    justify-content: flex-start;
+    margin-top: 2rem;
+    width: 100%;
+  }
+</style>
+
+<Remote {store} let:data={project} context="project">
```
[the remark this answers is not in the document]
rudolfs: Good catch!

https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/a20dd25a5733d5183189118e06af39804154ffd6

ui/src/project.ts
xla: What @FintanH said.

ui/Screen/Profile/Following.svelte
```diff
@@ -83,11 +83,11 @@
               </div>
               {#if hover}
                 <div transition:fade={{ duration: FADE_DURATION }}>
-                  <TrackToggle
+                  <FollowButton
                     expanded
                     warning
-                    tracking={true}
-                    on:untrack={() => cancelRequest(request.urn)} />
+                    following={true}
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/a20dd25a5733d5183189118e06af39804154ffd6

ui/Modal/ManageRemotes.svelte
```diff
@@ -0,0 +1,145 @@
...
+
+  let newPeer: PeerId;
+
+  $: if (newPeer === "") {
+    peerValidation.reset();
+  }
+
+  const submitPeer = async (projectUrn: Urn) => {
+    if (await addPeer(projectUrn, newPeer)) {
+      newPeer = "";
```
FintanH, line 28: Just to check, does the `peerValidation.reset()` above cover this case?

ui/Modal/ManageRemotes.svelte
```diff
@@ -0,0 +1,145 @@
...
+  }
+</style>
+
+<Remote {store} let:data={project}>
+  <div data-cy="remotes-modal" class="container">
+    <Illustration
+      style="margin-bottom: 1.5rem;"
+      variant={IllustrationVariant.Computer} />
+
+    <h1>Manage remotes</h1>
```
FintanH, line 80:
```suggestion
    <h1>Manage Peers</h1>
```

I think capital P makes sense since it's a header but also happy with "peers"

ui/Modal/ManageRemotes.svelte
```diff
@@ -0,0 +1,145 @@
+<script lang="ts">
```
FintanH, line 1: Should this file be called `ManagePeers.svelte`?

ui/Modal/ManageRemotes/Peer.svelte
```diff
@@ -0,0 +1,66 @@
+<script lang="ts">
```
FintanH, line 1: And this directory `ManagePeers/*`?

ui/Modal/ManageRemotes/PeerFollowRequest.svelte
```diff
@@ -0,0 +1,37 @@
...
+  import { Flex } from "../../DesignSystem/Primitive";
+  import { FollowToggle } from "../../DesignSystem/Component";
+
+  export let peer: User;
+  export let projectName: string;
+  export let projectUrn: Urn;
+
+  const dispatch = createEventDispatcher();
+
+  const firstSix = peer.peerId.substring(0, 7);
```
FintanH, line 16: This logic exists elsewhere as well, right? Should we put it in a shared location?

ui/Modal/ManageRemotes.svelte
```diff
@@ -0,0 +1,145 @@
...
+
+  import { Variant as IllustrationVariant } from "../src/illustration";
+  import {
+    addPeer,
+    pendingPeers,
+    peerSelection,
+    peerValidation,
+    project as store,
+    removePeer,
+  } from "../src/project";
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/53efd607ff814c41c341fae044ad1f2d282ff55d

ui/Modal/Search.svelte
```diff
@@ -17,7 +17,7 @@
   import { ValidationStatus } from "../src/validation";
 
   import { Icon, Input } from "../DesignSystem/Primitive";
-  import { Remote, TrackToggle } from "../DesignSystem/Component";
+  import { FollowToggle, Remote } from "../DesignSystem/Component";
```
FintanH, line 20: Should the component be called `Peer`?

ui/src/project.ts
```diff
@@ -94,12 +188,37 @@ interface FetchList extends event.Event<Kind> {
...
+  urn: urn.Urn;
+}
+
 interface FetchLocalState extends event.Event<Kind> {
   kind: Kind.FetchLocalState;
   path: string;
 }
 
-type Msg = ClearLocalState | Create | Fetch | FetchList | FetchLocalState;
+interface TrackPeer extends event.Event<Kind> {
```
FintanH, line 201: Should this be `FollowPeer`? :eyes:

ui/src/project.ts
```diff
@@ -321,6 +493,56 @@ export const repositoryPathValidationStore = (
   }
 };
 
+const VALID_PEER_MATCH = /^[\w\d]{54}$/;
+
+const checkPeerUniqueness = (peer: string): Promise<boolean> => {
```
FintanH, line 498: Just so you know, track and untrack are idempotent so they won't double track or anything. But if this is for UX then totaly ignore.

FintanH (commented):
We should definitely make an effort to make more sub-modules -- `project.ts` is blowing up with different concepts.

Nice work :clap: Feel free to dismiss some of my feedback if it's not relevant.

ui/DesignSystem/Component/List.svelte
```diff
@@ -6,6 +6,8 @@
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   export let items: any[];
 
+  export let styleHoverState: boolean = true;
```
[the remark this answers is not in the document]
rudolfs: I see where you're coming from. I also didn't like this approach.

The thing is, it doesn't make sense to extract the `<li>` styling outside of this component, because we want the list elements to look the same. So it makes sense that the styles are encapsulated in `<List>`.

What we could do is have two specialised components, instead of one generic: `<List>` and `<ClickableList>` for example.

ui/src/project.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/b5098322f2d4453b6229c85827c57814645aae52

ui/Modal/ManageRemotes.svelte
rudolfs: Correct, it does.

ui/Modal/Search.svelte
rudolfs: This `Remote` component is actually not related to our domain. It already existed in our system and is for remote data fetching. 🤷

ui/src/project.ts
rudolfs: Only if we go all the way and also rename the proxy endpoints. Not sure it's worth the effort at the moment.

ui/src/project.ts
rudolfs: Thanks, yeah, was aware of that. It's only for UX.

ui/Modal/ManageRemotes.svelte
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/8e5e847a1b0d3569a02b867b97c9b60a578f7f19

ui/Modal/ManageRemotes/Peer.svelte
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1014/commits/8e5e847a1b0d3569a02b867b97c9b60a578f7f19

ui/Modal/ManageRemotes.svelte
rudolfs: See: https://github.com/radicle-dev/radicle-upstream/pull/1014#issuecomment-717222964.

ui/Modal/ManageRemotes/PeerFollowRequest.svelte
rudolfs: I simplifier it to a one-liner. I think the duplication is okay for now.

rudolfs:
> We should definitely make an effort to make more sub-modules -- project.ts is blowing up with different concepts.

@FintanH yeah, I agree `project.ts` is getting out of hand, however I think we should attempt the refactoring in a separate PR. I bumped into circular dependency issues previously when trying to split this up.

FintanH (approved):
Nice work @rudolfs and @juliendonck :grin:

FintanH (approved):
../approve

FintanH (approved):
`assert!(has_tests, true)`

cloudhead (approved):

PR #1015, opened 2020-10-09
fix(proxy): peer browsing

### LINKED ISSUE #1012: Browse peer branch
When we replicate our view looks like:
```
|> tree ~/.local/share/radicle/git/refs
/home/xla/.local/share/radicle/git/refs
├── heads
├── namespaces
│   ├── hwd1yreb5addursjwiemqnaa4cq4satpnjk77mbdhnp61oms3hurcydysbw
│   │   └── refs
│   │       ├── heads
│   │       │   └── master
│   │       └── rad
│   │           ├── id
│   │           ├── ids
│   │           │   └── hwd1yrey9nwynotkkamguh8yy81qks83qb8uhw6p47q4fjh563omuakakre
│   │           ├── self
│   │           └── signed_refs
│   ├── hwd1yref58x5ceewffrdd3cd7euqsegp8toi75w9sez54djnrg1zm4j7n3w
│   │   └── refs
│   │       ├── rad
│   │       │   ├── id
│   │       │   ├── self
│   │       │   └── signed_refs
│   │       └── remotes
│   │           ├── hybh5cb7spafgs7skjg6qkssts3uxht31zskpgs4ypdzrnaq7ye83k
│   │           │   └── rad
│   │           │       └── id
│   │           └── hybwm6qummpzc6rk4h4ur437z1thfc4hqfkkntxtzfjk5k7c3o6hoa
│   │               ├── heads
│   │               │   └── master
│   │               └── rad
│   │                   ├── id
│   │                   ├── ids
│   │                   │   └── hwd1yreri919uggmg1pgg7tw9sot5jfmhfrtsc3uyj7ciqxtp88jajcmmkh
│   │                   ├── self
│   │                   └── signed_refs
│   └── hwd1yrey9nwynotkkamguh8yy81qks83qb8uhw6p47q4fjh563omuakakre
│       └── refs
│           └── rad
│               ├── id
│               ├── self
│               └── signed_refs
└── tags
```

As we can see we're now able to replicate the `hybwm6qummpzc6rk4h4ur437z1thfc4hqfkkntxtzfjk5k7c3o6hoa` remote's `master` branch. However, in the UI when we try to browse the project we try to look at our level of the tree where there is no `master`. We don't have a `master` because we haven't checked out the project and pushed any changes yet.

It seems like this could be a UI bug, but it would be good to test this scenario in proxy and see what the output getting revisions is.

FintanH:
Fixes #1012 

Previously, when initialising a `Browser` we expected the initial branch to be our own default branch. Now that we're getting into the replicated end of things, this won't always work. When we first replicate we are only tracking the project, and that means we don't have a copy of the default branch under our view of the project.

To fix this, we need to look at our peer's views, and in particular, we can use the first maintainer to initialise the `Browser`.

So the initialisation of a Browser now checks for our own branch first in the storage, if that doesn't exist then fall back to the first maintainer.


cloudhead:
I'm probably missing something, but shouldn't the browser by default show the project's default branch? And if we want to browse something else, it can be explicitly passed in? :thinking:

FintanH:
> I'm probably missing something, but shouldn't the browser by default show the project's default branch? And if we want to browse something else, it can be explicitly passed in?

This is the behaviour at the moment. The question is _who's_ default branch? If you look #1012 -- and the tree of refs displayed in there -- you'll see that it's not always the case the default branch lives in our own view of the project (we might only be tracking). So we need to create a `Browser` using some reference, but we need to find one given our list of remotes + ourselves.

Is that clearer?

But this is making me think that it might be more clear if this is split up into different components:
* `with_browser` takes a reference that it can initialise with
* `State` gives us a way of creating a reference given `Urn` which is this "best guess" for which default branch we could use.

Thoughts?

cloudhead:
My understanding was that _projects_ have a default branch. This is true according to the radicle link spec. Is this not the case?

On your second point, I would agree that that any kind of "guessing" logic should be moved to its own function, as it can be tested in isolation and keep `with_browser` simpler.

FintanH:
> My understanding was that projects have a default branch. This is true according to the radicle link spec. Is this not the case?

Maybe there's some confusion here. There is a default branch but each peer has their own view of that default branch. You and I might be working on `nonempty` and have different updates to the default branch. Then Kim replicates it but doesn't make any changes - he needs to initialise the `Browser` based off of one of our default branches.

 > On your second point, I would agree that that any kind of "guessing" logic should be moved to its own function, as it can be tested in isolation and keep with_browser simpler.

Great idea! :smile: Thanks for spurring on the thought!

proxy/coco/tests/source.rs
```diff
@@ -0,0 +1,85 @@
+use nonempty::NonEmpty;
+use pretty_assertions::assert_eq;
+
```
xla, line 3: good stuff

proxy/coco/src/state.rs
```diff
@@ -351,12 +352,56 @@ impl State {
     /// * If the passed `callback` errors.
     pub async fn with_browser<F, T>(&self, urn: RadUrn, callback: F) -> Result<T, Error>
     where
-        F: Send + FnOnce(&mut git::Browser) -> Result<T, source::Error>,
+        F: FnOnce(&mut git::Browser) -> Result<T, source::Error> + Send,
```
xla, line 355: That's the spirit!

xla (changes requested):
Agree with @cloudhead that the `with_browser` could be slimmed down.

cloudhead:
> Maybe there's some confusion here. There is a default branch but each peer has their own view of that default branch. You and I might be working on nonempty and have different updates to the default branch. Then Kim replicates it but doesn't make any changes - he needs to initialise the Browser based off of one of our default branches.

Gotcha, yeah I was confused, thought it was about the branch *name* :+1:

rudolfs (approved):
LGTM from what I can tell.

cloudhead (dismissed):

proxy/coco/src/state.rs
```diff
@@ -342,30 +343,111 @@ impl State {
...
 
+    /// This method helps us get a branch for a given [`RadUrn`] and optional [`PeerId`].
+    ///
+    /// If the `branch_name` is `None` then we get the project for the given [`RadUrn`] and use its
+    /// `default_branch`.
+    ///
+    /// # Errors
+    ///   * If the storage operations fail.
+    ///   * If the requested reference was not found.
+    pub async fn get_branch<P, B>(
```
xla, line 389: `<P, B>`

bubblegum

proxy/coco/src/state.rs
```diff
@@ -342,30 +343,111 @@ impl State {
...
+            Err(Error::MissingRef { reference })
+        }
+    }
+
+    /// This method helps us get the default branch for a given [`RadUrn`].
+    ///
+    /// It does this by:
+    ///     * First checking if the owner of this storage has a reference to the default
+    /// branch.
+    ///     * If the owner does not have this reference then it falls back to the first maintainer.
```
xla, line 427: Wonder if this leads to unexpected behaviour from the users perspective.

xla (approved):
Extra dope! Thanks for taking the time to put things in better places, this change was logical and necessary.

🆕 🐿 🍋 👰

rudolfs (approved):

rudolfs:
Does this bring us closer to being able to do [integration tests](https://github.com/radicle-dev/radicle-upstream/issues/908) for peers?

xla:
> Does this bring us closer to being able to do [integration tests](https://github.com/radicle-dev/radicle-upstream/issues/908) for peers?

Jein, we need running peers in the tests for that and we currently don't have a way of starting and managing them. The testnet story is the blocker for that.

proxy/coco/src/state.rs
FintanH: Ya, I think I'll create a follow-up ticket for this because I mostly "guessed" which way of getting a branch is correct. But I'd like to audit and understand the cases better.

PR #1019, opened 2020-10-12
feat(ui): use consistent shareable identifiers across UI

rudolfs:
Use user/project `URN` as shareable identifier in the UI.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/840.

- onboarding success screen
onboarding-success

- own profile screen
user-heading

- project screen
project-heading

- other user's profile screen
other-user-profile

- search modal
search-modal

- search modal following
search-modal-following

- search modal unfollow
search-modal-unfollow

- unavailable user/project screen
unavailable

- discovery screen
discovery

- profile following tab
following


cloudhead (approved):
:+1:

ui/Screen/Onboarding/Success.svelte
```diff
@@ -40,16 +40,13 @@
...
-      showCopyOnlyOnHover
-      notificationText={`Radicle ID copied to your clipboard.`} />
+    <ShareableIdentifier urn={id} />
 
     <p
       style="text-align: center; width: 23.13rem; margin: 1.75rem 0 1.75rem 0;
       color: var(--color-foreground-level-6);">
-      This is your Radicle ID — it's unique to you!<br /> Click to copy it and share
-      it with others so that they can find you.
+      This is your handle — it's unique to you!<br /> Click to copy it and share
```
xla, line 48: This has been changed in some other context, but quite confusing to refer to the id as a handle. Where in other products it's used for the "nickname". /cc @cloudhead

xla (approved):
🛅 🍦 🚷 🛩

ui/Screen/Onboarding/Success.svelte
cloudhead: Yeah, we've discussed this at length -- we don't have an exact equivalent to the "handle" on radicle, because our only unique identifier is a long random string. So we use the term handle as "unique id", in the same way `cloudhead` is my unique id/handle on social media.

Now the thing is, this choice was made in the context of an id that looked like a handle, ie. `@abcedf123`. Now that we're using a sort of hybrid of that + a URN on copy, the term "handle" doesn't work as well. I've been thinking about this and I don't like where we're at with all this. I think we'll have to revisit a lot of it after the B1.

PR #1020, opened 2020-10-12
cleanup(ui): deduplicated svgs that are part of twemoji

juliendonck:
Some of our illustration SVG's were straight up the same as the twemoji svg's, so I used the <Emoji> component inside of the <Illustration> component so we don't have duplicates.


rudolfs (approved):
Nice!

xla (approved):
🏂 🕰 🎈 🌿

PR #1024, opened 2020-10-13
TypeScript targets ES2018

geigerzaehler:
Set ES2018 as the target for the typescript compiler. This works because
electron supports this standard and results in less compilation overhead
and smaller/better sourcemaps


xla (approved):
🎸 🎲 🍱 🔉

rudolfs (approved):

PR #1025, opened 2020-10-13
fix(proxy): await request urn

xla:
Regression from #992 where the future of the call was not awaited,
unfortunately our tests didn't break.

Refs #984


FintanH (approved):
lol we're good at Rust

NunoAlexandre (approved):

xla:
> lol we're good at Rust

The reason the lint didn't catch is, because we assigned it to variable. :facepalm:

PR #1026, opened 2020-10-13
refactor: move request urn endpoint to PUT

xla:
Part of #984 

Instead of using POST this change converts the endpoint to PUT to
convey that the operation is idempotent. A new request for an already
requested URN should not lead to an error, instead just a noop and a
friendly message.


FintanH (approved):
Nice :+1: Thanks for separating the commits :smile:

PR #1030, opened 2020-10-13
feat(ui): check if project exists when searching

xla:
Up until now we assumed the provided urn is not present in the monorepo.
This change alters the behaviour to first check if a project exists
locally - because locally sourced is best source. If it exists locally
the result box offers to navigate directly to it -- also possible with a
hit of enter. Should it not exist a track toggle is presented, which
when pressed will fire a request for the project and show a
notification.

* check for project locally first
* navigate to existing project
* request urn from the network on track toggle
* validate input to be a valid urn

Refs #984


cloudhead:
Concept ACK

ui/App.svelte
```diff
@@ -22,7 +22,9 @@
   import Onboarding from "./Screen/Onboarding.svelte";
   import DesignSystemGuide from "./Screen/DesignSystemGuide.svelte";
   import Discovery from "./Screen/Discovery.svelte";
-  import Modal from "./Modal";
+  import ModalNewProject from "./Modal/NewProject.svelte";
+  import ModalSearch from "./Modal/Search.svelte";
+  import ModalShortcuts from "./Modal/Shortcuts.svelte";
```
NunoAlexandre, line 27: It would be separate such cosmetic changes into separates PRs. And I'd prefer to undo this.

ui/DesignSystem/Component/TrackToggle.svelte
```diff
@@ -1,6 +1,8 @@
-<script lang="ts">
+<script lang="typescript">
```
NunoAlexandre, line 1:
```suggestion
<script lang="ts">
```

Unless there's a good reason for the change, let's keep it consistent.

ui/DesignSystem/Component/TrackToggle.svelte
```diff
@@ -12,19 +14,14 @@
 
   let active: boolean = false;
 
-  enum TrackingEvent {
-    Track = "track",
-    Untrack = "untrack",
-  }
-
```
NunoAlexandre, line 19: Kind of unnecessary to move this to a new module with only this type when it is only used here...

ui/DesignSystem/Primitive/Input/Text.svelte
```diff
@@ -1,6 +1,4 @@
-<script lang="ts">
-  import { createEventDispatcher } from "svelte";
-
+<script lang="typescript">
```
NunoAlexandre, line 1:
```suggestion
<script lang="ts">
```

ui/Hotkeys.svelte
```diff
@@ -1,4 +1,4 @@
-<script lang="ts">
+<script lang="typescript">
```
NunoAlexandre, line 1:
```suggestion
<script lang="ts">
```

ui/Modal/Search.svelte
```diff
@@ -0,0 +1,171 @@
+<script lang="typescript">
```
NunoAlexandre, line 1:
```suggestion
<script lang="ts">
```

ui/Modal/Search.svelte
```diff
@@ -0,0 +1,171 @@
...
+    searchProject,
+    urnValidationStore,
+  } from "../src/search";
+  import { ValidationStatus } from "../src/validation";
+
+  import { Icon, Input } from "../DesignSystem/Primitive";
+  import { Remote, TrackToggle } from "../DesignSystem/Component";
+
+  let id: string;
+  let value: string;
```
NunoAlexandre, line 23: could you document what these are?

ui/src/search.ts
```diff
@@ -1,49 +1,88 @@
...
+  update
+);
 
-const requestStore = remote.createStore<boolean>();
-export const request = requestStore.readable;
-
-// TODO(sos): actual validation
-export const validation = writable({ status: ValidationStatus.Success });
+// URN validation.
+const VALID_URN_MATCH = new RegExp("^rad:git:[1-9A-HJ-NP-Za-km-z]{59}");
```
NunoAlexandre, line 79: FYI, it can also just be:

```suggestion
const VALID_URN_MATCH = /^rad:git:[1-9A-HJ-NP-Za-km-z]{59}/;
```

ui/src/search.ts
```diff
@@ -1,49 +1,88 @@
...
+const urnConstraints = {
+  format: {
+    pattern: VALID_URN_MATCH,
+    message: `Not a valid urn`,
```
NunoAlexandre, line 83: we are referring to URN the acronym, not to a [urn](https://en.wikipedia.org/wiki/Urn) :funeral_urn: 
and since it now only works with projects, I'd rather have it be more explicit:


```suggestion
    message: `Not a valid project URN`,
```

ui/DesignSystem/Component/TrackToggle.svelte
rudolfs, line 1: Just out of curiosity, what's the difference? So far we've been using `ts` everywhere.
Official docs also mention `ts`: https://svelte.dev/blog/svelte-and-typescript#1_Compiling_TypeScript.

ui/DesignSystem/Primitive/Input/Text.svelte
```diff
@@ -23,19 +21,11 @@
...
   export let autofocus: boolean = false;
 
-  const dispatch = createEventDispatcher();
-
   // Can't use normal `autofocus` attribute on the `inputElement`:
   // "Autofocus processing was blocked because a document's URL has a fragment".
   // preventScroll is necessary for onboarding animations to work.
   $: if (autofocus) inputElement && inputElement.focus({ preventScroll: true });
 
-  const onKeydown = (event: KeyboardEvent) => {
```
rudolfs, line 33: Why do you want to remove this from the input? I imagine this could be a pattern we'll use in other forms as well.

ui/Modal/Search.svelte
```diff
@@ -0,0 +1,171 @@
...
+
+<style>
+  .container {
+    width: 26.5rem;
+  }
+
+  .search-bar {
+    margin-bottom: 1rem;
+    position: relative;
+  }
```
NunoAlexandre, line 96:
```suggestion
  .container {
    width: 50vw;
    max-width: 800px;
    min-width: 500px;
  }

  .search-bar {
    margin-bottom: 1rem;
    position: relative;
    padding: var(--content-padding);
    background-color: var(--color-background);
    border-radius: 4px;
  }
```

Now the validation error message is displaying in red on top of the overlay which is dark, looking no good, imo:

 
x


This suggestion does two things:

1. Improves the responsiveness of the modal container, looking better in both small and large screens:

small
medium

Screenshot from 2020-10-14 09-25-07

2. Wraps the search bar with the appropriate background colour with nice padding:

Screenshot from 2020-10-14 09-32-38

xla: Thanks for the suggestion, @radicle-dev/product is aware of it and the work is tracked in #1031

ui/App.svelte
xla: This is not a cosmetic change, it's a structural one. Given the overhead of getting a  change reviewed, approved and kept in sync with mainline, I made the call to include it here. As I touched the part of the codebase already. It does not distract or obfuscate the other changes and it should be fine to approve it along the rest.

ui/DesignSystem/Component/TrackToggle.svelte
xla: Consumers rely on the string value of these variants to catch the right events, but I didn't find a way to share that meaningfully. This was an attempt to do so.

ui/DesignSystem/Component/TrackToggle.svelte
xla: Somehow my editor stack doesn't pick it up unless it's the full language name.

xla: https://github.com/radicle-dev/radicle-upstream/pull/1030#discussion_r504479002

ui/Modal/Search.svelte
xla: Could you expand on what kind of documentation you expect when stumbling upon these? Also would it help to rename them to not require documentation?

ui/App.svelte
NunoAlexandre: Sorry for the wrong categorization on my side. In any case, still not fond of this change. What's the context and motivation here?

ui/DesignSystem/Component/TrackToggle.svelte
rudolfs: Whirrd!

I think we should try to keep them consistent, so we should either fix your editor or update the rest of the codebase to use `typescript`.

ui/src/search.ts
xla: How you know there are no dead things inside our urn? :smile:

ui/App.svelte
xla: It's the removal of the unnecessary bookkeeping in `index.js` plus dropping the superfluous Modal suffix.

ui/DesignSystem/Component/TrackToggle.svelte
xla: #1032

ui/DesignSystem/Component/TrackToggle.svelte
FintanH: fwiw: I'm all for small modules with specific data :+1: Just because it's only used here now doesn't mean the tests of time will not force it to be used elsewhere or be added to.

ui/App.svelte
NunoAlexandre: Yeah, I don't dislike dropping the bookkepping but the namespacing I do appreciate. The prefix is staying around as you see, though less elegant. Happy to talk about this another time.

ui/DesignSystem/Component/TrackToggle.svelte
NunoAlexandre: Ah cool :+1:

ui/Modal/Search.svelte
NunoAlexandre: Yeah, renaming would probably be best. Expanding, I open this file and have no idea what `id` and `value` stand or are used for.

FintanH: Looking at the code I'd probably call this `rawUrn` or `unvalidatedUrn`.

Also is the `id` above actually the `urn`? I find the name `id` hard to discern whether it's a PeerId or a RadUrn. WHERE ARE MY TYPES AS DOCUMENTATION /_raises fist to sky_

ui/Modal/Search.svelte
```diff
@@ -0,0 +1,171 @@
...
+    requestProject,
+    searchProject,
+    urnValidationStore,
+  } from "../src/search";
+  import { ValidationStatus } from "../src/validation";
+
+  import { Icon, Input } from "../DesignSystem/Primitive";
+  import { Remote, TrackToggle } from "../DesignSystem/Component";
+
+  let id: string;
```
FintanH, line 22: Wait, what's the `id` used for? I see that upon validation you set it as the Urn minus `rad:git`, but I can't see where that's used :thinking:

ui/src/search.ts
```diff
@@ -1,49 +1,88 @@
...
-    case Kind.Update:
-      requestStore.loading();
+    case Kind.Reset:
+      projectSearchStore.reset();
+
+      break;
+
+    case Kind.RequestProject:
+      projectRequestStore.loading();
+      // FIXME(xla): This truly belongs in project.ts.
```
FintanH, line 46: Or `project/search.ts`

ui/Modal/Search.svelte
NunoAlexandre: Woudln't hurt to include this alrerady though as an temporary improvement.

ui/Modal/Search.svelte
xla: It's shown next to the track button in the search resullt.

ui/Modal/Search.svelte
xla: > Looking at the code I'd probably call this `rawUrn` or `unvalidatedUrn`.
It's validated.


It's the `id` part of of a RadUrn.

ui/Modal/Search.svelte
xla: We can make these changes when polishing for the beta release. Don't see much value in doing changes until @radicle-dev/product has settled on something.

NunoAlexandre (approved):

rudolfs (approved):

PR #1046, opened 2020-10-14
refactor(ui): move local peer status to store

### LINKED ISSUE #1039: Move local peer status into a store
Doing the data access in the component is not advisable, this should move into it's own ts file and get a proper API.

xla:
Closes #1039

### DIFF
--- a/ui/DesignSystem/Component/ConnectionStatusIndicator.svelte
+++ b/ui/DesignSystem/Component/ConnectionStatusIndicator.svelte
@@ -1,20 +1,12 @@
 <script>
-  export let peerState;
+  import { status as store, StatusType } from "../../src/localPeer.ts";
 
-  $: {
-    console.log(peerState);
-
-    if (peerState) {
-      state = peerState.type;
-    }
-  }
-
-  let state = "offline"; // stopped | offline | syncing | online
-
+  import Remote from "../Component/Remote.svelte";
   import { Icon } from "../Primitive";
+  import Tooltip from "./Tooltip.svelte";
+
   import Syncing from "./ConnectionStatusIndicator/Syncing.svelte";
   import Offline from "./ConnectionStatusIndicator/Offline.svelte";
-  import Tooltip from "./Tooltip.svelte";
 </script>
 
 <style>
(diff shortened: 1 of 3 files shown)


ui/src/localPeer.ts
```diff
@@ -0,0 +1,61 @@
...
+  status: Status;
+}
+
+export type PeerEvent = StatusChanged;
+
+// STATE
+const statusStore = remote.createStore<Status>();
+export const status = statusStore.readable;
+
+statusStore.start(() => {
```
rudolfs, line 51: Should we call this in `App.svelte`?

ui/DesignSystem/Component/ConnectionStatusIndicator.svelte
```diff
@@ -1,20 +1,12 @@
 <script>
-  export let peerState;
```
rudolfs, line 2: I wonder if this component should not live in `/ui/App/`, because it has data fetching built-in?

xla: Yeah it should, but then id shouldn't be imported by the sidebar. It's a bit odd still that the sidebar lives in the design system.

ui/src/localPeer.ts
xla: I think it's reasonable to only set up the subscription when at least one subscriber exists for the store. This way it doesn't require a global set up, which can be forgotten.

xla: It makes it more self-contained.

ui/DesignSystem/Component/ConnectionStatusIndicator.svelte
rudolfs: We can move that out too.

ui/src/localPeer.ts
rudolfs: Ah, I think I misunderstood what this was doing initially. Never mind.

ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -148,7 +135,7 @@
     </li>
     <li class="item indicator" data-cy="search">
       <!-- svelte-ignore a11y-missing-attribute -->
-      <ConnectionStatusIndicator {peerState} />
+      <ConnectionStatusIndicator />
```
NunoAlexandre, line 138: I'd expect `ConnectionStatusIndicator`, a Design component, to receive the state and not to access the store itself, meaning here we would do:

``` svelte
<Remote {store} let:data={status}>
  <ConnectionStatusIndicator {status} />
</Remote>
```

ui/DesignSystem/Component/ConnectionStatusIndicator.svelte
NunoAlexandre: Wouldn't this fix that? https://github.com/radicle-dev/radicle-upstream/pull/1046#discussion_r505229282

rudolfs: Yeah, that works even better @NunoAlexandre.

ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -9,19 +9,6 @@
   import ConnectionStatusIndicator from "./ConnectionStatusIndicator.svelte";
 
   export let identity = null;
-
-  let peerState;
-
-  // FIXME(rudolfs): cypress tests break when EventSource is initialised.
```
rudolfs, line 15: So, yeah, we need to get to the bottom of this. The tests fail now.

ui/src/localPeer.ts
```diff
@@ -0,0 +1,61 @@
+import * as remote from "./remote";
+
+// TYPES
+export enum StatusType {
```
FintanH, line 4:
```suggestion
export enum Status {
```

minor suggestion

ui/src/localPeer.ts
```diff
@@ -0,0 +1,61 @@
...
+  kind: StatusType.Started;
+}
+
+interface Syncing {
+  kind: StatusType.Syncing;
+  syncs: number;
+}
+
+interface Online {
+  kind: StatusType.Syncing;
```
FintanH, line 30:
```suggestion
  kind: StatusType.Online;
```

FintanH (changes requested):

ui/src/localPeer.ts
xla: We have a `Status` type tho.

FintanH: Ah, right I only scrolled down after that suggestion :sweat_smile:  DISREGARD...

Real talk though: these sum types are such boilerplate. Who hurt you, typescript?

xla: > Real talk though: these sum types are such boilerplate. Who hurt you, typescript?

It was javascript, never a good idea to be backwards compatible. Which for rescript.

ui/src/localPeer.ts
xla: :facepalm:

ui/src/localPeer.ts
FintanH: Bless this mess :pray:

FintanH (dismissed):
:metal:

NunoAlexandre (approved):

rudolfs (dismissed):

NunoAlexandre (approved):

rudolfs (approved):

PR #1048, opened 2020-10-15
refactor(proxy): APIs for delayed coco::State initialization

geigerzaehler:
This PR prepares the APIs around `coco::State` so that the HTTP layer cannot assume that `coco::State` is available. This change is necessary since `coco::State` is initialized with a `BoxedSigner` which will stop to be available at boot time. Instead the signer will only become available when the frontend provides a pass phrase to unlock the key storage.

To accomplish this, `Context::state` is now a mutable `Option` and the HTTP layer gets it through `Context::coco_state`, which potentially errors.

Out of scope for this PR are:
* Adding the actual error
* Initializing `Context::state` through the HTTP API

With this PR we also move the signer into `coco::State`. This simplifies some of the coco APIs since they do not need an explicit signer anymore.


geigerzaehler:
I ran into some issues with this approach. It needs to be improved. Will replace this with different PR.

PR #1051, opened 2020-10-15
fix: "run all specs" from cypress UI

rudolfs:
Wrap each spec file in its own context as recommended [here](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests.html#Run-all-specs) and [here](https://glebbahmutov.com/blog/run-all-specs/#solution).

Otherwise when clicking the `Run all specs` button the `before` and `beforeEach` blocks of each spec file would get executed in the beginning of the run.

### DIFF
--- a/cypress/integration/onboarding.spec.js
+++ b/cypress/integration/onboarding.spec.js
@@ -1,4 +1,4 @@
-context("identity creation", () => {
+context("onboarding", () => {
   const validUser = {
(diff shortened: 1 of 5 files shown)


xla (approved):
🛤 🏆 🙎 🚭

FintanH (approved):
:metal:

PR #1053, opened 2020-10-15
fix(proxy): clean up long-running notifications

xla:
When introducing long-running sse subscriptions for notifications the
server stopped responding to shutdown signals. This prevented a proper
reload of the stack, which cypress integration tests depend on. In order
to terminate handlers spinning on notification subscriptions the
shutdown forces a clear of all stored senders.


FintanH (approved):
:metal:

rudolfs (approved):
❤️ 🚀

PR #1057, opened 2020-10-16
fix(proxy): block on checkout

FintanH:
Since checkout uses `clone` under the hood, it's necessary that we block
on this IO operation. Otherwise we will end up in flakey test hell.


xla (approved):
🐇 🐘 👂 🕵

NunoAlexandre (approved):
:sweet_potato:

Issue #956, opened 2020-09-23
Waiting room purge and retry semantics


FintanH:
In the current plan (and implementation) for the waiting room, we don't have a way to remove any of the `Request`s from the `WaitingRoom` that are in a terminal state. For example, if a `Request` was cancelled then it will be stuck in the `Cancelled` state and we won't be able to retry it.

These further semantics need to considered and tested.

PR #1059, opened 2020-10-16
fix(proxy): persist TimedOut state

### LINKED ISSUE #1023: Timeout transition
Turns out that I didn't transition the state when timing out, but just return an error: https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/coco/src/request/waiting_room.rs#L225-L228.

We should insert the updated state and _then_ return the `Err`.

FintanH:
Fixes #1023 

We never persisted the TimedOut state when we encountered one, instead
returning it directly as an error.
This PR fixes that by first persisting the TimedOut state during a
transition and then returning it as an error. We can confirm the
persistence in the test cases for when we reach a timeout.

### DIFF
--- a/proxy/coco/src/request/waiting_room.rs
+++ b/proxy/coco/src/request/waiting_room.rs
@@ -14,9 +14,7 @@ use librad::{
     uri::{self, RadUrl, RadUrn},
 };
 
-use crate::request::{
-    sequence_result, Clones, Queries, Request, RequestState, SomeRequest, Status, TimedOut,
-};
+use crate::request::{Clones, Queries, Request, RequestState, SomeRequest, Status, TimedOut};
 
 /// The maximum number of query attempts that can be made for a single request.
 const MAX_QUERIES: Queries = Queries::new(5);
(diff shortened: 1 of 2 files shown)


proxy/coco/src/request/waiting_room.rs
```diff
@@ -183,14 +181,16 @@ impl<T, D> WaitingRoom<T, D> {
...
-                    },
-                    Either::Left(mismatch) => Err(Error::StateMismatch((&mismatch).into())),
-                }
+            Some(request) => match request.clone().transition(matcher, transition) {
+                Either::Right(Either::Right(next)) => {
+                    self.requests.insert(urn.id.clone(), next.into());
+                    Ok(())
+                },
+                Either::Right(Either::Left(timeout)) => {
+                    self.requests.insert(urn.id.clone(), timeout.clone().into());
```
cloudhead, line 190: Does this list every shrink?

FintanH: Not yet: https://github.com/radicle-dev/radicle-upstream/issues/956

xla (approved):
(*^ω^*)

xla (approved):
🎲 🐑 ⚰ 🐼

cloudhead (approved):

PR #1062, opened 2020-10-16
refactor(proxy): remove providers

FintanH:
We didn't end up using the providers approach. We should remove it as it
isn't needed and we don't want to encourage anyone to use it.

### DIFF
--- a/proxy/coco/src/state.rs
+++ b/proxy/coco/src/state.rs
@@ -2,10 +2,8 @@
 
 use std::{
     convert::TryFrom,
-    future::Future,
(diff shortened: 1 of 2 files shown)


xla (approved):
🔆 🍄 🕞 🏯

NunoAlexandre (approved):
Ouch :sob:

FintanH:
> Ouch

It was my fault sending you down that path :'(

NunoAlexandre:
> > Ouch
> 
> It was my fault sending you down that path :'(

No worries mate, having a working product is all that matters :heart:

PR #1064, opened 2020-10-16
refactor(proxy): use signer in `coco::State`

geigerzaehler:
We remove the `BoxedSigner` argument from all `coco` functions that act on `coco::State`. Instead we keep the `BoxedSigner`, which is provided when creating `coco::State`, around as a field.

The motivation for this is to centralize the signer management in `coco::State` which is crucial for the general key management operation.


proxy/coco/src/state.rs
```diff
@@ -39,6 +39,8 @@ pub use error::Error;
 pub struct State {
     /// Internal handle on [`PeerApi`].
     pub(crate) api: PeerApi<keys::SecretKey>,
+    /// Signer to sign artifacts generated by the user.
+    pub signer: signer::BoxedSigner,
```
FintanH, line 43: Should the signer be accessible publically? :eyes:

proxy/api/src/http.rs
```diff
@@ -123,22 +115,6 @@ fn with_owner_guard(ctx: context::Context) -> BoxedFilter<(coco::user::User,)> {
         .boxed()
 }
 
-/// Asserts presence of the signer and rejects the request early if missing.
-#[must_use]
-fn with_unsealed_guard(ctx: context::Context) -> BoxedFilter<()> {
```
xla, line 128: The motivation for the change does not outline why this was dropped. As we removed the signer from context it's not accessible anymore. Changing the semantics of the API silently because of that doesn't seem the correct approach. If there is an idea how this guard should work going forward it should be adjusted as part of this change.

xla (changes requested):
See comment about removal of the unsealed_guard.

FintanH (commented):
Just one question :)

proxy/api/src/http.rs
geigerzaehler: Yes, I should have mentioned this. I dropped this code because it checks for the presence of `ctx.signer` which does not exist anymore. At the moment there is no sensible replacement for this but similar logic will be re-introduced when `ctx.state` becomes optional (because it needs to be initialized with a signer). Note that this does not change the behavior of any APIs. Before, `Context` was constructed at boot time with `signer: Some(signer)` so the error code path was never triggered.

xla: Yeah it was always constructed `Some`, but would mean if it ever was `None` the "error" path would be hit, so it behaves correctly. If that is your plan, then this change should make `state` optional and the guard should check for that.

proxy/coco/src/state.rs
geigerzaehler: No, it does not need to be public. Will fix

proxy/api/src/http.rs
geigerzaehler: > If that is your plan, then this change should make `state` optional and the guard should check for that.

Unfortunately this is not an easy change and requires more adjustment. (For example, `Option<coco::State>` is not enough since we also need to mutate it.) This is why I’d prefer to separate the change, especially considering that it does not affect any behavior.

xla: Is your replacement design outlined somewhere, it's hard to approve these changes without the confidence in what will replace it. This is invasive as it might lead to the guard being removed and the replacement never land for whatever reason we are left with a strange in between state. If you can onboard reviewers onto your plan going forward that'd help.

geigerzaehler: > Is your replacement design outlined somewhere

You can find a draft of the change in #1065.

> This is invasive as it might lead to the guard being removed and the replacement never land for whatever reason we are left with a strange in between state.

I get you point. I would just like to avoid an approval of this (already fairly large PR) to hinge on a fleshed out design of unsealing. I will make sure that we come up with a satisfactory solution.

xla: > You can find a draft of the change in #1065.

This will suffice, thanks for the follow-up.

xla (approved):
🆖 🚝 🕙 🗻

FintanH (approved):
🤘

cloudhead (approved):
Awesome!

PR #1066, opened 2020-10-16
feat: introduce network screen

xla:
*KEEP CALM AND CARRY ON*


juliendonck:
Can we close this, @xla? There is no design or idea behind this yet and the latest commit is almost a month old. I'd rather do this properly later.

juliendonck:
> Can we close this, @xla? There is no design or idea behind this yet and the latest commit is almost a month old. I'd rather do this properly later.

ping @xla

PR #1067, opened 2020-10-17
fix: correct path in reset script

xla:
Looks to be a copy-paste error where the same path was deleted twice.


NunoAlexandre (approved):
good catch :dart:

CodeSandwich (approved):
Nice :+1:

PR #1071, opened 2020-10-17
fix(proxy): fix waiting room logic for next query

### LINKED ISSUE #1040: Requests for projects time out too quickly
When searching for a project on the network the TimedOut state is hit too quickly.

xla:
Fixes #1040

### DIFF
--- a/proxy/coco/src/request/waiting_room.rs
+++ b/proxy/coco/src/request/waiting_room.rs
@@ -401,7 +401,8 @@ impl<T, D> WaitingRoom<T, D> {
             .filter_by_state(RequestState::Requested)
             .find(move |(_, request)| {
                 request.elapsed(timestamp.clone()) >= self.config.delta.clone()
-                    && request.attempts().clones == Clones::new(0)
+                    || (request.attempts().clones == Clones::new(0)
+                        && request.attempts().queries == Queries::new(0))
             });
 
         created.or(requested).map(|(urn, _request)| urn)
@@ -643,10 +644,10 @@ mod test {
     #[test]
     fn cloning_fails_back_to_requested() -> Result<(), Box<dyn error::Error + 'static>> {
         const NUM_CLONES: usize = 5;
-        let mut waiting_room: WaitingRoom<(), ()> = WaitingRoom::new(Config {
+        let mut waiting_room: WaitingRoom<usize, usize> = WaitingRoom::new(Config {
             max_queries: Queries::new(1),
             max_clones: Clones::new(NUM_CLONES),
-            delta: (),
+            delta: 5,
         });
         let urn: RadUrn = "rad:git:hwd1yre85ddm5ruz4kgqppdtdgqgqr4wjy3fmskgebhpzwcxshei7d4ouwe"
             .parse()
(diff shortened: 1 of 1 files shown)


NunoAlexandre (approved):
Makes sense :ok_hand:

proxy/coco/src/request/waiting_room.rs
```diff
@@ -401,7 +401,7 @@ impl<T, D> WaitingRoom<T, D> {
             .filter_by_state(RequestState::Requested)
             .find(move |(_, request)| {
                 request.elapsed(timestamp.clone()) >= self.config.delta.clone()
-                    || request.attempts().clones == Clones::new(0)
+                    && request.attempts().clones == Clones::new(0)
```
FintanH, line 404: The `||` was saying, "If this is in a fresh state for cloning -- i.e. we haven't made an attempt yet -- then we should return it".

Here's a case analysis:
* elapsed < delta || clones == 0 ~= false || true => returns the request
* elapsed > delta || clones == 0 ~= true || true => returns the request
* elapsed < delta || clone > 0 ~= false || false => does not return request
* elapsed > delta || clone > 0 ~= true || false => returns the request

FintanH (changes requested):

proxy/coco/src/request/waiting_room.rs
NunoAlexandre: After reading this I realized I saw the diff the way around. The original issue states that "When searching for a project on the network the TimedOut state is hit too quickly.". How does this change address that, @xla? :thinking:

xla: > How does this change address that, @xla? thinking

By always returning the request even so its timeout hasn't elapsed yet. So it would be requiered immediately in short order until max queries are hit, meaning it would go into TimedOut as quick as we poll the waiting room. You can see that in the test case, which failed before this change, as it communicates the expectations and order of events clearly.

FintanH: I think the check for clones is completely wrong then and shouldn't even be in there. We don't want to prevent requests that fall back into `Requested` (by exhausting peer ids for clone attempts) to be tried again.

Could you add a test that does the following:
1. `let _ = waiting_room.request(&urn, 0);`
2. `queried`
3. `found`
4. `clone_failed`
5. `next_query` after delta amount of time

I expect that to fail with the check for clones in the code. If so, we should remove that line. If not, I'd like to try to understand why :smile:

FintanH: But ya, I was wrong about the `clones` check. Can't remember what I was optimising for there.

xla: I follow your logic and understand now how the oiginal check came to be, will add anothr test for the case when we wanna fall back to query after a clone failed and see if it holds.

proxy/coco/src/request/waiting_room.rs
```diff
@@ -401,7 +401,8 @@ impl<T, D> WaitingRoom<T, D> {
             .filter_by_state(RequestState::Requested)
             .find(move |(_, request)| {
                 request.elapsed(timestamp.clone()) >= self.config.delta.clone()
-                    || request.attempts().clones == Clones::new(0)
+                    || (request.attempts().clones == Clones::new(0)
```
FintanH, line 404: Wait, I was thinking there wouldn't be _any_ other checks here. I believe `queries` will always be `> 0` because once you move the `Created` into `Requested` by calling `queried` it will bump the value.
```suggestion
```

proxy/coco/src/request/waiting_room.rs
```diff
@@ -401,7 +401,8 @@ impl<T, D> WaitingRoom<T, D> {
...
+                        && request.attempts().queries == Queries::new(0))
```
FintanH, line 405:
```suggestion
```

FintanH (approved):
Very nice :+1: Great catch :bow:

NunoAlexandre (approved):

PR #1072, opened 2020-10-17
fix(proxy): replicate cloned project maintainers of cloned project

### LINKED ISSUE #1041: Entity not found when navigating to maintainer profile
When tracking a peer project locally and clicking in the peer selector to navigate to the profile of the maintainer it show entity not found.

xla:
Attempt to clone maintainers of a freshly cloned project to improve data
availability. This will enable use-cases like navigating to maintainers
profile when viewing a project. Needs to be understood as a stop-gap
solution as it should be addressed in the protocol and its replication
strategy.

Fixes #1041

***
For this to work in the current network layouts, this hinges on the seed node eagerly replicated maintainers of projects it gets to know about.


NunoAlexandre (approved):

rudolfs (approved):

FintanH:
I'd rather we understand why this isn't happening in `radicle-link`, and fix it there

xla:
> I'd rather we understand why this isn't happening in `radicle-link`, and fix it there

"This" is happening in link. I opened this PR to start the conversation, how we currently retrieve a user in proxy is by calling

https://github.com/radicle-dev/radicle-upstream/blob/a3aa1e28b99849232c0314136a795cf885027f0e/proxy/coco/src/state.rs#L322

when a project is cloned it does have it's maintainers replicated with it, but under the remotes of the projects. So we have a disconnect how we access user data for "profiles" in upstream and how maintainers are replicated. What's the way forward with this in your opinion? And generally what's the strategy to replicate users sans maintainership?

FintanH:
> when a project is cloned it does have it's maintainers replicated with it but under the remotes of the projects.

Ya, I thought they would be replicated as well. But maybe it's something I misunderstood during my asking questions of @kim and never properly verified on my side. I would have expected us to clone the maintainer's identity docs as well so that they stay up-to-date when listening for changes. For example, they update their key set.

FintanH:
> What's the way forward with this in your opinion? And generally what's the strategy to replicate users sans maintainership?

I was thinking there would be a test in `radicle-link` for seeing if the maintainers were cloned and implementing the clone fetch and clone there.

cloudhead:
I wouldn't consider this a "bug" in link though, I think both ways of solving this seem valid :shrug:

FintanH:
> I wouldn't consider this a "bug" in link though, I think both ways of solving this seem valid

Sure, but we've talked about having useful features happen in the protocol rather than the proxy and this seems like a useful feature. And again, I thought this was what was happening in link but didn't properly verify.

kim:
>  it does have it's maintainers replicated with it, but under the remotes of the projects

It is [_supposed_](https://github.com/radicle-dev/radicle-link/blob/dc23136019e5885e7599441c90c7e8ed5747253b/librad/src/git/types.rs#L308) to fetch `rad/id` and `rad/ids/*` as a top-level namespace, although I'm ofc not vouching for the legacy "entities". I **do not** think it is a good idea to _clone_ the entire repo of a delegate (fka "certifier" which is utterly wrong terminology) -- we'd have zero control over how much data to fetch, and if we'd abort after some made-up limit, we might still not have the data we're actually interested in (which is the identity document).

> And generally what's the strategy to replicate users sans maintainership

Another confusionary aspect of the "entities" drama: there is **no such thing** as maintainership or certifiers -- both users and projects **delegate** to other keys, where projects may indirectly delegate to a _set_ of keys by specifying a `user` identity instead. The **tl;dr** is: at the end of the day, delegations are just public keys, which are equivalent to device keys, which are equivalent to network identities (ie. `PeerId`s). What we are tracking is `<stable id> || <network id>`. Thusly, we ought to always track an identity's **delegations**.

xla:
> Sure, but we've talked about having useful features happen in the protocol rather than the proxy and this seems like a useful feature. And again, I thought this was what was happening in link but didn't properly verify

Full agreement, I didn't mentioned that in the description up top. Yet I wouldn't rush it in link and use this as a stop-gap. If we want to address it right now also happy to move the discussion into links context.

FintanH:
> Yet I wouldn't rush it in link and use this as a stop-gap.

I wasn't advocating for rushing :) My aim was to understand where the issues in the implementation or my own understanding were coming from. But I understand that you'd like to get things up and ready for the beta. I'm just being the devil's advocate here to make sure we try to find where the solution fits.

That being said, this works for fixing the go-to profile bug (I guess, have you tried it out?) so we can merge as-is and continue the investigation in link :grin:

xla:
> I wasn't advocating for rushing :) My aim was to understand where the issues in the implementation or my own understanding were coming from. But I understand that you'd like to get things up and ready for the beta. I'm just being the devil's advocate here to make sure we try to find where the solution fits.

:dog: - I do think addressing this in link would be great, as pointed out this change also hinges on the fact that the seed has the same behaviour, which means we have to do there as well.

Another option is to change the way how we "navigate" to maintainers, by not using the existing `get_user`, but rather get the identity through the project, would that make sense?

> That being said, this works for fixing the go-to profile bug (I guess, have you tried it out?) so we can merge as-is and continue the investigation in link grin

Have tested it and it fixes the issue.

kim (changes requested):
I veto cloning.

xla:
Superseded by #1076

PR #1073, opened 2020-10-18
chore(ui): drop old registry code

xla:
Continued improvements to use types in the ui code let to more and more
overhead keeping the registry related code in sync which lays barren.
This is a liability and should we be in need to resurrect any of the
valuable work we can always pick and choose from an older state.


NunoAlexandre (changes requested):
I am planning on doing this in the `feat/funding` branch as there is code here that's useful for us and I'd rather keep it just a little longer.

xla:
> I am planning on doing this in the `feat/funding` branch as there is code here that's useful for us and I'd rather keep it just a little longer.

I'm running into issues with our type checks, you can always resurrect in your feature branch. As funding is not part of ß1 this should have higher priority to remove the baggage and have less potential for breakage.

cloudhead:
> I am planning on doing this in the feat/funding branch as there is code here that's useful for us and I'd rather keep it just a little longer.

Could you instead extract the code you need somewhere?

NunoAlexandre:
I will just make a side copy of the current master. I've already deleted some of these files on the funding branch and renamed others so let's see how the conflicts go but that's no worries. :+1:

NunoAlexandre (dismissed):

NunoAlexandre:
@xla pushed a commit also deleting unused currency components.

rudolfs (approved):

xla:
@NunoAlexandre Great thanks, but you snuck in an eth related file. :>

rudolfs (approved):

NunoAlexandre:
> @NunoAlexandre Great thanks, but you snuck in an eth related file. :>

I did, fixed now, sorry :hand:

NunoAlexandre (approved):

PR #1080, opened 2020-10-20
refactor: improve project and source browsing

### LINKED ISSUE #970: Project readme flickers on branch switch

### LINKED ISSUE #1141: "Incorrect input" when opening project
* Followed a project
* Clicked on it in "Following" tab
* It shows up fine, but I get an "Incorrect input" notification, and the following log:
```
 ERROR api::http::error      > Rejection(State(MissingRef { reference: Reference { remote: None, category: Heads, name: RefLike("main"), _namespace: Hash(Multihash([160, 228, 2, 32, 144, 117, 148, 247, 128, 224, 111, 175, 161, 56, 59, 10, 98, 217, 79, 141, 110, 190, 147, 188, 176, 16, 3, 203, 53, 178, 248, 117, 106, 75, 186, 52])) } }))
```

In the project stats, I see `0 Branches`, but the branch shows up in the branch selector.

### LINKED ISSUE #1145: Switching between repos when file is selected breaks UI
```
FileSource.svelte:100 Uncaught (in promise) TypeError: Cannot read property 'split' of null
    at create_fragment$2l (FileSource.svelte:100)
    at init (index.mjs:1447)
    at new FileSource (FileSource.svelte:98)
    at Array.create_if_block$B (Source.svelte:125)
    at Array.create_default_slot$1w (Source.svelte:111)
    at create_slot (index.mjs:58)
    at Array.create_else_block$4 (Remote.svelte:34)
    at Array.create_if_block_2$5 (Remote.svelte:34)
    at create_fragment$1n (Remote.svelte:39)
    at init (index.mjs:1447)
```

xla:
Fixes #970
Fixes #1141 
Fixes #1145


ui/DesignSystem/Component/SourceBrowser/RevisionSelector.svelte
```diff
@@ -124,29 +101,27 @@
   }
 </style>
 
-<Overlay {expanded} on:hide={hideDropdown}>
+<Overlay {expanded} on:{hide}>
```
MeBrei, line 104:
```suggestion
<Overlay {expanded} on:hide={hide}>
```

ui/Screen/Project/Source.svelte
```diff
@@ -1,148 +1,132 @@
...
+  import type { Revision } from "../../src/source";
+  import * as screen from "../../src/screen";
+
+  import IconCommit from "../../DesignSystem/Primitive/Icon/Commit.svelte";
+  import IconHouse from "../../DesignSystem/Primitive/Icon/House.svelte";
+
+  import ActionBar from "../../DesignSystem/Component/ActionBar.svelte";
+  import HorizontalMenu from "../../DesignSystem/Component/HorizontalMenu.svelte";
+  import Remote from "../../DesignSystem/Component/Remote.svelte";
+  import RevisionSelector from "../../DesignSystem/Component/SourceBrowser/RevisionSelector.svelte";
```
MeBrei, line 24: See above. I find it easier to read and edit if they are grouped. What is your reason for importing them one by one?

ui/src/config.ts
```diff
@@ -1,4 +1,4 @@
 export const HIDDEN_BRANCHES = ["rad/contributor", "rad/project"];
-export const DEFAULT_BRANCH_FOR_NEW_PROJECTS = "main";
+export const DEFAULT_BRANCH_FOR_NEW_PROJECTS = "master";
```
MeBrei, line 2: Didn't we just change this to `main` ?

MeBrei (dismissed):
I had a first run through of the changes. 🏃 😅 ⛰️ ⭐ 
Could you summarize which part was refactoring and it's motivation and which part was fixing the bugs?

Some regressions I found:
- [x] peer selector has no effect
- [x] revision selector does not collapse on click outside of the dropdown
- [x] commits counter and page are not updated on branch switch
- [x] if I select a file on a branch and then switch to a branch that does not have that file, the file is still shown

ui/Screen/Project/Source/Commits.svelte
```diff
@@ -51,26 +52,23 @@
...
 
 <div class="commits-page" data-cy="commits-page">
-  <Remote {store} let:data={commits}>
-    {#each commits.history as history}
+  <Remote {store} let:data={{ history }}>
+    {#each history.history as group (group.time)}
       <div class="commit-group">
         <header>
-          <p>{formatTime(history.time * 1000)}</p>
+          <p>{formatTime(group.time * 1000)}</p>
```
FintanH, line 59: What's this `* 1000` logic?

ui/src/config.ts
FintanH, line 2: Why did you change this back? :thinking:

ui/src/screen/project.ts
```diff
@@ -0,0 +1,172 @@
+import { get, derived, Readable } from "svelte/store";
+
+import type { PeerId } from "../identity";
+import * as project from "../project";
+import * as remote from "../remote";
+import type { Urn } from "../urn";
+import * as validation from "../validation";
+
```
FintanH, line 8: There seems to be a mix of remote modal and peer selector logic in here, which I found a bit confusing when reviewing the code. Could we further split the functionality in separate modules?

ui/src/source.ts
```diff
@@ -1,18 +1,18 @@
 import { Readable, writable } from "svelte/store";
```
FintanH, line 1: What's the difference between `project/source.ts` and this `source.ts`?

FintanH (commented):
Mostly questions and observations. Otherwise, looks cleaner in a lot of places :clap: 

I'm going to do a run of loading up the application and seeing if everything looks legit.

proxy/api/src/http/source.rs
```diff
@@ -664,18 +589,29 @@ mod test {
 
         let urn = replicate_platinum(&ctx).await?;
 
-        let branch = git::Branch::local("master");
+        let branch_name = "master";
```
geigerzaehler, line 592: We should choose a branch that is not the default branch.

proxy/api/src/http/source.rs
```diff
@@ -664,18 +589,29 @@ mod test {
...
+            peer_id: None,
+        };
+        let query = super::CommitsQuery {
+            revision: Some(revision.clone()),
+        };
         let res = request()
             .method("GET")
-            .path(&format!("/commits/{}?branch={}", urn.clone(), branch.name))
+            .path(&format!(
+                "/commits/{}?branch={}",
```
geigerzaehler, line 603:
```suggestion
                "/commits/{}?{}",
```

I’m not sure why this did not trigger an error

proxy/api/src/http/source.rs
```diff
@@ -57,7 +53,7 @@ fn commit_filter(
         .and_then(handler::commit)
 }
 
-/// `GET /commits/<project_id>?branch=<branch>`
+/// `GET /commits/<project_urn>?branch=<branch>&peerId=<peer_id>`
```
geigerzaehler, line 56: I don’t think this is how the query string looks like.

geigerzaehler (commented):
First round of comments.

In the future we should try to keep the PRs small to avoid long review times and review fatigue.

FintanH:
Source browsing seems to be broken. I tracked `radicle-link` and I'm trying to browse it via the **Following** tab:

image

The following error in console tab:
```
message: "Internal Git error: reference 'refs/namespaces/hwd1yrey5s17crf8wrwgcmuo1ii3ug6isun3wqgb1ftw17tdedoxdbp1rqh/refs/heads/master' not found; class=Reference (4); code=NotFound (-3)"
```

Which indicates that the `PeerId` isn't be used to find the source file.

FintanH:
The branch `xla/handle-disconnect` is only showing up as `xla` and will error when selected:

**Branch selector**:
image

**Error**:
image

FintanH:
**Open folder** after checkout does nothing other than dismissing the notification.

FintanH:
Select a file in the source browser and then select another peer, the file is still selected but the view jumps to the README

rudolfs (changes requested):
The source browsing integration tests are failing. I'll wait for these to be fixed before digging deeper.

<details>
<summary>Failing integration tests</summary>

```
  17 passing (57s)
  2 pending
  5 failing

  1) project source browsing
       commit browsing
         commit history
           shows the commit history for another branch:
     AssertionError: Timed out retrying: Expected to find content: '8' within the element: <span.counter.typo-mono-bold.svelte-1g19mvv> but never did.
      at Context.eval (http://localhost:64604/__cypress/tests?p=cypress/integration/project/source_browsing.spec.ts:67:72)

  2) project source browsing
       source code browsing
         page view
           when we're looking at the project root
             when there is a README file
               shows the README file:
     AssertionError: Timed out retrying: Expected to find content: 'platinum / .i-am-well-hidden' within the element: <div.file-source.svelte-10wne91> but never did.
      at Context.eval (http://localhost:64604/__cypress/tests?p=cypress/integration/project/source_browsing.spec.ts:181:30)

  3) project source browsing
       source code browsing
         source-tree
           doesn't interfere with the horizontal menu item active state:
     AssertionError: Timed out retrying: expected '<span.file-name.svelte-7zedd6>' to have class 'active'
      at Context.eval (http://localhost:64604/__cypress/tests?p=cypress/integration/project/source_browsing.spec.ts:300:47)

  4) project source browsing
       source code browsing
         source-tree
           allows navigating the tree structure:
     AssertionError: Timed out retrying: expected '<span.file-name.svelte-7zedd6>' to have class 'active'
      at Context.eval (http://localhost:64604/__cypress/tests?p=cypress/integration/project/source_browsing.spec.ts:328:45)

  5) project source browsing
       source code browsing
         source-tree
           highlights the selected file:
     AssertionError: Timed out retrying: expected '<span.file-name.svelte-7zedd6>' to have class 'active'
      at Context.eval (http://localhost:64604/__cypress/tests?p=cypress/integration/project/source_browsing.spec.ts:341:54)




  (Results)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Tests:        24                                                                               │
  │ Passing:      17                                                                               │
  │ Failing:      5                                                                                │
  │ Pending:      2                                                                                │
  │ Skipped:      0                                                                                │
  │ Screenshots:  5                                                                                │
  │ Video:        false                                                                            │
  │ Duration:     57 seconds                                                                       │
  │ Spec Ran:     project/source_browsing.spec.ts                                                  │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
```
</details>

FintanH:
Go to **maintainer's profile**, select the project, and your local peer is the default for the browser. I imagine this could have already been an existing issue so I can create an issue for it. The flow seems weird to me that it would default to your own peer if you've just navigated via the maintainer.

xla:
> Go to **maintainer's profile**, select the project, and your local peer is the default for the browser. I imagine this could have already been an existing issue so I can create an issue for it. The flow seems weird to me that it would default to your own peer if you've just navigated via the maintainer.

Can you make sure you have the latest commit.

xla:
> Go to **maintainer's profile**, select the project, and your local peer is the default for the browser. I imagine this could have already been an existing issue so I can create an issue for it. The flow seems weird to me that it would default to your own peer if you've just navigated via the maintainer.

Existing issue and out-of-scope for this PR.

FintanH:
> Existing issue and out-of-scope for this PR.

_puts product management hat on_

cloudhead:
Amazing, the flicker is almost* gone! :clap: 

\* There is one new case of flicker when clicking on a `.md`, you briefly see the raw markdown source before it switches to the rendered HTML.

xla:
> In the future we should try to keep the PRs small to avoid long review times and review fatigue.

@geigerzaehler All in favour of, would love to see how this would have been broken up in smaller increments. I couldn't see one.

ui/src/config.ts
xla: Merge confusion.

ui/Screen/Project/Source.svelte
xla: I disagree, I find that extremely unproductive and necessitates maintaining indexs files, which require manual book-keeping.

ui/Screen/Project/Source/Commits.svelte
xla: Because every language defines epoch as seconds while Brendan Eich deemed it fitting to add some millis...

ui/src/source.ts
xla: `src/screen/project/source` inhabits all the state necessary to drive the Screen Fragment to be found under Project/Source. While `src/source.ts` inhabits all the types and API calls referring to the source domain of the upstream software project.

ui/src/screen/project.ts
xla: Could you expand on what confuses you?

ui/Screen/Project/Source/Commits.svelte
FintanH: Madness. I see we do it a lot, would it be worth having a helper?

ui/src/screen/project.ts
FintanH: It's hard for me to determine the responsibilities of the functions. Which UI features are they used for kind of thing.

Does that make sense?

ui/src/source.ts
FintanH: Ah, that's clear :+1:

ui/DesignSystem/Component/SourceBrowser/FileDiff.svelte
```diff
@@ -1,7 +1,7 @@
-<script>
+<script lang="typescript">
   import { Icon } from "../../Primitive";
 
-  export let file = null;
+  export let file;
```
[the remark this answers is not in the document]
xla: It should, we missed out on some juicy types here. Will ammend.

ui/src/screen/project.ts
xla: Yes and no, the stores in this file are highly interdepent and are exclusive to the `Screen/Project.svelte` and its `Modal/ManagePeers.svelte`.

FintanH: Ok, well then ignore me :)

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
```diff
@@ -1,17 +1,21 @@
...
+  import type { Person } from "../../../src/source";
 
   import { Icon } from "../../Primitive";
 
-  export let projectId = null;
-  export let commitMessage = null;
-  export let timestamp = null;
-  export let commitSha = null;
-  export let user = null;
+  export let message: string;
```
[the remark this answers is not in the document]
xla: Yes, thanks for calling it out. Turned out there was some juicy consolidation on `CommitHeader` here.

ui/Screen/Project/Source/Commits.svelte
```diff
@@ -20,59 +18,10 @@
...
-                var(--color-foreground-level-6); --commit-sha-color:
-                var(--color-foreground)"
-                timestamp={format(commit.committerTime * 1000)}
-                user={commit.author} />
-            </li>
-          {/each}
-        </ul>
-      </div>
-    {/each}
+    <History {history} on:seelct={onSelect} />
```
FintanH, line 25:
```suggestion
    <History {history} on:select={onSelect} />
```

FintanH (changes requested):

proxy/api/src/http/source.rs
xla: It did. :D

rudolfs:
When I load a bigger repo (radicle-upstream for example) and switch between projects, I can see the header portion of the project page to load twice causing a flicker. Seems that the problem of #970 moved a level up.
header-flicker

rudolfs:
When I create a project from an existing repository, the UI only shows the project headers. A manual refresh also doesn't seem to help, only when I go to the projects listing and select the project there, then I see the project and it's source code.

Screenshot 2020-11-16 at 10 01 40

ui/src/source.ts
```diff
@@ -324,13 +317,36 @@ export const fetchReadme = (
   });
 };
 
+export const formatCommitTime = (t: number): string => {
```
FintanH, line 320: I imagine this isn't specific to commits?

FintanH (commented):
Going to play around with the application now and can report back anything out of the ordinary

FintanH:
> When I create a project from an existing repository, the UI only shows the project headers. A manual refresh also doesn't seem to help, only when I go to the projects listing and select the project there, then I see the project and its source code.

I see this as well. You can also just switch between the **Commits** tab and back to make it appear.

Looking at the network calls the source code does get returned as pictured below. So I imagine this is in the UI code.

image

FintanH:
https://github.com/radicle-dev/radicle-upstream/commit/3858c7ded70c531a276ec178982fab7a1c386e84 fixed that issue :+1:

FintanH:
I'm getting an error when trying to open a checkout. Maybe someone else can confirm this just in case it's just my machine.

```
Error occurred in handler for 'IPC_OPEN_PATH': TypeError: electron.shell.openPath is not a function
    at /home/haptop/Developer/radicle-upstream/native/main.comp.js:109:20
    at electron/js2c/browser_init.js:5877:52
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3865:42)
    at WebContents.emit (events.js:210:5)
```

```
{
  "code": "UnhandledRejection",
  "message": "An unexpected error occured",
  "source": {
    "code": "UnknownException",
    "message": "Error invoking remote method 'IPC_OPEN_PATH': TypeError: electron.shell.openPath is not a function",
    "stack": "Error: Error invoking remote method 'IPC_OPEN_PATH': TypeError: electron.shell.openPath is not a function\n    at EventEmitter../lib/renderer/api/ipc-renderer.ts.ipcRenderer.invoke (electron/js2c/renderer_init.js:1125:19)",
    "details": {
      "name": "Error"
    }
  }
}
```

xla:
> I'm getting an error when trying to open a checkout. Maybe someone else can confirm this just in case it's just my machine.

@FintanH Did you make sure you have the latest dependencies: yarn install

cloudhead:
* Markdown loading fixed! :heavy_check_mark: 
* It's possible to have the file tree show `foo.rs` selected while the file view shows `bar.rs`. This happens if you click on `bar.rs` while `foo.rs` is loading. It will temporarily show `bar.rs`, and when `foo.rs` is done loading, it will show `foo.rs`, but the file tree will still show `bar.rs` as selected. :x:

ui/src/source.ts
```diff
@@ -379,29 +297,71 @@ export const revisionQueryEq = (
...
-    .catch(readme.error);
+export const fetchReadme = (
+  projectUrn: Urn,
+  peerId: PeerId,
+  revision: Revision,
+  tree: Tree
+): Promise<Readme | null> => {
+  const path = findReadme(tree);
+
+  return new Promise((resolve, _reject) => {
```
geigerzaehler, line 308: We should avoid explicit `Promise` construction as it is unidiomatic. You can make the function `async` and use the following body.
```typescript
if (!path) {
  return null;
}

let blob;
try {
  blob = await fetchBlob(projectUrn, peerId, path, revision, false);
} catch {
  return null;
}

if (blob && !blob.binary) {
  return blob;
} else {
  return null;
}
```

ui/src/source.ts
```diff
@@ -379,29 +297,71 @@ export const revisionQueryEq = (
...
+    if (!path) {
+      return resolve(null);
+    }
 
-  return readme.readable;
+    fetchBlob(projectUrn, peerId, path, revision, false)
+      .then(blob => (blob && !blob.binary ? blob : null))
+      .then(resolve)
+      .catch(() => resolve(null));
```
geigerzaehler, line 316: We should be more discriminate here: Only if the readme does not exist should we return `null`. Otherwise we should notify the user of an error.

FintanH:
> @FintanH Did you make sure you have the latest dependencies: yarn install

Fixed, thanks :+1:

ui/src/remote.ts
```diff
@@ -49,6 +54,10 @@ export const createStore = <T>(): Store<T> => {
   const internalStore = writable(initialState, set => {
     if (starter) {
       return starter(set);
+    } else {
+      return () => {
+        set({ status: Status.NotAsked });
+      };
```
geigerzaehler, line 60: This behavior might be very confusing and has the potential to break a lot of things in the app since it is so widely used. I’d propose to make this behavior opt-in or leave it out entirely.

FintanH (approved):
:metal:

ui/src/screen/project/source.ts
```diff
@@ -0,0 +1,281 @@
...
+      selectedRevision.name === revision.name
+    ) {
+      return;
+    }
+
+    Promise.all([
+      source.fetchCommits(project.urn, peer.peerId, revision),
+      source.fetchTree(project.urn, peer.peerId, revision, ""),
+    ])
+      .then(async ([history, tree]) => {
```
geigerzaehler, line 125: I’d prefer to make this function async instead of mixing `.then` and async.

geigerzaehler (commented):
I’ve only reviewed the non-svelte code. I leave the svelte code to other reviewers

ui/src/source/diff.ts
```diff
@@ -0,0 +1,117 @@
+export type Line = string;
+
+export enum LineDiffType {
+  Addition = "addition",
+  Context = "context",
+  Deletion = "deletion",
+}
+
+export interface Addition {
+  type: LineDiffType.Addition;
```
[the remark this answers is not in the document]
xla: We use `kind` for types defined in the UI, when data is serialised from the proxy we use `type` as we annotate serde with `tag = "type"`.

ui/src/source.ts
xla: > We should avoid explicit `Promise` construction as it is unidiomatic. You can make the function `async` and use the following body.

Could you expand what is unidiomatic about `Promise` construction? I fail to see how your proposal is an improvement, it introduces an additional variable and a handful more branches, likely harder to reason.

ui/src/source.ts
xla: Indeed!

ui/src/remote.ts
xla: Not resetting the store upon unsubscribe, has led to many issues with wrong/old state being rendered. It has been a bane and hard to debug. I'd argue the opposite is the case, not having this has led to unexpected behaviour. Given that the app is still functional and works as expected there is no apparent issue.

ui/src/source.ts
geigerzaehler: I personally find the control flow in my suggestion much clearer. For example we can write `return null` instead of `return resolve(null)`. Using `new Promise` introduces callback-style code for control flow which the whole JS ecosystem has moved away from because it is usually harder to reason about.

> it introduces an additional variable

I don’t think so. `blob` is also introduced in your code as a callback function argument.

> and a handful more branches,

This is not the case either. In your code the branches are just less visible (the ternary operator) and a `.catch` call.

ui/src/remote.ts
geigerzaehler: Can you provide a concrete example where this fixes an issue? I’m ok with this change if you have confidence in it but I believe this might be more an issue with us using global stores. Some examples should help us figure this out.

rudolfs (changes requested):
Another magnum opus! 📖 🎖️ 🧁 

I noticed that two design/UX changes slipped in, wondering if @juliendonck & @brandonhaslegs are aware of them:

- "Source" tab got renamed to "Code"
- the commit teaser got moved outside of FileSource

  Screenshot 2020-11-16 at 11 30 27

  vs

  Screenshot 2020-11-16 at 11 31 31

This leads to the source code and header being too close together (missing margin/padding):
Screenshot 2020-11-16 at 14 59 47


- [ ] There's a bug in the "Manage remotes" modal -- when unfollowing a remote, the other remote's button changes to "Follow":

  master-remotes

  vs

  broken-remotes

juliendonck:
> * the commit teaser got moved outside of FileSource

Should only be outside the file on the default view on the source page and show the latest commit to this repo. Otherwise the commit should be inside the and show the last commit that touched that file.

Is there a reason why it's not like that anymore @xla?

juliendonck:
> * the commit teaser got moved outside of FileSource

Should only be outside the file on the default view on the source page and show the latest commit to this repo. Otherwise the commit should be inside the and show the last commit that touched that file.

Is there a reason why it's not like that anymore @xla?

ui/src/remote.ts
xla: If you look at the parts in the source browsing that have been refactored, those relied heavily on a plethora of "reset" calls to enure that state would be reverted. These happen in many places away from the actual state handling and would rely on the ceremonial knowledge of the implementer to not be forgotten. If forgotten long winding debug session had to follow with patches putting a reset in the right place at the right time. Essentially this PR is a remedy of that larger issue by centralising state in screen specific stores and bundling data together where needed, while removing the need of consumers in components to know about the incantations required to avoid subtle breakages, double loading, weird flickering - all resulting from state being inconsistent.

cloudhead:
Switching between big files is working perfectly now :heavy_check_mark:

ui/src/remote.ts
geigerzaehler: Thanks for the explanation.

ui/src/source.ts
```diff
@@ -128,207 +124,138 @@ export interface Sha {
...
+  selected: string;
 }
 
-interface FetchCommit extends event.Event<Kind> {
-  kind: Kind.FetchCommit;
-  projectId: string;
-  peerId: string;
-  sha1: string;
+export interface SelectedRevision {
+  request: AbortController | null;
```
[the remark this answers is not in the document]
xla: It's in the works.

ui/Screen/Project/Source.svelte
xla: > @xla: IMHO it just shifts around where you do the bookkeeping.

It does not tho, when adding a component the implementer has to remember to add extra **entries** in the respective index file(s). None of that overhead is present when using the component directly.

ui/DesignSystem/Component/SourceBrowser/FileDiff.svelte
```diff
@@ -131,29 +106,29 @@
...
   <main>
-    {#if file.diff.type == 'plain' && file.diff.hunks.length > 0}
+    {#if file.diff.type === FileDiffType.Plain && file.diff.hunks.length > 0}
       <table class="diff">
         {#each file.diff.hunks as hunk}
           <tr class="diff-line">
-            <td colspan="2" class="diff-expand-action typo-mono" />
-            <td colspan="2" class="diff-expand-header typo-mono">
+            <td colspan={2} class="diff-expand-action typo-mono" />
+            <td colspan={2} class="diff-expand-header typo-mono">
```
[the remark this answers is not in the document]
xla: The typescripted dom elements define `colspan` as [number](https://github.com/microsoft/TypeScript/blob/79ffd03f8b73010fa03cef624e5f1770bc9c975b/src/lib/dom.generated.d.ts#L8438). Both `=2` and `="2"` are interpreted as strings. Here we don't require interpolation but a javascript expression, therefore the syntax choosen.

ui/DesignSystem/Component/SourceBrowser/FileView.svelte
```diff
@@ -0,0 +1,55 @@
...
+    <CommitTeaser
+      commit={$code.lastCommit}
+      on:select={onSelectCommit}
+      style="height: 100%" />
+  </div>
+
+  {#if view.kind === ViewKind.Blob}
+    <Blob {rootName} {view} on:root={onSelectRoot} />
+  {:else if view.kind === ViewKind.Root}
+    <Root {view} />
```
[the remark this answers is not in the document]
xla: Splitting components is not the same as indirection, usually the result of an introduced abstraction. Compartmentalization into isolaed components when units have been identified should improve the maintainability and composibaility of the codebase.

In this specific instance it helps the struggling svelte compiler to not trip over inference of typescript types.

xla:
> Should only be outside the file on the default view on the source page and show the latest commit to this repo. Otherwise the commit should be inside the and show the last commit that touched that file.
> 
> Is there a reason why it's not like that anymore @xla?

What is the default view? It's consistent to find in different places, even so the user is in the same context of the file browsing.

cloudhead:
> What is the default view? It's inconsistent to find the commit header in different places, even so the user is in the same context of the file browsing. No other forge seems to do that.

On the project home, the commit teaser is the latest commit of the repository, so it's outside of any file. When a file is selected, the teaser is the latest commit _on that file_, so the teaser is in the file.

xla:
> > 
> 
> On the project home, the commit teaser is the latest commit of the repository, so it's outside of any file. When a file is selected, the teaser is the latest commit _on that file_, so the teaser is in the file.

How come that all other forges have the commit element in a consistent position. Technically the commit is for the current preifx, which is "/" for the root, the directory path for sub-directories which we currently don't show and the file path for the selected file.

cloudhead:
Because other forges have a different layout, so they put the repository latest commit above the file listing, which we can't do. Regardless, we need to differentiate the latest repo commit from the latest commit of the readme, and currently they look the same.

MeBrei:
This following is still happening for me locally. reset my state after updating my branch to make sure.

> * [x]  There's a bug in the "Manage remotes" modal -- when unfollowing a remote, the other remote's button changes to "Follow":
>   master-remotes
>   vs
>   broken-remotes

xla:
> This following is still happening for me locally. reset my state after updating my branch to make sure.

Can you try deleting the last item and see if it exhibits the same issue?

FintanH (dismissed):
Looks good from what I can tell :+1:

cloudhead (approved):
The file browser UX is :ok_hand: :ok_hand: -- haven't checked the rest.

juliendonck (approved):
Source browsing feels super snappy, even on a repo as big as upstream. Super nice. 

One note (not a blocker): Once I click on any file (commit is in the file) I would expect to always be able to click back on "Files" and that resets the source view, meaning that the latest commit would be back above the file. Now I can only get back there by refreshing the app.

xla:
> One note (not a blocker): Once I click on any file (commit is in the file) I would expect to always be able to click back on "Files" and that resets the source view, meaning that the latest commit would be back above the file. Now I can only get back there by refreshing the app.

@cloudhead and I discussed this as well, I think most users expect that behaviour. Will see hard that is to put in.

rudolfs:
Revision selector looks wonky:

| revision selector | peer selector |
|-----------------|---------------|
| dropdown | peer |

- [ ] the expand icon should be visible in both expanded and collapsed states (like in peer selector)
- [ ] the expand icon is misaligned

xla:
> One note (not a blocker): Once I click on any file (commit is in the file) I would expect to always be able to click back on "Files" and that resets the source view, meaning that the latest commit would be back above the file. Now I can only get back there by refreshing the app.

This is implemented in 26d16bf

rudolfs:
Revision selector ordering is mixed up, the selected branch is always shown first in the list:
| master | xla/refactor-peer-selection |
|-----------------|---------------|
| master | xla |

- [ ] selecting a branch does not change the ordering in the list

xla:
> the expand icon should be visible in both expanded and collapsed states (like in peer selector)

This is how master works as well, can change it. Just wanna make sure that we want all selectors work the same.

> Revision selector ordering is mixed up, the selected branch is always shown first in the list:

This was suggested by @cloudhead and I agree that it makes sense with our current dropdown UX. We can review that once we make it more sophisticated: highlighting default branch, add filtering to the dropdown.

cloudhead:
Yeah it now behaves how dropdowns are supposed to behave :+1:

cloudhead:
>  the expand icon should be visible in both expanded and collapsed states (like in peer selector)

This seems strange, where do you see that?

cloudhead:
> This is implemented in [26d16bf](https://github.com/radicle-dev/radicle-upstream/commit/26d16bf9bf4d9d99003c35772a24898b9d0e3ccd)

Could we keep the behavior though when you click on it from the Commits view, so it preserves the file you were on, but if you're already on the Files view, then it goes to root? :grimacing:

xla:
> Could we preserve the behavior though when you click on it from the Commits view? So it preserves the file you were on, but if you're already on the Files view, then it goes to root? grimacing

Done in 597cb35

rudolfs:
> This seems strange, where do you see that?

> In Figma you can see that the expand button isn't there when expanded, which makes sense: https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=4147%3A7516

@cloudhead you're right. We changed the peer selector here: https://github.com/radicle-dev/radicle-upstream/pull/1261#issuecomment-728821854. Now that I look at it again, I think we should have just added a spacer that takes up as much space as the icon to avoid the "jumping", thanks for pointing out!

rudolfs (approved):
Awesome work! 🚀 


Found just one small thing:
- Figma reference 8px: 
  Screenshot 2020-11-19 at 09 33 42

- We have 12px:
Screenshot 2020-11-19 at 09 34 48

Not a blocker, can also be addressed later.

xla:
> Not a blocker, can also be addressed later.

Was triggering me as well, do we have a proper `rem` value for it?

rudolfs:
> Was triggering me as well, do we have a proper rem value for it?

@xla `0.5rem`

rudolfs (approved):

MeBrei (approved):
⭐ ⛰️ ⭐

juliendonck:
@xla merge it! quick! 
Screen Shot 2020-11-19 at 11 34 41

cloudhead (approved):
:birthday:

PR #1081, opened 2020-10-20
feature(proxy): list remotes

### LINKED ISSUE #1017: List peers
As part of #1007 the modal that appears has a list of the peers that you track for that particular project.

The view consists of yourself, the maintainers, and any of the other peers you track. This is relevant since we will want to be able to click "Unfollow" -- i.e. `untrack` -- and we SHALL not unfollow ourselves or the maintainers.

We also have the case where we have tracked the peer but still haven't found the peer's version of the project. So it's in a "still looking" state.

The goal here is to implement an endpoint for providing this list of peers, as from what I can see this doesn't exist yet.

Some food for thought:
```rust
// Capture the different kinds of peers in the list
enum Peer {
  Me { handle: String, peer_id: PeerId },
  Maintainer { handle: String, peer_id: PeerId },
  Fellow { handle: String, peer_id: PeerId },
  NotFound { peer_id: PeerId },
}

// The endpoint
/// `GET /<urn>/peers

// The function for getting peers. The list is NonEmpty because we should at least be in there
fn peers(context: Context) -> Result<NonEmpty<Peer>, Error>
```

To determine if a peer is a maintainer we can inspect the `rad/ids/*` of this project.

FintanH:
We refactor tracked so that it returns the most information as possible.
There are three cases in particular:
* Searching -- peer is tracked but no information is available for them
yet
* Maintainer -- peer is a maintainer of the project
* Tracking -- peer is a contributor to the project

Closes #1017


proxy/coco/src/project.rs
```diff
@@ -17,3 +20,81 @@ fn set_rad_upstream(repo: &git2::Repository, default_branch: &str) -> Result<(),
...
 }
+
+/// Relation of the peer to the project.
+#[derive(Copy, Clone, Debug, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub enum Role {
+    /// Replicating, but not participating.
+    Tracker,
+    /// Participated with unique changes.
+    Contributer,
```
rudolfs, line 31: Screenshot 2020-10-23 at 09 39 39

FintanH: lmao very good :smile:

proxy/coco/src/project.rs
```diff
@@ -13,6 +13,10 @@ pub use create::{Create, Repo};
 pub mod checkout;
 pub use checkout::Checkout;
 
+/// TODO(finto)
```
rudolfs, line 16: What's the TODO?

xla: It's module level doc. It was left to quiten the clippy. Should be addressed before approved.

FintanH: TODOcument

proxy/api/src/identity.rs
```diff
@@ -64,14 +64,22 @@ pub async fn get(state: &coco::State, id: coco::Urn) -> Result<Identity, error::
     Ok((state.peer_id(), user).into())
 }
 
+// TODO(finto): Check if this is used and if so, express more elegantly after
```
FintanH, line 67: This isn't used anywhere yet. It corresponds to the `GET identities/` endpoint. @MeBrei this was for the list of profiles work, right?

rudolfs (approved):
🔥

xla (approved):
❇️ 🚀 ⏬ 🎄

Issue #1029, opened 2020-10-13
Improve seed node input in setting


juliendonck:
As experienced in the integration party there's a need to improve the seed node input. There's an old commit lying around somewhere where it was implemented in a better way. Let's see if we can revert it to that or else I think we still have the designs somewhere.

This is what it is about:
<img width="1151" alt="Screenshot 2020-10-13 at 20 36 16" src="https://user-images.githubusercontent.com/2326909/95901512-c1efbd80-0d93-11eb-90d2-b587c69ea239.png">

juliendonck:
This is the commit: https://github.com/radicle-dev/radicle-upstream/pull/629/commits/457a086598ddee17abb01d2313b7c1e920572ef4

rudolfs:
As discussed with @cloudhead, @juliendonck and @brandonhaslegs.

We'll be going for [this design](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=5848%3A2641):

<img width="788" alt="Screenshot 2020-10-20 at 18 23 59" src="https://user-images.githubusercontent.com/158411/96709352-e0107b80-139a-11eb-9b2a-bd8967b346ae.png">
<img width="776" alt="Screenshot 2020-10-20 at 18 24 04" src="https://user-images.githubusercontent.com/158411/96709359-e141a880-139a-11eb-8488-ac97d40717f5.png">
<img width="777" alt="Screenshot 2020-10-20 at 18 24 11" src="https://user-images.githubusercontent.com/158411/96709361-e272d580-139a-11eb-9834-9223a6e15248.png">

rudolfs:
@xla for the "Restart" button I was thinking of wiring up a new IPC call to the Electron `main.js` to restart the proxy, but am open for other suggestions.

xla:
> @xla for the "Restart" button I was thinking of wiring up a new IPC call to the Electron `main.js` to restart the proxy, but am open for other suggestions.

Does electron support a restart? If so then it sounds like the correct approach.

rudolfs:
@xla in the end I had to go a different route, because currently we have two different ways of managing the proxy lifecycle: dev and production. I felt it would have been too big of a change to make it work for both.

Instead I opted for calling the `v1/control/reload` endpoint to reload the seed configuration.

xla:
[...]

PR #1082, opened 2020-10-21
feat(ui): implement new seed input form

rudolfs:
With this change:
  - we only allow adding and removing seeds one-by-one
  - allow the user to copy the seed to clipboard
  - perform validations
    - to make sure the seed is in a certain shape: `<peerId>@<host>:<port>`
    - to prevent adding duplicate seeds
  - ask the user to restart the app
    - the restart notification is made sticky and is always at the bottom of the notification stack
  - ~prompt the user to reload the peer so the configuration changes take effect~:
    - ~during the configuration reload we prevent the user from interacting with the UI by showing a busy cursor~
    - ~change the connectivity status indicator to "stopped" when the SSE channel is disconnected~

Closes: https://github.com/radicle-dev/radicle-upstream/issues/1029.

demo


cloudhead (approved):
Looks good :+1:

xla (changes requested):
Code and UX look great, as discussed we can't guarantee proper reload so we should remove it for now.

There is one minor edge-case which is not a big deal, when an existing seed address is entered and without clearing the input the entry that was conflicting was deleted the validation still shows the same error.

ui/src/session.ts
```diff
@@ -126,9 +129,87 @@ export const updateCoCo = (coco: CoCo): void =>
...
+
+  restartNotificationVisible = false;
+};
+
+const showRestartNotification = (): void => {
+  // The restart notification is sticky and will stay visible until the user
+  // clicks the "Restart" button, no need to show it multiple times.
+  if (restartNotificationVisible) return;
+
+  notification.error(
```
xla, line 184: I understand that we want to make sure the user doesn't miss this, but isn't it a bit of an abuse to show an error here? Can't we have something like a warning variant?

rudolfs: Good point. In the [style guide](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=586%3A0) we only have 2 levels so far.

Was actually thinking of renaming `error` in the code to `attention`.

rudolfs: Maybe @brandonhaslegs can let us know if we need a third variant?

brandonhaslegs: If we had persistent notifications (notifications that don't go away until you interact with them) then I would say keeping it black is fine. But this notification is **critical.** The user **must** restart to see results. Therefore, the user **must** see this notification. This is why I thought making it an error would be best. I don't see this as abuse, I see this as being helpful. If they miss this notification, they're confused, the app is "broken," they close the app and never come back.

Maybe I could settle for adding an "Attention" color. If we decide that's what we want to do, here's a design for it: https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=5863%3A47629

xla: Thanks for the context @brandonhaslegs - if I understand the code correctly the notification sticks around with this change, but maybe I'm off.

rudolfs:
@xla addressed all your comments.

ui/src/session.ts
rudolfs: @xla correct, we have persistent notifications now. Will change the notification type to `info` then, which will make the sticky notification black.

xla (approved):
🐎 🐑 🔱 💿

cloudhead (approved):

PR #1089, opened 2020-10-22
feat(proxy): streaming seed discovery

xla:
Alleviates the need to restart the app when the seed list in the settings was updated.


kim (dismissed):
lgtm

FintanH (approved):
:man_dancing:

cloudhead (commented):
N:ice_cube:

FintanH (approved):
:metal:

proxy/api/src/main.rs
```diff
@@ -86,9 +87,38 @@ async fn run(
...
+                let seeds = seed::resolve(&seeds).await.unwrap_or_else(|err| {
+                    log::error!("Error parsing seed list {:?}: {}", seeds, err);
+                    vec![]
+                });
+
+                if seeds == last_seeds {
+                    continue;
+                }
+
+                if seeds_sender.broadcast(seeds.clone()).is_err() {
```
NunoAlexandre, line 113: Would be helpful to `log.error!` here, no?

NunoAlexandre (approved):

rudolfs:
You might want to remove the notification then:

Screenshot 2020-10-22 at 15 59 42

https://github.com/radicle-dev/radicle-upstream/blob/67313e05a2e1ad08279e2ee70764b63d88a34cb6/ui/src/session.ts#L154

proxy/api/src/main.rs
xla: The error is not really helpful as it means all receivers are gone. It determines that this task should stop continuing its work. While we suffer from a complex graph of long-running routines for now, we should be most adamant about releasing resources when shutting down/reloading. Proxy's main binary is overdue for a more structured approach to plumbing taking the `Rigging` idea to the next level.

PR #1093, opened 2020-10-22
fix(proxy): avoid double discovery on startup

xla:
As the initial value is set when the watch channel is created any
subsequent calls will produce another message. So far the caching of the
seed to avoid continues sending of discovery updates was done in the
main binary. This led to besides the initial value another one was sent
on the first iteration of that task. Leaving the Peer in an offline
state after one of them got disconnected.

***
Addresses a regression introduced in #1089


rudolfs (dismissed):
LGTM 🚀

proxy/coco/src/config.rs
```diff
@@ -136,13 +136,18 @@ impl discovery::Discovery for StreamDiscovery {
...
-                    .into_iter()
-                    .map(|seed| (seed.peer_id, vec![seed.addr]))
-                {
+                if last_seeds == seeds {
+                    continue;
+                }
+
+                for pair in seeds.iter().map(|seed| (seed.peer_id, vec![seed.addr])) {
                     sender.send(pair).await.ok();
                 }
```
NunoAlexandre, line 148: I'd find this more readable :point_down: 

```suggestion
                for seed in seeds.iter() {
                    let pair = (seed.peer_id, vec![seed.addr]);
                    sender.send(pair).await.ok();
                }
```

rudolfs (approved):

NunoAlexandre (approved):

PR #1094, opened 2020-10-22
fix(proxy): correct handling of connections

xla:
Long overdue iteration on the Peers core run loop to correctly handle
events from the protocol pertaining to connections. At the same time
reacting appropriately to sync events which have been neglected so far.
There is a range of further improvements that could land subsequently.


rudolfs (approved):

proxy/coco/src/peer/run_state.rs
```diff
@@ -394,61 +424,74 @@ impl RunState {
...
             (_, ProtocolEvent::Disconnecting(peer_id)) => {
-                self.connected_peers.remove(&peer_id);
+                if let Some(counter) = self.connected_peers.get_mut(&peer_id) {
+                    *counter -= 1;
+
+                    if *counter == 0 {
+                        self.connected_peers.remove(&peer_id);
+                    }
+                } else {
+                    unreachable!();
```
FintanH, line 493: Why do we need this `else` branch?

proxy/coco/src/peer/run_state.rs
```diff
@@ -299,7 +301,7 @@ pub struct RunState {
     /// Confiugration to change how input [`Input`]s are interpreted.
     config: Config,
     /// Tracking remote peers that have an active connection.
-    connected_peers: HashSet<PeerId>,
+    connected_peers: HashMap<PeerId, usize>,
```
FintanH, line 304: I'm not sure I understood the motivation for the `HashMap`. Could you educate me, please? :)

FintanH (changes requested):

proxy/coco/src/peer/run_state.rs
xla: The peer can receive multiple connections from the same `peer_id`,  a vivid example is the extra connection that is established during syncs (git only exchange). If we just keep track of simply a binary connected status the following order of events can leave the peer in an offline state: `Connect(Peer1) -> Connect(Peer1) -> Disconnect(Peer1)`.

proxy/coco/src/peer/run_state.rs
xla: This branch should never been hit, hence the unreachable. Haven't found a way to encode this invariance.

proxy/coco/src/peer/run_state.rs
FintanH: Ahhhh very good :clap: Can you expand on the docs for `connected_peers` to outline this case? Some prose like that would have illuminated my mind immediately :bow:

proxy/coco/src/peer/run_state.rs
FintanH: But that branch doesn't need to be included. So could we log an error instead or something? Like:

```rust
log::error!("The impossible has happened, somehow we disconnected from '{}' without already being connected to them", peer_id);
```

proxy/coco/src/peer/run_state.rs
xla: 1c73907

FintanH (approved):
Nice one :metal:

rudolfs (approved):

PR #1096, opened 2020-10-23
refactor(proxy): Move binary code into `api` crate

### LINKED ISSUE #1052: Make it easier to detect unused code
At the moment all modules in the `api` crate (`proxy/api/src/lib.rs`) are public but only a fraction of these APIs are used for the proxy binary in `proxy/api/src/main.rs`. This results in unused code not being detected properly.

As a solution I’d suggest moving all the code from `proxy/api/src/main.rs` into a module in the crate and only export that module to build the binary.

geigerzaehler:
To benefit from all linting we move all code from `api/src/main.rs` to `api/src/lib.rs`.

Fixes #1052


xla (changes requested):
Conceptually this change is nice. There are a couple of changes that would put things in the right place.

* in a large crate like this the `lib.rs` should be purely `mod` declarations
* operational concerns still belong in `main.rs` translation of binary arguments, init of logging
* the rigging and run loop shenanigans might be best moved to something like `node` module
* the main in lib code should be thoroughly avoided as it is not a sensible API for anything but a binary

xla:
As inspiration the work currently done for the seed node in radicle-bins can be used as inspiration.

geigerzaehler:
> operational concerns still belong in `main.rs` translation of binary arguments, init of logging

What code specifically are you talking about? Code like `env::set_if_unset` comes from the crate. Using it in `main.rs` kind of defeats the purpose.

In addition, the `Args` struct is used, for example, by `rig` so we can’t just keep it in `main`. Let’s maybe try to not get caught up in getting the separation perfect.

> the rigging and run loop shenanigans might be best moved to something like `node` module

How do you feel about using a `process` module?

xla:
> In addition, the `Args` struct is used, for example, by `rig` so we can’t just keep it in `main`. Let’s maybe try to not get caught up in getting the separation perfect.

Everything that translates from and to OS APIs, primitives into domain

* ENV variables mapping to config
* reading args mapping it to config
* exit codes
* STDIN/STDOU/STDERR

naturally belong in the main where the translation into the domain types are happening.

> How do you feel about using a `process` module?

Sure, not too fuzzed about the actual name. :>

proxy/api/src/project.rs
```diff
@@ -140,15 +140,6 @@ impl Projects {
 
         Ok(projects)
     }
-
-    /// Give back an `Iter` that can be used to iterate over the projects. It first yields
-    /// contributed projects and then tracked projects.
-    pub fn iter(&self) -> Iter<'_> {
```
FintanH, line 146: I mean, it's unused but doesn't mean it's not useful for the future :thinking: How come it doesn't trigger on the `Iter` struct too?

geigerzaehler:
> Everything that translates from and to OS APIs, primitives into domain

Could you have another look? This should be addressed now.

FintanH (commented):
Overall, I'm digging the reorganisation :+1:

proxy/api/src/project.rs
geigerzaehler: This is the same code as for `<Projects as IntoIter>::into_iter()` which probably is used implicitly through a `for .. of`.

FintanH: Ah very good, thanks :grin:

FintanH (approved):

rudolfs (approved):

FintanH (approved):

xla (approved):
🏭 📱 ♉️ 🔉

PR #1097, opened 2020-10-23
refactor(proxy): Remove `enable_fixture_creation_flag`

geigerzaehler:
We remove the `enable_fixture_creation_flag` that was set to `args.test` and passed down through the API. Since #1086 this behavior is implemented by `ctx.test` in `http::api`.


xla (approved):
(*^o^*)

rudolfs (approved):

PR #1102, opened 2020-10-24
fix(ui): show correct projects for users

### LINKED ISSUE #1100: Wrong projects in remote user profile
We tested navigating to a user's profile and it was listing our local projects. For example, if I created a project "foo" and cloned a project "bar" from Baz, if I clicked on Baz's profile it will show up with "foo" instead of "bar".

xla:
The underlying issue was an automatic preloading on store initialisation
which fetched the owners projects. It would override the list of fetched
projects for the accessed user profile. To avoid similar subtle bugs
when building up data to satisfy the screen, all stores for the
UserProfile screen and its fragments are isolated into their own ts file
with a streamlined public api.

Fixes #1100

### DIFF
--- a/ui/Screen/Profile/Projects.svelte
+++ b/ui/Screen/Profile/Projects.svelte
@@ -1,10 +1,10 @@
-<script lang="ts">
+<script lang="typescript">
   import { getContext } from "svelte";
   import { push } from "svelte-spa-router";
 
   import * as modal from "../../src/modal";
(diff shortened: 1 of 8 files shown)


FintanH (approved):
Good catch :ok_hand:

NunoAlexandre (approved):

rudolfs (approved):

PR #1103, opened 2020-10-24
feat(proxy): list project requests

xla:
Exposes an endpoint to list existing requests for projects in
preparation for a deeper UI integration.

Part of #984


proxy/coco/src/peer/control.rs
```diff
@@ -11,15 +11,23 @@ use crate::request;
 use super::run_state::Status;
 
 /// Requests sent to the peer.
+#[allow(clippy::pub_enum_variant_names)]
 #[derive(Debug)]
 pub enum Request {
     /// Request the current status.
     CurrentStatus(oneshot::Sender<Status>),
+    /// Request a project request.
+    GetProjectRequest(
```
FintanH, line 20: Could name them `Project` and `Projects` or `GetProject(s)`. I quite like the idea of the enum use being `Request::Project` and `Request::Projects` since it sounds like the commands. Just a suggestion anyway :)

proxy/coco/src/peer/run_state.rs
```diff
@@ -608,10 +627,18 @@ impl RunState {
             },
             (_, RequestInput::Requested(urn, time, maybe_sender)) => {
                 let maybe_request = self.waiting_room.request(&urn, time);
+                let request = match maybe_request {
+                    Some(req) => req,
+                    None => self
```
FintanH, line 632: It'd be better to just return `Either` from `request` so that we don't have this `expect` business.

FintanH (changes requested):
Is the plan to add the call to the UI for the listing in another PR?

xla:
> Is the plan to add the call to the UI for the listing in another PR?

Yes: #1104

proxy/coco/src/peer/control.rs
xla: I do too. But the `Request` enum is not the thing as the "Project Reqests". If it would be named `Request::Projects` what would it the project be in that case. It's  a bit unfortunate that we refer to them as requests in the context of the waiting room and the commnication triggered by the peer control. #namingishard

proxy/coco/src/peer/run_state.rs
xla: I guess you mean `Option` here. #nohaskell And it wouldn't be better, as the caller then has to do another roundtrip to get the created request, rather argue that `waiting_room.request` ought to return a request in any case and not an `Option`.

FintanH: I think we're misunderstanding each other. The point I was trying to make is that request could return Either which is either the new request that is created or the existing request. That is, don't return Option here and return Either instead. This would mean we don't need to call get in the other branch. We would just call `into` to turn the new Created request into a SomeRequest.

Is that clearer or am I missing something?

xla: > Is that clearer or am I missing something?

Indeed, such clarification much types. Thanks that makes sense.

FintanH: Types as documentation 👌

proxy/coco/src/peer/control.rs
```diff
@@ -61,19 +80,48 @@ impl Control {
         receiver.await.expect("receiver is gone")
     }
 
+    /// Initiate a new reuest to fetch a project request.
```
rudolfs, line 83:
```suggestion
    /// Initiate a new request to fetch a project request.
```

FintanH (approved):
🤫

NunoAlexandre (approved):
LGTM with one remark: the terminology "owner" to mean "the current user" is confusing. I'd find, for example, "list_user_requested" clearer than "list_owner_requested". "Owner" leads to thinking of project ownership, amongst other things.

rudolfs (approved):

xla:
> LGTM with one remark: the terminology "owner" to mean "the current user" is confusing. I'd find, for example, "list_user_requested" clearer than "list_owner_requested". "Owner" leads to thinking of project ownership, amongst other things.

How do you refer to other users in this scenario? The reason we use owner explicitely is that the handling is specialised for the "owner" of the monorepo. Open to suggestions, but so ar nothing presented itself.

rudolfs (approved):

NunoAlexandre:
> the "owner" of the monorepo. Open to suggestions, but so ar nothing presented itself.

I see. I guess this is fine, shall anyone else in the future get confused with it we improve it then.

NunoAlexandre (approved):

PR #1104, opened 2020-10-25
feat(ui): list project requests

xla:
Ref. #984


rudolfs:
We probably shouldn't show the placeholder if we already have some projects in the queue:

Screenshot 2020-10-26 at 12 24 09

ui/DesignSystem/Component/HorizontalMenu/MenuItem.svelte
```diff
@@ -41,7 +41,7 @@
 </style>
 
 <!-- svelte-ignore a11y-missing-attribute -->
```
rudolfs, line 43: You don't need this anymore, now that you set the `{href}`.

ui/DesignSystem/Component/Remote.svelte
```diff
@@ -44,6 +44,10 @@
     <WithContext {data} name={context}>
       <slot {data} />
     </WithContext>
+  {:else if Array.isArray(data) && data.length === 0}
+    <slot name="empty" />
+  {:else if data === null}
+    <slot name="empty" />
```
rudolfs, line 50:
```suggestion
  {:else if (Array.isArray(data) && data.length === 0) || data === null}
    <slot name="empty" />
```

xla:
> We probably shouldn't show the placeholder if we already have some projects in the queue:

Yeah probably, not clear to me how the screen should behave for the different combinations of state. @radicle-dev/product Any opinion?

ui/DesignSystem/Component/SidebarLayout.svelte
```diff
@@ -1,12 +1,16 @@
...
+  import type { Authenticated } from "../../src/session";
+
   import Sidebar from "./Sidebar.svelte";
 
   export let dataCy = "";
   export let style = "";
 
-  const session = getContext("session");
+  const session: Authenticated = getContext("session");
+  const identity = session.identity || fallback;
```
geigerzaehler, line 13: If `session` is of type `Authenticated`, then `session.identity` is always set. But in actuality, `session` can be of type `Unauthenticated`. In this case We should probably not use the fallback but rather throw or show an error.

ui/DesignSystem/Component/Remote.svelte
```diff
@@ -44,6 +44,10 @@
     <WithContext {data} name={context}>
       <slot {data} />
     </WithContext>
+  {:else if Array.isArray(data) && data.length === 0}
```
geigerzaehler, line 47: Having a special handling for empty arrays might trip people up. It also limits the usability if we actually want to distinguish between the data not being there and an empty array. Is there a reason why we can’t have this guard at the site where `Remote` is used?

ui/DesignSystem/Component/SidebarLayout.svelte
xla: Was a leftover, thanks for spotting it. I introduced the vairants after this fallback was in place.

ui/DesignSystem/Component/Remote.svelte
xla: The pattern I observed was that whenever we had lists it would go as follows `{#if list.length > 0} content {:else} EmptyState {/if}`. Ideally the empty should be encoded in the `remote.Data`, because then the owner of the remote store can decide when to transition into that state. Didn't want to do that refactoring just yet.

geigerzaehler: Yes, it might make sense to extract the “if empty” pattern. But I don’t think that `Remote` is the right place for that. Since the pattern is quite small and not very complicated we could just leave it duplicated for now and tackle it properly later. I’ll leave it up to you, though.

geigerzaehler (approved):

rudolfs (approved):

Issue #1050, opened 2020-10-15
Restyle searchbox


brandonhaslegs:
The search box has lost some styles in the last week or so in the latest build.

- [x] the border radius should be .5rem
- [x] there should be no border on the field
- [x] the height of the field should be 3rem
- [x] the `.left-item-wrapper` should have a height of 3rem so it remains vertically centered in the field
- [x]  the background color of the form should not change on `:hover` or `:focus`

Currently looks like this:
![image](https://user-images.githubusercontent.com/4406983/96115177-d55f6d80-0ee6-11eb-925a-391a34706d6d.png)

Final result should look like this:
![image](https://user-images.githubusercontent.com/4406983/96116325-94685880-0ee8-11eb-8e68-a5c8c00ad824.png)

xla:
> [ ] the `.left-item-wrapper` should have a height of 3rem so it remains vertically centered in the field

This breaks the icon when the input has an error. Could we adjust the designs with invalid inputs in mind. Happy to fix the diverged styles then.

xla:
Refs #1031

brandonhaslegs:
> > [ ] the `.left-item-wrapper` should have a height of 3rem so it remains vertically centered in the field
> 
> This breaks the icon when the input has an error. Could we adjust the designs with invalid inputs in mind. Happy to fix the diverged styles then.

Sorry, what does centering the 🔍 icon have to do with errors?

xla:
> Sorry, what does centering the mag icon have to do with errors? Maybe centering can be achieved without changing the height of the `.left-item-wrapper`. All I'm saying is to make sure the icon stays vertically centered when the field height is increased.

Absolutely, just wanted to point out that the change you are referring to by itself won't fix the issue of centering.

Issue #1031, opened 2020-10-13
Expand Search design


xla:
While making the Search feature more robust a couple of things popped up for which designs are needed.

The input is now validated as we only expect valid urns at the moment:

![2020-10-13-233556_1662x1408_scrot](https://user-images.githubusercontent.com/1585/95918991-1489a380-0dad-11eb-8aa8-22d81dcd12d2.png)

When entering a urn it can be that the project is present locally, the output looks a bit barren at the moment:

![2020-10-13-233834_1662x1408_scrot](https://user-images.githubusercontent.com/1585/95919075-3aaf4380-0dad-11eb-8c05-786098ebc445.png)

Follow-up to #1030

xla:
Another issue with the current design is that we show a toggle in the result view. Which leads to behaviour which overlaps witth the toggles shown in the Following tab and makes the potential flows a user can go through quite complex and confusing. Instead we might be better served to just have an action button there which starts the search and navigates the user to the Following tab where the list of ongoing searches is presented.

xla:
https://github.com/radicle-dev/radicle-upstream/issues/1460#issue-757662431

juliendonck:
Copying the important info here from @cloudhead here:

One thing I found myself confused by recently is the functionality which allows you to jump to a project you already have. It looks like this currently:

![2020-12-05-123842_453x187_scrot](https://user-images.githubusercontent.com/40774/101241507-0134ed80-36f7-11eb-8681-9edc9b6db992.png)

I didn't realize I was already following upstream, so when I saw this, I thought "hmm it looks like the widget didn't load properly, where's the follow button?"

I think an improvement would be to include some information, when the project is available locally:

* Some kind of indicator that the project is already followed, eg. `[Following]` or `[local]`.
* Some kind of indicator that clicking on it or pressing enter will navigate to the project, like `-> [enter] to navigate` or just an arrow or something that says "go".

cloudhead:
Yeah same thing happened to me.
I'd actually show a proper project widget (like we have in the seed node)

brandonhaslegs:
Pretty straightforward, but I think we should just add the metadata in the search result. Here's a design:
![image](https://user-images.githubusercontent.com/4406983/119491897-c5eab180-bd5e-11eb-8787-598dfcea4e11.png)

PR #1105, opened 2020-10-26
feat: cancel project search

### LINKED ISSUE #736: proxy: cancel project search
If a user has been trying to locate a project for a long time and isn't having luck, we need a way to cancel the search.

xla:
Closes #736


rudolfs (commented):
Code looks good.

Was confused when testing it out though. After cancelling a follow request, the waiting area still contains the request even after an app refresh. If this functionality is not part of this PR, could you point me to a tracking issue/PR where this will be taken care of?

reload

xla:
@rudolfs You are right, this was an oversight and has been in addressed in 80027ec.

rudolfs (approved):

MeBrei (changes requested):
Maybe these are issues for another PR, but I found the following when testing:

- [ ] when searching for a project (for the first time), then following it and within the search modal hit "unfollow" again, it still shows up in the search list (after refreshing the page)
- [ ] when I cancel a all searches, the `Following` tab is blank (after refresh), ie I would expect the page empty-following-tab page with the look for a project button
-[ ] when I try to follow a project that I have canceled previously, it does not show up in the following list

xla:
> * [ ]  when searching for a project (for the first time), then following it and within the search modal hit "unfollow" again, it still shows up in the search list (after refreshing the page)

Showing the track toggle in the search modal context is confusing in general, it should be a one-time action that navigates away once the search has been started. @brandonhaslegs Can we do something about it?

> * [ ]  when I cancel a all searches, the `Following` tab is blank (after refresh), ie I would expect the page empty-following-tab page with the look for a project button

Addressed in e6ff7d7.

> * [ ]  when I try to follow a project that I have canceled previously, it does not show up in the following list

`Canceled` currently is a finite state for the seach. We have not yet expanded how we bring them back to life, there is a lot of work to be done on this feature including moving these concerns down into the protoocl. For this first iteration we might be okay with accepting that.

brandonhaslegs:
Just confirming here, @xla and I decided to change the follow button from a toggle to a button with two actions:
- tracks the project
- navigates the user to the Following tab on the user's profile.

And also now it's pink not blue.

xla:
Thanks @brandonhaslegs for the follow-up.

All improvements to the search box are out-of-scope for this PR and have been capture in #1031 and #1050

proxy/api/src/http/project/request.rs
```diff
@@ -0,0 +1,170 @@
...
+    ctx: context::Context,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    http::with_context(ctx)
+        .and(warp::get())
+        .and(path::end())
+        .and_then(handler::list)
+}
+
+/// Request handlers for initiating searches for projects on the network.
+mod handler {
```
FintanH, line 48: How would feel about having `project/request/handler.rs`? I've always found it a bit annoying to have these `mod`s in the same file and then I mess up where I import things.

rudolfs (approved):

proxy/coco/src/peer/run_state.rs
```diff
@@ -157,12 +157,25 @@ pub enum AnnounceInput {
...
 pub enum ControlInput {
-    /// Request a project request.
-    ProjectRequest(RadUrn, oneshot::Sender<Option<SomeRequest<Instant>>>),
-    /// Request the list of project requests.
-    ProjectRequests(oneshot::Sender<Vec<SomeRequest<Instant>>>),
     /// New status.
     Status(oneshot::Sender<Status>),
+
+    /// Cancel an ongoing project search.
+    CancelRequest(
```
FintanH, line 164: What about having a `Search` enum here which is embedded in `ControlInput`? Just to separate them out cognitively.

FintanH (approved):
Loving the renaming and separation into modules! I have some suggestions but definitely not blockers -- just surveying what you think :)

proxy/api/src/http/project/request.rs
xla: It's set up that way because it's the fullfilment of a single chain in a single file. The overhead we had to manage with this is the visibilty `pub`s and import `use` management, which I don't think is worth it.

xla: If we change it should be addressed in the entire `http` module.

proxy/coco/src/peer/run_state.rs
xla: Tried that, not enough variants for now, we can come to that later.

PR #1107, opened 2020-10-26
refactor(proxy): introduce Context variants

geigerzaehler:
We make the `Context` type into an enum so that the API is forced to handle the case where the keystore has not been unsealed yet and consequently the coco state and peer are not available.

Currently, the API is always intialized with an unsealed context. Making dynamic changes will be introduced in a follow-up.


xla (approved):
Very welcome and anticipated change.

🔓 📇 🏜 💱

proxy/api/src/http/session.rs
```diff
@@ -48,25 +48,30 @@ fn update_settings_filter(
...
 
     /// Fetch the [`session::Session`].
     pub async fn get(ctx: context::Context) -> Result<impl Reply, Rejection> {
-        let sess = session::current(ctx.state.clone(), &ctx.store).await?;
-
-        Ok(reply::json(&sess))
+        match ctx {
+            context::Context::Unsealed(unsealed) => {
+                let sess = session::current(unsealed.state.clone(), &unsealed.store).await?;
+
```
NunoAlexandre, line 65:
```suggestion
```

NunoAlexandre (approved):
Solid :ok_hand:

MeBrei (approved):
⭐

xla (dismissed):
😏 ◽️ 😄 💮

NunoAlexandre (dismissed):
:surfing_man:

proxy/api/src/http/session.rs
```diff
@@ -48,25 +48,29 @@ fn update_settings_filter(
...
 
         Ok(reply::with_status(reply(), StatusCode::NO_CONTENT))
     }
 
     /// Fetch the [`session::Session`].
     pub async fn get(ctx: context::Context) -> Result<impl Reply, Rejection> {
-        let sess = session::current(ctx.state.clone(), &ctx.store).await?;
-
-        Ok(reply::json(&sess))
+        match ctx {
```
FintanH, line 62: Why isn't this taking `Unsealed` and using the guard?

FintanH (commented):
Just one question :)

proxy/api/src/http/session.rs
geigerzaehler: Excellent catch! The code here is indeed a copy `with_context_unsealed`. I’ve pushed a fix

xla (approved):
💰 💒 👹 🐲

FintanH (approved):
:metal:

PR #1108, opened 2020-10-26
chore(ci): Check intra doc link resolution

geigerzaehler:
CI now checks that all the doc links resolve properly.


xla (approved):
Quality of Life!

🏪 👌 🎠 🎀

.buildkite/run.sh
```diff
@@ -33,7 +33,11 @@ cp .buildkite/.gitconfig "$HOME/"
 cat "$HOME/.gitconfig"
 
 echo "--- Run proxy docs"
-(cd proxy && time cargo doc --no-deps)
+(
+  cd proxy
+  export RUSTDOCFLAGS="-D intra-doc-link-resolution-failure"
+  time cargo doc --workspace --no-deps --all-features --document-private-items
```
rudolfs, line 39: We don't wanna do `&&` here?

rudolfs (approved):

.buildkite/run.sh
geigerzaehler: `&&` is not needed. Just like the main shell for this script the subshell (instantiated by `( ... )`) has the `set -e` option so that any failing command will abort the script.

PR #1110, opened 2020-10-27
fix(ui): revive avatar in profile

xla:
Regression introduced in #1104 - thanks @rudolfs for the spot.


rudolfs:
> thanks @rudolfs for the spot.

Was @brandonhaslegs actually who spotted it.

rudolfs (approved):

FintanH (approved):

Issue #1090, opened 2020-10-22
Unsealed keystore cookie


MeBrei:
To add some security, unsealing the keystore should set a cookie, which needs to be part of every subsequent request.  

# Flow
1. `POST /keystore` or `POST /keystore/unseal` request is sent to the API (via onboarding or unlock flow)
2. on success, a cookie is created, stored in the `Context` and sent in the header of the reply via `Set-Cookie`
3. Electron stores and sends the cookie on all subsequent requests
4. in proxy the `with_unsealed_guard` checks if the right cookie was provided, otherwise returns `FORBIDDEN` (it should maybe also seal the keystore, but this will not be part of the first iteration)

Depends on #1083

MeBrei:
Regarding the following warning mentioned in [#1120(comment)](https://github.com/radicle-dev/radicle-upstream/pull/1120#issuecomment-718845559)

<img width="434" alt="Screenshot 2020-10-29 at 16 51 29" src="https://user-images.githubusercontent.com/158411/97598265-0c657100-1a07-11eb-9e6c-b7d49460ac49.png">

The cookie currently set by `/v1/keystore` or  `/v1/keystore/unseal`. In order for the cookie to be sent with every request, the `Path` is set to `/`. Because of that, setting the cookie is blocked if we sent it with `SameSite=Lax` or `SameSite=Strict`. Setting `SameSite` and `Secure` as suggested above does not work, as we sent requests over `http` and not `https`.

Issue #1083, opened 2020-10-21
Keystore API


geigerzaehler:
We propose APIs for managing the keystore and starting the coco peer.

## Motivation

The proxy can only run a coco peer if it has access to the secret key for the peer. When the proxy starts the key is either encrypted or does not exist yet. In both cases user input is required to obtain the key. Once the secret key is available the current architecture demands that the proxy reloads. This means that the peer is started and the HTTP server is restarted.

A previous design (#1049) built on the existing `/identity` endpoint to provide this functionality. However, it is not possible to initialize the coco peer and create an identity in one request. 

## Endpoints

`GET /session`

Returns 200 (no content) with the settings and identity payload if the key pair exists, is unsealed, and an identity was created.
Returns 403 (forbidden) if there is a key pair but it is sealed
Returns 404 (not found) if there is no key pair or there is no identity

`POST /keystore`

Takes a pass phrase as data. Generates a key pair, encrypts it with the passphrase and stores it on disk. Then it reloads the server and initializes coco with the keypair. Returns 204 (no content).

`POST /keystore/unseal`

Takes a pass phrase as data. Decrypts the key stored on disk with the passphrase and initializes coco with the key pair and reloads the server. Returns 404 if there is no key pair.

## App startup

The keystore APIs are used when the app is started. First, the app issues a `/GET session` request.

* If the response code is 404, there is no peer key pair or identity yet and the onboarding flow is started.
* If the response code is 403, the unseal flow is started.
* If the response code is 200, the app is initialized with the session data.

*Onboarding flow*
1. The user inputs a password and a user handle
2. The frontend sends `POST /keystore` with the password
3. The frontend waits for the proxy to reload
4. The frontend sends `POST /identity` with the user handle
5. The frontend sends `GET /session` and initializes the app with the session data

If any of the steps fail, the onboarding flow is restarted.

*Unseal flow*

1. The user inputs a password
2. The frontend sends `POST /keystores/unseal` with the password
3. The frontend waits for the proxy to reload
4. The frontend sends `GET /session` and initializes the app with the session data

If any of the steps fail, the unseal flow is restarted.

## Alternatives

A previous proposal introduced another endpoint `GET /keystore` that was used to determine whether a key pair existed and was unsealed. This functionality is now merged into the `GET /session` endpoint to reduce the number of requests and intermediate states that are observable by the app.

geigerzaehler:
[...]

PR #1120, opened 2020-10-28
feat: require keystore unsealing

### LINKED ISSUE #997: Add sealed keystore filter guard

### LINKED ISSUE #998: Add and use unseal endpoint on app start

geigerzaehler:
With this PR we require the user to unseal the keystore with a password after the app is restarted. This closes #998. For the moment the password is hard-coded to `radicle-upstream` and not actually used in the key store. This will be implemented in #999.

The PR consist of two major changes. First, we reorganize the APIs according to #1083 and #1090. As a result of this we now always store the identity in the session, which necessitates changes to `api::session` and requires us to introduce a more granular `Session` type in `ui/src/session.ts`.

The second major change is the addition of `api::service` in the proxy. This module provides the necessary infrastructure for managing the coco peer and API configuration. It is now possible send changes to the configuration from the API to the top-level and have the peer restarted with a different configuration. (In our case this is the secret key.)

## Follow-ups

* Use the user-provided password to unseal the on-disk key (#999)
* Don’t reset session (and other state) when unsealing in tests and  re-enable skipped tests (#1124)
* Randomly generate auth token (issue TBD)


ui/src/session.ts
```diff
@@ -6,35 +6,43 @@ import * as event from "./event";
...
+import { Appearance, CoCo, Settings, defaultSetttings } from "./settings";
 
 import { createValidationStore, ValidationStatus } from "./validation";
 
 // TYPES
-export interface Authenticated {
-  identity: identity.Identity;
-  settings: Settings;
+export enum Status {
+  NoSession = "No_SESSION",
```
FintanH, line 15:
```suggestion
  NoSession = "NO_SESSION",
```
No_SeSsIoN :)

FintanH (commented):
I made it to the end!

ui/src/session.ts
MeBrei: Fixed in 8a81405

proxy/api/src/http/keystore.rs
```diff
@@ -0,0 +1,87 @@
...
+    use crate::{context, error};
+
+    /// Unseal the keystore.
+    pub async fn unseal(
+        mut ctx: context::Context,
+        input: super::UnsealInput,
+    ) -> Result<impl Reply, Rejection> {
+        // TODO(merle): Replace with correct password check
+        if input.passphrase != "radicle-upstream" {
+            return Err(Rejection::from(error::Error::KeystoreSealed));
```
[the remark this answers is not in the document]
MeBrei: Added in 8a81405

proxy/api/src/http/keystore.rs
```diff
@@ -0,0 +1,87 @@
+//! Endpoints for handeling the keystore.
```
rudolfs, line 1:
```suggestion
//! Endpoints for handling the keystore.
```

proxy/api/src/http/keystore.rs
```diff
@@ -0,0 +1,87 @@
...
+fn create_filter(
+    ctx: context::Context,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    warp::post()
+        .and(path::end())
+        .and(http::with_context(ctx))
+        .and_then(handler::create)
+}
+
+/// Keystore handlers for conversion between core domain and HTTP request fullfilment.
```
rudolfs, line 35:
```suggestion
/// Keystore handlers for conversion between core domain and HTTP request fulfilment.
```

proxy/api/src/http/keystore.rs
```diff
@@ -0,0 +1,87 @@
...
+        *cookie = Some("chocolate".into());
+        Ok(warp::reply::with_header(
+            reply::with_status(reply(), StatusCode::NO_CONTENT),
+            "Set-Cookie",
+            "auth-cookie=chocolate; Path=/",
+        )
+        .into_response())
+    }
+
+    /// Initialize the key store with a new key
```
rudolfs, line 66:
```suggestion
    /// Initialize the keystore with a new key.
```

cypress/integration/onboarding.spec.js
```diff
@@ -69,32 +69,36 @@ context("onboarding", () => {
...
       cy.pick("repeat-passphrase-input").type(validUser.passphrase);
       cy.pick("set-passphrase-button").click();
-      cy.pick("notification")
-        .contains(
-          /Could not create identity: the identity 'rad:git:[\w]{3}…[\w]{3}' already exists/
-        )
+
+      // Success screen.
+      cy.pick("urn")
+        .contains(/hwd[\w]{56}/)
```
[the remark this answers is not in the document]
geigerzaehler: I don’t know where this regex is coming from to be honest. I just mechanically split up this one test into three to better isolate the failing test case.

ui/Screen/Lock.svelte
```diff
@@ -0,0 +1,47 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    display: flex;
+    flex-direction: column;
+  }
+</style>
+
+<div class="lock">
+  <Input.Text
```
rudolfs: This should be `Input.Password`.

Screenshot 2020-10-29 at 10 05 13

vs

Screenshot 2020-10-29 at 10 42 02

proxy/api/src/http/keystore.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Unseal the keystore.
+    pub async fn unseal(
+        mut ctx: context::Context,
+        input: super::UnsealInput,
+    ) -> Result<impl Reply, Rejection> {
+        // TODO(merle): Replace with correct password check
+        if input.passphrase != "radicle-upstream" {
+            return Err(Rejection::from(error::Error::KeystoreSealed));
+        }
+        let key = coco::keys::SecretKey::new();
```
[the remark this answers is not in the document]
geigerzaehler: Yes, this is temporary and will be addressed in #999. I’ve added a comment in https://github.com/radicle-dev/radicle-upstream/pull/1120/commits/67726750ba09f3ef012a4b6ed8f0970d038a4e50 that makes this explicit.

rudolfs (changes requested):
Magnum opus!

Here's some blockers:

- Network activity indicator isn't shown after unseal nor after onboarding, a refresh reestablishes the SSE connection. We should probably only start the SSE connection when the backend is already unsealed.
Screenshot 2020-10-29 at 10 07 17


- We should block the user from interacting and show the waiting cursor when creating the identity (could use `screen.lock` from `screen.ts`)
Screenshot 2020-10-29 at 10 48 30

- Unsealing session takes a while and we just see a white screen, it should probably still show the unlock screen but block the user from interacting and show a waiting cursor as in the point above.

Screenshot 2020-10-29 at 10 53 55



And a few observations:
- the unlock screen doesn't know our color theme 😭 
- invalid password entries have no timeout -> brute force attacks?

ui/Screen/Lock.svelte
```diff
@@ -28,20 +28,28 @@
...
-  </Button>
+  <Emoji emoji={'🚪'} size="huge" style="margin-bottom: 0.75rem;" />
+  <h3>Unlock the app</h3>
+  <div class="form">
+    <Input.Password
+      autofocus
+      placeholder="Enter your passphrase"
+      bind:value={passphrase}
+      dataCy="passphrase-input"
+      on:keydown={onKeydown}
```
MeBrei, line 46: `Input.Password` already does the keydown-enter event handling, which we could use here.

So you could do here: `on:enter={onEnter}` and then the function in this file could be reduced to:
`const onEnter = () => {
    if (passphrase.length > 0) {
      unlock();
    }
  };`

cypress/integration/onboarding.spec.js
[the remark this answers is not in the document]
FintanH: I think it's used multiple times, so maybe we could make it a constant at least. But I believe the regex should be based on `z-base32` as documented here https://github.com/radicle-dev/radicle-link/blob/master/docs/spec/identities.md#radicle-urns

ui/Screen/Lock.svelte
MeBrei: Addressed in 51c9c77

proxy/api/src/http/keystore.rs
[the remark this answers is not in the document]
FintanH: Appreciate it :ok_hand:

ui/Screen/Lock.svelte
rudolfs: The <kbd>enter</kbd> key doesn't work for submission anymore.

geigerzaehler:
> invalid password entries have no timeout -> brute force attacks?

@rudolfs: When we use the real keystore (#999), trying to unlock will take some time that will make it practically impossible to brute-force the password

> Network activity indicator isn't shown after unseal nor after onboarding, a refresh reestablishes the SSE connection. We should probably only start the SSE connection when the backend is already unsealed.

Looking into this

cypress/integration/onboarding.spec.js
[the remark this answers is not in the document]
geigerzaehler: Addressed in https://github.com/radicle-dev/radicle-upstream/pull/1120/commits/fb4cd46565019c76b1812fa6a56c4212db3fadd7

cypress/integration/onboarding.spec.js
```diff
@@ -1,9 +1,11 @@
-context("onboarding", () => {
+context.only("onboarding", () => {
```
rudolfs, line 1: 👀

geigerzaehler: oops

MeBrei:
> * We should block the user from interacting and show the waiting cursor when creating the identity (could use `screen.lock` from `screen.ts`)

This will be addressed in a separate PR

> * Unsealing session takes a while and we just see a white screen, it should probably still show the unlock screen but block the user from interacting and show a waiting cursor as in the point above.
> 

90ae06b added a loading screen, which will be dressed up in a separate PR as well

juliendonck (approved):

juliendonck:
> [90ae06b](https://github.com/radicle-dev/radicle-upstream/commit/90ae06b4fa0c3fd580e068bfa913b2220924d143) added a loading screen, which will be dressed up in a separate PR as well

Two issues have been created for this: #1128 and #1127

rudolfs (changes requested):
<details>
<summary>
After onboarding for the first time our local peer isn't started automatically:
</summary>

Screenshot 2020-10-29 at 13 25 34


```
yarn run v1.22.5
$ RADICLE_UPSTREAM_EXPERIMENTAL=false yarn start:app
$ run-p --race svelte:watch proxy:start electron:start
$ yarn svelte:clean && rollup -c -w
$ cd proxy && cargo build --bin git-remote-rad && cargo run
$ wait-on ./public/bundle.js && wait-on ./native/main.comp.js && wait-on tcp:8080 && NODE_ENV=development electron .
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
$ rm -rf public/bundle.*
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
     Running `target/debug/api`
 INFO  coco::git_helper > Making sure git-remote-rad helper is set up
 INFO  api::process     > starting API
rollup v2.26.9
bundles ui/index.ts → public/bundle.js...
LiveReload enabled
created public/bundle.js in 10.6s
bundles native/main.js → native/main.comp.js...
created native/main.comp.js in 433ms

[2020-10-29 13:25:04] waiting for changes...
 ERROR api::http::error > Rejection([NoSession, MethodNotAllowed])
 INFO  proxy::http      > "GET /v1/session HTTP/1.1" 404 1.504316ms
 INFO  proxy::http      > "POST /v1/keystore HTTP/1.1" 204 982.538µs
 INFO  librad::net::protocol > Protocol::run; local.id=hyycnr1twd1c37wbzpa9a1gxksknhpaynqdjff6zphui9uhuhdzjte local.addr=0.0.0.0:56768
 INFO  api::process          > starting API
 INFO  api::process          > starting peer
 INFO  librad::net::protocol > Listening
 INFO  librad::git::storage  > Storage::create_repo; local.id=hyycnr1twd1c37wbzpa9a1gxksknhpaynqdjff6zphui9uhuhdzjte
 INFO  librad::git::types    > creating symbolic ref refs/namespaces/hwd1yreynm5jbt8s53yszui73hr9iqoa38meimj9ow7zqw6xx7c1or7ppga/refs/rad/self -> refs/namespaces/hwd1yreynm5jbt8s53yszui73hr9iqoa38meimj9ow7zqw6xx7c1or7ppga/refs/rad/id
 INFO  proxy::http           > "POST /v1/identities HTTP/1.1" 201 19.69224ms
 INFO  proxy::http           > "GET /v1/session HTTP/1.1" 200 777.083µs
 INFO  proxy::http           > "GET /v1/notifications/local_peer_events HTTP/1.1" 200 490.085µs
 INFO  proxy::http           > "GET /v1/projects/contributed HTTP/1.1" 200 3.866422ms
 INFO  proxy::http           > "GET /v1/projects/contributed HTTP/1.1" 200 3.927986ms
```
</details>

<details>
<summary>
After an app restart the peer is started as expected.
</summary>

Screenshot 2020-10-29 at 13 33 48


```
yarn run v1.22.5
$ RADICLE_UPSTREAM_EXPERIMENTAL=false yarn start:app
$ run-p --race svelte:watch proxy:start electron:start
$ cd proxy && cargo build --bin git-remote-rad && cargo run
$ yarn svelte:clean && rollup -c -w
$ wait-on ./public/bundle.js && wait-on ./native/main.comp.js && wait-on tcp:8080 && NODE_ENV=development electron .
$ rm -rf public/bundle.*
    Finished dev [unoptimized + debuginfo] target(s) in 0.28s
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/api`
 INFO  coco::git_helper > Making sure git-remote-rad helper is set up
 INFO  api::process     > starting API
rollup v2.26.9
bundles ui/index.ts → public/bundle.js...
LiveReload enabled
created public/bundle.js in 10.6s
bundles native/main.js → native/main.comp.js...
created native/main.comp.js in 433ms

[2020-10-29 13:31:02] waiting for changes...
 ERROR api::http::error > Rejection([KeystoreSealed, MethodNotAllowed])
 INFO  proxy::http      > "GET /v1/session HTTP/1.1" 403 1.409732ms
 INFO  proxy::http      > "POST /v1/keystore/unseal HTTP/1.1" 204 1.006691ms
 INFO  librad::net::protocol > Protocol::run; local.id=hyycnr1twd1c37wbzpa9a1gxksknhpaynqdjff6zphui9uhuhdzjte local.addr=0.0.0.0:54341
 INFO  api::process          > starting API
 INFO  api::process          > starting peer
 INFO  librad::net::protocol > Listening
 INFO  librad::net::protocol > Establishing connection remote.id=hynewpywqj6x4mxgj7sojhue3erucyexiyhobxx4du9w66hxhbfqbw
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  librad::net::protocol > New outgoing connection remote.id=hynewpywqj6x4mxgj7sojhue3erucyexiyhobxx4du9w66hxhbfqbw
 INFO  quinn::connection     > drive; id=0
 INFO  proxy::http           > "GET /v1/session HTTP/1.1" 200 593.703µs
 INFO  proxy::http           > "GET /v1/notifications/local_peer_events HTTP/1.1" 200 454.719µs
 INFO  proxy::http           > "GET /v1/projects/contributed HTTP/1.1" 200 3.345909ms
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
 INFO  quinn::connection     > drive; id=0
```
</details>

geigerzaehler:
@rudolfs Great work spotting these things :mag:! I’ve fixed the bug in https://github.com/radicle-dev/radicle-upstream/pull/1120/commits/9d18de99310087b6c0beab75584ae294fe25be08

rudolfs (approved):
🥳 🎉

NunoAlexandre (dismissed):
1. Great PR description, super helpful to be able to go through it
2. The engineering decisions look solid
3. Great work Merle & Thomas :ok_hand:

rudolfs:
Seeing this in the console, might be worth a follow-up issue:

Screenshot 2020-10-29 at 16 51 29

```
A cookie associated with a cross-site resource at http://localhost/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.
```

FintanH (dismissed):
Great stuff peeps :clap:

FintanH:
Oh no. Someone didn't GPG sign their commits

MeBrei:
> Oh no. Someone didn't GPG sign their commits

🤦 that are the two commits I made while radicle changed my git name and email address

proxy/api/src/http/keystore.rs
```diff
@@ -0,0 +1,91 @@
...
+        .into_response())
+    }
+}
+
+/// Bundled input data for unseal request.
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UnsealInput {
+    /// Passphrase to unlock the keystore.
+    passphrase: String,
```
kim, line 90: Should use `SecUtf8` (or zero memory manually on drop).

geigerzaehler: Good catch! Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1120/commits/94a3ff4e52db3b48c6e29829a95ad86688fcace8

FintanH (dismissed):
It's a secret :shushing_face:

rudolfs (approved):

cloudhead:
:partying_face:

proxy/api/src/context.rs
```diff
@@ -1,5 +1,9 @@
 //! Datastructure and machinery to safely share the common dependencies across components.
 
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+use crate::service;
 use coco::PeerControl;
```
xla, line 7:
```suggestion
use std::sync::Arc;

use tokio::sync::RwLock;

use coco::PeerControl;

use crate::service;
```

proxy/api/src/http/identity.rs
```diff
@@ -54,20 +54,23 @@ mod handler {
...
+                coco::state::Error::already_exists(session.identity.urn),
             )));
         }
 
         let id = identity::create(&ctx.state, &input.handle).await?;
 
-        session::set_identity(&ctx.store, id.clone())?;
+        session::initialize(&ctx.store, id.clone())?;
+        let mut cookie = ctx.auth_cookie.write().await;
+        *cookie = Some("chocolate".into());
```
xla, line 67: When are we going to create unique ones?

xla (changes requested):
Great work and massive feat. One thing that sticks out is the key and cookie handling inside of the http code, which I see as a sign that things are not landed yet in the right place. Wouldn't it be possible instead of passing the key to the method of the service handle, it takes the secstr which is the password and the attempted unsealing of the keystore is done in the service code, if the unsealing succeeds it creates a session secret which can be used as the `Set-Cookie` header in the response. If this is possible indeed, that would mean the cookie on the context doesn't need to be mutable as the service knows about it and can set it read only on the next rigging.

cypress/integration/onboarding.spec.js
```diff
@@ -4,6 +4,8 @@ context("onboarding", () => {
     passphrase: "curled",
   };
 
+  const radUrnRegex = /hwd[a-z0-9]{56}/;
```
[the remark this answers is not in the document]
geigerzaehler: Yes, the name for this variable is wrong. I’ve renamed it to `radIdRegex` in https://github.com/radicle-dev/radicle-upstream/pull/1120/commits/1c3a738bb1a1a70a504b689abfc7390d19b7ec26. We can’t match the full RadURN with the `rad:git:` prefix because the UI only shows the ID.

cypress/integration/lock.spec.js
```diff
@@ -0,0 +1,33 @@
+// TODO We skip these tests because sealing the proxy in test mode
```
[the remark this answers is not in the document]
geigerzaehler: It’s captured in #1124

cypress/support/commands.js
```diff
@@ -1,6 +1,20 @@
 Cypress.Commands.add("resetProxyState", async () => {
   console.log("Reset Proxy state");
   await fetch("http://localhost:8080/v1/control/reset");
+  await fetch("http://localhost:8080/v1/keystore/unseal", {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    credentials: "include",
```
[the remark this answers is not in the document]
geigerzaehler: This option is necessary to include cookies in the request. See the [MDN page on `fetch`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters) for a reference.

ui/src/session.test.ts
```diff
@@ -1,3 +1,5 @@
+import * as svelteStore from "svelte/store";
```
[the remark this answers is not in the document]
geigerzaehler: I agree with your reasoning and  I’m very much in favor of removing the tests.

proxy/api/src/http/identity.rs
geigerzaehler: @MeBrei is currently working on this as an immediate follow-up.

cypress/integration/onboarding.spec.js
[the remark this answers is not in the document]
rudolfs: I think this was my doing a while ago before I knew the exact encoding for our URNs. The `rad:git:` part is already stripped out as the test is looking at the UI.

We're validating peers with this in another part of the code:
`/[1-9A-HJ-NP-Za-km-z]{54}/`

I wonder if we could adapt that here as well for the URN hash: `/[1-9A-HJ-NP-Za-km-z]{56}/`?

ui/src/localPeer.ts
```diff
@@ -69,20 +70,29 @@ export type Event =
...
-  return (): void => source.close();
+let eventSource: EventSource | null = null;
+
+session.session.subscribe(sess => {
+  if (
+    sess.status === remote.Status.Success &&
+    sess.data.status === session.Status.UnsealedSession
+  ) {
+    if (eventSource === null || eventSource.readyState === EventSource.CLOSED) {
+      eventSource = new EventSource(
```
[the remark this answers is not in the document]
geigerzaehler: This is what I tried initially to tackle the problem. It turned out that this was more involved than expected and I opted for this version. Since this is only incidentally related and not a regression I’d prefer to tackle this in a separate issue.

ui/src/settings.ts
```diff
@@ -36,5 +36,17 @@ export const themeOptions: Option[] = [
   },
 ];
 
+export const defaultSetttings = (): Settings => ({
```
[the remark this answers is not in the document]
geigerzaehler: Indeed this is not ideal. One blocker for improving this is that both the settings and the identity are returned from the same `GET /session` endpoint. This makes it impossible to have the identity guarded by unsealing and the settings always available. To fix this, I’d suggest introducing a `settings` endpoint. I’d prefer to create an issue for this, though, and tackle it as a follow-up.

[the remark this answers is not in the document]
xla: Sounds good!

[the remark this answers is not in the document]
geigerzaehler: Here’s the issue: https://github.com/radicle-dev/radicle-upstream/issues/1137

geigerzaehler:
> Wouldn't it be possible instead of passing the key to the method of the service handle, it takes the secstr which is the password and the attempted unsealing of the keystore is done in the service code,

This is exactly the approach I’d like to take for the follow-up #999. The HTTP layer should indeed only talk through the context and service handle with the actual keystore.

>  if the unsealing succeeds it creates a session secret which can be used as the `Set-Cookie` header in the response. If this is possible indeed, that would mean the cookie on the context doesn't need to be mutable as the service knows about it and can set it read only on the next rigging.

This is a good suggestion.I’ll try to see if this is possible as part of #999.

xla (approved):
🐒 👊 👬 🏓

rudolfs (approved):

PR #1121, opened 2020-10-28
chore(ui): update ts-jest to eliminate warning

geigerzaehler:
We update ts-jest (and with it typescript) to eliminate a warning from `ts-jest` that it is not compatible with the chosen typescript version.


xla (approved):
🐊 ⏺ ㊗️ 🦂

MeBrei (approved):

PR #1122, opened 2020-10-28
fix(proxy): handle case for RequestInput::Failed

### LINKED ISSUE #1116: Failed requests are not rescheduled
When a proejct is found and a clone is attempted it's not rescheduled to be queried to find more peers. It also doesn't time out.

FintanH:
Fixes #1116 

We weren't marking a clone attempt as failed and that means it would
never go back to a Requested state. This change adds handling this case
and extracts out the TimedOut handling into a helper.


xla (commented):
Dope!

💋 📒 ▫️ 🍉

rudolfs (approved):

xla (approved):
⛑ 🕕 🍿 🔝

PR #1125, opened 2020-10-28
chore(proxy): extract avatar into a lib

rudolfs:
As requested by @cloudhead, extract `avatar.rs` into a lib so it can be shared between `radicle-upstream` and `radicle-bins`.


FintanH (dismissed):
Slick :smile:

proxy/api/src/http/avatar.rs
```diff
@@ -15,9 +15,10 @@ pub fn get_filter() -> BoxedFilter<(impl Reply,)> {
 
 /// Avatar handlers for conversion between core domain and http request fullfilment.
 mod handler {
+    use radicle_avatar as avatar;
     use warp::{http::StatusCode, reply, Rejection, Reply};
```
xla, line 19:
```suggestion
    use warp::{http::StatusCode, reply, Rejection, Reply};
    
    use radicle_avatar as avatar;
```

proxy/api/src/http/identity.rs
```diff
@@ -97,10 +97,11 @@ pub struct CreateInput {
 #[cfg(test)]
 mod test {
     use pretty_assertions::assert_eq;
+    use radicle_avatar as avatar;
     use serde_json::{json, Value};
     use warp::{http::StatusCode, test::request};
```
xla, line 102:
```suggestion
    use serde_json::{json, Value};
    use warp::{http::StatusCode, test::request};
    
    use radicle_avatar as avatar;
```

proxy/api/src/identity.rs
```diff
@@ -2,7 +2,8 @@
 
 use serde::{Deserialize, Serialize};
 
-use crate::{avatar, error};
+use crate::error;
+use radicle_avatar as avatar;
```
xla, line 6:
```suggestion
use radicle_avatar as avatar;

use crate::error;
```

xla (commented):
What are we going to do about the UI part? Seems it should go into the same repo and offer an npm package which can be used in conjunction.

cloudhead:
> What are we going to do about the UI part? Seems it should go into the same repo and offer an npm package which can be used in conjunction.

I think I would have this in a `radicle-svelte-components` repo instead, because there's a lot of ways to display these avatars, and it would be nice if this was just the core logic..

cloudhead (dismissed):

FintanH (dismissed):

FintanH (approved):

cloudhead (approved):

PR #1129, opened 2020-10-29
fix(proxy): partial project lists

### LINKED ISSUE #1118: Broken project state when creating project with existing name
Reproduce with:

1. New project
2. Create a new repository
3. Choose a path which has an existing project in it
4. Use an existing project name
5. Create project

You should get "Incorrect input" and the project dashboard doesn't work.

FintanH:
Fixes #1118 

When listing projects we don't want a single project to be a show-stopper
for the rest of the projects. So we capture the failures, logging them
as warnings, and create a failure list.

### DIFF
--- a/proxy/api/src/project.rs
+++ b/proxy/api/src/project.rs
@@ -53,7 +53,26 @@ pub struct Project {
     /// Attached metadata, mostly for human pleasure.
     pub metadata: Metadata,
     /// High-level statistics about the project
-    pub stats: coco::Stats,
+    pub stats: Option<coco::Stats>,
+}
+
+/// Construct a Project from its metadata and stats
+impl<ST> From<coco::Project<ST>> for Project
+where
+    ST: Clone,
+{
+    /// Create a `Project` given a [`coco::Project`] and the [`coco::Stats`]
+    /// for the repository.
+    fn from(project: coco::Project<ST>) -> Self {
+        let id = project.urn();
+
+        Self {
+            id: id.clone(),
+            shareable_entity_identifier: format!("%{}", id),
+            metadata: project.into(),
+            stats: None,
+        }
+    }
 }
 
 /// Construct a Project from its metadata and stats
@@ -70,7 +89,7 @@ where
             id: id.clone(),
             shareable_entity_identifier: format!("%{}", id),
             metadata: project.into(),
-            stats,
+            stats: Some(stats),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


proxy/api/src/project.rs
```diff
@@ -53,7 +53,26 @@ pub struct Project {
     /// Attached metadata, mostly for human pleasure.
     pub metadata: Metadata,
     /// High-level statistics about the project
-    pub stats: coco::Stats,
+    pub stats: Option<coco::Stats>,
```
FintanH, line 56: I'm not _too_ happy with this. I might split things out into `Partial` and `Project` where a `Partial` transitions to `Project` when it can fulfill the data requirements, i.e. `coco::Stats` in this case.

proxy/api/src/project.rs
```diff
@@ -161,6 +225,18 @@ impl Projects {
 
         Ok(projects)
     }
+
+    /// Helper method for logging the error as a warning and pushing the failure onto the failure
+    /// list.
+    fn handle_failure(
+        project: Project,
+        projects: &mut Self,
```
cloudhead, line 233: Any reason why this isn't `handle_failure(&mut self, ...)` ie. a method on `Self`?

proxy/api/src/project.rs
cloudhead: I think it's ok for now, until we have other states. But the stats are not crucial to the project, so it makes sense.

cloudhead:
LGTM!

xla:
> Fixes #1118

It doesn't fix the bug stated in the issue. It addresses the general problem of breaking views when one of the projects is broken.

proxy/api/src/project.rs
FintanH: Because I'm a silly billy

FintanH:
> It doesn't fix the bug stated in the issue. It addresses the general problem of breaking views when one of the projects is broken.

Sorry, this is true! I'll create a separate issue :)

FintanH:
Currently, the `failures` just get dropped. I suppose this is fine for now and can figure out how to present them in the UI after coming to a decision with @radicle-dev/product

FintanH:
Follow up ticket: https://github.com/radicle-dev/radicle-upstream/issues/1132

cloudhead (approved):

cloudhead:
Sick!

rudolfs (approved):

PR #1149, opened 2020-10-30
feat(proxy): Generate random cookie values

### LINKED ISSUE #1090: Unsealed keystore cookie

MeBrei:
Closes #1090


geigerzaehler (dismissed):
:clap:

One thing we could consider is using the term “(auth) token” instead of “cookie” or “cookie value”. This is a more widely used terminology. See for example [“What is token-based authentication”](https://stackoverflow.com/questions/1592534/what-is-token-based-).authentication

xla (changes requested):
The cookie value is a session secret, it would be favourable to move the creation of these secrets into the session module.

MeBrei:
> The cookie value is a session secret, it would be favourable to move the creation of these secrets into the session module.

I am not opposed to extracting functions for creating and setting the token. The token is somewhat detached from the session though. It get's set directly in the `Context` (session is set in the store) and when onboarding the token get's created before the session is initialized, as there are 2 requests now. First a request to create the key store (and setting the token) and a second one to create the identity. 
SO an alternative could be to move the functions into `context.rs`, but with the current set up I'd personally leave them in `keystore.rs`. Might be that #999 moves some things around.

xla:
> SO an alternative could be to move the functions into `context.rs`, but with the current set up I'd personally leave them in `keystore.rs`. Might be that #999 moves some things around.

As long as we put things in the right place in #999 I'm fine with it. It's just not a good sign to have logic/lib code in the http layer which should be only concerned with translating domain logic into and from HTTP.

xla (approved):
🐅 ⚪️ 🗄 🐍

geigerzaehler (approved):

rudolfs (approved):

PR #1153, opened 2020-10-30
feat(proxy): Use password from user to unseal key store

### LINKED ISSUE #999: Unseal key store with the user provided passphrase

geigerzaehler:
The proxy now uses the passphrase send from the API to encrypt the key store. This closes #999.

First, this requires a change to the key store API provided by `coco`. Before, the password was required when constructing the key store. Now we need to construct the key store without the password using just the configuration (memory or file with specific paths). The password is only provided when obtaining the key so that we can get information whether the password is correct.
To achieve this, we rewrite `coco::keystore` and introduce the `KeyStore` trait that provides the necessary APIs to the proxy.

The second change moves the construction of the key store into the `api::service` module. This is necessary because we need to persist the key store (specifically the in-memory key store) between iterations of the loop in `api::process::run`. Since constructing the file key store requires the coco paths we also need to create them when creating the service environment. Fortunately this simplifies some of the rigging code.

Follow-up
- [ ] Loading indicator while unsealing in the UI. (This is required because the unseal request now takes a lot of time while the actual reload does not)
- [ ] Consolidate auth token logic from `api::http::keystore` and keystore logic implemented on `api::context::Sealed`. (#1183)


geigerzaehler:
@kim, @FintanH, @xla: Could you take a look if the approach makes sense?

proxy/api/src/context.rs
```diff
@@ -95,6 +95,33 @@ pub struct Sealed {
...
+
+impl Sealed {
+    pub async fn unseal_keystore(
+        &mut self,
+        passphrase: coco::keystore::SecUtf8,
+    ) -> Result<(), crate::error::Error> {
+        let key_store = self.key_store.clone();
+        let key = tokio::task::spawn_blocking(move || key_store.unseal(passphrase))
+            .await
+            .expect("Task to create key was aborted")?;
```
FintanH, line 109:
```suggestion
            .expect("Task to unseal key was aborted")?;
```

proxy/coco/src/keystore.rs
```diff
@@ -1,145 +1,124 @@
...
-        let mut store = MemoryStorage::new(Pwhash::new(pw, *crypto::KDF_PARAMS_TEST));
-        let key = keys::SecretKey::new();
-        store.put_key(key)?;
+pub fn memory() -> impl KeyStore + Send + Sync {
+    MemoryStore {
+        key_and_passphrase: std::sync::Mutex::new(None),
+    }
+}
+
+struct MemoryStore {
```
FintanH, line 72: Why not use the `MemoryStorage` from `radicle-keystore`? The implementation seems somewhat similar

FintanH (commented):
Looking good to me

kim:
lgtm

proxy/coco/src/keystore.rs
geigerzaehler: Using `MemoryStorage` was my initial idea but that struct does not have the right APIs. Specifically, there is no way unseal a sealed in-memory password store and get information on whether the password is valid. This is due to the fact that `MemoryStorage` requires the password at construction and it can never be changed.

FintanH: Cool, thanks for the explanation :blush:

proxy/api/src/lib.rs
```diff
@@ -21,6 +21,7 @@
     clippy::multiple_crate_versions,
     clippy::or_fun_call,
     clippy::shadow_reuse,
+    clippy::clippy::option_if_let_else,
```
geigerzaehler, line 24: I’d like to highlight this change since it might warrant some discussion. I personally find [this lint](https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else) gets in the way. It is arguable whether `Option::map_or_else` is more concise or more readable. But using `Option::map_or_else` we definitely quickly run into limitations if we use control flow statements like `break`, the `?` try operator, or `.await`.

geigerzaehler:
I’ve updated the code and the description. Please take another look :bow:

proxy/api/src/http/error.rs
```diff
@@ -163,6 +163,27 @@ pub async fn recover(err: Rejection) -> Result<impl Reply, Infallible> {
                         )
                     },
                 },
+                error::Error::Keystore(key_store_err) => {
+                    if key_store_err.is_invalid_password() {
+                        (
+                            StatusCode::FORBIDDEN,
+                            "INCORRECT_PASSWORD",
+                            "incorrect password".to_string(),
```
MeBrei, line 171: I think the glossary suggests to call it `passphrase`. As this error is resurfaced in the ui I would keep it consistent with that.

proxy/api/src/service.rs
```diff
@@ -12,6 +12,8 @@ pub struct Environment {
     pub key: Option<coco::keys::SecretKey>,
     /// If set, we use a temporary directory for on-disk persistence.
     pub temp_dir: Option<tempfile::TempDir>,
+    pub coco_paths: coco::Paths,
+    pub key_store: Arc<dyn coco::keystore::KeyStore + Send + Sync>,
```
MeBrei, line 16: Missing documentation

proxy/api/src/http/keystore.rs
```diff
@@ -67,11 +63,13 @@ mod handler {
...
-        // hardcoded one.
-        let key = coco::keys::SecretKey::new();
-        ctx.service_handle().set_secret_key(key);
+    pub async fn create(
+        mut ctx: context::Context,
+        input: super::CreateInput,
+    ) -> Result<impl Reply, Rejection> {
+        if let context::Context::Sealed(ref mut sealed) = ctx {
+            sealed.create_key(input.passphrase).await?;
+        }
```
MeBrei, line 72: Does that mean on `NoSession`, it just returns `Ok` with the cookie set, but not actually create the keystorage?
I understood that when onboarding (ie no session and no keystore) the keystore then is created with a no-session-context.

cypress/integration/lock.spec.js
```diff
@@ -14,7 +14,7 @@ context("lock screen", () => {
     cy.pick("unlock-button").should("exist");
     cy.pick("passphrase-input").type("wrong-pw");
     cy.pick("unlock-button").click();
-    cy.contains(/Could not unlock the session: Passphrase incorrect/).should(
+    cy.contains(/Could not unlock the session: incorrect password/).should(
```
xla:
```suggestion
    cy.contains(/Could not unlock the session: incorrect passphrase/).should(
```

proxy/api/src/http/keystore.rs
geigerzaehler: I’m not sure what you mean by `NoSession`. In this implementation we create a key in the keystore if the proxy is still sealed and no key is available. If the proxy is unsealed already we don’t create another key. We simply set the cookie again.

This is probably _not_ what we want though. Instead we should error with a 409 (conflict) or silently re-create the key. The first option creates some issue, though: Without adding a `DELETE /keystore` endpoint we can’t realize the flow of clearing the session and creating a second identity. I’ll fix it to use the second option.

proxy/api/src/lib.rs
xla: Legit!

proxy/api/src/service.rs
xla, line 18:
```suggestion
    pub keystore: Arc<dyn coco::keystore::KeyStore + Send + Sync>,
```

proxy/coco/src/keystore.rs
```diff
@@ -1,145 +1,161 @@
...
 use librad::{keys, paths};
 pub use radicle_keystore::pinentry::SecUtf8;
 use radicle_keystore::{
     crypto::{self, Pwhash, SecretBoxError},
-    file, memory, FileStorage, Keystore, MemoryStorage, SecretKeyExt,
+    file, Keystore, SecretKeyExt,
 };
 
+/// Storage for one secret key.
+pub trait KeyStore {
```
xla, line 16:
```suggestion
pub trait Keystore {
```

proxy/api/src/http/error.rs
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1153/commits/751ddd213d26ef187ab6d84ea8b83ef22562df21

proxy/api/src/service.rs
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1153/commits/751ddd213d26ef187ab6d84ea8b83ef22562df21

proxy/api/src/http/keystore.rs
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1153/commits/e318d00817d752db4f837c0b4a834b9d4150fef2

xla (changes requested):
Great stuff. It feels the conclusions we had in https://github.com/radicle-dev/radicle-upstream/pull/1120#issuecomment-719492530 got lost in translation.

cypress/integration/lock.spec.js
geigerzaehler: Already fixed in https://github.com/radicle-dev/radicle-upstream/commit/751ddd213d26ef187ab6d84ea8b83ef22562df21

proxy/coco/src/keystore.rs
geigerzaehler: https://github.com/radicle-dev/radicle-upstream/pull/1153/commits/3506133f4bddcb890641c6f8e789287d21e9bcac

proxy/api/src/service.rs
geigerzaehler: https://github.com/radicle-dev/radicle-upstream/pull/1153/commits/3506133f4bddcb890641c6f8e789287d21e9bcac

geigerzaehler:
> It feels the conclusions we had in [#1120 (comment)](https://github.com/radicle-dev/radicle-upstream/pull/1120#issuecomment-719492530) got lost in translation.

I’ve decided to put this into a separate PR (#1183) because it there might be some different design discussions there.

MeBrei:
Regarding the onboarding test failures:

- `is possible to create a second identity`: We might have to `skip` this one for now, as we currently only deal with a single keystore
- `is not possible to create the same identity again`: this fails earlier than the test assumes. It checks for an error on identity creation, but it actually fails on trying to create the keystore. The test uses the `Clear Session` functionality in settings. For beta 1 we might have to add removing the keystore to actions on `Clear Session`

proxy/api/src/http/error.rs
```diff
@@ -163,6 +163,27 @@ pub async fn recover(err: Rejection) -> Result<impl Reply, Infallible> {
                         )
                     },
                 },
+                error::Error::Keystore(keystore_err) => {
+                    if keystore_err.is_invalid_password() {
```
FintanH, line 167: Should we keep with the glossary on function names as well?
```suggestion
                    if keystore_err.is_invalid_passphrase() {
```

proxy/coco/src/keystore.rs
```diff
@@ -1,145 +1,161 @@
...
-
-        assert!(
-            key.as_ref() == store.get().as_ref(),
-            "the stored key was not equal to the one retrieved"
-        )
+impl Error {
+    /// Returns `true` if the error indicates that an invalid password was used to decrypt the
+    /// secret key.
+    #[must_use]
+    pub const fn is_invalid_password(&self) -> bool {
```
FintanH, line 150:
```suggestion
    pub const fn is_invalid_passphrase(&self) -> bool {
```

FintanH (commented):
Maybe do a quick grep for password and use passphrase instead, just for consistency.

geigerzaehler:
Following the outcome of the discussion around #1188 I’ve removed the tests that try to create a second identity.

MeBrei (approved):
🔑 ⭐

xla (approved):
💙 🍝 🦄 ↔️

PR #1156, opened 2020-10-31
fix(proxy): check waiting room has urn

### LINKED ISSUE #1154: Waiting room tries to add on every gossip
We saw the waiting room adding RadUrn's thinking that it "Found" PeerIds, but it wasn't looking for those in the first place. We should add an API method to check that the Waiting Room `has` a `RadUrn`, and if it does then mark a `PeerId` as `Found`.

FintanH:
Fixes #1154 

During gossip announcements of Has, we would always mark the RadUrn as
Found in the WaitingRoom, but the WaitingRoom might not necessarily have
been looking for this RadUrn. So we first check that the WaitingRoom
indeed `has` the RadUrn before continuing on.

### DIFF
--- a/proxy/coco/src/peer/run_state.rs
+++ b/proxy/coco/src/peer/run_state.rs
@@ -564,6 +564,11 @@ impl RunState {
                     val: Gossip { urn, .. },
                 })),
             ) => {
+                // This message is uninteresting to the waiting room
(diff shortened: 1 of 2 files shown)


rudolfs (approved):

xla (approved):
😽 😉 👚 🚡

PR #1158, opened 2020-11-02
refactor(tests): Enforce success responses from test helpers

geigerzaehler:
We now enforce that the requests made in test helpers return with a success status code. Otherwise we throw an error and fail the tests.

This requires us to fix some of the routing tests.


rudolfs (approved):

xla (approved):
🚉 🌮 🍍 👰

Issue #947, opened 2020-09-22
eslint-plugin-no-only-tests doesn't catch skipped context


rudolfs:
`context.only()` doesn't produce an error on CI. It should.

We're using: https://github.com/levibuzolic/eslint-plugin-no-only-tests.

Refs: https://github.com/radicle-dev/radicle-upstream/pull/938.

PR #1160, opened 2020-11-02
chore(ci): Warn about focused tests

geigerzaehler:
We fail linting when a test is focused with `.only`.


rudolfs:
Fixes: #947.

rudolfs (changes requested):
We had a working husky/lint-staged setup which would prevent us even from committing skipped tests. That seems to be broken now (even on master). `eslint-plugin-no-only-tests` was supposed to take care of that, albeit it used to work only on `it.only` and not `context.only`.

Could we get lint-staged working again so we catch the focussed tests even before they reach CI?

geigerzaehler:
> `eslint-plugin-no-only-tests` was supposed to take care of that

I wasn’t aware of that plugin. I’m using it instead now. The only thing that was missing was to add it to `.eslintrc.js`.

rudolfs (approved):
Neat! 👍

NunoAlexandre (approved):

PR #1162, opened 2020-11-02
fix(ci): prevent concurrent builds on master branch

geigerzaehler:
I’ve properly enabled shared master caches in our CI builds. (Although this option was enabled in `.buildkite/pipeline.yaml` it was not enabled in the buildkite project config). To prevent issues with concurrent master builds we need to restrict concurrency. See
https://github.com/radicle-dev/infra/tree/master/ci#shared-master-cache for details.


xla (approved):
⬜️ 👌 🐿 🚁

rudolfs (approved):

PR #1164, opened 2020-11-02
style(ui): Improve loading screens + clean up Illustration component

juliendonck:
- Simplified `<Illustration />` and `<EmptyState />`
- Improved loading screen by using `<EmptyState />`
Screenshot 2020-11-02 at 14 42 23


ui/src/session.ts
```diff
@@ -105,7 +105,6 @@ export const unseal = async (passphrase: string): Promise<void> => {
     notification.error(`Could not unlock the session: ${error.message}`);
     return;
   }
-  notification.info("Unsealing the session...");
```
juliendonck, line 108: @MeBrei is there any reason why the loading screen wouldn't show and we'd need the notification as a "back up"?

rudolfs (approved):

ui/src/session.ts
MeBrei: I can't think of one. We can always add it back in again, when we notice it...

MeBrei (approved):

PR #1169, opened 2020-11-02
chore(proxy): update librad deps

FintanH:
With these changes we get some fixes, specifically urn_context.
We also get the changes to have a sub-workspace that is re-exported
as git_ext, and also the very useful macro `reflike!`


geigerzaehler (approved):

xla (approved):
😎 🕒 🚴 📎

PR #1170, opened 2020-11-02
fix(proxy): persist temp dir on unseal in test mode

### LINKED ISSUE #1124: Test unsealing by fixing full state reset
The tests (introduced in #1120) for unsealing the proxy _after_ an identity has been created don’t work. The reason for this is that unsealing in test mode will reset all persistent state. This needs to be fixed to re-enable the tests.

In detail, we should not recreate the temporary directory for the KV store in `api::process::rig` when we unseal the keystore. Instead the temporary directory should be retained between service reloads.

geigerzaehler:
In test mode unsealing or sealing the keystore does not recreate the temporary directory and reset all state anymore. This allows us to enable all tests that involve sealing.

We also fix the enabled tests and adjust the UI to make them work.

Fixes #1124.

Follow-up

- [ ] Remove `TEST_KEY` and `lazy_static` when tackling #999


proxy/api/src/service.rs
```diff
@@ -4,13 +4,43 @@ use futures::prelude::*;
...
+        #[from]
+        std::io::Error,
+    ),
+}
+
+impl Config {
+    /// Create a new default configuration.c:w
+    ///
+    /// If `test_mode` is `true` then `Config::temp_dir` is set for temporary on-disk persistence.
+    fn new(test_mode: bool) -> Result<Self, Error> {
```
FintanH, line 33: Might I suggest passing the `Option<TempDir>` as the argument here and forcing the layer above to handle determine test mode and creation of the directory?

With that change, I think resetting it below would consist of:
```rust
self.config = Config::new(self.config.temp_dir.map(|_| tempfile::tempdir()).transpose()?)
```

geigerzaehler: You make a good point about moving the decision whether to construct the temp dir from `Config` to one level above. One problem that I see with your approach is that the construction of `self.config` is not paralleled between `Manager::new` and the reset message and the later is harder to understand. One thing that I was considering was moving he `Config::new` logic to `Manager::new_config`. Of course, this would need to be a static method because we need to have `config` before we can construct `Manager`.

Another thing to consider is to rename `Config` to something else (maybe `Environment`) which would make it less strange that it cannot be constructed without side-effects and failures.

Any thoughts @FintanH?

FintanH: > One problem that I see with your approach is that the construction of self.config is not paralleled between Manager::new and the reset message and the later is harder to understand

Can you explain a bit more what you mean here? 

> Another thing to consider is to rename Config to something else (maybe Environment) which would make it less strange that it cannot be constructed without side-effects and failures.

Ya this seems like a nicer move imo. It lends itself to the concept of controlling things, where the test mode flag controls how we do the set up :+1:

cypress/integration/onboarding.spec.js
```diff
@@ -123,7 +120,9 @@ context("onboarding", () => {
 
       // Success screen.
       cy.pick("urn").contains(radIdRegex).should("exist");
+      cy.pick("go-to-profile-button").click();
 
+      cy.log("reset session");
```
rudolfs, line 125: This is cool 👍

I wonder if we should be using `.log` everywhere instead of adding comments in the specs about what's going on? 🤔

ui/Screen/Lock.svelte
```diff
@@ -5,11 +5,14 @@
...
-    session.unseal(passphrase);
+  const unlock = async () => {
+    const unlocked = await session.unseal(passphrase);
+    if (!unlocked) {
+      passphrase = "";
+    }
   };
 
-  const onEnter = () => {
+  const onEnter = async () => {
```
rudolfs, line 15: Does this really need to be `async` if you don't `await` in the body?

ui/Screen/Onboarding.svelte
```diff
@@ -41,10 +41,11 @@
       await session.createKeystore();
       // Retry until the API is up
       notification.info("Creating the identity...");
-      withRetry(() => createIdentity({ handle, passphrase }), 200).then(id => {
-        identity = id;
-        state = State.SuccessView;
-      });
+      identity = await withRetry(
```
rudolfs, line 44: Phew! 😌  Those `Promise#then()` were making me nervous.

rudolfs (dismissed):

xla (approved):
🗻 🏡 ⛸ 📿

proxy/api/src/service.rs
geigerzaehler: > > One problem that I see with your approach is that the construction of self.config is not paralleled between Manager::new and the reset message and the later is harder to understand
> 
> Can you explain a bit more what you mean here?

The `Config` value we construct in `Manager::new()` should be the same value as the one constructed when handling `Message::Reset` in `Manager::config()`. This is what resetting the service config means. In both cases we want a pristine value determined by some external parameters (in our case only `test_mode`). To make this clear in the code and more robust against changes to how a pristine `Config` value is constructed it’s beneficial to construct it with a simple functional call that can be placed in both locations. The downside of your proposal is that constructing `Config` is very dissimilar in the two cases but yields the same result.

> > Another thing to consider is to rename Config to something else (maybe Environment) which would make it less strange that it cannot be constructed without side-effects and failures.
> 
> Ya this seems like a nicer move imo. It lends itself to the concept of controlling things, where the test mode flag controls how we do the set up +1

Do you have a suggestion for a term? I find `Environment` to be ok, but it’s not hitting it on the nail.

cypress/integration/onboarding.spec.js
geigerzaehler: I think we can start by replacing `console.log` in the support commands. This would be very beneficial.

ui/Screen/Lock.svelte
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1170/commits/b0470946bc21f7a0d76e00d02cab481f60ef86b1

proxy/api/src/service.rs
FintanH: How about `Env`? :smile:

rudolfs (approved):

proxy/api/src/service.rs
geigerzaehler: @FintanH: I’ve renamed `service::Config` to `Environment`. Could you have another look?

FintanH: @geigerzaehler would it make more sense to have the `test_mode` a field of the `Environment` rather than the `Manager`?

geigerzaehler: That’s a good idea actually. It would allow us to drop the `args` argument from `run_rig`.

FintanH (approved):
Great stuff :metal:

rudolfs (approved):

PR #1176, opened 2020-11-03
fix(ui): prevent FollowToggle flickering

rudolfs:
When unfollowing a remote or when searching for a project and clicking `Follow`, the button would toggle to the opposite state for a brief time while the screen removes the component. This caused a visual flicker.

Before:
slomo-flicker

After:
slomo-noflicker

### DIFF
--- a/ui/DesignSystem/Component/FollowToggle.svelte
+++ b/ui/DesignSystem/Component/FollowToggle.svelte
@@ -8,6 +8,13 @@
   export let disabled: boolean = false;
   export let following: boolean = false;
 
+  // Set this to true if you don't want the button to toggle.
+  //
+  // Useful for when the button triggers an action that removes the component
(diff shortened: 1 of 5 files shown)


xla (changes requested):
I'd argue instead of making the toggle more complex and we don't want toggle behaviour  we should migrate to a proper button. For the search that is definitely on the horizon. We do a have action buttons we can use for these purposes.

rudolfs:
> I'd argue instead of making the toggle more complex and we don't want toggle behaviour we should migrate to a proper button. For the search that is definitely on the horizon. We do a have action buttons we can use for these purposes.

@xla I did it this way to avoid making a new button component for the "Following" -> "Unfollow" states as our action buttons don't support text changes on hover.

hover

Should I then create a new component for this type of button and undo the changes to `FollowToggle`?

Issue #1132, opened 2020-10-29
Partial projects view


FintanH:
In https://github.com/radicle-dev/radicle-upstream/pull/1129 we introduce projects that may fail. It does not, however, surface these failures to the UI.

We should surface them to the user so that they can know that something went wrong and possibly try and recover from that error, if possible.

It would be good to get @radicle-dev/product's view on how this should be surfaced.

FintanH:
Part of making our lives better, `Project` will become an `enum` where it's `Tracked`, `Contributed`, or `Failure` /cc @xla

PR #1180, opened 2020-11-03
chore(ui): show notifications for projects that can't be displayed

### LINKED ISSUE #1186: modifier key + click on links still open external links within the electron renderer
Happens if you <kbd>⌘</kbd>+click on a link in Upstream.
We should disable this.

Related to: https://github.com/radicle-dev/radicle-upstream/pull/634.

Screenshot 2020-11-04 at 13 33 29

rudolfs:
Shows an error notification for each broken project.

As per [design](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=6142%3A592).

Closes: https://github.com/radicle-dev/radicle-upstream/issues/1132.
Fixes #1186

Screenshot 2020-11-03 at 16 42 27


juliendonck:
How can I test this?

rudolfs:
> How can I test this?

@juliendonck you need to break a project. Easiest way would be to mangle your monorepo.

1. Create a new project (with new repository) via the app and note the project's URN *without* the `rad:git` part.
2. Then delete that project's `main` branch in the monorepo (replace `URN_GOES_HERE` with the actual urn!):
```
rm ~/Library/Application\ Support/xyz.radicle.radicle/git/refs/namespaces/URN_GOES_HERE/refs/heads/main
```
3. Go to the profile page, you should see an error notification about the mangled project.

ui/src/profile.ts
```diff
@@ -75,3 +76,14 @@ export const fetchFollowing = (): void => {
     return reqs.filter(req => req.type !== waitingRoom.Status.Cloned);
   });
 };
+
+export const showNotificationsForFailedProjects = async (): Promise<void> => {
+  const failedProjects = await project.fetchFailed();
+  failedProjects.forEach(project => {
+    notification.error(
+      `The project ${project.metadata.name} could not be loaded`,
```
FintanH, line 84: Would it make sense to have a way of displaying the type of failure? I'm thinking the more information we give the user about the failure, then the better the reports we get when they do fail.

FintanH (approved):
Looks good to me. I have one question about supplying the user with information about the failure. 

Also, could you create a follow-up ticket for the `enum` refactor that we discussed?

rudolfs:
> Also, could you create a follow-up ticket for the enum refactor that we discussed?

@FintanH here you go: https://github.com/radicle-dev/radicle-upstream/issues/1182.

juliendonck (dismissed):
LGTM 🐛

rudolfs:
Closes: https://github.com/radicle-dev/radicle-upstream/issues/1186.

rudolfs:
After discussing this with @FintanH, @juliendonck and @geigerzaehler, we decided to show concise error messages with additional context that can be copied to the user's clipboard and shared with us.

Screenshot 2020-11-04 at 17 18 21

The clipboard contents for this error are:
```
{
  "message": "The project newwww could not be loaded",
  "context": {
    "RepoType": {
      "New": "new",
      "Existing": "existing"
    },
    "Role": {
      "Contributor": "contributor",
      "Maintainer": "maintainer",
      "Tracker": "tracker"
    },
    "ReplicationStatusType": {
      "NotReplicated": "notReplicated",
      "Replicated": "replicated"
    },
    "PeerType": {
      "Local": "local",
      "Remote": "remote"
    },
    "creation": {},
    "peerSelection": {},
    "pendingPeers": {},
    "project": {},
    "projects": {},
    "tracked": {},
    "localState": {},
    "localStateError": {},
    "defaultBranch": {},
    "peerValidation": {}
  }
}
```

All other existing UI error handling code will be updated in: https://github.com/radicle-dev/radicle-upstream/issues/1187.

juliendonck (approved):
Worked for me

xla:
> `"type": "defaultBranch",`

Is this all we get from the backend for the error?

rudolfs:
> Is this all we get from the backend for the error?

@xla I'm still working on getting also an appropriate message from the backend. Was dealing with the `ipc.js` import errors so far...

rudolfs:
@xla, @FintanH: I had another look at the API code together with @geigerzaehler and we came to the conclusion that adding another string along with the error type which we already have (`"type": "defaultBranch"` | `"type": "stats"` | `"type": "signedRefs"`) wouldn't give us much. If a user runs into this problem and sends us this json blob, we already have enough information to debug this or instruct the user.

If we *really* want those nice messages, I propose we add the `error type` <-> `message` lookup in the front-end.

xla:
> @xla, @FintanH: I had another look at the API code together with @geigerzaehler and we came to the conclusion that adding another string along with the error type which we already have (`"type": "defaultBranch"` | `"type": "stats"` | `"type": "signedRefs"`) wouldn't give us much. If a user runs into this problem and sends us this json blob, we already have enough information to debug this or instruct the user.

Right because the failure is expressed in these three variants and doesn't carry the usual information like normal errors have. The shape of the `Failure` enum is awkward, given that we interleave the failure type and the project data. It would be clearer if it looked like:

``` rust
#[derive(Serialize)]
#[serde(rename_all = "camelCase", tag = "type")]
pub enum Failure {
    /// We couldn't get a default branch for the project.
    DefaultBranch { partial: Partial },
    /// We couldn't get the stats for the project.
    Stats { partial: Partial },
    /// We couldn't get the signed refs of the project, and so we can't determine if it's tracked
    /// or contributed.
    SignedRefs { full: Full },
}
```

geigerzaehler (approved):

FintanH:
@xla I'll follow up with the better representation of the `Project` and `Failure` modes :+1: This looks good to me otherwise

PR #1183, opened 2020-11-04
refactor(proxy): move auth token logic into context

geigerzaehler:
We remove the logic that validates and generates auth tokens from the HTTP layer and consolidate it in `api::context`.


rudolfs (approved):

xla (approved):
🈂️ 🔀 🚃 ⛎

PR #1185, opened 2020-11-04
fix(ci): don’t stall ci builds when tests fail

### LINKED ISSUE #1184: Failed builds stall
In #1167 we started seeing that builds stall because the proxy was not shutdown properly. We tried to fix this with [`cypress:run && kill pidof api`](https://github.com/radicle-dev/radicle-upstream/pull/1167/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L118). However this does not work if the tests fail and `cypress:run` exits with a non-zero status code. This again results in failing builds stalling.

geigerzaehler:
Fixes #1184

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "test:integration:debug": "TZ='UTC' run-p --race svelte:watch proxy:start:test-watch wait:debug",
     "test:unit": "jest",
     "test:unit:watch": "jest --watchAll",
-    "wait:test": "wait-on tcp:8080 && yarn svelte:build && yarn false; status=$?; [[ $CI == true ]] && kill `pidof api`; exit $status",
+    "wait:test": "wait-on tcp:8080 && yarn svelte:build && yarn false; status=$?; [ $CI == true ] && kill `pidof api`; exit $status",
     "wait:debug": "wait-on tcp:8080 && yarn cypress:open",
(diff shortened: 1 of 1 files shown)


geigerzaehler:
The first commit fails on purpose to show that it works.

FintanH (approved):
Nice :metal:

xla (approved):
🏰 🐓 ⏰ 📈

package.json
```diff
@@ -115,7 +115,7 @@
     "test:integration:debug": "TZ='UTC' run-p --race svelte:watch proxy:start:test-watch wait:debug",
     "test:unit": "jest",
     "test:unit:watch": "jest --watchAll",
-    "wait:test": "wait-on tcp:8080 && yarn svelte:build && yarn cypress:run && kill `pidof api`",
+    "wait:test": "wait-on tcp:8080 && yarn svelte:build && yarn cypress:run; status=$?; kill `pidof api`; exit $status",
```
rudolfs, line 118: Is there a way to run this workaround only on the Linux CI machine? I had to install `pidof` via brew on mac-n-cheese and my local machine for this to work. This might break running tests for other mac users like @MeBrei.

geigerzaehler: Let’s see if this works: https://github.com/radicle-dev/radicle-upstream/pull/1185/commits/d9d2f0654e2a869cfe958d199ffbdfdafd5b3fb6

geigerzaehler:
I got it to work. Please take another look @FintanH, @xla, @rudolfs

FintanH (approved):
LGTM

rudolfs (approved):

PR #1193, opened 2020-11-05
refactor(ui): convert ipc to TypeScript

geigerzaehler:
We convert `native/ipc.js` to TypeScript. This enables us to use it from
TypeScript files (not only Svelte files) in `./ui`. The reason why we
need to convert the file to TypeScript is that rollup and typescript
don’t seem able to handle JS files and spit out a warning. In general it
also seems to be a good idea to move everything to TypeScript.

One result of this change is that we can remove everything Babel
related.


native/ipc.ts
```diff
@@ -10,26 +13,27 @@ export const OPEN_PATH = "IPC_OPEN_PATH";
 //
 // The workaround is to use the electron native open dialog. As a bonus we
 // can configure it to allow users to create new directories.
-export const getDirectoryPath = () =>
+export const getDirectoryPath = (): void =>
```
rudolfs, line 16:
```suggestion
export const getDirectoryPath = (): string =>
```

It returns the file path that is chosen within the file dialog.

geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1193/commits/79e2b4b5d2cab8a3bc660aab9f9e7ee8a4522c5f

xla (dismissed):
🕹 🐢 🍀 🚊

rudolfs (dismissed):
This was [blocking my issue](https://buildkite.com/monadic/radicle-upstream/builds/6196#5be0e621-8e7d-40a5-ade1-e6d6fb02c9ac/294-320).

Thanks for fixing it! 🙏

NunoAlexandre (dismissed):
muy bueno!

rudolfs (approved):

NunoAlexandre (commented):
:ping_pong:

NunoAlexandre (approved):

PR #1194, opened 2020-11-05
feat(ui): lock screen on key store unseal

MeBrei:
Follow up of #999 

Locks the screen and disables the input and buttons while the key store is created/unsealed in the onboarding and unlock screen. Also removes the special loading screen for unsealing the session, as the more time consuming request that is unsealing the key store happens while the unlock screen is still shown.


ui/App.svelte
```diff
@@ -109,10 +109,6 @@
 
   <!-- TODO(julien): Dress up loading screen -->
```
juliendonck, line 110:
```suggestion

```

juliendonck (approved):
Nice!

ui/Screen/Lock.svelte
```diff
@@ -1,16 +1,23 @@
 <script lang="ts">
   import * as session from "../src/session";
+  import * as screen from "../src/screen";
```
xla, line 3:
```suggestion
  import * as screen from "../src/screen";
  import * as session from "../src/session";
```

ui/Screen/Onboarding.svelte
```diff
@@ -38,9 +40,11 @@
 
   const onCreateIdentity = async (handle, passphrase) => {
     try {
+      screen.lock();
+      requesting = true;
+      notification.info("Creating the identity...");
```
rudolfs, line 45: Can we remove the notification?
It feels to me like it's just adding noise because the identity creation in production builds is fairly fast.

prod-build

ui/Screen/Onboarding/EnterPassphrase.svelte
```diff
@@ -150,21 +153,23 @@
...
+        {disabled} />
     </div>
 
     <div class="buttons">
       <Button
         dataCy="back-button"
         variant="transparent"
         style="margin-right: 1rem;"
-        on:click={() => dispatch('previous')}>
+        on:click={() => dispatch('previous')}
```
xla, line 165:
```suggestion
        on:click={() => dispatch("previous")}
```

ui/Screen/Onboarding/EnterPassphrase.svelte
```diff
@@ -150,21 +153,23 @@
...
+        {disabled}>
         Back
       </Button>
 
       <Button
         dataCy="set-passphrase-button"
-        disabled={!allowNext}
+        disabled={!allowNext || disabled}
```
xla, line 172: This should go into the allowNext chain, so it's in one place.

xla (changes requested):
Very nice, felt awkward without it.

ui/Screen/Onboarding.svelte
geigerzaehler, line 45: Should we maybe get rid of this? For the release build this is rather quick and the notification may persist after the identity has been created.

ui/Screen/Lock.svelte
```diff
@@ -1,16 +1,23 @@
...
 
   import { Button, Emoji, Input } from "../DesignSystem/Primitive";
 
   let passphrase = "";
+  let requesting = false;
```
geigerzaehler, line 8: Suggestion: `unlockInProgress` sounds a bit more descriptive.

ui/App.svelte
```diff
@@ -109,10 +109,6 @@
...
   <div slot="loading" class="error">
-    {#if $location === path.lock()}
-      <EmptyState headerText="Unlocking the app..." emoji="🚪" text="" />
-    {:else}
-      <EmptyState headerText="Loading..." emoji="🕵️" text="" />
-    {/if}
+    <EmptyState headerText="Loading..." emoji="🕵️" text="" />
```
rudolfs, line 112: I wonder if there's a way to get rid of this intermediate loading screen completely and only show the unlock screen while unlocking the app?

loading

rudolfs (approved):
Nice work!

Added some non-blocking comments.

ui/Screen/Onboarding.svelte
MeBrei: The notification is c key store is created before the key store creation request, which takes a long time.

rudolfs: Jinx!

ui/App.svelte
MeBrei: The intermediate screen comes from the session store being in state loading. Before, the remote slot for that state was just empty ie a blank screen. So should the session request take longer, there is some description of what is happening. @juliendonck do you think the screen is unnecessary now that the requests should be quick?

rudolfs: Hm, I guess there is no way for the app to know what the intermediate state is before it's unlocked. I guess leaving the loading state then is probably the better option, in case the unlocking takes longer than normal, the user at least will know what's going on.

rudolfs (approved):

juliendonck (approved):

xla (approved):
🈁 🚜 🚛 🗻

PR #1204, opened 2020-11-06
chore(ui): enable "noEmit" in tsconfig.json

geigerzaehler:
This prevents us from messing up the source directory with build artifacts that are included instead of the proper source files when we run `tsc` from the command line.


xla (approved):
🍢 🚿 🐜 🖼

rudolfs (approved):
😚👌

PR #1205, opened 2020-11-06
fix(proxy): add include to project creation

### LINKED ISSUE #1201: Add/remove remotes in include on track/untrack
When we track and untrack a remote we should update the include file for that project so that the working copy has the remotes available for fetching.

As well as this, we don't create an include file for new projects. This was an oversight when we first introduced includes.

FintanH:
Fixes #1201 

~**WIP**: I have a suspicion this won't quite work. Need to test it out with the seed and a peer.~
Tested and twerks.

We add include setup during project creation to ensure that we have one
for when we need to update it.
Once we track and untrack peers we should update the include file.


FintanH:
Looks like the tests might be flakey: https://buildkite.com/monadic/radicle-upstream/builds/6261#080d374b-b207-4ed1-9c7a-1a1889cd068a

xla (dismissed):
Extra dope!

🛅 🕦 🏎 📀

xla (approved):
🈺 🔭 😎 ☺️

rudolfs (approved):

PR #1207, opened 2020-11-09
chore(build): cleanup all rollup artifacts before build

geigerzaehler:
To avoid a situation where we are using a stale rollup build artifact we’re also removing `native/main.comp.js` before building.


rudolfs (approved):

xla (approved):
⛰ 👈 😷 🖇

NunoAlexandre (approved):

PR #1211, opened 2020-11-09
refactor(ui): convert cypress tests to TypeScript

geigerzaehler:
We convert all Cypress test files to TypeScript.

To make this more manageable we export the commands as functions instead of adding them via `Cypress.Commands.add`. Otherwise we would need to extend the type of `cy` manually to make `cy.pick()`, say, type check.

We also switch to using `cy.request` instead of `fetch`. In this way we don’t need to convert promises to cypress "chainables". This also has the additional benefit that the requests are visible in the UI.


xla (approved):
🔬 🗡 🎋 💝

rudolfs (approved):

NunoAlexandre (approved):
:star2:

MeBrei (approved):

Issue #1189, opened 2020-11-04
Use a random free port to bind proxy to


geigerzaehler:
Currently, the proxy always binds to port 8080. This is a frequently used port. If the proxy fails to bind, the app becomes unusable.

To address this the app should find a random free port and bind the proxy to that port. An intermediate solution would be to bind to a fixed port of high order without implementing the check for a free port and passing this information to the frontend.

FintanH:
Can we set the port to `0` and let it be assigned by the system?

xla:
> Can we set the port to `0` and let it be assigned by the system?

This is not possible as we need to know what the port is from the ui. Instead the native should do a best effort random port check and pass that down when starting the proxy, if it happens to be occupied in between, we are out of luck.

geigerzaehler:
We are using a fixed random port now (#1212). The rest is not crucial for β1 so I’m removing it from the milestone.

Kaihuang724:
Closing this issue unless there's a clear need for it

PR #1212, opened 2020-11-09
feature(proxy): use port with less bind conflicts

geigerzaehler:
Instead of port 8080 we use port 17246 to bind the proxy to. This decreases the likelihood of another services binding to the port.


xla (changes requested):
Feels like this just moves the problem around. How hard would it be for the application container pass down a random port to the proxy?

geigerzaehler:
> How hard would it be for the application container pass down a random port to the proxy?

It’s a bit more effort since we need to use IPC to communicate the port the the app. We can tackle this later and it should not be a blocker for this PR. It’s all written down in #1189.

FintanH (dismissed):
Happy with this stop-gap since there's an issue to track the actual fix. I imagine this change will help stop people from running into the application not starting for the time being.

rudolfs (approved):
<3 the low-cost approach!

Screenshot 2020-11-09 at 17 36 55

rudolfs (approved):

cloudhead (approved):
I would normally bikeshed on the port number, but not this time :+1:

PR #1215, opened 2020-11-10
chore(ui): update svelte toolchain

geigerzaehler:
We update all svelte related packages. This requires us to fix some typechecking errors.


ui/src/session.ts
```diff
@@ -191,7 +191,7 @@ export const addSeed = async (seed: string): Promise<boolean> => {
   // running async remote validations in in the background. If we remove the
   // async then the seed input form will have to be submitted twice to take any
   // effect.
-  await seedValidation.validate(seed);
+  seedValidation.validate(seed);
```
xla, line 194: Is this safe, given the comment above?

rollup.app.js
```diff
@@ -23,7 +23,7 @@ export default {
       // we'll extract any component CSS out into
       // a separate file — better for performance
       css: css => {
-        css.write("public/bundle.css");
+        css.write("bundle.css");
```
xla, line 26: Why do we write to a different location?

ui/src/session.ts
geigerzaehler: Removing this wasn’t fully intentional. I’ve reverted it in https://github.com/radicle-dev/radicle-upstream/pull/1215/commits/2c8bc6d0b24a6a829e85863b12ddc7f7f908dc94

rollup.app.js
geigerzaehler: This was a [major breaking change](https://github.com/sveltejs/rollup-plugin-svelte/blob/master/CHANGELOG.md#600) in `svelte-rollup`. Now the path is relative to the output of the bundle.

xla (approved):
💶 ™️ 🐿 📇

rudolfs (dismissed):

xla (approved):
♎️ 🙎 🏎 🌓

MeBrei (approved):

rudolfs (approved):

Issue #1187, opened 2020-11-04
Give context to all Error notifications


juliendonck:
We're expanding the error notification to allow the users to copy the console logs to their clipboard so that they have more information they can share with us when they report a bug. @rudolfs started implementing this for partial projects errors in #1180. 

We have to go through the app and see what context we can give to other `Notification.Error` we have currently.

geigerzaehler:
This involves a couple of things
* Whenever we catch a JS exception in the UI we have to ensure that any additional information that is part of the error is exposed.
* We need to install a handler for [unhandled rejections](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onunhandledrejection) so we also show errors when we don’t explicitly catch them
* We should also log errors to the console with `console.error` so that devs and dev users can dig more

PR #1221, opened 2020-11-11
feat(ui): introduce structured error logging

geigerzaehler:
This change introduce structured error logging and applies it in a couple of places. Part of #1187.

The design of the `Error` interface is very similar to Rust errors and arises from the following needs:
* A stable identifier for errors (`code`)
* A message we can display to the user (`message`)
* Structured details that provide more information about the error (`details`)
* A stack trace if available (`stack`)
* Layering of error chains (`source`)

~~I’ve decided not to make `code` an enum. Having a centralized place that lists all errors does not help us with debugging. When debugging we need to look at the place where the error originated which can easily be found be grepping for the error code. In addition such a centralized place would make it more cumbersome to add new errors since the list of error would always need to be updated.~~

Errors are shown to the user as notifications (`error.show`). The error message is displayed to the user and the full error can be copied to the clipboard. In addition we also log the error to the console which allows us to debug it during development.

As a follow-up to this PR we will apply the error logging and showing in more places.

### DIFF
--- a/ui/App.svelte
+++ b/ui/App.svelte
@@ -3,9 +3,9 @@
 
   import * as hotkeys from "./src/hotkeys.ts";
   import "./src/localPeer.ts";
-  import * as notification from "./src/notification.ts";
   import * as path from "./src/path.ts";
   import * as remote from "./src/remote.ts";
+  import * as error from "./src/error.ts";
   import { fetch, session as store, Status } from "./src/session.ts";
 
   import {
@@ -78,8 +78,7 @@
       break;
 
     case remote.Status.Error:
-      console.error($store.error);
-      notification.error("Session could not be fetched");
+      error.show($store.error);
       break;
   }
 </script>
@@ -101,7 +100,7 @@
 <NotificationFaucet />
 <Theme />
 
-<Remote {store} context="session">
+<Remote {store} context="session" disableErrorLogging={true}>
   <Router {routes} />
 
   <div slot="loading" class="error">
--- a/ui/DesignSystem/Component/Remote.svelte
+++ b/ui/DesignSystem/Component/Remote.svelte
@@ -1,7 +1,6 @@
 <script lang="typescript">
   import type { Readable } from "svelte/store";
 
-  import * as error from "../../src/error";
   import * as notification from "../../src/notification";
   import * as remote from "../../src/remote";
 
@@ -11,27 +10,23 @@
   export let store: Readable<remote.Data<any>>;
   export let context: string | undefined = undefined;
 
-  // Shorthand for casting these states
-  type ErrorState = { status: remote.Status.Error; error: error.Error };
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   type SuccessState = { status: remote.Status.Success; data: any };
 
   // If no error slot was provided, svelte will instantiate the fallback div
   let noErrorSlotProvided: HTMLDivElement;
 
-  $: if ($store.status === remote.Status.Error && !!noErrorSlotProvided) {
-    console.error("Remote error", ($store as ErrorState).error);
-    notification.error(($store as ErrorState).error.message);
+  let storeValue: remote.Data<any>;
+  $: storeValue = $store;
+
+  $: if (storeValue.status === remote.Status.Error && !!noErrorSlotProvided) {
+    console.error("Remote error", storeValue.error);
+    notification.error(storeValue.error.message);
   }
 
   $: data =
     $store.status === remote.Status.Success
       ? ($store as SuccessState).data
-      : undefined;
-
-  $: remoteError =
-    $store.status === remote.Status.Error
-      ? ($store as ErrorState).error
       : undefined;
 </script>
 
@@ -49,8 +44,8 @@
   {:else}
     <slot {data} />
   {/if}
-{:else if $store.status === remote.Status.Error}
-  <slot name="error" error={remoteError}>
+{:else if storeValue.status === remote.Status.Error}
+  <slot name="error" error={storeValue.error}>
     <div bind:this={noErrorSlotProvided} />
   </slot>
 {/if}
--- a/ui/DesignSystem/Component/Remote.svelte
+++ b/ui/DesignSystem/Component/Remote.svelte
@@ -10,16 +10,23 @@
   export let store: Readable<remote.Data<any>>;
   export let context: string | undefined = undefined;
 
+  export let disableErrorLogging: boolean = false;
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   type SuccessState = { status: remote.Status.Success; data: any };
 
   // If no error slot was provided, svelte will instantiate the fallback div
   let noErrorSlotProvided: HTMLDivElement;
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let storeValue: remote.Data<any>;
   $: storeValue = $store;
 
-  $: if (storeValue.status === remote.Status.Error && !!noErrorSlotProvided) {
+  $: if (
+    storeValue.status === remote.Status.Error &&
+    !!noErrorSlotProvided &&
+    !disableErrorLogging
+  ) {
     console.error("Remote error", storeValue.error);
     notification.error(storeValue.error.message);
   }
--- a/ui/Screen/Onboarding.svelte
+++ b/ui/Screen/Onboarding.svelte
@@ -2,12 +2,11 @@
   import { fade, fly } from "svelte/transition";
 
   import { withRetry } from "../src/api.ts";
-  import * as notification from "../src/notification.ts";
   import { State } from "../src/onboarding.ts";
   import { createIdentity } from "../src/identity.ts";
   import * as screen from "../src/screen.ts";
   import * as session from "../src/session.ts";
-  import * as urn from "../src/urn.ts";
+  import * as error from "../src/error";
 
   import Welcome from "./Onboarding/Welcome.svelte";
   import EnterName from "./Onboarding/EnterName.svelte";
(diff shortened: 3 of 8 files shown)


MeBrei (approved):

ui/DesignSystem/Component/Remote.svelte
```diff
@@ -11,28 +10,31 @@
...
   type SuccessState = { status: remote.Status.Success; data: any };
 
   // If no error slot was provided, svelte will instantiate the fallback div
   let noErrorSlotProvided: HTMLDivElement;
 
-  $: if ($store.status === remote.Status.Error && !!noErrorSlotProvided) {
-    console.error("Remote error", ($store as ErrorState).error);
-    notification.error(($store as ErrorState).error.message);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let storeValue: remote.Data<any>;
```
FintanH, line 22: What's the `any` mean here? :thinking:

ui/Screen/Onboarding.svelte
```diff
@@ -46,12 +45,17 @@
...
       state = State.SuccessView;
-    } catch (error) {
+    } catch (err) {
       animateBackward();
       state = State.EnterName;
-      notification.error(
-        `Could not create identity: ${urn.shorten(error.message)}`
-      );
+      error.show({
+        code: "create-identity-failed",
```
FintanH, line 52: Are we eventually going to give these actual codes?

ui/src/error.ts
```diff
@@ -2,22 +2,48 @@ import * as notification from "./notification";
...
 import * as svelteStore from "svelte/store";
 
-export enum Variant {
-  EntityExists = "ENTITY_EXISTS",
-  GitError = "GIT_ERROR",
-  NotFound = "NOT_FOUND",
-}
-
 export interface Error {
+  // A code in kebab-case for easy identification of the error.
```
FintanH, line 6: Why this formulation for a code, out of interest?

ui/src/error.ts
```diff
@@ -2,22 +2,48 @@ import * as notification from "./notification";
...
+  // The error that caused this error
+  source?: Error;
 }
 
-export const show = (message: string, context: unknown): void => {
-  console.error({ message, context });
+// Turn a Javascript `Error` into our `Error`.
+//
+// Uses the code `unknown-exception`.
+export const fromException = (exception: globalThis.Error): Error => {
```
FintanH, line 23: This is cool :+1:

ui/src/profile.ts
```diff
@@ -80,9 +80,10 @@ export const fetchFollowing = (): void => {
 export const showNotificationsForFailedProjects = async (): Promise<void> => {
   const failedProjects = await project.fetchFailed();
   failedProjects.forEach(failedProject => {
-    error.show(
-      `The project ${failedProject.metadata.name} could not be loaded`,
-      failedProject
-    );
+    error.show({
+      code: "Project failed to load",
```
FintanH, line 84: This doesn't conform to kebab-case :trollface:

ui/src/error.ts
geigerzaehler: Are you referring to “kebab-case”. I just chose this because it is easiest to type compared to screaming snake case or others.

ui/src/profile.ts
geigerzaehler: This was just a test ;) https://github.com/radicle-dev/radicle-upstream/pull/1221/commits/8c334c1ada49d3678869da380952c32af992d743

ui/src/error.ts
FintanH: Ya, so I'd prefer to have some structured way of building an error code. For example, still have a string but have it be `E0001` and increment the codes. As opposed to people picking random sequences of words :smile: 

What are your thoughts?

geigerzaehler: I’m hesitant to use number codes. The primary purpose of these codes is to help developers identitfy an error when a user requests support or we see it during development. If we use descriptive strings as error codes it becomes much easier to identify the error without having to refer to the lookup from number to error description. Conversely, I don’t see an advantage of using numbers as error codes. We would still need to choose an adequate sequence of words to describe the error codes.

ui/DesignSystem/Component/Remote.svelte
geigerzaehler: The `any` type in TypeScript can be coerced into any other type. We are using it to work around the fact that components in svelte cannot be generic (unlike in react). So instead of having a type parameter `T` that is brought into scope by the component and used as `remote.Data<T>` we need to fall back to `any`.

FintanH: Ah, ok. Thanks for the explainer :+1:

ui/Screen/Onboarding.svelte
FintanH: Discussing in https://github.com/radicle-dev/radicle-upstream/pull/1221#discussion_r521412590

ui/src/error.ts
FintanH: > I’m hesitant to use number codes. The primary purpose of these codes is to help developers identitfy an error when a user requests support or we see it during development.

Me too. My main goal from this is to figure out how to make this not ad-hoc process. I can see errors happening like mispellings, copy-and-paste without changing the code, or it's the same error but gets copy-pasted and they have the potential to go out of sync.

> We would still need to choose an adequate sequence of words to describe the error codes.

Ya, I would say reasons and codes should be paired with each other.

In the description you said:
> I’ve decided not to make code an enum. Having a centralized place that lists all errors does not help us with debugging. When debugging we need to look at the place where the error originated which can easily be found be grepping for the error code.

What about having constants in a file, so something like: `const PROJECT_FAILED = name => { code: "project-request-failure", reason: The project {name} could not be loaded" }`?

We can then grep for `PROJECT_FAILED` to see where it's used.

FintanH (approved):
:100: :tada: :raised_hands: :seedling:

geigerzaehler:
After a brief chat, @FintanH and I decided that having a central place for all error codes is a good idea after all. https://github.com/radicle-dev/radicle-upstream/pull/1221/commits/107b6857ab0bd5240661752f976c6d50f2014788

ui/src/error.ts
geigerzaehler: Resolved in https://github.com/radicle-dev/radicle-upstream/pull/1221#issuecomment-725958625

ui/src/error.ts
```diff
@@ -42,7 +73,11 @@ export const setFatal = (fatalError: FatalError): void => {
 };
 
 ipc.listenProxyError(proxyError => {
-  console.error("Proxy process exited", { proxyError });
+  log({
+    code: Code.UnexpectedProxyExit,
+    message: "Proxy process exicted unexpectedly",
```
rudolfs, line 78:
```suggestion
    message: "Proxy process exited unexpectedly",
```

ui/App.svelte
```diff
@@ -101,7 +100,7 @@
 <NotificationFaucet />
 <Theme />
 
-<Remote {store} context="session">
+<Remote {store} context="session" disableErrorLogging={true}>
```
rudolfs, line 103: Why are we disabling the error logging here?

ui/src/session.ts
```diff
@@ -74,29 +74,27 @@ interface UpdateSettings extends event.Event<Kind> {
...
+        sessionStore.success({ status: Status.SealedSession });
+        return;
       }
-    });
-};
+    }
 
-const fetchSession = (): Promise<void> => {
-  return fetchSessionRetry().catch(sessionStore.error);
+    sessionStore.error({
```
rudolfs, line 92: Is the surfacing of this error to the user part of a follow-up issue?

ui/src/error.ts
geigerzaehler: Might have been very excited when I wrote this.

ui/App.svelte
geigerzaehler: If fetching the session errors, we show the blue screen. This means there is no need to show an additional notification.

ui/src/session.ts
geigerzaehler: What do you mean by “surfacing”? Setting the error her shows to the blue screen.

rudolfs: So I tried to manually break this by pointing the session request to a non-existent endpoint to simulate an error. The app would just go blank and show an error in the console, and when I refresh manually it shows the onboarding start screen.

I expected there to get a notification about the error that the session couldn't be fetched or something like that.

```diff
diff --git a/ui/src/session.ts b/ui/src/session.ts
index 0ed327af..8fb65d0b 100644
--- a/ui/src/session.ts
+++ b/ui/src/session.ts
@@ -76,7 +76,10 @@ type Msg = ClearCache | Fetch | UpdateSettings;

 const fetchSession = async (): Promise<void> => {
   try {
-    const ses = await api.withRetry(() => api.get<SessionData>(`session`), 200);
+    const ses = await api.withRetry(
+      () => api.get<SessionData>(`session-error-please`),
+      200
+    );
     sessionStore.success({ status: Status.UnsealedSession, ...ses });
   } catch (err) {
     if (err instanceof api.ResponseError) {
```

Screenshot 2020-11-12 at 16 42 13

Screenshot 2020-11-12 at 16 42 27

xla (dismissed):
🔬 💌 🛩 🚞

ui/src/session.ts
geigerzaehler: The first issue (blank screen with uncaught exception) is due to #1248. It has nothing to do with session errors per se.

The second issue is due to a 404 indicating that the session is missing and that the user needs to be onboarded. The problem is that we’re abusing transport errors (a 404) with application responses (session not available) and we’re unable to distinguish between the two. This is not something that we can address here.

rudolfs (dismissed):

rudolfs (approved):

xla (approved):
💞 📲 👻 🎺

PR #1228, opened 2020-11-12
feat(ui): reduce number of retries after proxy restart

geigerzaehler:
With #1153 the proxy restarts much quicker after the keystore is unsealed or created. This allows us to reduce the number of retries and fail earlier if the proxy does not start properly. (We now wait 5
seconds).

As part of this change we also adjust the signature of `withRetry`. Now the caller is able to control all timing related parameters.

### DIFF
--- a/ui/src/api.ts
+++ b/ui/src/api.ts
@@ -126,7 +126,7 @@ export const withRetry = async <T>(
     try {
       return await request();
     } catch (error) {
-      if (error.message !== "Failed to fetch" || retries > 0) {
+      if (error.message !== "Failed to fetch" || retries < 0) {
         throw error;
       }
     }
--- a/ui/src/api.ts
+++ b/ui/src/api.ts
@@ -119,13 +119,14 @@ const delay = (delay: number) => {
 
 export const withRetry = async <T>(
   request: () => Promise<T>,
-  delayTime: number
+  delayTime: number,
+  retries: number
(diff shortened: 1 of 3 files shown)


FintanH (dismissed):
:clock1:

xla (dismissed):
💋 🕜 📚 🍗

ui/src/api.ts
```diff
@@ -119,13 +119,14 @@ const delay = (delay: number) => {
...
+  delayTime: number,
+  retries: number
 ): Promise<T> => {
-  for (let retries = 0; ; retries++) {
+  for (; ; retries--) {
     try {
       return await request();
     } catch (error) {
-      if (error.message !== "Failed to fetch" || retries > 200) {
+      if (error.message !== "Failed to fetch" || retries > 0) {
```
MeBrei, line 129:
```suggestion
      if (error.message !== "Failed to fetch" || retries < 0) {
```

Or am I understanding the loop wrong?

MeBrei (changes requested):

ui/src/api.ts
geigerzaehler: Great catch!

MeBrei (approved):

FintanH (approved):
Nice catch @MeBrei!

rudolfs (approved):

PR #1238, opened 2020-11-12
refactor(proxy): improve project search semantics

### LINKED ISSUE #1148: Change project search semantics
Instead of timing out the project searches we should back off the queries and only cancel on user input. Also searches should be removed entirely when canceled so they can be reissued.

FintanH:
Closes #1148 

- [x] No upper bound on queries/clones
- [x] Add remove to the waiting room
- [x] Remove URNs from the waiting room on cancel
- [x] Have a back-off of request (~and clone attempts~ clone attempts get marked off as failed for peers)


rudolfs (dismissed):

proxy/coco/src/peer/run_state.rs
```diff
@@ -411,12 +411,16 @@ impl RunState {
...
-                ControlCommand::Respond(control::Response::CancelSearch(
-                    sender,
-                    self.waiting_room.canceled(&urn, timestamp),
-                )),
-            )],
+            ControlInput::CancelRequest(urn, timestamp, sender) => {
+                let request = self.waiting_room.canceled(&urn, timestamp).map(|()| {
+                    self.waiting_room
+                        .remove(&urn)
+                        .expect("URN was found for cancel")
```
xla, line 418: Can we avoid panics in the `RunState`. It's a sign that we didn't hold it right.

xla: Also why would we panic on a `None`, it's fine to make the `Option` part of the response, then the caller can decide if it was unexpected behaviour.

xla (changes requested):
Dope! One tiny thing.

proxy/coco/src/peer/run_state.rs
FintanH: Well it shouldn't _be_ `None` because we're in the `Ok` case of the call to `cancelled`. Happy to return the `Option` regardless.

FintanH:
Can we _not_ press **Update Branch** on each other's branches, please?

proxy/coco/src/peer/run_state.rs
FintanH: https://github.com/radicle-dev/radicle-upstream/pull/1238/commits/55e50aef3eb906e4026afd93e7894030164ef451

rudolfs:
> Can we not press Update Branch on each other's branches, please?

Sorry, won't touch that button anymore.

FintanH:
> Sorry, won't touch that button anymore.

Thanks, appreciate it :bow:

xla (approved):
➗ 🕗 🐒 🐔

rudolfs (approved):

PR #1247, opened 2020-11-12
feat(proxy): persist WaitingRoom

### LINKED ISSUE #1147: Persist project searches
Currently we don't persist project searches beyond app starts.

FintanH:
Closes #1147 

I did a little bit of reorganisation by introducing a `peer::waiting_room` module so that we can have persistence logic for the config and the WaitingRoom in there.

There is now a new command `PersistWaitingRoom` which we issue when we request a new URN. This could potentially leave the waiting room in previous state if we boot up again while not having requested a new URN.
Maybe we should persist at every point of `Found` and `Cloning`. It's a little hard to figure out the right place to do this because `RunState` has the `WaitingRoom` and `Subroutines` has the `kv::Store` :shrug: 

P.S. Would `CheckpointWaitingRoom` be a better name? :thinking:


rudolfs (changes requested):
Removing a request from the waitingroom by clicking the "Unfollow" button is not persisted. Does removing requests depend on https://github.com/radicle-dev/radicle-upstream/pull/1238?

Steps to repro:
1. start app, search a non-existant project `rad:git:hwd1yregz5rifajoy85aefeg81her8xxwmky936kytcnntsjftmxa87nuhe`
2. the project shows up in the "Following" tab in a "Still looking..." section
3. click on the "Unfollow" button of the search request. It should disappear from the list.
4. restart the app

Expected: to not see the request in the "Following" tab
Actual: the request is still there.

xla (changes requested):
Persisting the config doesn't make much sense imho, these params are not bound to the session of a user, as they are never surfaced or can be changed by user input. We can think abou tthat when it comes to that. Furthermore it's okay to move things out of `run_state` as it grows quite large, but then it should be part of the run_state module hierarchy, sanse the persistence parts.

To reiterate `RunState` is the side-effect free transition of the peer state machne. `subroutine.rs` is the side-effect management. Given this mantra things should be clear.

FintanH:
> Removing a request from the waitingroom by clicking the "Unfollow" button is not persisted. Does removing requests depend on #1238?

Yes it does depend on that PR :)

proxy/coco/src/peer.rs
```diff
@@ -29,13 +29,15 @@ pub mod gossip;
...
 mod run_state;
-pub use run_state::{AnnounceConfig, Config as RunConfig, Event, Status, SyncConfig};
+pub use run_state::{config as run_config, Config as RunConfig, Event, Status};
 
 mod subroutines;
 use subroutines::Subroutines;
 
 mod sync;
 
+pub mod waiting_room;
```
xla, line 39: What's the reason to `pub` this one?

proxy/coco/src/peer/run_state.rs
```diff
@@ -376,6 +165,10 @@ impl RunState {
         }
     }
 
+    pub fn set_waiting_room(&mut self, waiting_room: WaitingRoom<Instant, Duration>) {
```
xla, line 168: We can spare us that setter, if the runstate construction would take a `WaitingRoom` as argument.

xla: The reason this seems less favorable is that the mutability of the field might lead to unexpected runtime behaviour when the waiting room is swapped out under the arse of the state.

xla (changes requested):
Dope! Would love to see the setter go.

proxy/coco/src/peer.rs
FintanH: I don't think the functions inside need to be private. I find that it's easier to work with libraries when things global things are made public, like modules, and then if functions and fields need to hidden make those private.

Would you prefer it was private?

xla: It is preferrably to keep the exposed surface smaller as it reduces the coginitive overhead for consumers. In this case esp. the waiting room handling is mostly an internal concern to the peer and should not be general utility.

xla (approved):
🗽 ⛹ 🔭 😲

rudolfs (approved):

PR #1250, opened 2020-11-13
feat(ui): log unhandled exceptions

geigerzaehler:
As part of #1187 we show notifications for all unhandled exceptions.

### DIFF
--- a/ui/src/error.ts
+++ b/ui/src/error.ts
@@ -16,9 +16,11 @@ export interface Error {
 }
 
 export enum Code {
-  SessionFetchFailure = "SessionFetchFailure",
   ProjectRequestFailure = "ProjectRequestFailure",
+  SessionFetchFailure = "SessionFetchFailure",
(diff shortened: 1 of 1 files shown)


xla (approved):
🐤 🍤 🏂 🕜

ui/src/error.ts
```diff
@@ -16,9 +16,11 @@ export interface Error {
 }
 
 export enum Code {
-  SessionFetchFailure = "SessionFetchFailure",
   ProjectRequestFailure = "ProjectRequestFailure",
+  SessionFetchFailure = "SessionFetchFailure",
   UnexpectedProxyExit = "UnexpectedProxyExit",
```
FintanH, line 21: Do you want to do the rename you mentioned in the docs PR now?

FintanH (approved):
Nice!

ui/src/error.ts
geigerzaehler: Will do separately

PR #1251, opened 2020-11-13
feat(ui): use error logging for all `catch` statements

geigerzaehler:
As part of #1187 we add error logging statements to all places where we catch exceptions.


ui/src/ipc.ts
```diff
@@ -4,6 +4,15 @@ import * as ipcTypes from "../../native/ipc-types";
 
 export type { ProxyError } from "../../native/ipc-types";
 
+// `true` if we are running unit tests with Jest.
+const isNodeTestEnv = Boolean(
+  globalThis.process && globalThis.process.env["NODE_ENV"] === "test"
```
rudolfs, line 9: TIL: `globalThis`.

ui/src/ipc.ts
```diff
@@ -4,6 +4,15 @@ import * as ipcTypes from "../../native/ipc-types";
...
+);
+
+// `true` if the if this code is run by the Cypress test driver.
```
rudolfs, line 12:
```suggestion
// `true` if this code is run by the Cypress test driver.
```

rudolfs (commented):
Here's a couple of more places that could use the new error handling:

- [x] https://github.com/radicle-dev/radicle-upstream/blob/b7cfe1999ad133e1a7c415afcd36b4567d0795c0/ui/Screen/Project/Commit.svelte#L20

- [x] https://github.com/radicle-dev/radicle-upstream/blob/b7cfe1999ad133e1a7c415afcd36b4567d0795c0/ui/src/validation.ts#L179

- [x] https://github.com/radicle-dev/radicle-upstream/blob/b7cfe1999ad133e1a7c415afcd36b4567d0795c0/ui/src/session.ts#L136

- [ ] https://github.com/radicle-dev/radicle-upstream/blob/b7cfe1999ad133e1a7c415afcd36b4567d0795c0/ui/src/session.ts#L143

Possibly these as well (I'm not sure)?

- https://github.com/radicle-dev/radicle-upstream/blob/b7cfe1999ad133e1a7c415afcd36b4567d0795c0/ui/src/source.ts#L318

- https://github.com/radicle-dev/radicle-upstream/blob/b7cfe1999ad133e1a7c415afcd36b4567d0795c0/ui/src/source.ts#L405

ui/src/validation.ts
```diff
@@ -175,9 +175,7 @@ export const createValidationStore = (
     if (!inputStore) {
       inputStore = writable(input);
       inputStore.subscribe((input: string) => {
-        runValidations(input).catch(reason => {
-          console.error("runValidations() failed: ", reason);
-        });
```
geigerzaehler, line 180: This is now handled by the `unhandledrejection` event handler

ui/src/session.ts
```diff
@@ -132,17 +137,11 @@ const update = (msg: Msg): void => {
...
-
+      fetchSession();
       break;
 
     case Kind.UpdateSettings:
-      updateSettings(msg.settings).catch(reason => {
-        console.error("updateSettings() failed: ", reason);
-      });
-
+      updateSettings(msg.settings);
```
geigerzaehler, line 144: This is now handled by the `unhandledrejection` event handler

ui/src/session.ts
```diff
@@ -132,17 +137,11 @@ const update = (msg: Msg): void => {
   switch (msg.kind) {
     case Kind.Fetch:
       sessionStore.loading();
-      fetchSession().catch(reason => {
-        console.error("fetchSession() failed: ", reason);
-      });
-
+      fetchSession();
```
geigerzaehler, line 140: This is now handled by the `unhandledrejection` event handler

geigerzaehler:
@rudolfs: All remaining locations have been updated.

rudolfs:
Found one more:

- [ ] https://github.com/radicle-dev/radicle-upstream/blob/c7d99bb4afb13322b5ce2fa689d9299306c9a006/ui/src/session.ts#L134

geigerzaehler:
> Found one more:

Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1251/commits/872ffc46cf2285290b44708f6a80e3bd345de70b

ui/src/session.ts
```diff
@@ -127,11 +126,20 @@ export const createKeystore = (passphrase: string): Promise<null> => {
...
 };
 
-const updateSettings = (settings: Settings): Promise<void> =>
-  api
-    .set<Settings>(`session/settings`, settings)
-    .then(fetchSession)
-    .catch((err: error.Error) => notification.error(err.message));
+const updateSettings = async (settings: Settings): Promise<void> => {
+  try {
+    api.set<Settings>(`session/settings`, settings);
```
rudolfs, line 131: Don't we need to `await` here as well? For the 2 api calls happen serially.

geigerzaehler: Yeah, we do. Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1251/commits/98f5f6067046f1f39ed9cd6e4e013958476e691f

rudolfs:
Not sure what's going on here, but project search seems to be broken now:

| master | thomas/all-error-logging |
|--------|--------------------------|
| master | this |

I see only this in the console:
```
Failed to load resource: the server responded with a status of 404 (Not Found)
```

geigerzaehler:
I’ve fixed the problem and added an issue to add tests for the search modal (#1262).

rudolfs:
The behaviour is still different:

| master | thomas/all-error-logging |
|--------|--------------------------|
| master | th |

geigerzaehler:
:sob: I think I got it now. Thanks for staying with me on this.

rudolfs (approved):
LGTM 🚀

MeBrei (approved):

Issue #1133, opened 2020-10-29
Resolve electron security warnings


rudolfs:
We get these when we start the app in dev mode:

```
electron/js2c/renderer_init.js:15 (electron) Security Warning: webFrame.executeJavaScript was called without worldSafeExecuteJavaScript enabled. This is considered unsafe. worldSafeExecuteJavaScript will be enabled by default in Electron 12.
```

```
electron/js2c/renderer_init.js:135 Electron Security Warning (Insecure Content-Security-Policy) This renderer process has either no Content Security
    Policy set or a policy with "unsafe-eval" enabled. This exposes users of
    this app to unnecessary security risks.

For more information and help, consult
https://electronjs.org/docs/tutorial/security.
This warning will not show up
once the app is packaged.
```

<img width="701" alt="Screenshot 2020-10-29 at 16 57 02" src="https://user-images.githubusercontent.com/158411/97599006-d83e8000-1a07-11eb-9c16-8531f31bf1d0.png">

MeBrei:
The first warning seems to be resolved.

PR #1253, opened 2020-11-13
chore(ui): update ui deps

rudolfs:
As part of https://github.com/radicle-dev/radicle-upstream/issues/1133 @MeBrei and I went through the [Electron security documentation](https://www.electronjs.org/docs/tutorial/security). One point was that dependencies should be kept up-to-date.

Did a smoke test on macOS for the development environment as well as the packaged production version -- didn't find any issues. Would be nice if somebody would test it briefly on Linux.


xla (approved):
😀 😊 ⏳ 🍣

MeBrei (approved):

PR #1264, opened 2020-11-17
chore(ui): add basic tests for project following

### LINKED ISSUE #1262: Add tests for search modal
We don’t have any tests for the search modal.

geigerzaehler:
We add a test for following and unfollowing a project.

I’m not sure whether this allows us to close #1262. Maybe I’m missing some test cases.

### DIFF
--- a/cypress/integration/project/follow.spec.ts
+++ b/cypress/integration/project/follow.spec.ts
@@ -1,0 +1,36 @@
+import * as commands from "../../support/commands";
+
+context("project following", () => {
+  const projectId =
+    "hwd1yre8dchhkyd34h9f5h7ymbnj5tkzoezb9kuthi885ecbuu5ey69cipw";
+
+  beforeEach(() => {
+    commands.resetProxyState();
+    commands.onboardUser("cloudhead");
+    cy.visit("./public/index.html");
+  });
+
+  it("follows and unfollows", () => {
+    commands.pick("Following").click();
+    commands.pick("primary-action").contains("Look for a project").click();
+    commands.pick("search-input").type(`rad:git:${projectId}`);
+    commands.pick("follow-toggle").should("contain", "Follow");
+    commands.pick("follow-toggle").click();
+    commands
+      .pick("notification")
+      .should("contain", "You’ll be notified when this project has been found");
+
+    // The “following” list doesn’t update automatically
+    cy.reload();
+
+    commands
+      .pickWithContent("undiscovered-project", projectId)
+      .trigger("mouseenter")
+      .within(() => {
+        commands.pick("follow-toggle").should("contain", "Following");
+        commands.pick("follow-toggle").click();
+      });
+
+    commands.pick("empty-state").should("exist");
+  });
+});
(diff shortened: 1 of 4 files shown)


cypress/integration/project/follow.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    commands.pick("primary-action").contains("Look for a project").click();
+    commands.pick("search-input").type(`rad:git:${projectId}`);
+    commands.pick("follow-toggle").should("contain", "Follow");
+    commands.pick("follow-toggle").click();
+    commands
+      .pick("notification")
+      .should("contain", "You’ll be notified when this project has been found");
+
+    // The “following” list doesn’t update automatically
+    cy.reload();
```
rudolfs, line 24: Why does this work without an explicit reload outside of the tests?

noreload

geigerzaehler: I don’t know. It looks to me that outside of the tests events from the peer trigger an implicit update to the screen.

xla: Do you observe the relevant `localPeer` events in testing?

rudolfs: Ah, that makes sense, we haven't bothered to set up the SSE stuff to work within cypress. I wonder how much work that would be?

xla: @rudolfs Are you saying the UI doesn't establish an SSE connection when run with cypress>

rudolfs: @xla correct: https://github.com/cypress-io/cypress/issues/2747.

rudolfs: I added a debug log to `localPeer.ts`:
```diff
diff --git a/ui/src/localPeer.ts b/ui/src/localPeer.ts
index ae3ca89c..9b4c2fc6 100644
--- a/ui/src/localPeer.ts
+++ b/ui/src/localPeer.ts
@@ -83,6 +83,7 @@ session.session.subscribe(sess => {
         { withCredentials: true }
       );
       eventSource.addEventListener("message", msg => {
+        console.log(msg);
         const event: Event = JSON.parse(msg.data);
         eventStore.set(event);
       });
```

And it seems like we only get the first event and no subsequent events through the `EventSource` when running within cypress. Might be worth investigating: https://github.com/cypress-io/cypress/issues?q=sse.

xla: To avoid any red herrings, we are not using SSE from cypress itself, so we should not be affected by the conversations of having sse support in cypress itself.

rudolfs: @xla we are running the our app within the Electron shell of Cypress.

geigerzaehler: It has nothing to do with cypress. I started the app with `proxy --test` and had the same issue.

geigerzaehler:
Can I get another look on this?

rudolfs (approved):
Looks good!

Can we create a follow-up issue to get the SSE stuff working in tests?

cypress/integration/project/follow.spec.ts
geigerzaehler: Created an issue: https://github.com/radicle-dev/radicle-upstream/issues/1279

rudolfs: 🙏

xla (approved):
🛄 🚣 👡 🚾

PR #1277, opened 2020-11-18
fix(proxy): surf branches

FintanH:
There was an issue in `radicle-surf` where some branches would not parse correctly, for example `xla/handle-disconnect`.

This is now fixed in `surf` and pushed to crates.io as v0.5.1. So I relaxed the versioning in upstream so it's easier to get hot fixes like this in v0.5.


xla (approved):
🔥 📱 🏂 🚨

geigerzaehler (approved):

PR #1280, opened 2020-11-18
fix(ui): specify List keys

### LINKED ISSUE #1269: Add/remove items in waiting area is inconsistent
- When adding items to the waiting area, the list is updated inconsistently leading to duplicates.
- Items are being removed from the bottom, instead of the one which is being clicked on.

If the app is refreshed manually, then the list shows the correct items.

|adding|removing|
|-------|---------|
| adding | removing |

Might be due to a missing [keyed each](https://svelte.dev/docs#each).

MeBrei:
Closes #1269

### DIFF
--- a/ui/DesignSystem/Component/List.svelte
+++ b/ui/DesignSystem/Component/List.svelte
@@ -10,6 +10,7 @@
 
   export let dataCy = "";
   export let style = "";
+  export let key = "";
 </script>
 
 <style>
@@ -50,7 +51,7 @@
 <div class="list-container" {style}>
   {#if items.length > 0}
     <ul data-cy={dataCy}>
-      {#each items as item}
+      {#each items as item (key.length > 0 ? item[key] : item)}
         <li
           class:hover={styleHoverState}
           on:click={() => dispatch('select', item)}>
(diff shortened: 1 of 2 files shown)


rudolfs (approved):

ui/DesignSystem/Component/List.svelte
```diff
@@ -10,6 +10,7 @@
 
   export let dataCy = "";
   export let style = "";
+  export let key = "";
```
xla, line 13:
```suggestion
  export let key: string | null = null;
```

This should express optionality not default. The caller should be empowered to determine if the items are keyed or not.

ui/Screen/Profile/Following.svelte
```diff
@@ -76,7 +76,7 @@
             exist.
           </p>
         </div>
-        <List items={data.requests} let:item={request}>
+        <List items={data.requests} let:item={request} key="urn">
```
xla, line 79: If a key was passed the items should be keeyed, otherwise not.

xla (changes requested):
Smart little change.

xla (approved):
🍢 🚒 ⌚️ ❔

rudolfs (approved):

PR #1281, opened 2020-11-19
fix(ui): fetch list of requested projects after search

### LINKED ISSUE #1279: Following a project does not update the “Following” tab
Following a project does not always update the “following” tab on a user’s profile. The followed project should be added to the list displayed in the “following” tab after a user has clicked the “follow” button.

I believe the problem is reliably reproducible if the proxy is not connected to any peer. This can be simulated with the `--test` flag. To reproduce this issue take a look at the cypress tests for following which include a reload that forces the update which should happen automatically.

geigerzaehler:
We fetch the list of requested projects again after searching for a project. This includes the searched for project in the project list.

This includes a separate commit that simplifies the `search` code by removing the use of events.

Fixes #1279


xla (changes requested):
Do we have an understanding why we don't observe the events as intended and working before?

geigerzaehler:
> Do we have an understanding why we don't observe the events as intended and working before?

I’ve posted the result of my investigation on the issue https://github.com/radicle-dev/radicle-upstream/issues/1279#issuecomment-730328933.

To me the behavior of the proxy is correct and this is the right fix on the frontend.

rudolfs (dismissed):

xla:
> I’ve posted the result of my investigation on the issue [#1279 (comment)](https://github.com/radicle-dev/radicle-upstream/issues/1279#issuecomment-730328933).
> 
> To me the behavior of the proxy is correct and this is the right fix on the frontend.

That's a good find and definitely indicates that the original implementation was done under the happiest of conditions. The general approach to managing the lifecycle of the items in the list is through the reactivity to significant events concerning the waiting room. Therefore it is perceivable to remedy the issue by introducing a `RequestCreated` variant which will take the place of the `RequestQueried` variant, if emitting under all connectivity scenarios it should correctly address the faulty behaviour.

geigerzaehler:
I think the fix in this PR is more robust and simpler and I don’t see a clear advantage of the other approach. We’ve already seen that updates based on events are quite brittle since the relationship between the events and data on the API is quite indirect.

xla:
> I think the fix in this PR is more robust and simpler and I don’t see a clear advantage of the other approach. We’ve already seen that updates based on events are quite brittle since the relationship between the events and data on the API is quite indirect.

The change introduces a coupling of the search feature to the profile code, so it couples it more tightly. Why I think this is a bandaid than a proper fix is the fact that the profile page will only properly work if the code path in `search.ts` has been hit. What if in the future the request wasn't created through those means? What if we change the behaviour to where we navigate after a request has been issued from the search modal?

If the following tab behaves correctly independently of spooky action at a distance future extensions will be more robust. If we introduce the new event variant, it will not only make the surface of the waiting room more complete, but also enable future implementations on top of to be more correct. Positive side-effect, less coupling between components.

geigerzaehler:
I’ve updated the PR according to your suggestions @xla.

xla (approved):
♑️ 🔢 🔭 🚠

FintanH (approved):
Makes sense to me :+1:

rudolfs (approved):

PR #1287, opened 2020-11-19
chore(ui): update ui deps

rudolfs:
Follow-up of: https://github.com/radicle-dev/radicle-upstream/pull/1253.

Most notably this upgrades [Electron v10 -> v11](https://www.electronjs.org/releases/stable#release-notes-for-v1101).

Smoke tests passed on:
  - [x] macOS catalina
  - [x] Linux

Would be great if we could test this on:
   - [ ] macOS Big Sur on Apple silicon
   - [x] macOS Big Sur on Intel hardware

**Note to self**: remind everyone to run `yarn cache clean && rm -rf node_modules && yarn` due to `twemoji` fixup.


geigerzaehler (commented):
Could you also run `yarn upgrade` to upgrade the transitive dependencies?

rudolfs:
> Could you also run yarn upgrade to upgrade the transitive dependencies?

Good call! I didn't realise that `yarn upgrade-interactive --latest` doesn't do that automatically. TIL.

geigerzaehler (approved):

geigerzaehler (approved):
I’ve played around with the app and did not find any issues

xla (approved):
🚣 🦀 🔴 🍊

MeBrei (approved):

rudolfs:
Moved this to Draft, so it doesn't get merged accidentally before we test it on macOS Big Sur on Apple silicon.

xla:
> macOS Big Sur on Apple silicon

The target audience being affected by this will unlikely be significant, esp. for the first beta. Instead we might be better served extracting this as a follow-up for ß2 and bring this PR to completion.

rudolfs:
> Will we have a device to do this any time soon?

In the process of ordering one now.

> The target audience being affected by this will unlikely be significant, esp. for the first beta. Instead we might be better served extracting this as a follow-up for ß2 and bring this PR to completion.

I guess this is low prio and can be pushed to ß2.
If the laptop arrives before the ß1 release we might be able to test it though.

xla:
> If the laptop arrives before the ß1 release we might be able to test it though.

As we want to cut the release today/tomorrow let's not depend on that event.

PR #1288, opened 2020-11-19
refactor(proxy): simplify `CreateRequest` control input

geigerzaehler:
This is something I discovered while investigating #1279.

`input::Control::CreateRequest` now always holds a response sender instead of it being optional. The `None` case was only used in the tests and the variant is now aligned with the other `input::Control` variants.

This also allows us to simplify the tests. We remove any assertions on the command output being empty,


xla (approved):
🔔 🐘 🔃 ☝️

rudolfs (approved):

PR #1290, opened 2020-11-19
refactor(ui): change checkout button copy

### LINKED ISSUE #1239: Distinguish between checkout and fork
Currently, doing a checkout is a public action: it pushes your own refs. If we want to preserve this behavior, we should rename this to "Fork", and have another way of doing a regular checkout. Otherwise, the behavior should be changed so that a checkout is a purely local action, and to "fork", one has to push to `rad`. Regardless, the copy and instructions currently don't match the behavior.

cloudhead:
Make it depend on whether we are a contributor already or not (ie. whether we already have a fork).

Closes #1239 

Looks like this now, if you don't already have a fork (otherwise looks the same as before):

2020-11-19-184153_444x315_scrot

### DIFF
--- a/ui/Screen/Project.svelte
+++ b/ui/Screen/Project.svelte
@@ -43,6 +43,10 @@
     selectPeer(peer);
   };
 
+  const isContributor = (userUrn: Urn, users: User[]): boolean => {
+    return users.map(u => u.identity.urn).includes(userUrn);
+  };
+
   // Initialise the screen by fetching the project and associated data.
   fetch(urn);
 </script>
@@ -70,6 +74,9 @@
         </div>
       </div>
     </Header.Large>
-    <Source {project} {selectedPeer} />
+    <Source
+      {project}
+      {selectedPeer}
+      isContributor={isContributor(session.identity.urn, peerSelection)} />
   </Remote>
 </SidebarLayout>
--- a/ui/Screen/Project.svelte
+++ b/ui/Screen/Project.svelte
@@ -4,7 +4,7 @@
 
   import * as modal from "../src/modal";
   import * as path from "../src/path";
-  import { isMaintainer } from "../src/project";
+  import { isMaintainer, isContributor } from "../src/project";
   import type { User } from "../src/project";
   import { fetch, selectPeer, store } from "../src/screen/project";
   import type { UnsealedSession } from "../src/session";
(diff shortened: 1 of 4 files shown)


ui/Screen/Project.svelte
```diff
@@ -43,6 +43,10 @@
     selectPeer(peer);
   };
 
+  const isContributor = (userUrn: Urn, users: User[]): boolean => {
```
xla, line 46: You can express that without any extra information. In the user array if there is one element which is of `peerType` local and has either one of two roles: `Contributor`, `Maintainer` - we know the project has been "forked" or started by the local user.

While your approach is likely to work with how the current peer selection is assembled this can break subtly, as the heuristics that filters out a local peer which only tracks might be changed. Additionally to keep components - ideally -  concerned with presentation; it would be great to move a function like this one close to the types (`project.ts`), to then be imported here.

ui/Screen/Project/Source/CheckoutButton.svelte
```diff
@@ -39,7 +42,9 @@
 <Overlay {expanded} on:hide={hide}>
   <div class="clone-dropdown" hidden={!expanded}>
     <p style="margin-bottom: 0.5rem;">
-      Checkout a working copy to your local disk
+      {#if isContributor}
+        Checkout a working copy to your local disk.
+      {:else}Fork this project and checkout a working copy.{/if}
```
xla, line 47: Would stick it into a `const`, to reduce boolean logic inside markup.

ui/Screen/Project/Source/CheckoutButton.svelte
```diff
@@ -7,8 +7,11 @@
 
   import { dismissRemoteHelperHint, settings } from "../../../src/session.ts";
 
+  export let isContributor = false;
```
xla, line 10: It;s a bit odd to see mention of contributor here, might be better expressed with `isFork` or something along those lines.

xla (changes requested):
Dope, left some feedback.

juliendonck (dismissed):
Looks good from a UX perspective 👻

xla (approved):
🏵 🎺 🌺 📪

rudolfs:
The notification message says _"checked out"_ while doing a fork.
Don't think it's a blocker though.

Screenshot 2020-11-23 at 11 27 47

rudolfs (approved):

cloudhead:
> The notification message says "checked out" while doing a fork.

Yeah I think that's still correct, but we could say "forked and checked out" if we wanted. Problem is right now the UI doesn't show your new remote until you refresh, so we'd want to fix that first.

Issue #1223, opened 2020-11-11
Following project shows no branches


FintanH:
![image](https://user-images.githubusercontent.com/3503528/98827199-7c73ee00-242e-11eb-9f52-6a496dc50b0f.png)

This is likely down to something happening with the stats. Probably the fact that these are remote branches and not local branches in the monorepo.

cloudhead:
Since this is potentially too much to get in for B1, a quick hack that would be acceptable is to either:

1. Show a `?` if the number is zero, or
2. Not show the branch stat for followed projects, or when the number is zero

rudolfs:
I'm gonna go with 2. for now -> #1291.

FintanH:
Is this bug still here? Do we know if it's a wrong call to the daemon or should `radicle-source` be fixed?

Kaihuang724:
Closing this issue as I believe it has been fixed

PR #1291, opened 2020-11-20
fix(ui): don't show stats that equal zero

rudolfs:
This is a workaround for #1223, until we figure out the underlying issue in surf.

- missing branches in project screen
  Screenshot 2020-11-20 at 06 12 22
- missing branches in "Following" tab
  Screenshot 2020-11-20 at 06 13 03
- all stats available in "Following" tab
  Screenshot 2020-11-20 at 06 13 14
- all stats available in project screen
  Screenshot 2020-11-20 at 06 13 18

### DIFF
--- a/ui/DesignSystem/Component/Header/Large.svelte
+++ b/ui/DesignSystem/Component/Header/Large.svelte
@@ -108,19 +108,25 @@
         {/if}
         {#if stats}
           <div class="project-stats" data-cy="project-stats">
-            <div class="project-stat-item">
-              <Icon.Branch />
-              <p>
-                {stats.branches === 1 ? `1 Branch` : `${stats.branches} Branches`}
-              </p>
-            </div>
-            <span class="typo-mono-bold project-stat-separator">•</span>
-            <div class="project-stat-item">
-              <Icon.User />
-              <p>
-                {stats.contributors === 1 ? `1 Contributor` : `${stats.contributors} Contributors`}
-              </p>
-            </div>
+            {#if stats.branches > 0}
+              <div class="project-stat-item">
+                <Icon.Branch />
+                <p>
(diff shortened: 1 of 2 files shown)


juliendonck:
> * missing branches in project screen

Can't we just fake `1 branch` here? THere's always 1 branch, right?

rudolfs:
> Can't we just fake 1 branch here? THere's always 1 branch, right?

There's always at least one, but it could also be more than one.
I think it's worse if we always show "1 branch", but the revision selector has three, for example.

We should really fix the underlying issue in surf if we want to do this proper.

geigerzaehler (approved):

juliendonck (approved):
Ok makes sense

xla (changes requested):
Can we avoid the heavy bools in the markup and have a single flag to show or not show stats?

rudolfs:
> Can we avoid the heavy bools in the markup and have a single flag to show or not show stats?

@xla you mean show all stats or show none? Does it make sense if we can show the other stats, and considering that this is a short-lived workaround which we can remove once the underlying issue is fixed?

xla:
> @xla you mean show all stats or show none? Does it make sense if we can show the other stats, and considering that this is a short-lived workaround which we can remove once the underlying issue is fixed?

Is there ever the case where we get 0 branches, but have a number for the contributors?

rudolfs:
> Is there ever the case where we get 0 branches, but have a number for the contributors?

Yes, that's why I added the screenshots in the description ☝️ 

The branches are 0 in this case:
  - Screenshot 2020-11-20 at 11 24 04
  - Screenshot 2020-11-20 at 11 24 09

xla (approved):
🏦 📜 🍼 🏓

cloudhead (approved):
Nice!

PR #1294, opened 2020-11-20
fix(ui): increase clickable are in sidebar

geigerzaehler:
We increase the clickable area in the sidebar so that it cover the whole
`li` element. The `li` element provides the affordance via the hover
style and the cursor pointer.

### DIFF
--- a/ui/DesignSystem/Component/Sidebar.svelte
+++ b/ui/DesignSystem/Component/Sidebar.svelte
@@ -142,7 +142,7 @@
       class="item indicator"
       data-cy="settings"
       class:active={path.active(path.settings(), $location)}>
-      <Tooltip value="Settings">
+      <Tooltip value="Settings" style="width: 100%; height: 100%">
         <a href={path.settings()} use:link>
           <Icon.Settings />
         </a>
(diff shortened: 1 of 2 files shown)


ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -142,7 +142,7 @@
       class="item indicator"
       data-cy="settings"
       class:active={path.active(path.settings(), $location)}>
-      <Tooltip value="Settings">
+      <Tooltip value="Settings" style="width: 100%; height: 100%">
```
xla, line 145: Wouldn't be simpler to wrap the `li` in the tooltip?

geigerzaehler: That would not be valid HTML since only `li` elements can be nested inside `ul` elements

juliendonck (changes requested):
It took some magic for this to be consistent and I'd be wary of changing it. In your fix it already pushing just the settings tooltip more to the left.
If you get it to work properly, I'm all for it though.

Wrong positioning
Screen Shot 2020-11-20 at 10 22 33

Right positioning
Screen Shot 2020-11-20 at 10 23 16

geigerzaehler:
After talking with @juliendonck about this we decided that it some more discussion. We have a ticket in #1296

PR #1305, opened 2020-11-20
fix(proxy): fix announcemnets

### LINKED ISSUE #1299: Delta announcements seemingly broken for project with many branches
When adding upstream as a project with 10-20 branches locally checked out, the proxy continously advertises new `Have`s for all branches, indicating that the delta updates we persist are incorrect.

xla:
Fixes #1299


cloudhead (approved):
:1st_place_medal:

rudolfs (approved):

PR #1313, opened 2020-11-23
chore(ui): update copy

abbey-titcomb:
@cloudhead and I did a full walkthrough of the app on Friday and came up with a couple of copy edits that are ready to review in this PR.

In addition, there were two outstanding items that I wanted to note:

1. We decided that it would be best to include a subtitle on the `Lock` page and in the `Manage Remotes` modal, to better inform the user about what they need to do. **These need styling** - we think a margin-top: 0.5rem on each would work, but to be safe, I wanted to check with @juliendonck & @brandonhaslegs 

- [x]  Add styling for `Lock` and `Manage Remotes` (@juliendonck and/or @brandonhaslegs ?)

@cloudhead had some notes on the search bars vs. input bars:

- The `Paste an <X> here...` format makes sense for the `Search` field but not for the `Manage Remotes` field or seed address in the `Settings` page as they are input forms. We should remove the `...` in those fields.

- Also, the point was raised that the `⌘``V` should only be shown when something is copied in a user's clipboard, since you can also type an address. And maybe the text should reflect that as well by replacing`Paste` with `Enter` in all fields.

- [x] Remove `...` in input fields in `Manage Remotes` and `Settings`

- [x] Replace `Paste` with `Enter`

- [ ] Remove  `⌘``V` from all input bars (adding this to new issue to unblock this PR)

Maybe @brandonhaslegs can chime in here with thoughts on what of the above changes we can make for beta? @cloudhead can elaborate on his thoughts.


brandonhaslegs:
> We decided that it would be best to include a subtitle on the Lock page and in the Manage Remotes modal

Can you share the text so I can put it in the design? Shouldn't be a problem. I added them to the design file.

---

> `⌘ + V`

honestly not sure why we need to tell someone they can paste something into a field anyway. you can paste into any field. i think we can remove the `⌘ + V` tips everywhere.

ui/Screen/Lock.svelte
```diff
@@ -43,11 +43,12 @@
...
 <div class="lock">
   <Emoji emoji={'🚪'} size="huge" style="margin-bottom: 0.75rem;" />
-  <h3>Unlock the app</h3>
+  <h3>Knock, knock</h3>
+  <p>Enter your passphrase to unlock your Radicle secret key.</p>
   <div class="form">
     <Input.Password
       autofocus
-      placeholder="Enter your passphrase"
+      placeholder=""
```
juliendonck, line 51: Why remove the placeholder? It feels like bad UX to me.

ui/Screen/Lock.svelte
```diff
@@ -43,11 +43,12 @@
 
 <div class="lock">
   <Emoji emoji={'🚪'} size="huge" style="margin-bottom: 0.75rem;" />
-  <h3>Unlock the app</h3>
+  <h3>Knock, knock</h3>
+  <p>Enter your passphrase to unlock your Radicle secret key.</p>
```
juliendonck, line 47:
```suggestion
    <p style="margin-top: 0.5rem; color: var(--color-foreground-level-6);">Enter your passphrase to unlock your Radicle secret key.</p>

```

ui/Modal/ManagePeers.svelte
```diff
@@ -83,6 +83,7 @@
     <Emoji emoji={'💻'} size="huge" style="margin-bottom: 1.5rem;" />
 
     <h1>Manage remotes</h1>
+    <p>Add a user's Device ID to collaborate with them on this project</p>
```
juliendonck, line 86:
```suggestion
        <p style="margin-top: 0.5rem; color: var(--color-foreground-level-6);">Add a user's Device ID to collaborate with them on this project</p>

```

ui/Screen/Lock.svelte
juliendonck: Also just looks broken

Screen Shot 2020-11-23 at 14 34 45

ui/Screen/Settings.svelte
```diff
@@ -170,11 +170,11 @@
       <div class="section-item-single">
         <div class="info">
           <p class="typo-text-bold">
-            Seeds help you find more projects and people on the network.
+            Seeds help you find more projects and users on the network.
```
juliendonck, line 173: peers?

ui/Screen/Onboarding/EnterPassphrase.svelte
```diff
@@ -127,10 +127,11 @@
...
 
     <p>
-      This decrypts your key so you can access and publish to the Radicle
-      network. Be sure to store this in a safe place —
+      This protects your Radicle keypair in case your device is compromised.
+      It's used to sign and publish to the Radicle network. Be sure to store
+      this in a safe place —
       <span class="typo-text-bold">
-        there is no way to recover lost keys yet!
+        there is no way to recover your keys without it!
```
juliendonck, line 134: This insinuates that you can recover your keys? That's not true, right? This copy was saying that if you lose your password, there's nothing you can do and that's still true.

juliendonck (dismissed):

juliendonck:
> * Add styling for `Lock` and `Manage Remotes` (@juliendonck and/or @brandonhaslegs ?)

Addressed this in my review

@abbey-titcomb how are you expecting the other work to be addressed?

ui/Screen/Settings.svelte
brandonhaslegs: Why not people?

ui/Screen/Lock.svelte
brandonhaslegs: I'd rather say what entering the password does, rather than what it technically does. It unlocks the app. The user doesn't know about a 🔮 secret 🗝 key 🌝

Like my question when I read this text is: Once I enter my password here and I've unlocked my secret key, how do I use that key to unlock the app? Why even mention it? Your password unlocks the app, that's it.

ui/Screen/Lock.svelte
cloudhead: It's written right above the field, so it's redundant.

ui/Screen/Settings.svelte
cloudhead: Whether they are human isn't really the point, "peers" or "users" works for me.

ui/Screen/Onboarding/EnterPassphrase.svelte
cloudhead: "Recover" is meant here as recover from their encrypted form, not recover from loss. We could also say "There is no way to access your keys without it"

brandonhaslegs:
Just also looking around for copy things, and noticed in the New Project Modal, the tooltip reads, `The project name is taken from the repository you selected` Just should keep the verbs consistent. Above we say choose, below in the tooltip we say selected.

Suggest we change that tooltip to  `The project name is taken from the chosen repository`

ui/Screen/Lock.svelte
brandonhaslegs: Agree

ui/Screen/Settings.svelte
abbey-titcomb: I've been trying to use `users` consistently throughout the app and documentation to represent people. `Users` can will be able to manage multiple `peers`

ui/Modal/ManagePeers.svelte
brandonhaslegs: Add a period to the end of this sentence

ui/Screen/Lock.svelte
brandonhaslegs: I don't think we have any fields without placeholders. It's a must.

ui/Screen/Onboarding/EnterPassphrase.svelte
juliendonck: Yes I understand. But the sentence that was replaced wasn't talking about that, it was talking about what happens when you lose your password. So if you're changing the copy of that sentence, it should still talk about the same topic. Here you changed the copy and the topic, meaning that that original message is lost.

ui/Screen/Lock.svelte
cloudhead: My question as a user would then be: why is the app locked?

ui/Screen/Onboarding/EnterPassphrase.svelte
cloudhead: Right, because your keys are stored on disk, but your passphrase isn't, so you have to remember it or back it up explicitly, vs. the keys that are already saved locally.

ui/Screen/Lock.svelte
abbey-titcomb:
```suggestion
      placeholder="Enter your passphrase"
```

oki, changing it back.

ui/Screen/Onboarding/EnterPassphrase.svelte
abbey-titcomb: But I think it does say the same thing. 

We're saying, make sure to store this passphrase in a safe place, because you can't recover (as in decrypt or access) your keys without it. 

If you lose your keys (delete the file etc...) then you won't be able to recover those either, but you wouldn't be able to recover them with a passphrase, that needs multi-device support. 

So I say we sub 'recover' with 'access' to make this clearer and add another note to the Docs, but besides that, I think it works as is:

```suggestion
        there is no way to access your keys without it!
```

ui/Screen/Lock.svelte
abbey-titcomb: I see both sides. While the passphrase unlocks the app, it is also explicitly associated with a user's keys. Therefore, I think they *should* know about the keys. But, we can make this statement more user friendly.

Here is my iteration taking into consideration the fact that we will have `Enter your passphrase` in the input field:

```suggestion
  <p>Unlock the app by granting access to your keys with your passphrase</p>
```

abbey-titcomb: What I gauged from @cloudhead was that the passphrase unlocks the app sure, but it also gives you publishing access to the Radicle Network, so it's not exactly just an 'app' passphrase, since you also add it when you push. Since the passphrase *does* decrypt a keypair so they can be used by the user, I think they should associate it with keys.

Here's my iteration:

' Decrypt your keys with your passphrase to unlock the app.`

> > * Add styling for `Lock` and `Manage Remotes` (@juliendonck and/or @brandonhaslegs ?)
> 
> Addressed this in my review
> 
> @abbey-titcomb how are you expecting the other work to be addressed?

Will update the TODOs once we resolve the conversations and ask people to implement from there.

cloudhead: I like where this is going, I'd propose an iteration on the above:

"Unlock the app by granting access to your Radicle secret key"
[ Enter your passphrase ] [Unlock]

ui/Screen/Onboarding/EnterPassphrase.svelte
cloudhead: :+1:

ui/Screen/Lock.svelte
```diff
@@ -43,7 +43,9 @@
 
 <div class="lock">
   <Emoji emoji={'🚪'} size="huge" style="margin-bottom: 0.75rem;" />
-  <h3>Unlock the app</h3>
+  <h3>Knock, knock</h3>
+    <p style="margin-top: 0.5rem; color: var(--color-foreground-level-6);">Enter your passphrase to unlock your Radicle secret key.</p>
```
abbey-titcomb, line 47:
```suggestion
    <p style="margin-top: 0.5rem; color: var(--color-foreground-level-6);">Unlock the app by granting access to your Radicle secret key.</p>
```

ui/Screen/Lock.svelte
abbey-titcomb: @brandonhaslegs - all good with this?

MeBrei:
@abbey-titcomb It seems like your pre-commit hook didn't run `prettier` correctly, resulting in the failing build. Let me know if I can help with resolving it!

ui/Screen/Lock.svelte
brandonhaslegs: Ah cool, I like the update. Now the placeholder is not repeating information. Cool!

abbey-titcomb:
> @abbey-titcomb It seems like your pre-commit hook didn't run `prettier` correctly, resulting in the failing build. Let me know if I can help with resolving it!

Definitely need

brandonhaslegs (approved):

xla (approved):
🔩 ✍ 🍞 🔇

xla (dismissed):
🚃 🖖 🎁 💪

cloudhead (approved):

ui/Screen/Settings.svelte
juliendonck, line 173: I wonder if we need to think more about this based on this comment: https://github.com/radicle-dev/radicle-link/pull/436#issuecomment-732407411 @abbey-titcomb @cloudhead

xla (approved):
🎉 ⤴️ 🏖 🌙

ui/Screen/Settings.svelte
cloudhead: Yeah, although in the context of the app we can talk about users. It was changed in the protocol because protocols don't really have "users".

abbey-titcomb: Chatting with Kim about this now. Will open an issue if we need to change about it. But I think what @cloudhead captures it.

PR #1315, opened 2020-11-23
chore(proxy): repeatedly stream configured seeds

### LINKED ISSUE #1310: Repeatedly feed configured seeds into discovery
To make sure the app reconnects when a disconnect from the seed happnes, we emit the entries for the configured seeds periodically. Mostly a stop-gap until we figure out better heuristics, ideally in the protocol layer (Gelbe Seiten).

xla:
Fixes #1310


FintanH (approved):
Looks good to me :+1:

proxy/coco/src/config.rs
```diff
@@ -133,22 +135,39 @@ impl discovery::Discovery for StreamDiscovery {
     type Stream = mpsc::Receiver<(PeerId, Vec<SocketAddr>)>;
 
     fn discover(mut self) -> Self::Stream {
+        let last_seeds: Arc<Mutex<Vec<seed::Seed>>> = Arc::new(Mutex::new(vec![]));
         let (mut sender, receiver) = mpsc::channel(1024);
 
+        {
+            let last_seeds = last_seeds.clone();
+            let mut sender = sender.clone();
+            tokio::spawn(async move {
```
geigerzaehler, line 144: Is there a reason why we poll the streams explicitly instead of using stream combinators like `flat_map` and `scan` to return the result stream? Using the combinators would be more concise and we wouldn’t need to worry about the task if the consumer terminates.

xla: We have to spawn this task explicitely as we need to return from the method. Not quite what you explicitely would change heere.

kim (commented):
I can't find this a particularly smart change. What this does is to brute force that the `proxy` is _always_ connected to the bootstrap nodes. This may seem desirable at the moment, but certainly not in the longer run.

Since `StreamDiscovery` is already polling a `watch` channel, why is it not possible to thread around an mpsc sender which gets tickled iff the connection count drops?

proxy/coco/src/config.rs
geigerzaehler: I’d suggest the following implementation
~~~rust
fn discover(mut self) -> Self::Stream {
    let interval = tokio::time::interval(Duration::from_secs(5)).map(Either::Left);
    let seed_updates = self.seeds_receiver.map(Either::Right);

    futures::stream::select(interval, seed_updates)
        .scan(vec![], |last_seeds, item| {
            let seeds = match item {
                Either::Left(_) => last_seeds.clone(),
                Either::Right(seeds) => {
                    if *last_seeds == seeds {
                        vec![]
                    } else {
                        *last_seeds = seeds;
                        last_seeds.clone()
                    }
                },
            };
            futures::future::ready(Some(seeds))
        })
        .flat_map(|seeds| {
            futures::stream::iter(
                seeds
                    .into_iter()
                    .map(|seed| (seed.peer_id, vec![seed.addr])),
            )
        })
        .boxed()
}
~~~

xla:
> I can't find this a particularly smart change. What this does is to brute force that the `proxy` is _always_ connected to the bootstrap nodes. This may seem desirable at the moment, but certainly not in the longer run.

It's not meant as a permanent solution and would be looked at when we addressing radicle-dev/radicle-link#245

> Since `StreamDiscovery` is already polling a `watch` channel, why is it not possible to thread around an mpsc sender which gets tickled iff the connection count drops?

Whiped up an alternative solution in #1316 - which has the downside that it won't reconnect to seeds when connected to a non-seed.

proxy/coco/src/config.rs
xla: Elegant solution!

xla:
Superseded by #1316

PR #1316, opened 2020-11-23
chore(proxy): stream seeds when disconnected

### LINKED ISSUE #1310: Repeatedly feed configured seeds into discovery

xla:
Fixes #1310


proxy/api/src/process.rs
```diff
@@ -171,7 +150,9 @@ async fn run_rigging(
                         .await
                         .expect("Failed to read session store");
 
-                    if seeds == last_seeds {
+                    let current_status = peer_control.current_status().await;
+
+                    if seeds == last_seeds && current_status != coco::PeerStatus::Started {
```
geigerzaehler, line 155: Shouldn’t we broadcast seeds when the status is `Offline`?

xla: You are right.

xla, line 155:
```suggestion
                    if seeds == last_seeds && current_status != coco::PeerStatus::Offline {
```

geigerzaehler (approved):

rudolfs (approved):

kim:
> has the downside that it won't reconnect to seeds when connected to a non-seed

I don't understand... do you mean that it won't do anything unless completely disconnected?

xla:
> I don't understand... do you mean that it won't do anything unless completely disconnected?

@kim In the way we track being connected, we don't distinguish if we are connected to a seed or another peer. So as long as there is at least one connection we consider ourselvees connected/online.

kim:
Ya, but that's what we want, no? There should never be a reason to force a connection to a statically configured bootstrap node _unless_ we lost all connectivity (in which case those are the only "known good" ones). I get that you may want a way to make an apparently-stuck `upstream` node reconnect without having to restart the application -- what about storing a modification time in the session, which could be tickled by user interaction? That [loop](https://github.com/radicle-dev/radicle-upstream/pull/1316/files#diff-a7d7a60c7f96df2291d5a133387f8e3be51bc1176ba89f8a4704ef6febbc8aaaR172) would then fire on _any_ modification, even if the value is the same as before.

xla:
> Ya, but that's what we want, no? There should never be a reason to force a connection to a statically configured bootstrap node _unless_ we lost all connectivity (in which case those are the only "known good" ones). I get that you may want a way to make an apparently-stuck `upstream` node reconnect without having to restart the application -- what about storing a modification time in the session, which could be tickled by user interaction? That [loop](https://github.com/radicle-dev/radicle-upstream/pull/1316/files#diff-a7d7a60c7f96df2291d5a133387f8e3be51bc1176ba89f8a4704ef6febbc8aaaR172) would then fire on _any_ modification, even if the value is the same as before.

You like to tickle! Yes we want that behaviour to not force a conenction to a static bootstrap node, yet currently we depend on these nodes for proper replication, until we **punch** all the Löcher. I thknk we are good for now with the current behaviour, where we force reconnect for a stuck upstream, this should address the original UX hurdle.

PR #1319, opened 2020-11-24
feat(ui): offer theme toggle shorcut

NunoAlexandre:
Offer a new shortcut `mod` `+` `t` to toggle the selected theme.

Spontaneous PR as I've felt the need for this myself :surfing_man:

### DIFF
--- a/ui/Hotkeys.svelte
+++ b/ui/Hotkeys.svelte
@@ -1,12 +1,13 @@
 <script lang="typescript">
   import { location, pop, push } from "svelte-spa-router";
 
+  import * as hotkeys from "./src/hotkeys";
+  import { isDev } from "./src/ipc";
   import * as modal from "./src/modal";
   import * as path from "./src/path";
   import * as screen from "./src/screen";
-  import { isMac } from "./src/settings";
-  import * as hotkeys from "./src/hotkeys";
-  import { isDev } from "./src/ipc";
+  import { settings, updateAppearance } from "./src/session";
+  import { isMac, toggleTheme } from "./src/settings";
 
   const toggle = (destination: string) => {
     if (path.active(destination, $location)) {
@@ -63,6 +64,12 @@
         break;
       case hotkeys.ShortcutKey.NewProjects:
         toggleModal(path.newProject());
+        break;
+      case hotkeys.ShortcutKey.ToggleTheme:
+        updateAppearance({
+          ...$settings.appearance,
+          theme: toggleTheme($settings.appearance.theme),
+        });
         break;
     }
   };
--- a/ui/src/hotkeys.ts
+++ b/ui/src/hotkeys.ts
@@ -23,6 +23,7 @@ export enum ShortcutKey {
   NewProjects = "n",
   Search = "p",
   Settings = ",",
+  ToggleTheme = "t",
 }
 
 export interface KeyboardShortcut {
@@ -41,6 +42,11 @@ export const shortcuts: KeyboardShortcut[] = [
   },
   { description: "Search", key: ShortcutKey.Search, modifierKey: true },
   { description: "Settings", key: ShortcutKey.Settings, modifierKey: true },
+  {
+    description: "Toggle Theme",
+    key: ShortcutKey.ToggleTheme,
+    modifierKey: true,
+  },
 ];
 
 export const devShortcuts: KeyboardShortcut[] = [
--- a/ui/src/settings.ts
+++ b/ui/src/settings.ts
@@ -50,3 +50,12 @@ export const defaultSetttings = (): Settings => ({
 
 // gives back the OS you're using in hotkeys.svelte & shortcuts.svelte
 export const isMac: boolean = navigator.platform.includes("Mac");
+
+export const toggleTheme = (current: Theme): Theme => {
+  switch (current) {
+    case Theme.Light:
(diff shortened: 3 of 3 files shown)


xla:
We've come full circle.

MeBrei (approved):

geigerzaehler (approved):

rudolfs (approved):

juliendonck (approved):
🌝 🌚

brandonhaslegs:
Is this a standard behavior? Never seen it. Kinda cool, but do other apps do this? I feel like I associate `mod + t` with navigating to new _something_. Not creating, but navigating to something. Almost like our search (what we have set to `mod + p` at the moment).

For example in browsers it opens a new tab. In Slack it opens navigating to a new channel.

juliendonck:
> Is this a standard behavior? Never seen it. Kinda cool, but do other apps do this? I feel like I associate `mod + t` with navigating to new _something_. Not creating, but navigating to something to have to search for. Almost like our search (what we have set to `mod + p` at the moment).
> 
> For example in browsers it opens a new tab. In Slack it opens navigating to a new channel.

Are you making the case that it's not necessary or that `t` isn't the right key @brandonhaslegs ?

NunoAlexandre:
> Is this a standard behavior? Never seen it. Kinda cool, but do other apps do this? I feel like I associate `mod + t` with navigating to new _something_. Not creating, but navigating to something to have to search for. Almost like our search (what we have set to `mod + p` at the moment).
> 
> For example in browsers it opens a new tab. In Slack it opens navigating to a new channel.

I get your point regarding the familiarity of using `cmd` `+` `t` to navigate to something new. Would you have another key in mind? Open to change, of course, though I'd rather using `t`heme than some unrelated, not-as-intuitive key. Happy either way :)

brandonhaslegs:
Well, if we want the shortcut, I'd definitely suggest another key since users will be very used to that key doing something else in other apps. But also I do question the necessity of having a shortcut that toggles the theme. Why do you want to swap between themes so regularly?

Related to this: I do think we should enhance the theme functionality to be more robust. Eventually we should have an option to follow the system OS setting, and another option to follow sunrise/sunset. You can see this in the "Future" design screen for Settings.

image

cloudhead:
This is nice, but I'd prefer if we didn't pick such a "top-level" key binding for something that isn't that common to do. I'd reserve `cmd-*` for workflow-related commands. Perhaps `cmd-shift-t` or one of the function keys?

juliendonck:
On second thought, I'd argue that this doesn't deserve a shortcut. It is not something you'd do multiple times a day. If we wanna do anything with this, it would be to connect it to your UI settings (eg on mac if you have dark mode when it's dark outside the UI of the app should switch)

xla:
> On second thought, I'd argue that this doesn't deserve a shortcut. It is not something you'd do multiple times a day. If we wanna do anything with this, it would be to connect it to your UI settings (eg on mac if you have dark mode when it's dark outside the UI of the app should switch)

That's the rationale we remoed it originally. Just to contextualise my full circle comment above.

NunoAlexandre:
> Well, if we want the shortcut, I'd definitely suggest another key since users will be very used to that key doing something else in other apps. But also I do question the necessity of having a shortcut that toggles the theme. Why do you want to swap between themes so regularly?

> 
> Related to this: I do think we should enhance the theme functionality to be more robust. Eventually we should have an option to follow the system OS setting, and another option to follow sunrise/sunset. You can see this in the "Future" design screen for Settings.
> 
> image

It's not necessarily about wanting to change the theme so regularly, it's about having full and ultra easy control over it. I really dislike the sunrise/sunset feature because one day I might want something different, have more or fewer lights on around me, the sun gets hidden behind some clouds (#lifeinthenetherlands), etc. It's also super useful to have this for development.

NunoAlexandre:
> On second thought, I'd argue that this doesn't deserve a shortcut. It is not something you'd do multiple times a day. If we wanna do anything with this, it would be to connect it to your UI settings (eg on mac if you have dark mode when it's dark outside the UI of the app should switch)

Following that rationale, we could argue that creating a new project or changing some settings is neither something a user will do multiple times a day :man_shrugging: 

@cloudhead's suggestion to make it `cmd-shift-t` seems good to me.

brandonhaslegs:
> I really dislike the sunrise/sunset feature because one day I might want something different

Would just be an option, you don't have to have it on. I would set this setting to `System` by default. So it looks like the rest of your OS theme.

`cmd` `shift` `t` is how you reopen a recently closed tab by the way. I use that often. If we ever implement a `cmd` + `t` then `cmd` `shift` `t` would be related to that. I can see us implementing windows at some point too, since it's a mac app and you can't have multiple tabs open.

xla (changes requested):
Whatever the outcome, it should land after `0.1.0`.

cloudhead:
> `cmd` + `shift` + `t` is how you reopen a recently closed tab by the way. I use that often. If we ever implement a `cmd` + `t` to open a new window, then `cmd` + `shift` + `t` would be related to that. I can see us implementing windows at some point too, since it's a standalone app (not a web app) and you can't have multiple tabs open.

Good point. I think we need to think a bit more holistically about shortcuts and look at the ones we already have too.

PR #975, opened 2020-09-30
[wip] feat(ui): Funding Pool v0

NunoAlexandre:
:construction: :building_construction: Under construction, please do not disturb :construction_worker_man: :hammer_and_wrench: 

Closes #974 
Superseds #952

PR #1320, opened 2020-11-24
chore: git-ignore ./sandbox

NunoAlexandre:
We use the `sandbox` for local development artifacts in [`feat/funding`](#975), where it is git-ignored. At the moment, we place a `.local-ethereum-account` in there with the environment dev address in use for development/playing around. Calling it `sandbox` seemed both appropriate and generic enough to use for other needs.

It would be handy to have it already git-ignored on master to avoid having to move it in and out of the repository every time I switch branches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ cypress/fixtures/example.json
 cypress/workspace
 proxy/**/target
 proxy/**/*.rs.bk
+sandbox


xla:
What is that directory used for?

NunoAlexandre:
> What is that directory used for?

>> We use the sandbox for local development artifacts in feat/funding, where it is git-ignored.

xla (approved):
🍬 💸 🏘 🌃

rudolfs (approved):

PR #1303, opened 2020-11-20
build: rename binary to "Radicle Upstream"

rudolfs:
This renames the app binary from `radicle-upstream` to `Radicle Upstream`.
The package filename remains unchanged.

Refs: https://github.com/radicle-dev/radicle-docs/pull/26#discussion_r527079630.

Screenshot 2020-11-20 at 08 54 32

PR #1325, opened 2020-11-24
chore: update reset script

rudolfs:
Related to the change we did in: https://github.com/radicle-dev/radicle-upstream/pull/1303.

### DIFF
--- a/scripts/reset-state.sh
+++ b/scripts/reset-state.sh
@@ -18,9 +18,8 @@ elif [ "$platform" == "Linux" ]; then
 
     dirs=(
         "$HOME/.radicle"
-        "$HOME/.config/Radicle Upstream"
         "$config_home/radicle"
-        "$config_home/radicle-upstream"
+        "$config_home/Radicle Upstream"
         "$data_home/radicle"
         "$data_home/radicle-upstream"
     );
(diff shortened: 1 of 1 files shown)


scripts/reset-state.sh
```diff
@@ -18,6 +18,7 @@ elif [ "$platform" == "Linux" ]; then
 
     dirs=(
         "$HOME/.radicle"
+        "$HOME/.config/Radicle Upstream"
         "$config_home/radicle"
         "$config_home/radicle-upstream"
```
xla, line 23: This is obsolete now.

xla (changes requested):

scripts/reset-state.sh
rudolfs: Right. Should this perhaps be `"$config_home/Radicle Upstream"` and the one on L21 be deleted?

xla: yus

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1325/commits/d641c203dab85cb2d918feba4b5c922e50176ec9

xla (approved):
🚪 🕴 🕷 🌓

FintanH (approved):
utACK

PR #1330, opened 2020-11-24
refactor(proxy): Avoid confusing errors in http route matching

geigerzaehler:
We apply the following rules to the order of HTTP routing filters:

1. Path matchers come first
2. Path matchers must be terminated with `path::end()` or `path::tail()`
3. Method matchers come after path matchers
4. State guards (like `with_unsealed_context`) come after method matchers.

This prevents the following issues:

* We always get a 404 when a path does not match. Without the proper ordering we could get a 405 (method not allowed) or a 403 (not authorized) which would mask the fact that the path was wrong.
* We run the expensive state guards at most once per request since the path-method combination is now uniquely matched. Before, if the state guard was at the beginning we would run it for every route that was combined with `or`.
* We enforce exact matches on paths so that we don’t have redundant segments in our path. Before, for example, `/projects/<urn>/checkout/garbage` was matched.

### DIFF
--- a/proxy/api/src/http/control.rs
+++ b/proxy/api/src/http/control.rs
@@ -20,6 +20,7 @@ fn create_project_filter(
     ctx: context::Context,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     path!("create-project")
+        .and(warp::post())
         .and(super::with_context_unsealed(ctx.clone()))
         .and(super::with_owner_guard(ctx))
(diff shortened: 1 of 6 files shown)


MeBrei (approved):
⭐

xla (changes requested):
Nice, blocking until after `0.1.0`.

NunoAlexandre (approved):
:ice_skate:

PR #1332, opened 2020-11-24
fix(proxy): query peer refs after tracking

### LINKED ISSUE #1242: Fetch on stale gossip
At the moment the app considers the tracking of a peer for project successful only if the tracked peer pushes a commit after it has been tracked. The expected behavior is that a peer should be considered tracked after we were able to obtain its refs for the project.

## Background

The state of peer tracking is determined by the `GET /projects/<urn>/peers` endpoint which uses [`coco::State::list_project_peers`](https://github.com/radicle-dev/radicle-upstream/blob/9bc9fe55f7a3b7218f834cf8823e501f0572ddd0/proxy/coco/src/state.rs#L649). We consider a peer as successfully tracked if the `status` field of a peer in the list returned by `list_project_peers` is set to [`Replicated`](https://github.com/radicle-dev/radicle-upstream/blob/9bc9fe55f7a3b7218f834cf8823e501f0572ddd0/proxy/coco/src/project/peer.rs#L177). This is the case if the following ref exists in the monorepo:

    refs/namespaces/<project_id>/refs/remotes/<peer_id>/self

This check is implemented [here](https://github.com/radicle-dev/radicle-upstream/blob/9bc9fe55f7a3b7218f834cf8823e501f0572ddd0/proxy/coco/src/state.rs#L618-L619).

## Solution

To achieve the expected behavior the local peer should try to obtain the refs given above after tracking the remote peer.

The right way for this seems to emit a `Want` for the project URN with the tracked peer ID as the [`origin`](https://github.com/radicle-dev/radicle-link/blob/e5e1a1b3a251ae9f37dcb2c3211efcc186ea2503/librad/src/net/peer/types.rs#L108). However, at the moment, `librad` [does not apply](https://github.com/radicle-dev/radicle-link/blob/e5e1a1b3a251ae9f37dcb2c3211efcc186ea2503/librad/src/net/peer.rs#L590-L592) the corresponding have requests if the `rev` field is `None`.

geigerzaehler:
When we track a peer for a given project we query the tracked peers refs for the project so that it is considered “tracked”.

Fixes #1242

### DIFF
--- a/proxy/coco/src/peer/announcement.rs
+++ b/proxy/coco/src/peer/announcement.rs
@@ -116,6 +116,8 @@ pub async fn run(state: &State, store: &kv::Store) -> Result<Updates, Error> {
     let old = load(store)?;
     let new = build(state).await?;
     let updates = diff(&old, &new);
+    log::warn!("{:?}", old);
+    log::warn!("{:?}", new);
 
     announce(state, updates.iter()).await;
 
--- a/proxy/coco/src/peer/announcement.rs
+++ b/proxy/coco/src/peer/announcement.rs
@@ -116,8 +116,6 @@ pub async fn run(state: &State, store: &kv::Store) -> Result<Updates, Error> {
     let old = load(store)?;
     let new = build(state).await?;
     let updates = diff(&old, &new);
-    log::warn!("{:?}", old);
-    log::warn!("{:?}", new);
 
     announce(state, updates.iter()).await;
 
--- a/proxy/coco/src/peer/gossip.rs
+++ b/proxy/coco/src/peer/gossip.rs
@@ -21,25 +21,14 @@ pub async fn announce(state: &State, urn: &RadUrn, rev: Option<Oid>) {
 }
 
 /// Emit a [`Gossip`] request for the given `urn`.
-pub async fn query(state: &State, urn: RadUrn) {
-    let protocol = state.api.protocol().clone();
-    protocol
+pub async fn query(state: &State, urn: RadUrn, origin: Option<PeerId>) {
+    state
+        .api
+        .protocol()
         .query(Gossip {
             urn,
             rev: None,
-            origin: None,
-        })
-        .await;
-}
-
-/// Emit a [`Gossip`] request for the given `urn` from the given `peer_id`.
-pub async fn query_with_origin(state: &State, urn: RadUrn, peer_id: PeerId) {
-    let protocol = state.api.protocol().clone();
-    protocol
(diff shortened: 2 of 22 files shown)


proxy/coco/src/peer/announcement.rs
```diff
@@ -116,6 +116,8 @@ pub async fn run(state: &State, store: &kv::Store) -> Result<Updates, Error> {
     let old = load(store)?;
     let new = build(state).await?;
     let updates = diff(&old, &new);
+    log::warn!("{:?}", old);
+    log::warn!("{:?}", new);
```
xla, line 120:
```suggestion
```

proxy/coco/src/peer/gossip.rs
```diff
@@ -30,3 +31,15 @@ pub async fn query(state: &State, urn: RadUrn) {
         })
         .await;
 }
+
+/// Emit a [`Gossip`] request for the given `urn` from the given `peer_id`.
+pub async fn query_with_origin(state: &State, urn: RadUrn, peer_id: PeerId) {
```
xla, line 36: Why wouldn't we extend the query function above with an optional peer_id?

xla (changes requested):
Great stuff! Would be nice to resolve the merge conflicts. Latest link `maint` should have all the changes.

xla (approved):
👓 🎮 🚝 ✅

proxy/coco/src/peer/gossip.rs
```diff
@@ -20,13 +21,14 @@ pub async fn announce(state: &State, urn: &RadUrn, rev: Option<Oid>) {
 }
 
 /// Emit a [`Gossip`] request for the given `urn`.
-pub async fn query(state: &State, urn: RadUrn) {
-    let protocol = state.api.protocol().clone();
-    protocol
+pub async fn query(state: &State, urn: RadUrn, origin: Option<PeerId>) {
```
FintanH, line 24:
```suggestion
pub async fn query(state: &State, urn: RadUrn, origin: impl Into<Option<PeerId>>) {
```

This allows us to pass a `PeerId` without wrapping it in `Some`. Not a blocker, but I do like the pattern

FintanH (approved):
This is sick! Nice :grin:

proxy/coco/src/peer/gossip.rs
geigerzaehler: I’m going to leave it as it is since the other code in this file does not use the pattern.

PR #1371, opened 2020-11-25
fix(proxy): remote tracking semantics

### LINKED ISSUE #1243: Wrong default peer
During testing we noticed the following:
* A peer had tracked two other peers -- a maintainer and a contributor
* The contributor had the most up to date changes
* The default view was selecting the contributor rather than the maintainer on refreshing

FintanH:
When we successfully track a peer but they haven't pushed any branches they will appear in the peer selector but switching to them will fail due to the lack of branches.

We need a way to display the fact that you have tracked a peer but they have no code branches.

***

* order peers in the order described in #1243 
* don't show your own local remote when just tracking (all projects in the Following tab)
* show remote tracking peers as disabled with a tooltip
* make peer dropdown consistent with revision dropdown keeping the currently selected at the top of the list

Fixes #1243


package.json
```diff
@@ -107,6 +107,8 @@
     "start": "RADICLE_UPSTREAM_EXPERIMENTAL=false yarn start:app",
     "start:experimental": "RADICLE_UPSTREAM_EXPERIMENTAL=true yarn start:app",
     "start:app": "run-p --race rollup:watch proxy:start electron:start",
+    "start:release": "run-p --race rollup:watch proxy:start:release electron:start",
```
FintanH, line 110: :raised_hands: 

Although, I think it should be the other way around. `start` should be release and then `start:debug` is the debug build, or whatever you want to call it.

FintanH (commented):
Slick :raised_hands:

rudolfs:
Checkout modal text overflow broken:

Screenshot 2020-11-26 at 19 56 33

rudolfs:
Missing padding makes the dropdown jump around:

| fintan/remote-tracking | master |
|------------------------|--------|
| this| master |

rudolfs (changes requested):
Good work! 🚀 
Tested the functionality, it's all there.

- [x] expanded dropdown corners are not rounded

| master | fintan/remote-tracking |
|--------|------------------------|
| Screenshot 2020-11-27 at 10 03 19 | Screenshot 2020-11-27 at 10 04 31 |

Probably forgot to copy these to the new component:
```css
  .peer:first-child {
    border-radius: 0.1875rem 0.1875rem 0 0;
  }
  .peer:last-child {
    border-radius: 0 0 0.1875rem 0.1875rem;
  }
```

As a side note: I still find the new dropdown UX frustrating (at least now it's consistent with the other one). If you have a larger list of things, it's hard to go through the items one by one, because the list order changes every time you select something. I think we should reconsider this behaviour (not part of this PR, however). \cc @cloudhead, @juliendonck, @brandonhaslegs.

package.json
rudolfs: Is this so that the unlocking and pushing works faster? Do we lose anything by not having proxy in debug mode? If so we could run it by default as @FintanH suggests.

ui/DesignSystem/Component/PeerSelector.svelte
```diff
@@ -0,0 +1,168 @@
...
+        style="vertical-align: bottom; fill: var(--color-foreground-level-4)" />
+    </div>
+  </div>
+  <div class="peer-dropdown-container">
+    <div class="peer-dropdown" hidden={!expanded}>
+      {#each orderPeers(peers) as peer (peer.peerId)}
+        <Entry
+          active={peer.role !== Role.Tracker}
+          on:click={() => onSelect(peer)}
+          selected={peer.identity.peerId == selected.identity.peerId}>
```
MeBrei, line 135:
```suggestion
          selected={peer.identity.peerId === selected.identity.peerId}>
```

ui/DesignSystem/Component/PeerSelector/Entry.svelte
```diff
@@ -0,0 +1,41 @@
+<script lang="typescript">
+  export let active: boolean = true;
```
MeBrei, line 2:
```suggestion
  export let disabled: boolean = true;
```

`active` seems a bit misleading as it refers more to something being selected or not.

ui/src/screen/project.ts
```diff
@@ -166,10 +166,58 @@ const filterPeers = (peers: project.Peer[]): project.User[] => {
...
+      }
+
+      if (
+        a.role === project.Role.Contributor &&
+        b.role === project.Role.Tracker
+      ) {
+        return -1;
+      }
+      if (
+        a.role == project.Role.Tracker &&
```
MeBrei, line 199:
```suggestion
        a.role === project.Role.Tracker &&
```

MeBrei (changes requested):
⭐  Functionality is there, just some small changes.

MeBrei:
I also added an issue that is kind of a follow up #1394 1394

package.json
FintanH: We lose time every time we wait for a git operation :upside_down_face:

MeBrei (approved):

xla (approved):
✨ 🙏 📒 ⏳

PR #1377, opened 2020-11-26
fix(proxy): only save most recent logs from proxy

### LINKED ISSUE #1359: Proxy crashes in long-running app
When leaving the app running (couple of hours) the proxy will crash, likely from the captured logs filling up

```
Proxy process exited with status code ERR_CHILD_PROCESS_STDIO_MAXBUFFER and signal null
```

geigerzaehler:
We change how the proxy process is spawned. Instead of saving all output of the process in a buffer we only save the 200 most recent lines. In addition we forward all of the proxy’s output to the main process output. This allows us to see the proxy logs when starting the app from the terminal.

To implement this we introduce the `ProxyProcessManager` which encapsulates this behavior. The logic for line buffering the output of the process suffers from duplication. I’ve experimented with some abstractions but they were all to complicated and did not add enough value.

Fixes #1359

### DIFF
--- a/native/main.ts
+++ b/native/main.ts
@@ -7,7 +7,7 @@ import {
   shell,
 } from "electron";
 import path from "path";
-import { execFile, ChildProcess } from "child_process";
+import { ProxyProcessManager } from "./proxy-process-manager";
 import { RendererMessage, MainMessage, MainMessageKind } from "./ipc-types";
 
 const isDev = process.env.NODE_ENV === "development";
@@ -91,7 +91,7 @@ class WindowManager {
 }
 
 const windowManager = new WindowManager();
-let proxyProcess: ChildProcess | undefined;
+const proxyProcessManager = new ProxyProcessManager(proxyPath, [], !isDev);
 
 ipcMain.handle(RendererMessage.DIALOG_SHOWOPENDIALOG, async () => {
   const window = windowManager.window;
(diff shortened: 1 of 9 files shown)


native/main.ts
```diff
@@ -145,50 +145,25 @@ const openExternalLink = (url: string): void => {
...
+app.on("ready", () => {
+  proxyProcessManager.run().then(({ status, signal, output: stdout }) => {
     windowManager.sendMessage({
       kind: MainMessageKind.PROXY_ERROR,
       data: {
         status,
         signal,
         stdout,
-        stderr,
+        stderr: "",
```
rudolfs, line 163: Dev leftover?
I don't see the proxy error logs anymore when I kill the proxy.

Screenshot 2020-11-26 at 13 18 26

native/proxy-process-manager.ts
```diff
@@ -0,0 +1,108 @@
...
+  // Throws an error if a process is already running.
+  async run(): Promise<ProcessResult> {
+    if (!this.enabled) {
+      return new Promise(() => undefined);
+    }
+
+    if (this.childProcess !== undefined) {
+      throw new Error("Proxy process already started");
+    }
+    const outputBuffer = new CircularBuffer<string>(Array, 200);
```
rudolfs, line 47: Is there a reason why we keep it so low? Would opt for `10000`.

rudolfs (changes requested):

native/proxy-process-manager.ts
xla: `10000` might be a bit much, these logs are meant to be posted in issues and so on. Maybe somewhere around 500-1000 is managable.

geigerzaehler:
I’ve addressed the issue, @rudolfs, and set the line limit to 1000.

rudolfs:
> I’ve addressed the issue, @rudolfs, and set the line limit to 1000.

https://github.com/radicle-dev/radicle-upstream/pull/1377/files#diff-e9aad0d8c74cb96ae9573c3e3594346000c337b0eab3e680fc0da53b2596e5c8R98

500 !== 1000

native/proxy-process-manager.test.ts
```diff
@@ -0,0 +1,63 @@
+import * as lodash from "lodash";
+import { ProxyProcessManager } from "./proxy-process-manager";
+
+["stdout", "stderr"].forEach(output => {
+  it(`collects and writes ${output}`, async () => {
+    const redirection = output == "stderr" ? "1>&2" : "";
```
rudolfs, line 6:
```suggestion
    const redirection = output === "stderr" ? "1>&2" : "";
```

rudolfs (approved):

xla (approved):
👝 🕠 🃏 💴

PR #1382, opened 2020-11-26
fix(ui): make stats component reactive

geigerzaehler:
We make the stats component reactive so that changes in the props are reflected in the rendered dom.

### DIFF
--- a/ui/DesignSystem/Component/Stats.svelte
+++ b/ui/DesignSystem/Component/Stats.svelte
@@ -1,13 +1,13 @@
-<script>
+<script lang="typescript">
   import { Icon } from "../Primitive";
 
-  export let contributors = null;
-  export let branches = null;
-  export let commits = null;
+  export let contributors: number;
+  export let branches: number;
+  export let commits: number;
 
-  export let style = null;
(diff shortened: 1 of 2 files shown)


xla (approved):
👼 ♊️ 🚅 ⤴️

MeBrei:
When are the stats actually updated without refresh? 
Tried to test it with creating a branch to a project I had open in the app, but the changes were not visible without refreshing.

xla:
> When are the stats actually updated without refresh?

This a chore for the actual ui rendering not for rolling in backend changes afaik.

MeBrei (approved):

geigerzaehler:
> When are the stats actually updated without refresh?
> Tried to test it with creating a branch to a project I had open in the app, but the changes were not visible without refreshing.

Here’s how you can reproduce the bug:
1. Make sure you have at least one project in the “Following” list
2. Follow another project and wait for it to be tracked successfully
3. If the new project is not added to the end of the list in “Following” it will show the stats from the item it replaced. Same for all other projects that were moved by the new project being inserted.

PR #1385, opened 2020-11-26
fix(proxy): validate default git config

### LINKED ISSUE #1358: Git config user set up
As we found out in https://github.com/radicle-dev/radicle-upstream/issues/1348, if there's no global git config set up with user details, we won't be able to create projects as we can't create commits.

We should add this to validations and figure out a flow as to how we want the user to set this up.

FintanH:
Fixes #1358 

To ensure that we can add a first commit when creating a project we must
make sure that the user has a git config set up. We check this before
creating a new project and give back an error if it isn't the case.


xla (approved):
🍆 🎙 🔨 🎱

MeBrei (approved):

Issue #1296, opened 2020-11-20
Fix tooltips on sidenav indicators


brandonhaslegs:
The hover area for the sidenav items should show the tooltip any time the indicator is shown. See below that the indicator (the line to the left) is activated but the tooltip is not shown. (The tooltip is currently only shown on hover of the bounds of the icon.)

![image](https://user-images.githubusercontent.com/4406983/99795836-ba5acb80-2b2c-11eb-972b-73ac68972413.png)

geigerzaehler:
In addition the clickable area (red) for the settings and profile icon are smaller than the hover and cursor pointer area (green).
![sidebar](https://user-images.githubusercontent.com/3919579/99797898-5cc87e00-2b30-11eb-8d89-f77b4efc3a1c.png)

rudolfs:
Seems like this got fixed in #1387.

![fixed](https://user-images.githubusercontent.com/158411/100746845-8cee0780-33e1-11eb-9da8-170b260918b4.gif)

PR #1387, opened 2020-11-26
fix(ui): Increase hit area for sidebar buttons

juliendonck:
This increases the clickable area of sidebar items to the size where the indicator shows up.  (closes #664 and #1296)
sidebar-hover

### DIFF
--- a/ui/DesignSystem/Component/ConnectionStatusIndicator.svelte
+++ b/ui/DesignSystem/Component/ConnectionStatusIndicator.svelte
@@ -50,26 +50,26 @@
   <div>
     {#if status.type === StatusType.Online}
       <Tooltip value={`You’re connected to ${peerCount(status.connected)}`}>
-        <div class="item indicator" data-cy="search">
+        <div class="item indicator" data-cy="connection-status">
           <Icon.Network />
         </div>
       </Tooltip>
     {:else if status.type === StatusType.Syncing}
       <Tooltip
         value={`Syncing with ${peerCount(status.syncs)} to get new content from your network`}>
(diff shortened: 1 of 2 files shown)


ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -111,9 +111,11 @@
     <li
```
juliendonck, line 111: @cloudhead I could wrap the tooltip around the `<li>` and change the `ul` and `li` to divs so the markup is correct. But it's complicated cause some of the markup is replicated in the `ConnectionStatusIndicator` component.

juliendonck:
The last commit move the tooltip to outside the sidebar. I can live with this, don't wanna overcomplicate the tooltip logic for this one use case.

ui/DesignSystem/Component/ConnectionStatusIndicator.svelte
```diff
@@ -18,45 +18,60 @@
...
   }
 </style>
 
 <Remote {store} let:data={status}>
   <div>
     {#if status.type === StatusType.Online}
       <Tooltip value={`You’re connected to ${peerCount(status.connected)}`}>
-        <!-- svelte-ignore a11y-missing-attribute -->
-        <a>
+        <div class="item indicator" data-cy="search">
```
MeBrei, line 53:
```suggestion
        <div class="item indicator" data-cy="connection-status">
```

also for the changes below

MeBrei (approved):
Looking good!

xla (approved):
💶 👝 👜 🕸

rudolfs (approved):

PR #1434, opened 2020-12-01
fix(ui): open only one modal at a time

### LINKED ISSUE #1417: shortcuts: when navigating to settings via shortcuts it doesn't hide modal
When having a modal open and the user navigates to the settings page via the shortcut the modal is not closed.

MeBrei:
Closes #1417 

Any opened modal or the design system guide will be closed if another one is opened via shortcuts.

### DIFF
--- a/ui/Hotkeys.svelte
+++ b/ui/Hotkeys.svelte
@@ -13,9 +13,15 @@
       pop();
     }
     push(destination);
+    modal.hide();
   };
 
-  const toggleModal = (destination: string) => modal.toggle(destination);
+  const toggleModal = (destination: string) => {
+    if (path.active(path.designSystemGuide(), $location)) {
+      pop();
+    }
+    modal.toggle(destination);
+  };
 
   const onKeydown = (event: KeyboardEvent) => {
     const modifierKey = isMac ? event.metaKey : event.ctrlKey;
@@ -47,13 +53,6 @@
     });
 
     if (!shortcut) return;
-
-    // Close any opened modal, before openening the next one.
-    modal.hide();
-    // Close the design system guide if opnened
-    if (path.active(path.designSystemGuide(), $location)) {
-      pop();
-    }
 
     switch (shortcut.key) {
       case hotkeys.ShortcutKey.Help:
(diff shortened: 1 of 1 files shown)


ui/Hotkeys.svelte
```diff
@@ -48,6 +48,13 @@
 
     if (!shortcut) return;
 
+    // Close any opened modal, before openening the next one.
+    modal.hide();
```
xla, line 52: This leads to a slight flicker, it might be more effective to move this logic into the toggle and only hide the modal if the new destination isn't one.

xla (changes requested):

ui/Hotkeys.svelte
MeBrei: For me the flicker only happens, when I have a modal opened and the switch to a fullscreen (settings or design system guide). I don't see a flicker when switching between modals.
For the case modal-fullscreen case I guess we have to decide on the order, either close modal than navigate (that is how it is now) or first navigate and then close the modal.

xla: Sorry the case where we navigate to a fullscreen page is the case I meant. The second option you layed out seems reasonable and is what I tried to say with checking if it is a modal that we navigate to before hiding.

xla (approved):
🎓 👆 😊 ☁️

rudolfs (approved):
Looks good to me.

One thing that we could maybe improve is the settings toggle.

To me, it makes sense for modals (fullscreen like the design system or overlays), but having the settings screen toggleable always feelt weird.

If I press <kbd>⌘</kbd>+<kbd>,</kbd> it jumps between settings and the previous screen I was on.

Not sure if it's in the scope of this PR though.

MeBrei:
> Looks good to me.
> 
> One thing that we could maybe improve is the settings toggle.
> 
> To me, it makes sense for modals (fullscreen like the design system or overlays), but having the settings screen toggleable always felt weird.
> 
> If I press ⌘+, it jumps between settings and the previous screen I was on.
> 
> Not sure if it's in the scope of this PR though.

Moved it into a new issue #1435

PR #1450, opened 2020-12-03
fix(ui): revert to default after untrack

### LINKED ISSUE #1210: Untrack selected peer
Setup: I have tracked a peer and they appear in my manage remotes modal and peer selector.

1. Select this peer in the project view
2. Open manage remotes and untrack them

The selector will stay on the peer and you will still see their project. The only reason we still see their project is because we do not prune refs in librad (yet, it is implemented in `next`).

Note that the manage remotes modal does correctly update.

When I untrack a peer while I've selected them the view stays on them.

xla:
If the untracked peer/remote is the currently selected this change
reverts back to the default one, to avoid confusion.

Fixes #1210

### DIFF
--- a/ui/src/screen/project.ts
+++ b/ui/src/screen/project.ts
@@ -193,8 +193,21 @@ export const addPeer = async (
   return true;
 };
 
-export const removePeer = (projectId: Urn, remote: PeerId): void => {
-  untrackPeer(projectId, remote);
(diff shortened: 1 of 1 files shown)


rudolfs (approved):
Bye 🐞 !

NunoAlexandre (approved):

MeBrei (approved):

PR #1453, opened 2020-12-03
fix(ui): avoid notify on internal request handling

### LINKED ISSUE #1424: Error message `The user aborted a request` when adding a remote
When I added a remote I got an error twice even though everything worked as expected.
Error: `The user aborted a request`

Screen Shot 2020-11-30 at 13 08 36

xla:
Closes #1424

### DIFF
--- a/ui/src/error.ts
+++ b/ui/src/error.ts
@@ -24,6 +24,7 @@ export enum Code {
   ProjectCreationFailure = "ProjectCreationFailure",
   ProjectRequestFailure = "ProjectRequestFailure",
   RemoteStoreError = "RemoteStoreError",
+  RequestAbortError = "RequestAbortError",
   SessionFetchFailure = "SessionFetchFailure",
   SessionSettingsUpdateFailure = "SessionSettingsUpdateFailure",
   UnhandledError = "UnhandledError",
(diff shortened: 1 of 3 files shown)


NunoAlexandre (approved):

ui/src/error.ts
```diff
@@ -38,8 +39,14 @@ export const fromException = (exception: globalThis.Error): Error => {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const details = Object.assign({}, exception) as any;
   details.name = exception.name;
+
+  let code = Code.UnknownException;
+  if (details.name === "AbortError") {
```
rudolfs, line 44: What is this `AbortError`? And where does it come from?

xla: The [`AbortError`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException#exception-AbortError) is part of the AbortController API which can be used to abort a `fetch`. We use that to avoid request amplification and out-of-order returns leading to invalid state.

MeBrei (approved):
How can this be tested, ie how can I reproduce the error on master?

The changes look good!

xla:
@MeBrei On a project with multiple remotes, remove and readd one of them. Sooner or later the errors will pop up. Alternatively you can inject a custom AbortController signal into a request thay supports it like fetchPeers and abort it manually.

geigerzaehler (approved):

cloudhead (approved):

xla:
Done in 05bd5aa9

PR #1463, opened 2020-12-05
feat: thank you next

### LINKED ISSUE #1078: Upgrade to tokio v1.0
Reminder that [tokio v0.3.0](https://github.com/tokio-rs/tokio/releases/tag/tokio-0.3.0) introduces [cancellable JoinHandle](https://github.com/tokio-rs/tokio/pull/2474)s, which would obsolete [`SpawnAbortable`](https://github.com/radicle-dev/radicle-upstream/blob/master/proxy/coco/src/spawn_abortable.rs).

Other changes may or may not be compatible, and may or may not be immediate improvements.

### LINKED ISSUE #1163: SpawnAbortable does not release resources immediately when aborted or dropped
Our `SpawnAbortable` implementation does not release the resource it holds immediately when it is aborted or dropped. Instead, resources are only released at a later point when the task executor polls the abortable task a final time.

For us this causes concrete issues with regard to the `kv::Store`: `coco::peer::Subroutines` holds a reference to a `kv::Store` which in turn holds a file lock on the database on disk. When restarting the proxy service `api::process::run_rigging` returns and drops the task that is running the peer. Because of nested `SpawnAbortable`s this does not drop the `kv::Store` reference before `run_rigging` returns. This means that in next iteration of the loop in `api::process::run` when `rig` is called the old `kv::Store` has still not been  dropped and creating a new `kv::Store` instance will fail because of the file lock. This is mitigated by using `tokio::time::delay_for` in the `run` loop.

We’ll have to investigate if this is fixed by using tokios native implementation in v0.3. (See #1078)

### LINKED ISSUE #1226: Delete coco::user
As we move over to the new Identity dos we don't need this band-aid anymore.

### LINKED ISSUE #1236: Migrate coco::state to identity docs
When migrating to the `next` branch the module that will be affected the most is the `coco::state` module. A lot of the functionality there refers to the `Entity` code. 

While migrating we should try to remove as many of the wrapper functions as possible, e.g. `listen_addr`, `peer_id`.

Since the API for `librad` has changed so that a lot of the functionality is split by purpose the `State` code will need to change as well. It would be a good opportunity to try to follow suit and have separated modules for the functionality, i.e. `replicate`, `track`, `user`, `project`.

xla:
Closes #1236 
Closes #1226 

**Edit**: Closes #1163
**Edit**: Closes #1078


NunoAlexandre:
@FintanH @xla given the size of this PR, to ease the review, I would like to know if you made any technical decisions worth noting or what kind if interesting problems you've faced :)

FintanH:
> given the size of this PR, to ease the review, I would like to know if you made any technical decisions worth noting or what kind of interesting problems you've faced :)

Good thinking, @NunoAlexandre. So I would sum up the majority of this PR as being mechanical changes. Mostly revolving around the new data types `Project`, `Person`, and `LocalIdentity` which replace the old entity data types `Project` and `User`.

Some interesting changes that needed to be made were involving the working copy creation (checkout and create). The interactions with git -- known as local transport because they are local interactions between the monorepo and a working copy -- are all done through `Remote`s now. This removes the need for registering local transports and waiting on results.

The removal of cypress tests is due to some interactions being idempotent (taking the same action twice has no effect). So creating the same project or person twice does nothing, and the same goes for checking out a project.

One notable change in the subroutine code was that @xla added an error variant for the default owner having to be set. This is because a lot of interactions need a default owner. Aside, we'd like to explore a more correct-by-construction approach to that post PR. 

That's all I can think of for now, but if anything comes up I'll add it in a follow-up comment.

NunoAlexandre:
> > given the size of this PR, to ease the review, I would like to know if you made any technical decisions worth noting or what kind of interesting problems you've faced :)
> 
> Good thinking, @NunoAlexandre. So I would sum up the majority of this PR as being mechanical changes. Mostly revolving around the new data types `Project`, `Person`, and `LocalIdentity` which replace the old entity data types `Project` and `User`.
> 
> Some interesting changes that needed to be made were involving the working copy creation (checkout and create). The interactions with git -- known as local transport because they are local interactions between the monorepo and a working copy -- are all done through `Remote`s now. This removes the need for registering local transports and waiting on results.
> 
> The removal of cypress tests is due to some interactions being idempotent (taking the same action twice has no effect). So creating the same project or person twice does nothing, and the same goes for checking out a project.
> 
> One notable change in the subroutine code was that @xla added an error variant for the default owner having to be set. This is because a lot of interactions need a default owner. Aside, we'd like to explore a more correct-by-construction approach to that post PR.
> 
> That's all I can think of for now, but if anything comes up I'll add it in a follow-up comment.

Thank you, Fintan! Will give it a review tomorrow morning if that works for you.

rudolfs:
Looking real good!

Gave it a spin already to familiarise myself with the new backend and noticed a couple of things:

- we should probably remove the default seed node address from settings ` hynewpywqj6x4mxgj7sojhue3erucyexiyhobxx4du9w66hxhbfqbw@seedling.radicle.xyz:12345` so it doesn't try to connect to it, because seedling still runs the old backend code
- something is off with checkouts, steps to repro:
  1. create a new project
  2. via the UI check out the project source to a different folder
  3. via the terminal cd into the project
  4. expected to be able to do `git log` and see the initial commit, instead got `fatal: your current branch 'master' does not have any commits yet` -- it seems like the project is created with `main` as the default branch, but the checkout defaults to `master`
- there's an issue with showing a replicated project, steps to repro:
  1. set up 2 machines on a local network, let's call them machine A and machine B
  2. add A's peer address to B's seed nodes in B's settings (`PEER_ID@IP:PORT`), wait until the network icon show a 1 connection
  3. create a new project on machine A and note the project URN
  4. on machine B search for the URN and click "Follow"
  5. once the project is replicated on machine B, navigate to it
  6. expected to see the project contents, instead an error message is shown; if we select the maintainer's view in the repo selector, we can see the repo contents just fine
```
{
"code": "RemoteStoreError",
"message": "Incorrect input",
"source": {
"code": "UnknownException",
"message": "Incorrect input",
"stack": "Error: Incorrect input\n    at http (file:///Users/rudolfs/work/radicle-upstream-2/public/bundle.js:3457:19)\n    at async Promise.all (index 0)\n    at async fetch$6 (file:///Users/rudolfs/work/radicle-upstream-2/public/bundle.js:58889:31)",
"details": {
"response": {},
"name": "Error"}}}
``` 

own
maintainers-view

proxy/coco/tests/working_copy.rs
```diff
@@ -0,0 +1,57 @@
+use coco::RunConfig;
+
+use pretty_assertions::assert_eq;
+
+#[macro_use]
+mod common;
+use common::{build_peer, init_logging, shia_le_pathbuf};
+
+#[tokio::test]
+async fn upstream_for_default() -> Result<(), Box<dyn std::error::Error>> {
```
NunoAlexandre, line 10: Sorry for my ignorance, but I don't quite understand what this is testing. The name would suggest that "upstream" would be tested as a project's default branch but that doesn't seem to be it.

NunoAlexandre (commented):
Looks solid overall!

Some questions/thoughts:

> The removal of cypress tests is due to some interactions being idempotent (taking the same action twice has no effect). So creating the same project or person twice does nothing, and the same goes for checking out a project.

- This is great to have correct by design! However, I'm wondering if it wouldn't be better to have the tests in place to validate such property?

- What is the rationale behind the name `LocalIdentity`? I see that wraps a `VerifiedPerson`, which seems rather easier and less obscure to understand. Understanding it might help me build a better mental model.

proxy/api/src/http.rs
```diff
@@ -112,8 +112,8 @@ fn with_owner_guard(ctx: context::Context) -> BoxedFilter<(coco::user::User,)> {
                 .state
                 .get_user(session.identity.urn)
                 .await
-                .expect("unable to get coco user");
-            let user = coco::user::verify(user).expect("unable to verify user");
+                .expect("failed to get local identity")
+                .expect("the local identity is missing");
```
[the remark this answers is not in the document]
kim: Haha. The first one is unwrapping a `Result`, the second an `Option` -- it is actually quite accurately worded, as the first is some kind of error, while a `None` value means "not found".

kim:
>     * What is the rationale behind the name `LocalIdentity`? I see that wraps a `VerifiedPerson`, which seems rather easier and less obscure to understand

A `LocalIdentity` is not just any person, but one that was created using the local keypair, or has a verifiable delegation to the local key, **and** has been selected as the identity to use for a certain operation (you can have any number of identities associated with the local keypair).

NunoAlexandre:
> A `LocalIdentity` is not just any person, but one that was created using the local keypair, or has a verifiable delegation to the local key, **and** has been selected as the identity to use for a certain operation

That makes sense reading, but I am specifically talking about the type definition:

``` rust
pub struct LocalIdentity(VerifiedPerson);

// where
pub type VerifiedPerson = VerifiedIdentity<PersonDoc>;
```

> (you can have any number of identities associated with the local keypair).

Having the type definitions in mind, I am not sure how you mean this or how the two come together, as I'd assume from that sentence that `LocalIdentity` would be a union type, not exclusively a `VerifiedPerson`.

kim:
@NunoAlexandre `LocalIdentity` is a newtype with a "smart constructor", if you will. You can only construct it such that the aforementioned invariants hold (one of which is the verification status, which is captured in the type system), or by unsafe coercion. What's unclear about that?

> [...] as I'd assume from that sentence that LocalIdentity would be a union type, not exclusively a VerifiedPerson.

Why is that? A union of what? Look:

From the local perspective, there is one and only one keypair to use with `radicle-link`. Associating an "identity" (that is, a blob of data containing some metadata) with that key is an explicit interaction -- it is not even required that you do this, it is perfectly fine to work with an "anonymous" key. 

Likewise, there is nothing which prevents you (or anyone else) from creating multiple identity documents and adding that key to it, for example on another device -- to prevent someone else from impersonating as you, there is the additional constraint that the `LocalIdentity` must be signed by the local key, which I forgot to mention.

It is not that you gain anything by doing this -- after all, anyone can correlate them by the key, so it doesn't make sense to e.g. create "nuno" and "nuno-work" with the expectation that this would be privacy-preserving. However, if we were to enforce on the protocol level that any given public key can appear in one and only one identity, every peer would have to keep track of the mapping key -> identity, and if a duplicate is detected... do what? Stop replicating anything signed by that key?

You see that this would create additional complexity with dubious benefits -- for example, how do you bubble up to the user that such a conflict was detected, and therefore some data which they'd expect to be there is not there? How do you recover from such a situation?

Since it is not **harmful** per-se to have multiple personalities (it may only confuse the people you interact with), we thus allow key -> identity to have a `0..*` cardinality.

NunoAlexandre:
> @NunoAlexandre `LocalIdentity` is a newtype with a "smart constructor", if you will. You can only construct it such that the aforementioned invariants hold (one of which is the verification status, which is captured in the type system), or by unsafe coercion. What's unclear about that?
> 
> > [...] as I'd assume from that sentence that LocalIdentity would be a union type, not exclusively a VerifiedPerson.
> 
> Why is that? A union of what? Look:
> 
> From the local perspective, there is one and only one keypair to use with `radicle-link`. Associating an "identity" (that is, a blob of data containing some metadata) with that key is an explicit interaction -- it is not even required that you do this, it is perfectly fine to work with an "anonymous" key.
> 
> Likewise, there is nothing which prevents you (or anyone else) from creating multiple identity documents and adding that key to it, for example on another device -- to prevent someone else from impersonating as you, there is the additional constraint that the `LocalIdentity` must be signed by the local key, which I forgot to mention.
> 
> It is not that you gain anything by doing this -- after all, anyone can correlate them by the key, so it doesn't make sense to e.g. create "nuno" and "nuno-work" with the expectation that this would be privacy-preserving. However, if we were to enforce on the protocol level that any given public key can appear in one and only one identity, every peer would have to keep track of the mapping key -> identity, and if a duplicate is detected... do what? Stop replicating anything signed by that key?
> 
> You see that this would create additional complexity with dubious benefits -- for example, how do you bubble up to the user that such a conflict was detected, and therefore some data which they'd expect to be there is not there? How do you recover from such a situation?
> 
> Since it is not **harmful** per-se to have multiple personalities (it may only confuse the people you interact with), we thus allow key -> identity to have a `0..*` cardinality.


We are discussing this at different levels.

My confusion comes from a place were, given that "Identity", in Link, can take the form of a Person, a Project, etc, calling `LocalIdentity` to a `VerifiedPerson` seems confusing. So I was trying to understand if there was a reason for that broad scope in the naming when, at the type level, and as you said, it can only hold a "Person" that meets the mentioned criteria.

Using an analogy to ease this out, say that we'd have the notion of `Shape = Circle Float | Rectangle Float Float` (standing for "Identities") and we'd have `SpecialRectangle` (standing for `VerifiedPerson`) that represents a `Rectangle` whose _length == π*width_. Now, we'd have `LocalShape(SpecialRectangle)`. That's where my confusion emerged: if this can only be a `SpecialRectangle`, why the broad naming.

proxy/api/src/http.rs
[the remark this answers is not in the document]
NunoAlexandre: ah that is right :D

kim:
> Shape = Circle Float | Rectangle Float Float

Your confusion seems to be that `VerifiedPerson` has the broader meaning. There is a partial isomorphism between `LocalIdentity` and `VerifiedPerson`: a `LocalIdentity` can always be converted to a `VerifiedPerson`, but not the other way round. In your example, circles and rectangles are of the same type, but verified and local identities are not -- you must prove that the identity you got is a local one (it's a refinement type in that sense).

`Person`s and `Project`s unify elsewhere, but there is no need for unifying `Person` and `LocalIdentity` (because of the prism).

kim:
Oh I get it: what throws you off is "identity". If the type was called `LocalPerson`, it would check out with you. I think at that level, I would just recommend import aliasing :shrug:

NunoAlexandre:
> Oh I get it: what throws you off is "identity". If the type was called `LocalPerson`, it would check out with you. I think at that level, I would just recommend import aliasing 🤷

Exactly :) it's not a big deal, I wanted to understand if there was more it that I should learn about. And there was :) thanks for bearing!

proxy/coco/tests/working_copy.rs
FintanH: > Sorry for my ignorance, but I don't quite understand what this is testing

Perfectly valid question.

There's a concept of an upstream or tracking branches in git -- see here https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches.

We're confirming the behaviour of setting the upstream branch since we had to work around `git2` in this scenario and write to the git config directly.

proxy/coco/tests/working_copy.rs
```diff
@@ -0,0 +1,57 @@
...
+        .map_err(radicle_surf::vcs::git::error::Error::from)?;
+    let remote = repo.branch_upstream_remote("refs/heads/it")?;
+
+    assert_eq!(remote.as_str().unwrap(), "rad");
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn can_checkout() -> Result<(), Box<dyn std::error::Error>> {
```
FintanH, line 34: @NunoAlexandre you mentioned validating the properties, here is the relevant one for `radicle-upstream`. For project and person creation, these are properties in `librad` and I believe are tested there.

proxy/coco/tests/working_copy.rs
NunoAlexandre: Ah, I see! Cool, the test now makes sense to me :) thank you!

FintanH:
> we should probably remove the default seed node address from settings hynewpywqj6x4mxgj7sojhue3erucyexiyhobxx4du9w66hxhbfqbw@seedling.radicle.xyz:12345 so it doesn't try to connect to it, because seedling still runs the old backend code

Good shout :+1: 

> something is off with checkouts, steps to repro: 

So the default `HEAD` value always seems to be `refs/heads/master`. It's a simple fix to set the head to our default branch.

> there's an issue with showing a replicated project, steps to repro: 

We weren't going far enough when checking that a peer is a contributor. We were only checking that `rad/signed_refs` exists, however, we should have been checking that the `heads` in there were empty if `rad/signed_refs` did exist.

All fixes for those issues have been pushed now :)

proxy/coco/src/peer/subroutines.rs
```diff
@@ -349,28 +354,36 @@ async fn query(urn: RadUrn, state: State, mut sender: mpsc::Sender<Input>) {
...
+        .send(Input::Request(input::Request::Cloning(
+            urn.clone(),
+            remote_peer,
+        )))
         .await
         .ok();
 
-    match state.clone_project(url.clone(), None).await {
+    match state
+        .clone_project(urn.clone(), remote_peer, None, None)
```
FintanH, line 367: @xla: I opted for defaulting the `fetch::Limit` -- introduced here https://github.com/radicle-dev/radicle-upstream/pull/1463/commits/ddb6dae3d973151a07061bcabcf8cffde7fcca22 -- but it would be nice to be able to change this at runtime as well.

Any thoughts on how this could be injected into the subroutines at runtime? Another channel perhaps?

proxy/coco/src/project/create/validation.rs
```diff
@@ -239,13 +259,15 @@ impl Repository {
                 let path = path.join(name);
                 log::debug!("Setting up new repository @ '{}'", path.display());
                 let repo = Self::initialise(path, description, &default_branch)?;
+                log::debug!("REPO PATH: {}", repo.path().display());
```
cloudhead, line 262: :eyes:

proxy/coco/src/peer/subroutines.rs
xla: Could you expand on what potential scenarios look like that would constitute runtime configuration changes? The reason I'm asking is that neither this comment or the original commit message reason about the why such a capability is desried.

> Any thoughts on how this could be injected into the subroutines at runtime? Another channel perhaps?

Yeah usually configuration changes over time would come through some form of channel. For the configured seeds we use a `watch` channel to provide last value semantics. Would need to look deeper into this use-case to give proper recommendation. There might be a general need for a latest configuration made to be known to the internals of the peer.

FintanH: So a use case would be the following scenario.

We ask to clone the Linux project, but as we're doing so the fetcher throws an error and says we've exceeded the configured limits. We're sure that this really is the Linux project and not a bunch of spammy data, so we configure our limits to go higher, try the replication again, and once we've got the project we can reset the limit.

Does that help shed some light on the question here?

FintanH:
Note to self:
* Tracking and finding a user didn't add them to the include file
* When creating a project there's a message saying the upstream is gone

FintanH:
> When creating a project there's a message saying the upstream is gone

Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1463/commits/bcad1d925e9efab80bcc76bfcba886bee7c3cd9f

proxy/coco/src/peer/subroutines.rs
xla: - [ ] Somewhat. Will the limit go down again after the linux kernel has been replicated from `pc-torvalds01`. By the sound of it this has to do with trust and spam protection. Which seems odd to be bound to runtime configuration changes. As in where would the signal come from that the fetch limit can be increased. I don't want to challenge the scenario too specifically, just curious to see how the concerns align

FintanH: > Somewhat. Will the limit go down again after the linux kernel has been replicated from pc-torvalds01.

Ya you would want to turn it down again.

> By the sound of it this has to do with trust and spam protection.

Exactly, it's exactly a trust mechanism.

> Which seems odd to be bound to runtime configuration changes. As in where would the signal come from that the fetch limit can be increased.

As @kim and I deliberated on it, we landed on "sane" default values and should be configurable depending on the scenario -- giving the malkovich the final say when using `replicate`.

How are you thinking about it?

rudolfs:
Did a small test run with 2 laptops and setzling and bumped into some issues.
I'm not sure whether they're known issues or something new, thought it would be worth mentioning.

Steps:
  - onboarded both machines
  - created a project on machine A
  - saw the project appear on setzling
  - then replicated the project on machine B and did a local checkout via the `Fork` button
  - created a new commit on machine A and pushed it to the `rad` remote
  - commit appeared in the project on machine A
  - waited around 5 mins, but the commit didn't appear on machine B
  - added machine B as a remote to the project on machine A, waited 5 min, the remote was still in the waiting area
  - restarted Upstream on both machines, waited 5 min, but the commit still didn't appear on machine B, however the remote showed up as added on machine A
  - on machine A selected machine B in the project's peer selector and got an error

<details>
<summary>
Error messages
</summary>

UI:
```
{
  "code": "RemoteStoreError",
  "message": "Incorrect input",
  "source": {
    "code": "UnknownException",
    "message": "Incorrect input",
    "stack": "Error: Incorrect input\n    at http (file:///Users/rudolfs/work/radicle-upstream/public/bundle.js:1991:19)\n    at async Promise.all (index 0)\n    at async fetch$6 (file:///Users/rudolfs/work/radicle-upstream/public/bundle.js:143163:31)",
    "details": {
      "response": {},
      "name": "Error"
    }
  }
}
```
Proxy:
```
 ERROR api::http::error         > Rejection(State(MissingRef { reference: Reference { remote: Some(PeerId(PublicKey(VerificationKeyBytes("d5b92a9d8d6c335f639a88f01193c296f3d8ec868f684dd04e524ce19ddd9499")))), category: Heads, name: RefLike("master"), namespace: Some(Namespace(Urn { id: Oid(783793840e886b19fc8d301df4be190bc9ce96dc), path: None })) } }))
```
Screenshot 2021-02-04 at 09 26 38
</details>

<details>
<summary>
Machine A monorepo:
</summary>

```
Application Support/xyz.radicle.radicle-link » tree
.
├── active_profile
└── d9a6c997-c481-4e0b-888f-6b0e30792458
    ├── git
    │   ├── FETCH_HEAD
    │   ├── HEAD
    │   ├── config
    │   ├── description
    │   ├── hooks
    │   │   └── README.sample
    │   ├── info
    │   │   └── exclude
    │   ├── objects
    │   │   ├── 04
    │   │   │   └── 612ea1862481ac2b936ca2d6f482709859ef72
    │   │   ├── 1a
    │   │   │   └── 7f34ee9e09d604630292973a41e09a581f99d7
    │   │   ├── 24
    │   │   │   ├── 25d9839bbaac941047038c6b2e00d8d90e4461
    │   │   │   └── 2ba95a2c8b60144e7d27d1628e127b042b12fe
    │   │   ├── 29
    │   │   │   └── ae94c5cb1201b85ae6e054f8f7fbe4b92dd67a
    │   │   ├── 32
    │   │   │   └── fe621707c650f6f6e302f90c6911653f87e4a7
    │   │   ├── 33
    │   │   │   └── f428753e20bf09b16a72bc27ff162385d1572f
    │   │   ├── 3e
    │   │   │   └── 640a3d6cb7d114e532d20f08999185dd61eebd
    │   │   ├── 45
    │   │   │   └── 9113266b6690a9086c929f8cedf3f34baa48fe
    │   │   ├── 68
    │   │   │   └── 64c20f006de8f2ee3f8129de8d68690549565a
    │   │   ├── 69
    │   │   │   └── 304ea4b1b500d8ceccccb96ac23e2b12bb87eb
    │   │   ├── 6f
    │   │   │   └── bea447d0c202a3153c612b0cc6bd8c7a1d50bb
    │   │   ├── 78
    │   │   │   └── 3793840e886b19fc8d301df4be190bc9ce96dc
    │   │   ├── 7e
    │   │   │   └── d378e2fdfeebc3e684ad31b7d46cc7423047fb
    │   │   ├── 88
    │   │   │   └── 3b84e17bd280fa83c4e06ac80a787b3c6ef4ac
    │   │   ├── 8b
    │   │   │   └── d6bd0bdca19938d09936cf0ec94549e4ed47b2
    │   │   ├── 8c
    │   │   │   └── f95616a22dc7e90777d5730487d15d669a0459
    │   │   ├── 90
    │   │   │   └── 8d891149eb1ad2e288bbb7259b5f2f3e02951b
    │   │   ├── 97
    │   │   │   ├── 70fc68874662b0709fc43ddf6d21477865c7dc
    │   │   │   └── d0a8d8cf2491889e5a142f71f42ae2fbd15cdd
    │   │   ├── a3
    │   │   │   └── 949341992a6853e93eeec8c677e03f23f1b5c8
    │   │   ├── ad
    │   │   │   └── 74a118d941897401e596ceb9b095a49d34e155
    │   │   ├── cb
    │   │   │   └── 2438b389488201a3030e33dc7403923cb8a95a
    │   │   ├── cf
    │   │   │   └── e4ff04675474f90498cf30933148ff9b0fc60b
    │   │   ├── d1
    │   │   │   └── 9e047c534404193de7514263fcdfde58fbe6cb
    │   │   ├── db
    │   │   │   └── a885d983566603f023f51ec4cb865785d6e4d2
    │   │   ├── dd
    │   │   │   └── ba3c73d2889f9324af5b254ce96fd89e934bc9
    │   │   ├── e9
    │   │   │   └── 023890699d385c232e4aaa5e5b485dc2328c92
    │   │   ├── ee
    │   │   │   └── b010f312eabfabfe80b390d66abe0a39f930b9
    │   │   ├── f3
    │   │   │   └── 3b283de52fd955fcd48368d562934a5fce0991
    │   │   ├── info
    │   │   └── pack
    │   │       ├── pack-1f080b0f30be160399a894e8d87639d3d54780c5.idx
    │   │       ├── pack-1f080b0f30be160399a894e8d87639d3d54780c5.pack
    │   │       ├── pack-c8a58d53531d768a77c318b233acd3099cb501ea.idx
    │   │       └── pack-c8a58d53531d768a77c318b233acd3099cb501ea.pack
    │   └── refs
    │       ├── heads
    │       ├── namespaces
    │       │   ├── hnrk8ophuoo8eo4a391guy8xwzacoz1qq15qy
    │       │   │   └── refs
    │       │   │       ├── heads
    │       │   │       │   └── master
    │       │   │       ├── rad
    │       │   │       │   ├── id
    │       │   │       │   ├── ids
    │       │   │       │   │   └── hnrknejq3oqp5imrwnbdo8ddmfayptseqetoo
    │       │   │       │   ├── self
    │       │   │       │   └── signed_refs
    │       │   │       └── remotes
    │       │   │           ├── hydddxdbqoj6azkq39637eso6inuxw99tpo9h8aqgnkracfaufy9jn
    │       │   │           │   └── rad
    │       │   │           │       ├── id
    │       │   │           │       ├── ids
    │       │   │           │       │   └── hnrknejq3oqp5imrwnbdo8ddmfayptseqetoo
    │       │   │           │       ├── self
    │       │   │           │       └── signed_refs
    │       │   │           └── hydk51kw7tisdgz5dukrxyrcuakmx8s8co48souqoj3jr3ac75skj1
    │       │   │               └── rad
    │       │   │                   ├── id
    │       │   │                   ├── ids
    │       │   │                   │   └── hnrknejq3oqp5imrwnbdo8ddmfayptseqetoo
    │       │   │                   ├── self
    │       │   │                   └── signed_refs
    │       │   └── hnrknejq3oqp5imrwnbdo8ddmfayptseqetoo
    │       │       └── refs
    │       │           └── rad
    │       │               ├── id
    │       │               ├── self
    │       │               └── signed_refs
    │       └── tags
    ├── git-includes
    │   └── hnrk8ophuoo8eo4a391guy8xwzacoz1qq15qy.inc
    └── keys
        └── librad.key

56 directories, 59 files
```
</details>

<details>
<summary>
Machine B monorepo state:
</summary>

```
Application Support/xyz.radicle.radicle-link » tree
.
├── 59652cfc-8de1-4928-a4b5-78ad406d1dd5
│   ├── git
│   │   ├── FETCH_HEAD
│   │   ├── HEAD
│   │   ├── config
│   │   ├── description
│   │   ├── hooks
│   │   │   └── README.sample
│   │   ├── info
│   │   │   └── exclude
│   │   ├── objects
│   │   │   ├── 27
│   │   │   │   └── 94615c11b607b294e5ec78e9c8e6f44f6d2cf2
│   │   │   ├── 2c
│   │   │   │   └── 2b685f5af48b9c55dc19d2f5b83f1560d9f5dc
│   │   │   ├── 2d
│   │   │   │   └── 81d98333ba4677beb41d24dcce6235a396d32f
│   │   │   ├── 35
│   │   │   │   └── c241d67118d01dad47d188ea196b0751883b32
│   │   │   ├── 48
│   │   │   │   └── 0109e51ce02f9e21fc68e430e80df27e7172e4
│   │   │   ├── 4f
│   │   │   │   └── 2f0fe8786382cba3b849e1e92faeebf75907c8
│   │   │   ├── 64
│   │   │   │   └── dd30923ff40e49afcde6cc5cd326af9c526379
│   │   │   ├── 67
│   │   │   │   └── d03ae07efd4c844ea0ce41e4b59c0e252cd17b
│   │   │   ├── 80
│   │   │   │   └── 78c1aa2bacc9629f1da091bb6883fe658b1240
│   │   │   ├── 93
│   │   │   │   └── 51c27897f8032d4f339461933da8df7aa57937
│   │   │   ├── 9a
│   │   │   │   └── 652cb6b54a72a953d4cdecb3a73921d416a160
│   │   │   ├── a4
│   │   │   │   └── 8d107dc0eac5a3f91aefe1a36b321200e88eb6
│   │   │   ├── ab
│   │   │   │   └── bd01223a0db5daf819fcf4ed1909dfcd4c4d6e
│   │   │   ├── ad
│   │   │   │   └── 3af707da2a02437007c59877356e88bad5e1c9
│   │   │   ├── af
│   │   │   │   └── b7a993211ab388e4c05e9d40fdf4c56d8ae851
│   │   │   ├── b3
│   │   │   │   └── f0ffb9e2b475cd5056383b0fadd96f18a89629
│   │   │   ├── e0
│   │   │   │   └── 1b3b8422d66c713f42bbde050c8b09cbea799c
│   │   │   ├── e3
│   │   │   │   └── 8fc677d89fc5fa4e715e3ab9b50607f116ea19
│   │   │   ├── info
│   │   │   └── pack
│   │   │       ├── pack-8a0c7cf50ac057dc54f417ff5a3b945692d5cc08.idx
│   │   │       ├── pack-8a0c7cf50ac057dc54f417ff5a3b945692d5cc08.pack
│   │   │       ├── pack-a7e92c4dd37bf2459956377e47b6c6f40eeed56e.idx
│   │   │       ├── pack-a7e92c4dd37bf2459956377e47b6c6f40eeed56e.pack
│   │   │       ├── pack-c0a22e4f510e6368ea8abdf8e31c2485a35aaf84.idx
│   │   │       └── pack-c0a22e4f510e6368ea8abdf8e31c2485a35aaf84.pack
│   │   └── refs
│   │       ├── heads
│   │       ├── namespaces
│   │       │   ├── hnrk8ophuoo8eo4a391guy8xwzacoz1qq15qy
│   │       │   │   └── refs
│   │       │   │       ├── heads
│   │       │   │       │   └── master
│   │       │   │       ├── rad
│   │       │   │       │   ├── id
│   │       │   │       │   ├── ids
│   │       │   │       │   │   └── hnrknejq3oqp5imrwnbdo8ddmfayptseqetoo
│   │       │   │       │   ├── self
│   │       │   │       │   └── signed_refs
│   │       │   │       └── remotes
│   │       │   │           ├── hybz9gfgtd9d4pd14a6r66j5hz6f77fed4jdu7pana4fxaxbt369kg
│   │       │   │           │   └── rad
│   │       │   │           │       └── ids
│   │       │   │           └── hydddxdbqoj6azkq39637eso6inuxw99tpo9h8aqgnkracfaufy9jn
│   │       │   │               ├── heads
│   │       │   │               │   └── master
│   │       │   │               └── rad
│   │       │   │                   ├── id
│   │       │   │                   ├── self
│   │       │   │                   └── signed_refs
│   │       │   ├── hnrkey6gbiei431mnuhq4brp5pnb9h3cmnjyy
│   │       │   │   └── refs
│   │       │   │       └── rad
│   │       │   │           ├── id
│   │       │   │           ├── self
│   │       │   │           └── signed_refs
│   │       │   └── hnrknejq3oqp5imrwnbdo8ddmfayptseqetoo
│   │       │       └── refs
│   │       │           └── rad
│   │       │               └── id
│   │       └── tags
│   ├── git-includes
│   │   └── hnrk8ophuoo8eo4a391guy8xwzacoz1qq15qy.inc
│   └── keys
│       └── librad.key
└── active_profile

49 directories, 46 files
```
</details>

FintanH:
Seems like the seed isn't replicating `heads` when someone forks the project. This would explain @rudolfs error messages above.

proxy/api/src/http/source.rs
```diff
@@ -346,21 +323,19 @@ mod test {
...
         };
-        let path = "text/arrows.txt";
-        let default_branch = ctx.state.find_default_branch(urn.clone()).await?;
-        let want = ctx
-            .state
-            .with_browser(default_branch, |mut browser| {
-                coco::blob(&mut browser, Some(revision.clone()), path, None)
-            })
-            .await?;
+        let arrows = "text/arrows.txt";
```
FintanH, line 330: The renaming here was clippy giving out about name shadowing of `path`.

proxy/coco/src/peer.rs
```diff
@@ -48,21 +48,60 @@ pub enum Error {
...
+
+    Ok(peer)
 }
 
 /// Local peer to participate in the radicle code-collaboration network.
-pub struct Peer {
-    /// Peer [`librad::net::peer::RunLoop`] to advance the network protocol.
-    run_loop: RunLoop,
-    /// Underlying state that is passed to subroutines.
-    state: State,
```
FintanH, line 65: `net::peer::Peer` is the new `State`. It's essentially what was the `PeerApi`, but a simplified and less footgun-y version

proxy/coco/src/peer/run_state.rs
```diff
@@ -74,21 +84,27 @@ impl MaybeFrom<&Input> for Event {
             Input::Announce(input::Announce::Succeeded(updates)) => {
                 Some(Self::Announced(updates.clone()))
             },
-            Input::Peer(event) => match event {
-                PeerEvent::GossipFetch(FetchInfo {
```
FintanH, line 78: There's no `PeerEvent` anymore so we moved this logic to here https://github.com/radicle-dev/radicle-upstream/pull/1463/files#diff-463e58bd670bb12315d14a2f26fbfc88e97a22551753a25ecd20fe3056fe9de3R89

proxy/coco/src/peer/run_state.rs
```diff
@@ -112,10 +128,12 @@ pub enum Status {
     /// Phase where the local peer tries get up-to-date.
     #[serde(rename_all = "camelCase")]
     Syncing {
-        /// Number of completed syncs.
-        synced: usize,
-        /// Number of synchronisation underway.
-        syncs: usize,
+        /// The set of failed syncs.
```
FintanH, line 131: We now keep track of the `PeerId`s we've interacted with for the syncing process on startup

proxy/coco/src/peer/run_state.rs
```diff
@@ -140,9 +158,10 @@ pub struct RunState {
     /// `Connected(Peer1) -> Connected(Peer1) -> Disconnecting(Peer1)`
     //
     // FIXME(xla): Use a `Option<NonEmpty>` here to express the invariance.
-    connected_peers: HashMap<PeerId, usize>,
```
FintanH, line 143: We no longer get messages about `Connected` and `Disconnected` so we don't need this `HashMap`. Instead, we can poll for the set of connected peers

proxy/coco/src/peer/run_state.rs
```diff
@@ -140,9 +158,10 @@ pub struct RunState {
...
+    connected_peers: HashSet<PeerId>,
     /// Current internal status.
     pub status: Status,
+    stats: net::protocol::event::downstream::Stats,
```
FintanH, line 164: `Stats` is what we can poll for `Peer` information, such as the connected peers

proxy/coco/src/peer/run_state.rs
```diff
@@ -254,165 +275,77 @@ impl RunState {
...
-                ..
-            }) => vec![Command::Include(urn)],
-            PeerEvent::GossipFetch(_) => vec![],
-        }
-    }
-
     /// Handle [`input::Sync`]s.
     fn handle_peer_sync(&mut self, input: &input::Sync) -> Vec<Command> {
-        if let Status::Syncing { synced, syncs } = self.status {
+        if let Status::Syncing {
```
FintanH, line 280: The logic is changing here because of the move from `usize` to `HashSet<PeerId>`

proxy/coco/src/peer/run_state.rs
```diff
@@ -254,165 +275,77 @@ impl RunState {
...
         vec![]
     }
 
     /// Handle [`ProtocolEvent`]s.
     #[allow(clippy::wildcard_enum_match_arm)]
-    fn handle_protocol(&mut self, event: ProtocolEvent<Gossip>) -> Vec<Command> {
+    fn handle_protocol(&mut self, event: ProtocolEvent) -> Vec<Command> {
         match (&self.status, event) {
-            // Go from [`Status::Stopped`] to [`Status::Started`] once we are listening.
-            (Status::Stopped { .. }, ProtocolEvent::Listening(_addr)) => {
```
FintanH, line 300: `Listening` doesn't exist anymore and it's now `Endpoint::Up`, with `Endpoint::Down` being its inverse

proxy/coco/src/peer/run_state.rs
```diff
@@ -254,165 +275,77 @@ impl RunState {
...
+            (Status::Stopped, ProtocolEvent::Endpoint(upstream::Endpoint::Up { .. })) => {
                 self.status = Status::Started;
                 self.status_since = SystemTime::now();
 
                 vec![]
             },
-            (state, ProtocolEvent::Connected(peer_id)) => {
```
FintanH, line 306: This all moved to `handle_stats` due to these events not existing anymore: https://github.com/radicle-dev/radicle-upstream/pull/1463/files#diff-463e58bd670bb12315d14a2f26fbfc88e97a22551753a25ecd20fe3056fe9de3R414

proxy/coco/src/state.rs
```diff
@@ -1,799 +1,878 @@
 //! Utility to work with the peer api of librad.
 
-use std::{convert::TryFrom as _, net::SocketAddr, path::PathBuf, sync::Arc, time::Duration};
+use std::{convert::TryFrom as _, net::SocketAddr, path::PathBuf};
```
FintanH, line 3: This file easily has the worst diff to review. We removed the `State` struct and made all the functions top-level. They now take `&net::peer::Peer` to interact with the underlying `Storage`. The changes were mechanical and the functions should be on par with what was there.

FintanH (commented):
@geigerzaehler @MeBrei @rudolfs I've left my inline summary of portions of the code that are probably worth more attention. Thank you and good luck :bow:

proxy/api/src/context.rs
```diff
@@ -204,21 +204,26 @@ impl Unsealed {
...
+            let coco_peer =
+                coco::bootstrap(config, disco, store.clone(), RunConfig::default()).await?;
+            let peer = coco_peer.peer.clone();
+
+            let peer_control = coco_peer.control();
+            // FIXME(finto): As we've seen `tokio::spawn` without explicit aborting of handling
+            // could mean we end up in resource locks. This is probably safe due to everything
+            // acting in a tmp dir, but it would be better to have something like `WithUnsealed`
+            // holding the handle which aborts when dropped.
+            tokio::spawn(coco_peer.into_running());
```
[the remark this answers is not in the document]
geigerzaehler: Thanks for pointing this out. I’ve created [an issue](https://github.com/radicle-dev/radicle-upstream/issues/1648) for this with more details. Feel free to remove the comment or link to the issue. Happy to provide more details if anything is unclear.

geigerzaehler:
I’m taking a look at everything in `proxy/coco`.

cypress/integration/project/checkout.spec.ts
```diff
@@ -106,25 +106,6 @@ context("project checkout", () => {
               expect(result.stdout).to.equal(`rad`);
             }
           );
-
-          // Make sure we can't check out a project to the same directory twice.
```
MeBrei, line 110: Is this removed because the behavior changed?

cypress/integration/project/creation.spec.ts
```diff
@@ -378,28 +378,6 @@ context("project creation", () => {
             .contains(`Project ${repoName} successfully created`)
             .should("exist");
           commands.pick("notification").contains("Close").click();
-
-          // Make sure we can't add the same project twice.
```
MeBrei, line 382: same question as above

cypress/integration/project/checkout.spec.ts
FintanH: We saw that `checkout` is idempotent, as confirmed by this test in `working_copy.rs` https://github.com/radicle-dev/radicle-upstream/pull/1463/files#diff-940a798aab5c5c9a4058c1e8c8b384b600a22e234ece99249c8bd028a5be74dbR44

rudolfs (commented):
I built the release dmg package on my macOS Big Sur machine via `yarn dist` and it seems to be broken due to some code signing issues in libraries linked in proxy. When I build the package on master, it just works, so something must have changed in our dependencies. 🤷 

Screenshot 2021-03-03 at 17 01 23

```
dyld: Library not loaded: /usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib
  Referenced from: /Applications/Radicle Upstream.app/Contents/Resources/radicle-proxy
  Reason: no suitable image found.  Did find:
	/usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib: code signature in (/usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib) not valid for use in process using Library Validation: mapped file has no cdhash, completely unsigned? Code has to be at least ad-hoc signed.
```

We should test our Linux AppImage package on this branch as well!

native/main.ts
```diff
@@ -31,13 +31,14 @@ if (isDev) {
...
+    "hybz9gfgtd9d4pd14a6r66j5hz6f77fed4jdu7pana4fxaxbt369kg@setzling.radicle.xyz:12345"
+  );
 } else {
   // Packaged app, i.e. production.
   proxyPath = path.join(__dirname, "../../radicle-proxy");
-  proxyArgs = [
-    "--default-seed",
-    "hynewpywqj6x4mxgj7sojhue3erucyexiyhobxx4du9w66hxhbfqbw@seedling.radicle.xyz:12345",
-  ];
+  proxyArgs = [];
```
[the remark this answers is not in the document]
FintanH: It's in my brain stack

geigerzaehler (commented):
I’m about 70% through. Here are my (minor) comments so far.

cypress/integration/project/checkout.spec.ts
FintanH: Thanks for the thoroughness @rudolfs :bow: So what's happening is that the project with the same name and description has a different URN to the same name and _no description_. This is because they have different hash outputs. So, with this in mind, it's probably best that we don't allow to use the same working copy for checkout and guard against it.

FintanH: Actually, this is a bit more sinister than I thought. Before we were able to tell if an `Entity` ended up with the same `URN`. But we can only know the URN once we create the project, which is idempotent. So we can't easily tell if the project was created or just didn't do anything. Need to think about that more.

kim: > Before we were able to tell if an `Entity` ended up with the same `URN`.

Wait what, how is this different now?

FintanH: Is it possible to determine the `Urn` before calling `create`? :thinking:

kim: In theory yes: https://github.com/radicle-dev/radicle-link/blob/d5f2e37cb6a751c6f8e18bc1c977ad27d4e9c8f5/librad/src/identities/git.rs#L485

FintanH: Right, was just looking at that :sweat_smile: I'll create a proof for theory tomorrow

proxy/coco/tests/working_copy.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    let _ = state::init_project(
+        &alice_peer,
+        &alice,
+        shia_le_pathbuf(alice_repo_path.clone()),
+    )
+    .await?;
+
+    let repo = git2::Repository::open(alice_repo_path.join("just"))
+        .map_err(radicle_surf::vcs::git::error::Error::from)?;
```
geigerzaehler, line 33: This has been confusing to me in existing tests, too. `alice_repo_path` is not the repo path but the path that the repo is created in. We could just use `alice_tmp_dir.path()` for that. It’s also a bit unclear where the `just` comes from. Maybe it would make sense to extract it from the return value of `shia_le_pathbuf` or excplicitly construct `project::Create` here.

proxy/coco/tests/working_copy.rs
```diff
@@ -0,0 +1,87 @@
...
+    assert_eq!(remote.as_str().unwrap(), "rad");
+
+    let branch = repo.find_branch("rad/it", git2::BranchType::Remote);
+    assert!(branch.is_ok(), "could not find `rad/it`");
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn can_checkout() -> Result<(), Box<dyn std::error::Error>> {
```
geigerzaehler, line 45: Wouldn’t `checkout_twice_fails` be a more descriptive name?

proxy/coco/tests/replication.rs
```diff
@@ -193,43 +220,46 @@ async fn can_fetch_project_changes() -> Result<(), Box<dyn std::error::Error>> {
...
+                &repo,
+                LocalPushspec::Matching {
+                    pattern: RefLike::try_from(format!(
+                        "refs/heads/{}",
+                        project.subject().default_branch.clone().unwrap()
+                    ))
+                    .unwrap()
+                    .into(),
+                    force: Force::False,
+                },
```
geigerzaehler, line 236: I’m not a hundred percent certain but using `LocalPushspec::Configured` would work too and would be a lot easier to read.

proxy/coco/tests/replication.rs
```diff
@@ -393,43 +460,61 @@ async fn can_create_working_copy_of_peer() -> Result<(), Box<dyn std::error::Err
     Ok(())
 }
 
+#[allow(clippy::needless_collect)]
```
geigerzaehler, line 463: Doesn’t seem like this case has a needless collect.

proxy/coco/tests/replication.rs
```diff
@@ -292,55 +333,70 @@ async fn can_create_working_copy_of_peer() -> Result<(), Box<dyn std::error::Err
...
 
     let commit_id = {
-        let alice_peer_id = alice_state.peer_id();
-        let path = bob_state
-            .checkout(project.urn(), alice_peer_id, bob_repo_path)
-            .await?;
+        let alice_peer_id = alice_peer.peer_id();
+        log::debug!("CHECKING OUT");
+        let path = state::checkout(&bob_peer, project.urn(), alice_peer_id, bob_repo_path).await?;
+        log::debug!("CHECKED OUT");
```
geigerzaehler, line 392: Should probably be removed.

geigerzaehler (commented):
Some more things I discovered

proxy/coco/tests/gossip.rs
```diff
@@ -34,21 +28,26 @@ async fn can_announce_new_project() -> Result<(), Box<dyn std::error::Error>> {
...
         .expect("unable to init project");
 
-    let announced = alice_events
-        .into_stream()
+    let announced = async_stream::stream! { loop { yield alice_events.recv().await } }
         .filter_map(|res| match res.unwrap() {
             coco::PeerEvent::Announced(updates) if updates.len() == 1 => future::ready(Some(())),
-            _ => future::ready(None),
+            res => {
+                println!("What? {:?}", res);
```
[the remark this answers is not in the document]
FintanH: _Probably_ :P

proxy/coco/tests/replication.rs
FintanH: It's been a while _but_ I _think_ I tried that. I could try again and see

FintanH: Didn't work, but I could make it look a bit nicer at least

proxy/coco/src/project/checkout.rs
```diff
@@ -66,78 +82,137 @@ pub enum Ownership {
...
-        }
+        let rad = {
+            // Create a fetchspec `refs/heads/*:refs/remotes/rad/*`
+            let fetchspec = Refspec {
+                src: GenericRef::<_, RefLike, _>::heads(Flat, None),
+                dst: refspec_pattern!("refs/remotes/rad/*"),
+                force: Force::True,
+            };
+            let mut rad = Remote::rad_remote(url, fetchspec);
+            rad.save(&repo)?;
```
geigerzaehler, line 203: Is it necessary to add the source as a remote in the git config? Isn’t this what the git includes are for?

proxy/coco/src/project/checkout.rs
```diff
@@ -206,15 +284,18 @@ where
...
-        let repo = ownership.clone(self.urn, &self.default_branch, &project_path, &mut builder)?;
+        let (repo, rad) =
+            ownership.clone(open_storage, self.urn, &self.default_branch, &project_path)?;
 
         // Set configurations
-        super::set_rad_upstream(&repo, &self.default_branch)?;
+        super::set_upstream(&repo, &rad, self.default_branch.clone())?;
         include::set_include_path(&repo, self.include_path)?;
+        repo.set_head(Qualified::from(self.default_branch).as_str())?;
+        repo.checkout_head(None)?;
```
geigerzaehler, line 298: I’m not sure this plays well with existing repos that have a dirty working tree, a different branch checked out, etc.

geigerzaehler (commented):
A couple more questions and suggestions.

proxy/coco/src/project.rs
```diff
@@ -16,12 +20,51 @@ pub use checkout::Checkout;
...
+/// Check that the `path` provided is either:
+///  * an empty directory
+///  * a non-existent directory
+/// If these checks pass then the path is returned as `Ok(Some(path))`.
+///
+/// If the `path` is a file or is a directory with contents, then it will return `Ok(None)`.
+///
+/// # Errors
+///   * I/O error in reading the directory contents
+pub fn ensure_directory(path: &Path) -> Result<Option<&Path>, io::Error> {
```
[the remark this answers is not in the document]
FintanH: This will be ripe for refactoring once it lives in radicle-link :)

proxy/coco/src/project/checkout.rs
```diff
@@ -66,78 +82,137 @@ pub enum Ownership {
...
+{
+    let repo = git2::Repository::init(path)?;
+    remote.save(&repo)?;
+    for (reference, oid) in remote.fetch(storage, &repo, LocalFetchspec::Configured)? {
+        let msg = format!("Fetched `{}->{}`", reference, oid);
+        log::debug!("{}", msg);
+
+        let branch: RefLike = OneLevel::from(reference).into();
+        let branch = branch.strip_prefix(remote.name.clone())?;
+        let branch = branch.strip_prefix(reflike!("heads")).unwrap_or(branch);
```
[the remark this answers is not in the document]
FintanH: Ya, you're right I think :+1: I'm going to leave a FIXME

proxy/coco/src/project/checkout.rs
```diff
@@ -66,78 +82,137 @@ pub enum Ownership {
...
+        let _remote_branch = repo.reference(
+            reflike!("refs/remotes")
+                .join(remote.name.clone())
+                .join(branch.clone())
+                .as_str(),
+            oid,
+            true,
+            &msg,
+        )?;
+        let _local_branch = repo.reference(Qualified::from(branch).as_str(), oid, true, &msg);
```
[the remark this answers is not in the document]
FintanH: Yes, and yes :) There's `into_qualified` if one wants to specify the category

proxy/coco/src/project/checkout.rs
FintanH: It's necessary for the checkout to perform correctly. It could be that we could use the `include` here though.

proxy/coco/src/project/checkout.rs
FintanH: It shouldn't get here if it's already a project though https://github.com/radicle-dev/radicle-upstream/pull/1463/files/cd99c84c65040be8565cd0b400655f877f88bf95#diff-3362a0bfada6f68cdf7b17ea16658dbbfe028d7ce27b96bf1a4fd6e37318b8c8R288

PR #1467, opened 2020-12-06
fix(ui): list border overflow (#1447)

### LINKED ISSUE #1447: Border radius glitch when hovering over projects in the "Still looking…" section on the Following tab of your profile
image

if you don't see it, it's happening on the corners. some inner element is probably not rounded 
image

finleygn:
Fixes #1447

Border radius of lists now just handled by the parent with overflow, rather than having border radius on children too.

image

### DIFF
--- a/ui/DesignSystem/Component/List.svelte
+++ b/ui/DesignSystem/Component/List.svelte
@@ -19,10 +19,12 @@
     max-width: var(--content-max-width);
     padding: 0 var(--content-padding);
   }
+
   ul {
     width: 100%;
     border: 1px solid var(--color-foreground-level-2);
     border-radius: 8px;
+    overflow: hidden;
   }
 
   li {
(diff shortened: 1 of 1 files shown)


geigerzaehler (dismissed):
Thanks for the contribution!

xla (changes requested):
Thanks for the contirbution. We aim to attribute all contributors properly, see our [DCO](https://github.com/radicle-dev/radicle-upstream/blob/master/DCO). For that could you repush your commit with [signoff](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff).

finleygn:
@xla This projects really cool so happy to help out, learning rust and svelte at the moment so starting out small :)

re-pushed on here and #1468

xla (approved):
🍤 ✔️ 🍺 ⌚️

geigerzaehler (approved):

PR #1379, opened 2020-11-26
fix(ci): timeout for proxy tests

geigerzaehler:
At the moment some of the tests in the proxy test suite may stall (#1378). To avoid the build only failing after 60 minutes we put a timeout on the proxy tests.

Issue #1378, opened 2020-11-26
Proxy tests on CI stall on `can_fetch_project_changes`


geigerzaehler:
Sporadically the `can_fetch_project_changes` test that is part of the coco replication test suite stalls on CI. This leads to the builds running until they time out after 60 minutes.

Some things we can do about this
* Use `timeout` command for the proxy tests
* Use `timeout` in proxy tests
* Add logging to proxy tests

@FintanH mentioned that this might be addressed by upcoming chnages in `librad`.

geigerzaehler:
Even with the forced timeouts added in #1379 the builds still time out: https://buildkite.com/monadic/radicle-upstream/builds/6930#d6fcf5c5-54ea-4aec-aa7c-50e2c84a4a72

geigerzaehler:
Ok, it looks to me like third party builds always stall at the point above. The cause of this is probably that we’re using the kata container environment for these builds. We need to investigate further.

geigerzaehler:
The tests were moved out of the repo in #1753.

PR #1474, opened 2020-12-08
fix(ci): put timeout on proxy tests

geigerzaehler:
We’re using the `--ensure-time` option with `cargo test` to notify us when a test takes longer than the specified time. However, this does not abort the test suite and thus does not address #1378. To actually abort the test suite we are using the `timeout` command.

In addition we have the following changes
* We remove the `timeout` command from the lint step. This was committed by me accidently in #1379. We have never experienced issues with clippy timing out.
* We’re replacing the `--ensure-time` option with `--report-time`. We want to know how long tests take but not fail them since we have no expectations for the time they should take.

### DIFF
--- a/.buildkite/run.sh
+++ b/.buildkite/run.sh
@@ -45,7 +45,7 @@ echo "--- Run proxy fmt"
 echo "--- Run proxy lints"
 (
   cd proxy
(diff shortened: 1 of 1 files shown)


rudolfs (dismissed):

xla (dismissed):
🔬 🚵 🦂 👋

rudolfs (approved):

Issue #1345, opened 2020-11-24
Improve QA documentation


rudolfs:
**Fixups**:
- ~~when QA instructions are copied to issue, the "this script" link breaks~~
- ~~After passphrase input the identity is created and a copyable URN is
provided  -- should be Device ID instead of URN~~
- ~~test that " " is replaced with "-" in onboarding and project creation~~
- ~~current QA still mentions `radicle-upstream` as password~~
- ~~remove the section about dancing round Apple Gatekeeper, our package is signed now~~
- ~~"Peer entries" -> "Network -> seeds"~~

**Add new sections**:
  - ~~replicate a project from seed~~
    - ~~checkout~~
    - ~~add changes~~
    - ~~publish changes~~
    - ~~add other remotes~~
    - pull changes from a different remote
  - peer selector
  - ~~network indicator~~
  - ~~search~~
    - ~~follow a project~~
    - ~~waiting area~~

NunoAlexandre:
we also miss having a setup git as a pre-requisite. I failed to create a repo because:

```
{
  "code": "ProjectCreationFailure",
  "message": "Could not create project: the author name for creating the project could not be found - have you configured your git config?",
  "source": {
    "response": {}
  }
}
```

PR #1485, opened 2020-12-10
docs: tidy up release docs

rudolfs:
Trying to strike a balance here between testing everything that is not covered by our test suite and keeping the QA checklist short simple enough to be doable for every release.

This PR:
  - adds a release checklist so we don't miss a step
  - adds templates for community announcements
  - shuffles the QA checklist around so it can be completed in one go without jumping around between sections

**Note:** I left out testing the peer selector and pulling changes from a new remote, because that would need extra coordination and would make the QA process even longer. Hope we can cover this case by tests soon.

Closes: #1345.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -502,10 +502,10 @@ which contains a QA checklist. Before publishing packages for a wider audience
 someone from the team goes through the checklist and manually tests the app,
 afterwards the team can evaluate whether the release is up to our standards.
 
-**Title:** `QA: v0.0.11 macOS`\
+**Title:** `QA: vX.X.X macOS`\
 **Body** [QA.md][qa]
 
-**Title:** `QA: v0.0.11 Linux`\
+**Title:** `QA: vX.X.X Linux`\
 **Body:** [QA.md][qa]
 
 ## Publishing a release
@@ -522,7 +522,7 @@ following commands:
 
 ```
 cd radicle-upstream
-git checkout v0.0.11
+git checkout vX.X.X
 
 CSC_NAME="Monadic GmbH (XXXXXXXXXX)" \
 APPLE_ID="XXXXXXX@monadic.xyz" \
@@ -543,7 +543,7 @@ Once the package is notarized you can upload it to the web using
 [`gsutil`][gs]:
 
 ```bash
-gsutil cp dist/radicle-upstream-0.0.11.dmg gs://releases.radicle.xyz
+gsutil cp dist/radicle-upstream-X.X.X.dmg gs://releases.radicle.xyz
 ```
 
 To be able to upload packages to the GCS bucket you will need the appropriate
@@ -554,8 +554,8 @@ After this you also need to obtain and upload the Linux package.
 FIXME(rudolfs): this doesn't work, we don't build images off of tags.
 
 ```bash
-curl -sSLO https://builds.radicle.xyz/radicle-upstream/v0.1.2/radicle-upstream-0.1.2.AppImage
-gsutil cp radicle-upstream-0.1.2.AppImage gs://releases.radicle.xyz
+curl -sSLO https://builds.radicle.xyz/radicle-upstream/vX.X.X/radicle-upstream-X.X.X.AppImage
+gsutil cp radicle-upstream-X.X.X.AppImage gs://releases.radicle.xyz
 ```
 
 After all the packages are uploaded, update the links to those binaries on the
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -519,62 +532,96 @@ yarn dist
...
-  https://matrix.radicle.community/#/room/#support:radicle.community
-
-If you encounter a bug, please open an issue here:
-  https://github.com/radicle-dev/radicle-upstream/issues
-```
-
-[an]: #setting-up-apple-notarization
+  - https://radicle.community/c/announcements
+    - subject:
+        Radicle Upstream vX.X.X is out! 🎉`
```
NunoAlexandre, line 570:
```suggestion
        Radicle Upstream vX.X.X is out! 🎉
```

QA.md
```diff
@@ -1,186 +1,191 @@
...
-    - is at least 4 characters long
-    - the repeated passphrase is equal to the first one
-    - all of the following example pairs should show a validation error:
-      `1`/`1`, `12`/`12`, `123`/`123`, `supersecret`/`supersceret`
-- [ ] After passphrase input the identity is created and a copyable URN is
-      provided
-- [ ] After completion we land on our profile page which contains a placeholder
-      with instructions on how to create your first project
+- [ ] Can create a new project with a new repository
+  - [ ] Validations work
```
NunoAlexandre, line 47: Would be good to list here examples of invalid inputs. At the same time, this is all UI-tested so ya...

NunoAlexandre (approved):
Looks great to me :broom:

DEVELOPMENT.md
```diff
@@ -420,15 +437,11 @@ Prerequisites:
     - [Certificates Add][ca] -> Developer ID Application
       **Note:** this can only be created via the company account holder
 
+Once you've created the _Developer ID Application_ certificate, download it
+locally and add it to your keychain by double cliking on the file.
```
MeBrei, line 441:
```suggestion
locally and add it to your keychain by double clicking on the file.
```

QA.md
```diff
@@ -1,186 +1,191 @@
...
+```
 
-### Proxy Error
 
-- [ ] Killing the proxy process while the app is running shows a blue error
-      screen with the proxy logs.
 
-[re]: https://github.com/radicle-dev/radicle-upstream/blob/master/CHANGELOG.md
+[rs]: https://raw.githubusercontent.com/radicle-dev/radicle-upstream/master/scripts/reset-state.sh
+[bu]: https://releases.radicle.xyz/radicle-upstream-0.1.5.dmg
```
MeBrei, line 191: Would this link have to be updated with every release?

MeBrei (dismissed):

QA.md
rudolfs: Yea, we had those (see a couple of lines above), I decided to remove them because they're all covered by our cypress test suite and they took quite a while to check manually.

My thinking here is that this is just a very simple smoke test that's fast to execute. Like when you delete the input contents it shows the validation, nothing more.

QA.md
rudolfs: Correct. Should probably make it 0.X.X.dmg 🤔

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1485/commits/f6a324c2a870a6e7c48f4170d0afe189a5c9291a

DEVELOPMENT.md
rudolfs: Good catch!

The formatting was also busted, fixed in:
https://github.com/radicle-dev/radicle-upstream/pull/1485/commits/c0c438543ebd746977f3f3cb7ba67a70dc75d7de
https://github.com/radicle-dev/radicle-upstream/pull/1485/commits/48229e541009d56b5c3eed2525b8ab4d40a06cd0

Screenshot 2020-12-16 at 14 56 38

DEVELOPMENT.md
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1485/commits/b529e238f3600b6d41b6d4ec3770c2b923fb0df2

QA.md
rudolfs: And to answer your question: the link would have to be updated when we create the release QA issues and paste in the contents of this template.

QA.md
NunoAlexandre: Yeah I am aware of that, what I meant to say is that maybe having an example of an invalid input instead of "validation work" would be better after all since both want to test the same but the latter makes it implicit what to test for.

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1485/commits/faf154b3a93a3bd007a95c23ce09dd471669e122

NunoAlexandre (approved):

MeBrei (approved):

PR #1489, opened 2020-12-16
feat!(proxy): Consolidate all directories under .radicle

### LINKED ISSUE #1199: Allow the user to start the app with a different directory
To have a solution for a forgotten passphrase, but also to be able to later build rotating your passphrase, having multiple identities and having completely separate environments for dev and production, we want to be able to start the app with a different directory (different state). 

Currently all files are created in `XDG_CONFIG_HOME`, `XDG_DATA_HOME` or `HOME`, when building from source there is the possibility to delete all files created by the app via `yarn reset:state`. dev and production mode store their data files in different directories but share the same config files.

So far the following scenarios have been discussed:
- start the app from console, setting `XDG_CONFIG_HOME`, `XDG_DATA_HOME` or `HOME` to a new directory
- the directory is set "behind the scenes" and a button is created in `Settings` (like the former clear session button) which provokes switching to a new directory and therefor a new identity/key store has to be created

MeBrei:
❗ Breaking change ❗
I am keeping this as a draft PR for now, so it is not accidentally merged.

All files are now under `$HOME/.radicle`. Files that are specific to an identity are stored under `$HOME/.radicle/identities/current` which is a symlink to a specific random hash folder.  Managing of the files (creation, symlink) was moved into `main.ts` (electron) as it is run before the proxy in production mode. Before, in dev mode the order was the other way around, but to keep things consistent and not having to manage the files in two places, the order was also switched in dev mode.
The new file system looks as follows:

```
~/.radicle » tree -L 3
.
├── bin
│   └── git-remote-rad
└── identities
    ├── cb8a9782-2641-4947-9349-64e41a1925a0
    │   ├── electron
    │   ├── git
    │   ├── git-includes
    │   ├── keys
    │   └── store
    ├── current -> /Users/rudolfs/.radicle/identities/cb8a9782-2641-4947-9349-64e41a1925a0
    └── e35c9925-54e8-49f5-9e52-448d175e2da1
        ├── electron
        ├── git
        ├── git-includes
        ├── keys
        └── store
```
This allows to use different identities. For both actions, the app needs to be stopped:
- to create a new identity, `rm` or `unlink` the symlink: `$HOME/.radicle/identities/current`
- to switch to a different identity, point the symlink to a different identity folder

Closes #1199 

**ToDo**
- [x] create folder with random hash for each identity
- [x] move `electron` into specific identity folder (this requires to create an empty random hash folder with symlink on start if there is no symlink yet)
- [x] make it work on `production` ie the current implementation assumes that `process.rs` runs before `main.ts` which is not the case when in production mode
- [x] script that let's you switch between identities or create a new one (ie removing the symlink)
- [x] use environment variable `RAD_HOME` with `HOME` as default for where to store all data
- [x] test on non-macos setup
- [ ] migration script to move your existing radicle data into the new file system (probably shared outside of  the repository)
- [ ] update `scripts/reset-state.sh` script


rudolfs:
Force pushed because my commits weren't verified...

package.json
```diff
@@ -115,12 +115,13 @@
     "wait-on": "^5.2.0"
   },
   "scripts": {
-    "start": "RADICLE_UPSTREAM_EXPERIMENTAL=false yarn start:app",
-    "start:experimental": "RADICLE_UPSTREAM_EXPERIMENTAL=true yarn start:app",
-    "start:app": "run-p --race rollup:watch proxy:start electron:start",
-    "start:release": "run-p --race rollup:watch proxy:start:release electron:start",
-    "start:experimental:release": "RADICLE_UPSTREAM_EXPERIMENTAL=true yarn start:release",
-    "start:test": "run-p --race rollup:watch proxy:start:test electron:start",
+    "start": "RADICLE_UPSTREAM_EXPERIMENTAL=false RADICLE_UPSTREAM_PROXY_PATH=../proxy/target/debug/radicle-proxy yarn run:dev:proxy",
```
FintanH, line 118: While we're here can we default `start` to be release? :pray: :innocent: :bow:

package.json
```diff
@@ -131,18 +132,16 @@
...
     "dist:notarize": "NOTARIZE=true yarn dist",
-    "electron:start": "wait-on ./public/bundle.js && wait-on ./native/main.comp.js && wait-on tcp:17246 && NODE_ENV=development electron .",
+    "electron:start": "wait-on ./public/bundle.js ./native/main.comp.js && NODE_ENV=development electron .",
     "svelte:check": "svelte-check",
     "svelte:check:watch": "svelte-check --watch",
     "rollup:clean": "rm -rf public/bundle.* && rm -f native/main.comp.js",
     "rollup:build": "yarn rollup:clean && rollup -c",
     "rollup:watch": "yarn rollup:clean && rollup -c -w",
     "typescript:check": "tsc --noEmit && tsc --noEmit --project cypress && svelte-check",
-    "proxy:build": "cd proxy && cargo build --all-features --all-targets",
```
FintanH, line 141: Why are these flags being removed?

proxy/api/src/config.rs
```diff
@@ -21,10 +19,11 @@ pub enum Error {
...
 
-/// Returns the directories to locate all application state.
-#[must_use]
-pub fn dirs() -> ProjectDirs {
-    ProjectDirs::from("xyz", "radicle", "radicle-upstream").expect("couldn't build dirs")
+/// Returns the directories to locate all application state specific to the current identity.
+pub fn id_dirs() -> Result<path::PathBuf, Error> {
+    let home_dir = std::env::var("HOME")?;
+
+    Ok(path::Path::new(&home_dir).join(".radicle/identities/current"))
```
FintanH, line 26: Actually, now that I'm thinking about it, why did we opt for `~/.radicle` instead of putting `identities` under the project dir?

FintanH (commented):
Looking good! Just a few questions

package.json
rudolfs: They're not used anywhere.

package.json
rudolfs: Hm, I don't see a reason why not.

package.json
FintanH: `--all-targets` seems fine to remove since we're building at the top of the workspace, so I believe it will build all the workspace targets.

`--all-features` I'm a little hesitant about. I suppose we don't have any features _now_ but if we did introduce feature flags later that would be different. No biggy for now though.

proxy/api/src/config.rs
rudolfs: My thinking was that since the git helper has to be in `~/.radicle/bin` anyway (because of https://github.com/radicle-dev/radicle-upstream/issues/763#issuecomment-671374819), we could have everything in one place. Would also be easier to instruct users, because the directory would be in the same place for every platform.

I know this is not _the right way_, so I'm open for counter-arguments.

rudolfs: This would make it also easier to back up or remove all radicle data, since it wouldn't be spread around the filesystem: https://github.com/radicle-dev/radicle-upstream/blob/master/scripts/reset-state.sh.

FintanH: Alrighty, thanks for the clarification :+1:

package.json
rudolfs: Hm, I think I might have misunderstood what these flags do. Thanks for pointing it out, I added them back, I guess they do no harm and prevent us from tripping over in the future if we change something in proxy.

https://github.com/radicle-dev/radicle-upstream/pull/1489/commits/4cc40f5437174a0f7162a4e9e4d7641d70ddc738

package.json
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1489/commits/e4325829a97673b72553c87de86a78c676b62e37

package.json
```diff
@@ -115,12 +115,13 @@
...
-    "start:release": "run-p --race rollup:watch proxy:start:release electron:start",
-    "start:experimental:release": "RADICLE_UPSTREAM_EXPERIMENTAL=true yarn start:release",
-    "start:test": "run-p --race rollup:watch proxy:start:test electron:start",
+    "start": "RADICLE_UPSTREAM_EXPERIMENTAL=false RADICLE_UPSTREAM_PROXY_PATH=../proxy/target/release/radicle-proxy yarn run:release:proxy",
+    "start:dev": "RADICLE_UPSTREAM_EXPERIMENTAL=false RADICLE_UPSTREAM_PROXY_PATH=../proxy/target/debug/radicle-proxy yarn run:dev:proxy",
+    "start:experimental": "RADICLE_UPSTREAM_EXPERIMENTAL=true RADICLE_UPSTREAM_PROXY_PATH=../proxy/target/release/radicle-proxy yarn run:release:proxy",
+    "start:experimental:dev": "RADICLE_UPSTREAM_EXPERIMENTAL=true RADICLE_UPSTREAM_PROXY_PATH=../proxy/target/debug/radicle-proxy yarn run:dev:proxy",
+    "run:test": "run-p --race rollup:watch proxy:start:test electron:start",
+    "run:dev:proxy": "yarn proxy:build && yarn run-p --race rollup:watch electron:start",
+    "run:release:proxy": "yarn proxy:build:release && yarn run-p --race rollup:watch electron:start",
```
geigerzaehler, line 124: I’m getting more and more confused by the number of commands and what they do. I have a couple of suggestions.
1. Eleminate the `*:experimental:*` commands. They can easily be replicated by setting the environment variable. E.g. `RADICLE_UPSTREAM_EXPERIMENTAL=true yarn:start`
2. Some commands, e.g. `run:dev:proxy` don’t work if you call them directly. Either we remove them and inline their code or we clearly mark them as internal by changing their name.
3. Let’s try to make the naming more consistent. It’s hard to disinguish `start:dev` from `run:dev:proxy`. And the `proxy` part is confusing because it really runs more than the proxy.

geigerzaehler (commented):
:+1: on the general approach

One issue that we still have is that when you run the app in development mode, say with `yarn start` it uses the real identity by default. Is this something that you intend on improving in a follow-up?

package.json
geigerzaehler: Could we also update the docs in `DEVELOPMENT.md`?

MeBrei:
> 👍 on the general approach
> 
> One issue that we still have is that when you run the app in development mode, say with `yarn start` it uses the real identity by default. Is this something that you intend on improving in a follow-up?

yes, this will be addressed in a follow up. We have 2 options (as long as we don't have an ui for switching identities)
1) we have a script/command that you run when you want to switch your identity (basically just changing the symlink)
2) we pass a variable to `yarn start` etc. with which you can set which identity should be used.

package.json
MeBrei: I agree that it needs cleaning up. I regularly forget about the very useful `yarn start:test`. I think it is outside of the scope of this PR though.

native/main.ts
```diff
@@ -6,18 +6,52 @@ import {
...
+    path.join(__dirname, process.env.RADICLE_UPSTREAM_PROXY_PATH!)
+  : path.join(__dirname, "../../radicle-proxy");
 
 // The default value of app.allowRendererProcessReuse is deprecated, it is
 // currently "false".  It will change to be "true" in Electron 9.  For more
 // information please check https://github.com/electron/electron/issues/18397
 app.allowRendererProcessReuse = true;
 
+const home = app.getPath("home");
+const identities_path = `${home}/.radicle/identities`;
```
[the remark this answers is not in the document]
MeBrei: changed in c95ea30

package.json
rudolfs: Good points, I could take it over and do it for this PR.

package.json
rudolfs: Rolling this back as @geigerzaehler pointed out this would slow down proxy recompilation when working on integrating proxy <-> UI.

geigerzaehler (approved):

FintanH:
We discussed in Matrix an environment variable to decide where this `.radicle` home will live, the default being `$HOME`. Is that still the case?

MeBrei:
> We discussed in Matrix an environment variable to decide where this `.radicle` home will live, the default being `$HOME`. Is that still the case?

I added it to the ToDos in the description. Thanks for bringing it up again!

package.json
rudolfs: @geigerzaehler addressed all points, PTAL!

FintanH (approved):

proxy/api/src/config.rs
```diff
@@ -21,10 +19,11 @@ pub enum Error {
...
 
-/// Returns the directories to locate all application state.
-#[must_use]
-pub fn dirs() -> ProjectDirs {
-    ProjectDirs::from("xyz", "radicle", "radicle-upstream").expect("couldn't build dirs")
+/// Returns the directories to locate all application state specific to the current identity.
+pub fn id_dirs() -> Result<path::PathBuf, Error> {
+    let home_dir = std::env::var("RAD_HOME").unwrap_or(std::env::var("HOME")?);
+
+    Ok(path::Path::new(&home_dir).join(".radicle/identities/current"))
```
geigerzaehler, line 26: Suggestion: There is no immediate benefit but it would be nice to get in the habit of always using `join` to avoid hard-coding platform-specific behavior.

```suggestion
    Ok(path::Path::new(&home_dir).join(&[".radicle", "identities", "current"].iter().collect())
```

scripts/manage-identity.sh
```diff
@@ -0,0 +1,32 @@
...
+        echo "Removed the symlink"
+        break
+        ;;
+    * )
+        ln -sfn $id $home/.radicle/identities/current
+        echo "Switched to identity $id"
+        break
+        ;;
+esac
+done
```
geigerzaehler, line 32: Could we add a new-line at the end to avoid unnecessary diff noise when this is edited again.

scripts/manage-identity.sh
```diff
@@ -0,0 +1,32 @@
...
+    echo "You need to stop the proxy before switching the identity!"
+    exit 1
+fi
+
+echo "Select the identity you want to switch to or opt to create a new one:"
+
+shopt -s extglob
+select id in $home/.radicle/identities/!(current)/ "NEW IDENTITY"
+do
+case "$id" in
```
geigerzaehler, line 20: Is there a reason we don’t use an `if` statement instead of `case`?

geigerzaehler (commented):
I’m a bit confused about `RAD_HOME`. I would expect the environment variable to control the location of `$HOME/.radicle`. Don’t let this be a blocker, though.

FintanH:
> I’m a bit confused about RAD_HOME. I would expect the environment variable to control the location of $HOME/.radicle. Don’t let this be a blocker, though.

I can probably shed some light here. The environment variable for `radicle-link` is `RAD_HOME` so we're utilising this in Upstream as the main home for identities

geigerzaehler:
> I can probably shed some light here. The environment variable for `radicle-link` is `RAD_HOME` so we're utilising this in Upstream as the main home for identities

It seems to me that in upstream `librad` never utilizes `RAD_HOME` because we’re calling `Paths.::from_root` and not `Paths::from_env`. So it seems like this environment variable is only used by `git-remote-rad`.

Moreover the rad remote helper stores keys in `$RAD_HOME/keys` whereas upstream configures `Paths` so that the keys are stored in `$RAD_HOME/.radicle/identities/current/keys`.

Sorry to add to the confusion. Maybe we can have a quick chat about this after the break.

scripts/manage-identity.sh
MeBrei: I think the code looks neater this way and we can easily extend the selection, if we want additional or more specific behavior.

scripts/manage-identity.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
xla, line 1: Parts of this script re-implement logic from the rust code, could this be an additional `bin` inside the proxy codebase?

xla: Alternatively as the rest of the id management lives inside the native app shell, it could be moved there.

package.json
```diff
@@ -115,55 +115,49 @@
...
+    "_private:rollup:clean": "rm -rf public/bundle.* && rm -f native/main.comp.js",
+    "_private:prettier": "prettier \"**/*.@(js|ts|json|svelte|css|html)\" --ignore-path .gitignore",
+    "_private:proxy:build": "cd proxy && cargo build --all-features --all-targets",
+    "_private:proxy:build:release": "cd proxy && cargo build --release",
+    "_private:proxy:clean": "cd proxy && cargo clean",
+    "_private:proxy:start:test": "cd proxy && cargo build --bin git-remote-rad && cargo run -- --test",
+    "_private:proxy:start:test:watch": "cd proxy && cargo build --bin git-remote-rad && cargo watch -x 'run -- --test'",
+    "_private:rollup:build": "yarn _private:rollup:clean && rollup -c",
+    "_private:rollup:watch": "yarn _private:rollup:clean && rollup -c -w",
+    "postinstall": "patch-package"
```
xla, line 149: Is this a normal amount of behaviour in the scripts section of a package.json. It seems rather peculiar that so much operational knowledge is captured in here. Are there plans to have a more sustainable way to manage these scripting needs?

kim:
Just catching up on this.

Iiuc, the idea here is to move away from the XDG scheme (and its lesser-defined counterparts on other platforms). I am not convinced that this is a good idea -- XDG is widely adopted among both Linux and macOS users, not least because it makes it easy to back up configuration files, and share them between machines. It also greatly reduces the clutter in `HOME`. Admittedly, the macOS and Windows counterparts as implemented by the `directories` crate are a bit obscure, which is rooted in those platforms typically discouraging users from directly interacting with configuration and application state directly.

There is an aspect of XDG which may have caused some confusion here: XDG does not specify a common root directory, but splits different concerns under different hierarchies (`~/.config`, `~/.local/share`, `~/.cache`, ...). This is a good thing: `XDG_CONFIG_HOME` is owned by users, and package managers / installers should leave it alone, while the others can safely be removed, f.ex. when uninstalling.

However, this turns out a bit annoying for test setups, because **all** XDG environment variables have to be defined to confine temporary state + configuration. This is why `Paths::from_root` and `RAD_HOME` exist -- the former because overriding `HOME` (or any other environment variable) for multiple in-process instances is not practical, and the latter because remote helpers (by definition separate processes) can't take any arguments.

While the single root directory approach works well for throwaway environments, what seems to be the goal here is to support multiple **persistent** configurations, from which users may eventually be able to choose. Clearly, modifying `XDG_*` doesn't work for this, nor does `HOME/.radicle` (for the reasons outlined above).

I think the approach of introducing another sub-hierarchy is reasonable, but it is unclear to me why this should be a random identifier, when the key id (`PeerId`) would do. I would also prefer a _file_  which _contains_ the key id, rather than a symlink, because of portability. As for terminology, "profile" comes to mind (instead of "identity"), to avoid overloading.

To summarise, I would suggest to:

* Keep the XDG convention
* Rename `RAD_HOME` to `RAD_PROFILE`, whose value is a valid `PeerId`
* Introduce a file `$XDG_CONFIG_HOME/radicle/profile` (respectively the platform-specific config directory), whose contents is a valid `PeerId`
* If `RAD_PROFILE` is set, it takes precedence over the profile file. If a `radicle` application takes a CLI argument pertaining the profile, this value takes precedence over both `RAD_PROFILE` and the profile file.
* If either `RAD_PROFILE` is set to a valid value, or the profile file exists and contains a valid value, paths are looked up relative to `prefix`, where `prefix` is the platform-equivalent of `$XDG_*/radicle/$RAD_PROFILE` (e.g. `$XDG_DATA_HOME/radicle/$RAD_PROFILE/git/`). It is an error if the `RAD_PROFILE` component of the respective path doesn't exist, or the value does not parse as a `PeerId`.
* If neither `RAD_PROFILE` is set, nor the profile file exists, the layout is assumed to be as before. This may or may not trigger a migration.

MeBrei:
Thank you for your input! I added some comments/questions below to make sure I understand your suggestion. 

> To summarise, I would suggest to:
> 
> * Keep the XDG convention
> * Rename `RAD_HOME` to `RAD_PROFILE`, whose value is a valid `PeerId`
> * Introduce a file `$XDG_CONFIG_HOME/radicle/profile` (respectively the platform-specific config directory), whose contents is a valid `PeerId`

- If I understand you correctly, this would be the "default" user? Would we allow the user to change this?
- The random identifier was chosen, as the profile specific `electron` files are created before the onboarding process is started. 
- Not only the keys are specific to the profile, but also data that is not stored under `$XDG_CONFIG_HOME`, so we would have to have the `profile`/`Peer_id` sub-directories in all those places. The full list of files can be found [here](https://github.com/radicle-dev/radicle-upstream/issues/1199#issuecomment-738669752)
> * If `RAD_PROFILE` is set, it takes precedence over the profile file. If a `radicle` application takes a CLI argument pertaining the profile, this value takes precedence over both `RAD_PROFILE` and the profile file.

The idea down the line is to support choosing your profile in the app on start up. So I am not sure if basing the switching of profiles via CLI argument/env variable will work for that. You would also have to instruct users to set the env variable manually when doing eg `git rad push` with a profile other than their default one.
> * If either `RAD_PROFILE` is set to a valid value, or the profile file exists and contains a valid value, paths are looked up relative to `prefix`, where `prefix` is the platform-equivalent of `$XDG_*/radicle/$RAD_PROFILE` (e.g. `$XDG_DATA_HOME/radicle/$RAD_PROFILE/git/`). It is an error if the `RAD_PROFILE` component of the respective path doesn't exist, or the value does not parse as a `PeerId`.
> * If neither `RAD_PROFILE` is set, nor the profile file exists, the layout is assumed to be as before. This may or may not trigger a migration.

Additional question:
- With this approach, how can the user trigger a new creation of a profile? (The current implementation triggers the onboarding flow, when the `current` symlink does not exist. So for now, a new profile creation can be triggered with removing the symlink)

package.json
rudolfs: We could extract this into a separate script and call it from `package.json`. For that we would have to write some process management boilerplate, or use the `wait-on` and `npm-run-all` packages programmatically in that script. @geigerzaehler also pointed out that this could be a nice refactoring.

I didn't want to do it in this issue to keep the scope down. Should I open an issue about it?

scripts/manage-identity.sh
MeBrei: This is just an intermediate script, which we also want to provide to user who are not building from source. In the long term, it should be replaced with an in-app interface to select your current identity.

kim:
>     * If I understand you correctly, this would be the "default" user? Would we allow the user to change this?

Sure. I was just suggesting to use a proper file instead of a symlink -- symlinks are not portable, so will cause issues. 


>     * The random identifier was chosen, as the profile specific `electron` files are created before the onboarding process is started.

Fair point. From the comments, I don't see how chromium state is significant in this context, though.



>     * Not only the keys are specific to the profile, but also data that is not stored under `$XDG_CONFIG_HOME`, so we would have to have the `profile`/`Peer_id` sub-directories in all those places.

Yes, that's exactly the point about sticking to XDG. As a user, I would want control over configuration files and key material, but e.g. `git` and database state should not be interacted with directly.



> The idea down the line is to support choosing your profile in the app on start up. So I am not sure if basing the switching of profiles via CLI argument/env variable will work for that. 

`upstream` will not be the only application down the road -- as you can see already with the remote helper. So any solution must work in a many-app world. I understood that this was the reason for storing the "current" profile on-disk as opposed to `upstream` local storage -- other applications can use this shared state as well.

> You would also have to instruct users to set the env variable manually when doing eg `git rad push` with a profile other than their default one.

What's the problem with that? The symlink approach is isomorphic to a "profile" file, which could be manipulated manually or via tooling.



>     * With this approach, how can the user trigger a new creation of a profile? (The current implementation triggers the onboarding flow, when the `current` symlink does not exist. So for now, a new profile creation can be triggered with removing the symlink)

There is no difference between a symlink and a file in this respect.

MeBrei:
We decided to stay with the `XDG` directory scheme. New work can be found here: https://github.com/radicle-dev/radicle-upstream/pull/1527, https://github.com/radicle-dev/radicle-upstream/issues/1528, https://github.com/radicle-dev/radicle-link/issues/488

Issue #1475, opened 2020-12-08
Network integration testing


rudolfs:
## Rationale

To increase confidence in our release process and avoid wasting time performing manual QA, we want to have test coverage for most of the features in Upstream. However, currently we don't have a way to test any of the networking related features, e.g.:
  - peer selector
  - adding/removing remotes
  - network activity indicator
  - searching/following projects
  - fetching project changes
  - replication/syncing
  - etc.

See https://github.com/radicle-dev/radicle-upstream/issues/1345 and [QA.md](https://github.com/radicle-dev/radicle-upstream/blob/master/QA.md) for an exhaustive list of what needs to be automated.

## Requirements

Ideally we would have something that can:
  - simulate a swarm of nodes/seeds
    - in many different network configurations (e.g. 2 isolated nodes connecting to 1 seed; 3 nodes on the same LAN, etc)
  - run on CI
  - run on local **Linux** and **macOS** dev machines natively
  - run from the proxy rust test suite
  - run from the UI cypress test suite

### Docker
This would allow us to create different network configurations with `docker-compose`. Possibly also allow us to isolate nodes on a network level from one another. This way we could do proper integration tests where we don't have to modify `proxy` at all.

Drawbacks:
  - our current CI setup already runs in a container and running docker within docker [might be messy](https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/)
  - multicast UDP between containers might be difficult or impossible to set up
  - we would also have to re-configure our buildkite nodes to support that

### Bake in test harnesses into proxy
In this scenario we would change `proxy` in such a way that we can run multiple `proxy` binaries on the same host (or container). Then we could control the additional `proxy` processes via their HTTP APIs from cypress or rust tests to set up different network configurations for each test.

For this to work, we would need to add a couple of additional CLI options to the binary, so that we can:
  - bind to a specific `interface:port` for incoming data connections (we already allow specifying a port for the HTTP API)
  - specify a custom data directory (similar to how we do it on the seed node via `--root`): https://github.com/radicle-dev/radicle-upstream/issues/1199
  - specify allow/deny lists of peers/seeds the process is allowed to connect to directly, to simulate different network configurations

Drawbacks:
  - different network configurations are only simulated, this might work differently when clients talk to each other in the real world over the wire

Open questions:
  - can two processes bound to the same interface, but different ports discover each other over mDNS?

[...]

PR #1492, opened 2020-12-17
test: add network integration testing framework

rudolfs:
This PR:
- implements a Cypress plugin which allows starting and interacting with multiple `radicle-proxy` instances
- demonstrates the plugin by testing project replication between two Upstream clients
- adds a `--default-seed` CLI option to `radicle-proxy` for configuring default seed addresses
- provides a way to pass the backend HTTP API `address:port` pair to the UI via a query string

The `--default-seed` option will also come in handy when we split up our seedling into dev/prod environments.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/1475.

https://user-images.githubusercontent.com/158411/105336831-fc11fc00-5bd9-11eb-87eb-c143acce92a1.mov


cypress/integration/networking.spec.ts
```diff
@@ -0,0 +1,65 @@
+import type { OnboardedNode } from "../plugins/index";
+
+import * as commands from "../support/commands";
+import * as path from "path";
+import { ipcStub } from "../support";
+
+context("networking", () => {
+  it("replicates a project from one node to another", () => {
+    cy.task("startNode", 17000);
```
FintanH, line 9: I wonder if we could do something similar to `radicle-link`s approach where we can specify the number of nodes to boot up and get back a list of them. I'd also wonder if we should randomly generate the set of `NodeId`s and maybe probe to see if that port is taken yet.

DEVELOPMENT.md
```diff
@@ -314,17 +314,40 @@ Read more about the colors used in Upstream in the [Color System post][cg].
...
-`native/main.js`. When running `yarn dist` it is bundled together into an
+`native/main.ts`. When running `yarn dist` it is bundled together into an
 application package by [electron-builder][eb].
 
+### CLI options
+
+```
+$ cd proxy && cargo run --bin radicle-proxy -- --help
+
+Usage: radicle-proxy [--test]
```
geigerzaehler, line 333: I fear that this will become outdated very quickly. We’ll probably forget to update this. Since people can just run the proxy with `--help` by themselves I’d prefer to remove this section.

ui/App.svelte
```diff
@@ -43,6 +44,11 @@
   import ModalCollect from "./Modal/Funding/Pool/Collect.svelte";
   import UserProfile from "./Screen/UserProfile.svelte";
 
+  // We receive the backend address from index.ts before the app loads and
+  // svelte-spa-router can mess with `window.location`.
+  export let backendAddress = "localhost:17246";
```
FintanH, line 49: I don't understand the relation between this value and its comment above, could you elaborate?

FintanH (commented):
Nice work :) I'm excited about all the tests we'll have :grin:

proxy/api/src/process.rs
```diff
@@ -11,14 +14,24 @@ use coco::{convert::MaybeFrom as _, peer::run_config, seed, signer, Peer, RunCon
...
-const PORT: u16 = 17246;
-
 /// Flags accepted by the proxy binary.
-#[derive(Clone, Copy)]
+#[derive(Clone, FromArgs)]
 pub struct Args {
-    /// Put proxy in test mode to use certain fixtures.
+    /// put proxy in test mode to use certain fixtures
+    #[argh(switch)]
     pub test: bool,
```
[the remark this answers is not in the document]
rudolfs: It still controls whether we run tests in a tempdir or not.

cypress/integration/networking.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    cy.task("onboardNode", 17000);
+
+    cy.task("startNode", 18000);
+    cy.task("onboardNode", 18000);
+
+    cy.task("connectNodes", [17000, 18000]);
+
+    cy.task<OnboardedNode>("withNode", 17000).then(node => {
```
geigerzaehler, line 17: I find that using string-based APIs on has a couple of shortcomings. The most obvious one is that we don’t get any help from TypeScript when it comes to the arguments of functions and we rely on the user to put in the return argument type. This also makes it harder to create better abstractions on top of the basic API. For example, it seems that we always run `startNode` and `onboardNode` in conjunction so maybe we want to abstract this away to make the API easier to use.

proxy/api/src/process.rs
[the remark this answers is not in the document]
FintanH: Cool, thanks for clarifying :grin:

cypress/integration/networking.spec.ts
geigerzaehler, line 17: It looks to me, like this is all boilerplate that will be repeated for every test with two nodes. So how about encapsulating this in one function and do some like.

~~~js
withTwoNodes().then(([node1, node2]) => ...)
~~~

ui/index.ts
```diff
@@ -1,8 +1,19 @@
 import App from "./App.svelte";
 
+const parseBackendAddress = (): string | undefined => {
+  const search = window.location.search;
+
+  if (search.includes("?backend=")) {
+    const match = search.match(/\?backend=(.*)/);
```
geigerzaehler, line 7: An alternative would be to use [`qs`](https://www.npmjs.com/package/qs)

ui/index.ts
```diff
@@ -1,8 +1,19 @@
...
+    if (match) {
+      return match[1];
+    }
+  }
+};
+
 const app = new App({
   target: document.body,
-  props: {},
+  props: { backendAddress: parseBackendAddress() },
```
geigerzaehler, line 16: Is there an advantage to passing this down via `App` as opposed to calling `backendAddresssStore.set()` here? The latter would involve fewer components.

ui/src/api.ts
```diff
@@ -1,4 +1,7 @@
 import qs from "qs";
+import { get as getStore, writable } from "svelte/store";
+
+export const backendAddressStore = writable("");
```
geigerzaehler, line 4: How about encoding the default address here?

```suggestion
export const backendAddressStore = writable("localhost:17246");
```

proxy/api/Cargo.toml
```diff
@@ -13,14 +13,14 @@ default-run = "radicle-proxy"
 
 [dependencies]
 anyhow = "1.0"
+argh = "0.1.4"
```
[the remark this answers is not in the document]
rudolfs: 👌 😑 👌 
https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/67de4613702e2d5eaeae4a05d9fe19af171765b1

cypress/plugins/index.ts
```diff
@@ -0,0 +1,242 @@
...
+    // This command waits on the proxy process to start listening on its HTTP
+    // API port before returning.
+    async startNode(id: NodeId) {
+      console.log(`  ### startNode(${id})`);
+
+      const node: Node = {
+        id: id,
+        httpListen: id,
+        peerListen: id,
+        workspacePath: path.join(CYPRESS_WORKSPACE_PATH, uuid.v4()),
```
geigerzaehler, line 89: With support for `RAD_HOME` landing in the proxy (#1527) we should use this we should set it to this directory. Otherwise it looks to me like the proxies will share state?

proxy/api/src/context.rs
```diff
@@ -32,6 +32,22 @@ impl Context {
...
+    /// Returns the [`net::SocketAddr`] where the HTTP API is bound to.
+    pub const fn http_listen(&self) -> net::SocketAddr {
+        match self {
+            Self::Sealed(sealed) => sealed.http_listen,
+            Self::Unsealed(unsealed) => unsealed.http_listen,
+        }
+    }
+
+    /// Returns the default seeds that will be written to the settings kv store.
+    pub const fn default_seeds(&self) -> &Vec<String> {
```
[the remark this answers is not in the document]
rudolfs: Changing this leads to:
```
    Checking api v0.1.0 (/Users/rudolfs/work/radicle-upstream/proxy/api)
error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> api/src/context.rs:45:9
   |
45 | /         match self {
46 | |             Self::Sealed(sealed) => &sealed.default_seeds,
47 | |             Self::Unsealed(unsealed) => &unsealed.default_seeds,
48 | |         }
   | |_________^

error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
  --> api/src/context.rs:46:37
   |
46 |             Self::Sealed(sealed) => &sealed.default_seeds,
   |                                     ^^^^^^^^^^^^^^^^^^^^^

error: aborting due to 2 previous errors
```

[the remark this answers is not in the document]
FintanH: Ah because of the `const`. No biggy really. I'm not even sure what the benefits are to be quite honest.

proxy/api/src/process.rs
```diff
@@ -1,6 +1,9 @@
 //! Provides [`run`] to run the proxy process.
+// Otherwise clippy complains about FromArgs
+#![allow(clippy::default_trait_access)]
```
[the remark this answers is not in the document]
rudolfs: > I think it's complaining because of the generated code.

Most likely. This probably should be addressed in the `argh` codebase.

cypress/plugins/index.ts
```diff
@@ -0,0 +1,242 @@
...
+      await waitOn({ resources: [`tcp:${HOST}:${id}`] });
+      console.log(`    [${id}] node started successfully`);
+
+      nodes.push(node);
+
+      return null;
+    },
+    // Onboard a node by requesting and storing an auth token and creating
+    // an identity.
+    async onboardNode(id: NodeId) {
```
geigerzaehler, line 139: It does not seem necessary to me to run this and other tasks (except `startNode` and `killAllNodes`) on the backend side of cypress. It should be possible to just turn these into regular functions and not use the `task` mechanism at all.

geigerzaehler (commented):
Great work @rudolfs. This will be immensely helpful!

proxy/api/src/process.rs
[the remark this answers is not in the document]
FintanH: Yaw, sure. Was just commenting on my findings :) This is a `pedantic` complaint anyway. It's not something that would majorly affect our understanding of the code :+1:

cypress/plugins/index.ts
FintanH: They wouldn't share state because of the random UUID afaict, but incorporating `RAD_HOME` is a good shout regardless because one of our tests later should be about seeing new commits. Env variables can be per process, right? :thinking:

geigerzaehler: > They wouldn't share state because of the random UUID

If I understand this code correctly a random UUID is only used in the workspace path but the workspace path itself is never used. But isolation is provided through the `--test` flag so maybe we just don’t need the workspace path.

FintanH: Mmmm good point, I actually didn't trace where this path was used!

cypress/plugins/index.ts
```diff
@@ -0,0 +1,242 @@
...
+  _config: Cypress.PluginConfigOptions
+): void => {
+  on("task", {
+    // Kill any nodes that were started by this plugin.
+    async killAllNodes() {
+      console.log("  ### killAllNodes()");
+
+      nodes.forEach(async node => {
+        try {
+          await exec(`kill -9 ${node.pid}`);
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/e7e4b1f0eabee325f09259cd15ac7e279fc7afc7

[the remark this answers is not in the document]
FintanH: :relaxed: :hocho:

cypress/plugins/index.ts
```diff
@@ -0,0 +1,242 @@
...
+
+const sleep = async (ms: number) => {
+  await new Promise(resolve => setTimeout(resolve, ms));
+};
+
+type NodeId = number;
+
+interface Node {
+  id: NodeId;
+  httpListen: number;
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/47671ba4bb39d6d0748b58b4b5c1f52a3b6f658e

DEVELOPMENT.md
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/e709a488ac28155a690c1b40068ae3da35aae615

cypress/plugins/index.ts
rudolfs: > With support for RAD_HOME landing in the proxy (#1527) we should use this we should set it to this directory. Otherwise it looks to me like the proxies will share state?

The intention of `CYPRESS_WORKSPACE_PATH` is to store anything that is needed for a test-case scoped to a running node. It's for storing newly created project or project checkout working directories, so that we can inspect the contents of those directories from within a test.

This is _not_ the path where we store the internal monorepo/keys/sled/electron things. The internal state is kept in a tempdir when we run tests via the `--test` flag.

We keep the `CYPRESS_WORKSPACE_PATH` within the `radicle-upstream` repository, so that it would work both locally as well as on CI.

Does this clear things up?

> If I understand this code correctly a random UUID is only used in the workspace path but the workspace path itself is never used. But isolation is provided through the --test flag so maybe we just don’t need the workspace path.

The workspace path is used in the test: https://github.com/radicle-dev/radicle-upstream/pull/1492/files#diff-c1135bec4b86c0f227a805cabfd0d9933c669329ede40a2ef36a2c031c066264R21-R22.

I wonder if there's a way to make this more clear, perhaps a comment?

geigerzaehler: Thanks for the explanation @rudolfs. That clears things up.

As an alternative (since the workspace path and the node are not intrinsically linked) we could also prepare the temporary directory independently of the node in the test case. Similar to what we do in [other tests](https://github.com/radicle-dev/radicle-upstream/blob/929f4593b70a419a88141b26d65751a35c064768/cypress/integration/project/checkout.spec.ts#L5).

ui/index.ts
rudolfs: Good call! I wasn't sure whether the store would work before the `App` component is bound to the DOM. But it does, and the result is much cleaner.

https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/fa20392eb28f7b53d4ef4712a8c6350b61c0b42a

ui/App.svelte
rudolfs: Is this description clearer?
https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/fa20392eb28f7b53d4ef4712a8c6350b61c0b42a

ui/index.ts
rudolfs: Even cleaner with `qs`: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/4420975ab7cfc41d091792d380503bc29ab62e4b.

ui/src/api.ts
rudolfs: 👍  https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/4420975ab7cfc41d091792d380503bc29ab62e4b

ui/index.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/4420975ab7cfc41d091792d380503bc29ab62e4b

cypress/integration/networking.spec.ts
rudolfs: That was my initial approach, I wanted to start the proxy listening on random ports via `radicle-proxy --http-listen 127.0.0.1:0 --peer-listen 0.0.0.0:0` and then read out the ports that the OS picked from the logs on STDOUT/STDERR. But I got stuck when I couldn't figure out a way to read out the actual port that `warp` bound to.

So I opted for specifying the ports manually as a first step to move forward with the implementation.
I'd love to work on this as a follow up, but will need some help figuring out how to get the bound ports.

cypress/plugins/index.ts
```diff
@@ -0,0 +1,242 @@
...
+import waitOn from "wait-on";
+import * as uuid from "uuid";
+import * as fs from "fs-extra";
+
+const exec = util.promisify(childProcess.exec);
+
+const HOST = "127.0.0.1";
+const ROOT_PATH = path.join(__dirname, "../../");
+const CYPRESS_WORKSPACE_PATH = path.join(ROOT_PATH, "cypress/workspace");
+const PROXY_PATH = path.join(ROOT_PATH, "proxy/target/debug/radicle-proxy");
```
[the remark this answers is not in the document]
rudolfs: We could extract this as a parameter and then specify from within the test. But at the moment I don't see the need for it yet. Let's cross the bridge when we get there.

ui/App.svelte
FintanH: Ya super clear :ok_hand:

cypress/integration/networking.spec.ts
FintanH: Ya, I was thinking more like:
```
getRandomPort() => number {
  let available = false;
  let port = random(9000, 10_000);
  while !available { 
    available = test_port(port);
  }
}
```

Where there's some utility for checking a port is taken by the OS. But definitely makes sense to do it in a follow-up

rudolfs: Yea, that could also work. I just wanted to avoid writing the utility and eliminate a possible race condition. Will keep this in mind for the follow-up. Marking this as resolved for now.

cypress/plugins/nodeManager/commands.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+export const asNode = (node: OnboardedNode, callback: () => void): void => {
+  cy.setCookie("auth-token", node.authToken);
+  // NB: it is important that we pass `localhost` here instead of `127.0.0.1`.
+  // I haven't figured out why Cypress would treat it differently, but it does.
+  cy.visit(`./public/index.html?backend=localhost:${node.httpPort}`);
+
+  callback();
+};
```
geigerzaehler, line 51: Since there is no cleanup associated with this function we don’t need a callback. In a test case we can just do
~~~javascript
cmd.asNode(node1)
// do whatever
~~~

cypress/plugins/nodeManager/plugin.ts
```diff
@@ -0,0 +1,294 @@
...
+}
+
+const nodeManager = new NodeManager();
+
+export const nodeManagerPlugin = {
+  [Commands.StartNode]: async (options: StartNodeOptions): Promise<null> => {
+    await nodeManager.startNode(options);
+
+    return null;
+  },
```
geigerzaehler, line 271: Is there a particular reason we return `null` instead of not returning anything and using `Promise<void>` as the return type?

cypress/plugins/nodeManager/commands.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const withTwoConnectedNodes = (
+  callback: (node0: OnboardedNode, node1: OnboardedNode) => void
+): void => {
+  startNode({ id: 17000 });
+  onboardNode({ id: 17000 });
+
+  startNode({ id: 18000 });
+  onboardNode({ id: 18000 });
```
geigerzaehler, line 33: We could return `OnboardedNode` from this function. This would eliminate the need for `Commands.GetOnboardedNodes` (which is susceptible to exceptions if the number of onboarded nodes does not match the number we expect`.

cypress/integration/networking.spec.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/e46870ae78be463f7244faa979625f2a267cc92b

rudolfs: Added cleaner separation between the plugin "backend" and "frontend". Commands are properly typed now.
https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/e46870ae78be463f7244faa979625f2a267cc92b

cypress/plugins/index.ts
rudolfs: Added the tempdir wrapper to our existing `commands.ts`: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/e46870ae78be463f7244faa979625f2a267cc92b.

cypress/plugins/nodeManager/plugin.ts
rudolfs: Yes, it's documented in the Cypress [task API](https://docs.cypress.io/api/commands/task.html#Usage):

> The task plugin event handler can return a value or a promise. The command will fail if undefined is returned or if the promise is resolved with undefined. This helps catch typos or cases where the task event is not handled.
>
> If you do not need to return a value, **explicitly return null to signal that the given event has been handled**.

FintanH (dismissed):
All good from my point of view

cypress/plugins/nodeManager/commands.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/788844a40a36860f296e65631c61cb99733deda2

geigerzaehler (dismissed):
Very cool :sunglasses:!

cypress/support/nodeManager/commands.ts
```diff
@@ -0,0 +1,28 @@
+import type { OnboardedNode } from "../../plugins/nodeManager/shared";
+import { Commands } from "../../plugins/nodeManager/shared";
+
```
geigerzaehler, line 3: The nesting of this file in the `nodeManager` directory seems kind of redundant. We could just name the file `cypress/support/NodeManager.ts`.

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/21b8021537085b85d2b39bb8e6c86bc19caeb581

geigerzaehler (dismissed):

cypress/plugins/nodeManager/commands.ts
rudolfs: I don't want to spend more time refactoring before we know a bit more about what real life test-cases will look like.
While a callback wrapping approach might be more type-safe, it might make testing non-linear process lifecycles more obfuscated.

I suggest we write a couple of tests and do a follow-up if necessary.

cypress/plugins/index.ts
rudolfs: See https://github.com/radicle-dev/radicle-upstream/pull/1492#discussion_r565382281.

cypress/plugins/nodeManager/plugin.ts
```diff
@@ -0,0 +1,373 @@
...
+
+  // StartedNode
+  process: childProcess.ChildProcess;
+
+  // OnboardedNode
+  authToken: AuthToken;
+  peerAddress: PeerAddress;
+}
+
+type NodeState = ConfiguredNode | StartedNode | OnboardedNode;
```
geigerzaehler, line 78: Neat!

geigerzaehler (approved):
I only have one suggestoin. Feel free to ignore it

FintanH (approved):

cypress/plugins/nodeManager/plugin.ts
```diff
@@ -0,0 +1,373 @@
...
+// Because it's not possible to mix tagged union types and extending
+// interfaces, we have to repeat the "inherited" attributes in each state.
+interface ConfiguredNode {
+  kind: "configured";
+
+  // ConfiguredNode
+  id: NodeId;
+  httpPort: number;
+  peerPort: number;
+  proxyBinaryPath: string;
```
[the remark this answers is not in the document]
rudolfs: That is a great idea!
https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/c2bfc221bb348be7a3fafcbbeabab2f738ebc01a

cypress/plugins/nodeManager/plugin.ts
```diff
@@ -0,0 +1,345 @@
...
+  log(message: string) {
+    const indentation = " ".repeat(this.indentationLevel * 2);
+    console.log(indentation + this.prefix + message);
+  }
+}
+
+// Because it's not possible to mix tagged union types and extending
+// interfaces, we have to repeat the "inherited" attributes in each node state.
+interface ConfiguredNode {
+  kind: "configured";
```
FintanH, line 39: Not a blocker, but isn't common to create an enum for these `kind`s?

cypress/plugins/nodeManager/plugin.ts
```diff
@@ -0,0 +1,345 @@
...
+    if (this.state.kind !== "onboarded") {
+      throw new Error("Can't get peerAddress before node is onboarded");
+    }
+
+    return this.state.peerAddress;
+  }
+
+  get authToken(): AuthToken {
+    if (this.state.kind !== "onboarded") {
+      throw new Error("Can't get peerAddress before node is onboarded");
```
FintanH, line 75: This OOP state stuff feels off to me. I'd expect a cleaner solution out of having state transition functions:
```
configure : ConfiguredNode
start : ConfiguredNode -> Result<StartedNode>
onboard: StartedNode -> Result<OnboardedNode>
```

then your `OnboardedNode` functions would _only_ work on those. I dunno how easy this is to represent in TS but I feel like it's possible. I imagine the `Node` would be generic over `T` where that would be filled in by our different node states. This rules out these state-specific errors.

Food for thought and maybe worth a follow-up ticket.

FintanH (approved):
Ya looks good to me. Just some comments to think about :+1:

cypress/plugins/nodeManager/plugin.ts
rudolfs: Not in Typescript, if I understand the [docs](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) correctly:

> A common technique for working with unions is to have **a single field which uses literal types** which you can use to let TypeScript narrow down the possible current type.

Maybe @geigerzaehler knows more?

cypress/plugins/nodeManager/plugin.ts
rudolfs: > This OOP state stuff feels off to me.

I agree, maybe we can remove the inner Node class, but then the NodeManager would have to keep an array for each different state.

Will do a follow up once I know how this can be done better in Typescript.

cypress/plugins/nodeManager/plugin.ts
FintanH: Here's an example: https://github.com/radicle-dev/radicle-upstream/blob/master/ui/src/project.ts#L22-L36

cypress/plugins/nodeManager/plugin.ts
FintanH: Ya, that's a good point. But could be interesting in that it only needs to inspect the specific arrays when want to promote nodes :)

> Will do a follow up once I know how this can be done better in Typescript.
:ok_hand:

FintanH (dismissed):

cypress/plugins/nodeManager/plugin.ts
rudolfs: Is there any added benefit in wrapping the sate in an enum? Typescript already guards against using the wrong string.

Screenshot 2021-01-28 at 14 23 13

Screenshot 2021-01-28 at 14 21 44

It seems the enum version is slightly more verbose, and if there is no added benefit, I'd rather go with where I have to type less.

rudolfs: I guess using enums is a convention in this codebase, so here it is: https://github.com/radicle-dev/radicle-upstream/pull/1492/commits/f9abb6d96640f4c930325d129e14ff730c331b82.

FintanH: Ya, the linter saves us, but having an `enum` reduces the cognitive load -- you can jump to the enum and see all the cases laid out for you :) Thanks for improving it :100:

FintanH (approved):
For real this time :+1:

geigerzaehler (approved):

rudolfs:
Closed via 9770b08a5c0df7738702db96caae07b9c34a4f5a.

PR #1493, opened 2020-12-18
Add ethers-cli dependency

CodeSandwich:
Fixes `yarn start:ethereum`


NunoAlexandre (changes requested):
Thanks for this Igor but could you place it within the `devDependencies` instead?

NunoAlexandre:
Thanks man :)

PR #1501, opened 2021-01-04
refactor(ui): use TypeScript in settings screen

geigerzaehler:
We use TypeScript in the settings screen and the `SegmentedControl` component.

We also remove the use of `onMount`. It is not necessary to delay the call to `getVersion` until after the component has mounted.


rudolfs (approved):

xla (dismissed):
🐰 📣 ㊙️ 🔱

PR #1506, opened 2021-01-05
chore(release): 0.1.6

rudolfs:


FintanH (dismissed):

FintanH (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [0.1.6](https://github.com/radicle-dev/radicle-upstream/compare/v0.1.5...v0.1.6) (2021-01-05)
+
+
+### Bug Fixes
+
+* **ci:** fail after rollup warnings ([#1502](https://github.com/radicle-dev/radicle-upstream/pull/1502))
+* **ui:** make logo text color work in dark/light mode ([#1491](https://github.com/radicle-dev/radicle-upstream/issues/1491)) ([34e09fe](https://github.com/radicle-dev/radicle-upstream/commit/34e09fe2721e3c329c3827c0db9a718269491edb))
+* **ci:** put timeout on proxy tests ([#1474](https://github.com/radicle-dev/radicle-upstream/pull/1474)) ([c111d6e](https://github.com/radicle-dev/radicle-upstream/commit/c111d6eaa7963bac9d30ad0708115103d8267d61)), closes [#1378](https://github.com/radicle-dev/radicle-upstream/issues/1378) [#1379](https://github.com/radicle-dev/radicle-upstream/issues/1379)
```
cloudhead, line 12: Should CI fixes be included in the changelog? It's not part of the product/release itself, just the development process..

rudolfs: I think we never bothered to change the default config for [`standard-version`](https://github.com/conventional-changelog/standard-version). Wanna open an issue?

I also remember discussing this with @abbey-titcomb and @brandonhaslegs at some point and if I remember correctly the decision was to leave the auto-generated changelog as is and write a more "human friendly" status update manually for bigger releases.

rudolfs: I guess we could explicitly define the list in `package.json`, something like: 
```json
...
  "standard-version": {
    "skip": {
      "tag": true
    },
    "types": [
      {"type": "feat", "section": "Features"},
      {"type": "fix", "section": "Bug Fixes"},
      {"type": "chore", "hidden": true},
      {"type": "docs", "hidden": true},
      {"type": "style", "hidden": true},
      {"type": "refactor", "hidden": true},
      {"type": "perf", "hidden": true},
      {"type": "test", "hidden": true}
      {"type": "ci", "hidden": true}
    ]
  },
...
```


According to: https://github.com/conventional-changelog/conventional-changelog-config-spec/blob/HEAD/versions/2.1.0/README.md#types.

abbey-titcomb: Yeah, I think keeping the rolling releases announcements focused on the changelog is good, but if we can remove irrelevant information, then I'm all for cleaning it up — which is what it seems like you did with the above ^

These should be low-touch and something that doesn't require too much coordination to cut and push public.

rudolfs: Hm, just realised that the configuration is correct, but the commit messages looked like this:
```
fix(ci): fail after rollup warnings
fix(ci): put timeout on proxy tests
```

`ci` comes from the "scope" part of the commit message. So I guess we should write down what are the allowed scopes that we should use and make everyone stick to the convention.

Will clean up the changelog manually this time.

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1506/commits/951a44891d37d4669223afa4d7be8072d833af29

\cc @geigerzaehler

PR #1509, opened 2021-01-06
chore(ui): update cypress

geigerzaehler:
Update Cypress from v5 to v6. The only breaking change we needed to fix was that the `not.visible` assertion now throws if the element does not exist. We change the affected occurrences to use the `not.exists` assertion.


rudolfs (approved):

MeBrei (approved):
⭐

PR #1511, opened 2021-01-07
fix(ui): use alternative support channel link

### LINKED ISSUE #1483: Button "Reach out to support" redirects to a broken link
Hey guys, 

just tried to get radicle working on macOS and while getting errors I tried to use the `Reach out to support` button which redirected me to https://matrix.to/#/%23support:radicle.community , which is a page to create sharable links. Looks like the link is broken:

Bildschirmfoto 2020-12-10 um 12 20 26

MeBrei:
The second `#` in the url still get's encoded, but the link works as expected nevertheless. 
This is a quick fix. As we allow to open urls from eg readme's of projects, this might cause other problems. After timeboxed investigation on this I was not able to find out why this is not properly decoded or how the encoding could be prevented. 

Closes #1483 

Signed-off-by: Merle Breitkreuz <merle@monadic.xyz>

### DIFF
--- a/ui/Screen/Bsod.svelte
+++ b/ui/Screen/Bsod.svelte
@@ -27,7 +27,8 @@
   };
 
   const support = () => {
-    window.location.href = "https://matrix.to/#/#support:radicle.community";
+    window.location.href =
+      "https://matrix.radicle.community/#/room/#support:radicle.community";
   };
 </script>
 


rudolfs (approved):
LGTM 🚀 


https://user-images.githubusercontent.com/158411/103919788-c0befa00-5110-11eb-88a0-57053a9e283b.mov

abbey-titcomb (approved):
LGTM. Thanks Merle!

ui/Screen/Bsod.svelte
```diff
@@ -27,7 +27,8 @@
   };
 
   const support = () => {
-    window.location.href = "https://matrix.to/#/#support:radicle.community";
+    window.location.href =
+      "https://matrix.radicle.community/#/room/#support:radicle.community";
```
xla, line 31: Should we account for the potential rename to `radicle:support`?

MeBrei: For the `general` channel, only the name was changed, but not the address. The link is still https://matrix.radicle.community/#/room/#general:radicle.community
So if we do the same for the support channel, the link would not have to be changed.
Screenshot 2021-01-08 at 10 59 00

MeBrei:
Closed via 8bbf4614a9a547961f99e994237d847cad07a2b3

PR #1519, opened 2021-01-13
refactor(ui): notifications with multiple actions

### LINKED ISSUE #1270: Divider should be full height on notifications
The divider line between the message and the action in notifications should be full-height. Right now there's padding around it. See design file [here](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=586%3A0).

image

geigerzaehler:
* Notifications can now have multiple actions
* The action divider in notifications now covers the whole notification height. Fixes #1270.
* The `notification.info` and `notification.error` functions now take parameter objects as a single argument instead of multiple arguments. This provides a cleaner API with respect to default values.
* We eliminate the event logic from notifications to simplify the code

### DIFF
--- a/ui/Screen/DesignSystemGuide.svelte
+++ b/ui/Screen/DesignSystemGuide.svelte
@@ -131,6 +131,24 @@
       value: 2,
     },
   ];
+
+  const infoNotification = params => {
+    return {
+      level: notification.Level.Info,
+      message: params.message,
+      showIcon: params.showIcon || false,
+      actions: params.actions || [],
+    };
+  };
+
+  const errorNotification = params => {
+    return {
+      level: notification.Level.Error,
+      message: params.message,
+      showIcon: params.showIcon || false,
+      actions: params.actions || [],
+    };
+  };
 </script>
 
 <style>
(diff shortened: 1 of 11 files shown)


ui/Screen/DesignSystemGuide.svelte
```diff
@@ -131,6 +131,24 @@
       value: 2,
     },
   ];
+
+  const infoNotification = params => {
+    return {
+      level: notification.Level.Info,
+      message: params.message,
+      showIcon: params.showIcon === undefined ? false : params.showIcon,
```
NunoAlexandre, line 139:
```suggestion
      showIcon: params.showIcon || false,
```

I find this syntax more concise but it's up to you

NunoAlexandre (approved):
This looks great! :ok_hand: really like the change from having all the notifications attributes gathered.

What is the context of this work though? Haven't heard about this until now :smile:

geigerzaehler:
> What is the context of this work though? Haven't heard about this until now

This will be needed for the app update work

Screenshot from 2021-01-13 18-08-04

NunoAlexandre:
> > What is the context of this work though? Haven't heard about this until now
> 
> This will be needed for the app update work
> 
> Screenshot from 2021-01-13 18-08-04

Sweet!

ui/Screen/DesignSystemGuide.svelte
```diff
@@ -131,6 +131,24 @@
...
+      showIcon: params.showIcon || false,
+      actions: params.actions || [],
+    };
+  };
+
+  const errorNotification = params => {
+    return {
+      level: notification.Level.Error,
+      message: params.message,
+      showIcon: params.showIcon === undefined ? false : params.showIcon,
```
NunoAlexandre, line 148:
```suggestion
      showIcon: params.showIcon || false,
```

ui/src/notification.ts
```diff
@@ -1,155 +1,83 @@
...
-const show = (
-  level: Level,
-  showIcon: boolean,
-  message: string,
-  actionText: string | false,
-  actionHandler: ActionHandler | false
-): void => {
+const show = (level: Level, params: NotificationParams): void => {
   const id = Math.random();
+  const showIcon = params.showIcon === undefined ? false : params.showIcon;
```
NunoAlexandre, line 46:
```suggestion
  const showIcon = params.showIcon || false;
```

NunoAlexandre (commented):
These two left:

MeBrei (approved):
Looks good!

A nice addition would be, to have at least one example in the design system guide with the `Close` button showing.

NunoAlexandre (approved):

geigerzaehler:
> A nice addition would be, to have at least one example in the design system guide with the `Close` button showing.

I’ve replaced the “close” button with the more generic action because no special “close” button exists anymore.

PR #1520, opened 2021-01-13
feat(infra): upload release info as part of release process

### LINKED ISSUE #1518: Information about latest release
To show update prompts to the user the app should be able to query information about the latest release. 

### Proposal

Information about the latest version will be stored in a JSON file in the same GCS bucket as the release artifacts. This means the file can be downloaded via HTTP from `https://builds.radicle.xyz/latest.json`. As part of the release process this file will be updated.

The JSON file have the following content

```json
{
  "version": "0.1.6",
  "announcementUrl": "https://radicle.community/t/radicle-upstream-v0-1-6-is-out"
}
```

The announcement URL will be used by the app to link to the release.

### Tasks

- [ ]  Create a script that updates the release JSON file and document its use in the release process
- [ ]  Fetch the file in the app

### Alternatives

*Git Tags*. We would need either need to query Github or Radicle. Both are more elaborate than an simple HTTP request. In addition, existence of a tag does not mean an new version has been released. It still has to be QA’ed.

*Github releases*. We would like to avoid creating an additional dependency on Github.

geigerzaehler:
We add a step to the release process that uploads information about the latest release to `builds.radicle.xyz`. Fixes #1518.


scripts/set-latest-release.ts
```diff
@@ -0,0 +1,85 @@
...
+// that has permissions to write to the `builds.radicle.xyz` bucket.
+
+const fetch = require("node-fetch");
+const path = require("path");
+const fs = require("fs");
+const childProcess = require("child_process");
+
+const pkg = require("../package.json");
+
+const fileName = "latest.json";
```
rudolfs, line 16: Should we add this to `.gitignore`?

DEVELOPMENT.md
```diff
@@ -622,6 +632,7 @@ for all of the required steps:
   - [ ] deploy docs.radicle.xyz
 - [ ] announce new release on radicle.community
 - [ ] announce new release on the matrix #general:radicle.community channel
+- [ ] update release information
```
rudolfs, line 635:
```
- [ ] trigger an in-app notification about the latest app release
```

DEVELOPMENT.md
```diff
@@ -602,6 +602,16 @@ sure to update all the versions and links**):
         Radicle Upstream vX.X.X is out! 🎉
         https://radicle.community/t/radicle-upstream-vX-X-X-is-out
 
+Finally, update the information about the latest release by running
```
rudolfs, line 605:
```suggestion
Finally, trigger an in-app notification about the new release by running:
```

scripts/set-latest-release.ts
```diff
@@ -0,0 +1,85 @@
...
+      `https://radicle.community/t/radicle-upstream-v${versionDash}-is-out`
+    );
+    if (!response.ok) {
+      throw new Error(
+        `Announcement url ${annoucementUrl} does not exist. Response status is ${response.status}`
+      );
+    }
+
+    const latestPath = path.join(tempDir, fileName);
+    await fs.promises.writeFile(
```
rudolfs, line 37: TIL `fs.promises` 🔥

package.json
```diff
@@ -91,6 +92,7 @@
     "jest": "^26.6.3",
     "lint-staged": "^10.5.3",
     "lodash": "^4.17.20",
+    "node-fetch": "^2.6.1",
```
rudolfs, line 95: Nice, I'll also need this for my p2p testing.

scripts/set-latest-release.ts
geigerzaehler: This file is created in a temporary directory so we don’t need to add it to `.gitignore`.

rudolfs (dismissed):

CodeSandwich (dismissed):
:package:!

PR #1523, opened 2021-01-13
fix(ui): preserve NewProject Modal layout

### LINKED ISSUE #1515: Dismissing the hint in the New Project modal breaks the modal layout
Close the hint about adding to your shell configuration to see the modal's layout break.

image

NunoAlexandre:
Fixes #1515


ui/Modal/NewProject.svelte
```diff
@@ -153,6 +153,7 @@
     flex-direction: column;
     justify-content: center;
     text-align: center;
+    width: -webkit-fill-available;
```
geigerzaehler, line 156: The following has the same behavior but uses standardized CSS functionality so I’d prefer it.

```suggestion
    align-self: stretch;
```

NunoAlexandre: TIL :cowboy_hat_face: 

The same behaviour can also be achieved using `width: 100%;`, I picked `width: -webkit-fill-available;` as 1) I found it more idiomatic and 2) because I see it being used in other places already.

Which of the 3 do you prefer?

geigerzaehler: I think `width: 100%` would be the most obvious one.

NunoAlexandre: Done!

brandonhaslegs (approved):

rudolfs (approved):

geigerzaehler (approved):

NunoAlexandre:
Merged in https://github.com/radicle-dev/radicle-upstream/commit/f41a7599dfe6603627120c32be3fc5f19d7228e9

PR #1525, opened 2021-01-15
refactor(tests): strict type-checking for cypress code

geigerzaehler:
The TypeScript project configuration for the cypress code now extends the main project configuration to achieve parity. This means that `strict` type-checking is now enabled, which requires some fixes.


rudolfs (approved):
LGTM 🚀

MeBrei (approved):

PR #1526, opened 2021-01-18
feat(ui): more improved notifications

geigerzaehler:
Some more improvements no notifications

* Notifications can now have the `primary` style variant. This is required for update notifications
* The clickable area of notification actions is increased and covers the whole notification height. In addition we now use a `button` element to better indicate things are clickable.
* Simplify CSS selectors

### DIFF
--- a/ui/DesignSystem/Component/Notification.svelte
+++ b/ui/DesignSystem/Component/Notification.svelte
@@ -1,15 +1,25 @@
 <script lang="typescript">
-  import { Level } from "../../src/notification";
+  import type { SvelteComponent } from "svelte";
   import type { Notification } from "../../src/notification";
   import { Icon } from "../Primitive";
 
   export let notification: Notification;
   export let style = "";
 
-  const icon =
-    notification.level === Level.Info
-      ? Icon.InfoCircle
-      : Icon.ExclamationCircle;
+  const variantClass = `variant-${notification.variant.toLowerCase()}`;
+
+  let icon: typeof SvelteComponent | null;
+  switch (notification.icon) {
+    case "InfoCircle":
+      icon = Icon.InfoCircle;
+      break;
+    case "ExclamationCircle":
+      icon = Icon.ExclamationCircle;
+      break;
+    case null:
+      icon = null;
+      break;
+  }
 </script>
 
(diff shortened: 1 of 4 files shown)


ui/DesignSystem/Component/Notification.svelte
```diff
@@ -6,6 +6,10 @@
   export let notification: Notification;
   export let style = "";
 
+  const variantClass = notification.styleVariant
+    ? `variant-${notification.styleVariant}`
+    : "";
+
```
MeBrei, line 12: Could we instead just introduce another `level`? This would simplify the setup (only one parameter influencing the color)

ui/DesignSystem/Component/Notification.svelte
```diff
@@ -65,9 +70,15 @@
   .info .action-divider {
     background-color: var(--color-foreground-level-6);
   }
+
+  .variant-primary .action-divider {
+    background-color: var(--color-primary-level-1);
+  }
```
MeBrei, line 76: Is this the desired color? It looks brighter than for the other notifications:
Screenshot 2021-01-21 at 13 05 30

ui/DesignSystem/Component/Notification.svelte
geigerzaehler: You’re right. I think this is a better approach. But then I’d move away from the term “level” and use “variant” because it’s usually used for logging levels.

@brandonhaslegs do you have a better suggestion than calling the variant “primary”?

ui/DesignSystem/Component/Notification.svelte
geigerzaehler: I0ve checked. This is the correct color, taken directly from the figma design.

rudolfs (approved):

ui/DesignSystem/Component/Notification.svelte
MeBrei: @brandonhaslegs Maybe you can take a look if this needs work (in a different PR). In the Dark-theme, the vertical dividers also are not all the same.

MeBrei (approved):

PR #1537, opened 2021-01-20
fix(ui): resolve 0.1.7 visual regressions

NunoAlexandre:
Fixes the visual regressions found in https://github.com/radicle-dev/radicle-upstream/issues/1535#issuecomment-763782263

### DIFF
--- a/ui/DesignSystem/Component/SidebarLayout.svelte
+++ b/ui/DesignSystem/Component/SidebarLayout.svelte
@@ -14,7 +14,7 @@
 <style>
   .container {
     position: relative;
-    margin-left: var(--sidebar-width);
+    left: var(--sidebar-width);
     width: calc(100vw - var(--sidebar-width));
     min-width: 60rem;
     height: 100%;
--- a/ui/DesignSystem/Component/SidebarLayout.svelte
+++ b/ui/DesignSystem/Component/SidebarLayout.svelte
@@ -15,7 +15,8 @@
   .container {
     position: relative;
     margin-left: var(--sidebar-width);
-    min-width: calc(100vw - var(--sidebar-width));
+    width: calc(100vw - var(--sidebar-width));
+    min-width: 60rem;
     height: 100%;
   }
 </style>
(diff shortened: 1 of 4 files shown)


ui/DesignSystem/Component/SidebarLayout.svelte
```diff
@@ -15,7 +15,8 @@
   .container {
     position: relative;
     margin-left: var(--sidebar-width);
-    min-width: calc(100vw - var(--sidebar-width));
+    width: calc(100vw - var(--sidebar-width));
+    min-width: 60rem;
```
NunoAlexandre, line 19: This reverts the changes introduces in bcfb9e012d921ddb3c53f82dab13ab258b23deec. That means that the funding feature won't look proper on a small viewport but that is fine for now. 

This way we have things working as well as before for our users and I will find a viable way to have it working well for all cases.

juliendonck: Was looking through these changes and reverts and saw that this was a bit different originally, using `left: ` in stead of `margin-left: `. Thought it might make sense to revert it to the original to avoid running into other issues.
```suggestion
    position: relative;
    left: var(--sidebar-width);
    width: calc(100vw - var(--sidebar-width));
    height: 100%;
    min-width: 60rem;
```

NunoAlexandre: I have only seen Left used when using position: absolute, thus that change. If it works with margin-left I wouldn't regress that change.

ui/DesignSystem/Component/SidebarLayout.svelte
```diff
@@ -15,7 +15,8 @@
   .container {
     position: relative;
     margin-left: var(--sidebar-width);
```
NunoAlexandre, line 17:
```suggestion
    left: var(--sidebar-width);
```

MeBrei (approved):

geigerzaehler (approved):

rudolfs (approved):

NunoAlexandre:
Merged in 903fb72ce8426eaceeefefa366cbe94588840ae0.

PR #1541, opened 2021-01-21
Improve funding-related bits

NunoAlexandre:
Three small funding-related improvements:

1. Validate that a given input value can be converted to `BigNumber` before doing so.
     To reproduce, on master go to any place where you can input/edit a funding-related numeric value (top up,  withdraw, edit support) and enter `9f` e.g. The app will crash as there will be an underlying attempt to convert that input to `BigNumber` which fails by throwing. We fix that here by first validating the input to be a valid BigNumber.

2. We make the transaction hash in the Transaction modal copyable. This is helpful to verify the status of a tx in, for example, etherscan.

3. Other small things like avoiding calling toString on a string value.

### DIFF
--- a/ui/src/funding/pool.ts
+++ b/ui/src/funding/pool.ts
@@ -339,7 +339,7 @@ export function isValidBigNumber(value: string): boolean {
   try {
     BigNumber.from(value);
     return true;
-  } catch (_) {
+  } catch {
     return false;
   }
 }
--- a/ui/src/funding/pool.ts
+++ b/ui/src/funding/pool.ts
@@ -294,7 +294,10 @@ export const balanceValidationStore = (
 ): validation.ValidationStore => {
   return validation.createValidationStore(constraints.topUpAmount, [
     {
-      promise: amount => Promise.resolve(BigNumber.from(balance).gte(amount)),
+      promise: amount =>
+        Promise.resolve(
+          isValidBigNumber(amount) && BigNumber.from(balance).gte(amount)
+        ),
       validationMessage: "Insufficient balance",
     },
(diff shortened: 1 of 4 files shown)


ui/src/funding/pool.ts
```diff
@@ -328,3 +331,15 @@ function newSetOfReceivers(current: Receivers, changes: Receivers): Receivers {
     [...merged].filter(([_address, status]) => status != ReceiverStatus.Removed)
   );
 }
+
+// Check whether the given value can be converted to `ethers.BigNumber`.
+// N.B: ethers.BigNumber.isBigNumber doesn't work as expected:
+// https://github.com/ethers-io/ethers.js/blob/master/packages/bignumber/src.ts/bignumber.ts#L285
+export function isValidBigNumber(value: string): boolean {
```
NunoAlexandre, line 338: It doesn't seem ideal to have this here but I don't find a better place, given we don't have a general `util` module and that this function is only used within the scope of the funding pool.

ui/src/funding/pool.ts
```diff
@@ -328,3 +331,15 @@ function newSetOfReceivers(current: Receivers, changes: Receivers): Receivers {
...
+  try {
+    BigNumber.from(value);
+    return true;
+  } catch (_) {
```
rudolfs, line 342:
```suggestion
  } catch {
```

rudolfs (approved):

geigerzaehler (approved):

NunoAlexandre:
merged in dc8b347b4c0656434daf029967dd4809bfef1531

PR #1542, opened 2021-01-21
fix(dev): use locally installed ethers-cli

geigerzaehler:
We don’t require `ethers` to be in PATH to run `./scripts/ethereum-dev-node.sh`. Instead we use the binary from the installed node modules.

### DIFF
--- a/scripts/ethereum-dev-node.sh
+++ b/scripts/ethereum-dev-node.sh
@@ -5,7 +5,7 @@
 
 set -eumo pipefail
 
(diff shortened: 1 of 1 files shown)


NunoAlexandre (approved):
TIL on `npx` :+1:

xla (approved):
🎏 🕊 ♎️ 🔗

geigerzaehler:
Closed in d6a7e75f2cdbb82a4447f61ebb0a0800a20366fd

PR #1548, opened 2021-01-22
chore(dev): update rust and node toolchain

geigerzaehler:
A bunch of changes that update our toolchains and fix some issues with how they were used.

* Switch to rust nightly-2021-01-21
* Switch to Node v14. Node v12 is not actively maintained anymore.
* Add `.nvmrc` to indicate the node version we are using
* Re-run `cargo fmt`
* Fix clippy issues for new lints
* Ignore `clippy::large_types_passed_by_value` lint. This limits us in the design of our APIs and may force _less efficient_ code if the function eventually requires ownership of the value.
* Improvements to `Dockerfile` and CI scripts
  * Add `--no-install-recommends` to `apt get` in docker file to reduce number of dependencies.
  * Consistently use recommended curl flags `-sfLS`
  * Don’t install `cargo-deny`, it’s unused
  * Don’t overwrite `RUSTUP_HOME` in `.buildkite/env.sh`. Otherwise we don’t use the rustup toolchain from the docker image but are forced to re-install it. This also should reduce the cache size.


rudolfs:
> Switch to rust nightly-2021-01-21

Can we do this independently of radicle-link and radicle-surf?

> Add .nvmrc to indicate the node version we are using

What part of the toolchain picks this up? I'm using `asdf` for version management which uses `.node-version`.

geigerzaehler:
> > Switch to rust nightly-2021-01-21
> 
> Can we do this independently of radicle-link and radicle-surf?

Yes, we’re not in sync anyways.


> > Add .nvmrc to indicate the node version we are using
> 
> What part of the toolchain picks this up? I'm using `asdf` for version management which uses `.node-version`.

I’m using [nvm](https://github.com/nvm-sh/nvm). I think it makes sense to also create a `.node-version`. (We can symlink `.nvmrc` to that file.)

proxy/coco/src/project/create/validation.rs
```diff
@@ -173,8 +173,8 @@ impl Repository {
                 let repo = git2::Repository::open(path.clone())
                     .or_matches(git_ext::is_not_found_err, || Err(Error::NotARepo(path)))?;
 
-                let _ = Self::existing_branch(&repo, &default_branch)?;
-                let _ = Self::existing_remote(&repo, &url)?;
+                Self::existing_branch(&repo, &default_branch)?;
```
FintanH, line 176: Is this part of clippy/fmt? I tend to prefer knowing when a value that isn't `()` is returned but is being ignored. But that's a personal preference and not a hill I'll die on :)

FintanH (approved):

proxy/api/src/lib.rs
```diff
@@ -22,7 +22,8 @@
     clippy::or_fun_call,
     clippy::shadow_reuse,
     clippy::clippy::option_if_let_else,
```
xla, line 24:
```suggestion
    clippy::option_if_let_else,
```

Seems to be bogus otherwise.

proxy/coco/src/git_helper.rs
```diff
@@ -22,7 +22,7 @@ pub const GIT_REMOTE_RAD: &str = "git-remote-rad";
 ///   * Could not get the current working directory.
 ///   * Could not create the path to binary directory.
 ///   * Could not copy helper executable to the binary directory.
-pub fn setup(src_dir: &path::PathBuf, dst_dir: &path::PathBuf) -> Result<(), Error> {
+pub fn setup(src_dir: &path::Path, dst_dir: &path::Path) -> Result<(), Error> {
```
xla, line 25: What's the rationale for this change?

proxy/coco/src/project/create/validation.rs
xla: I second this sentiment.

xla (commented):
> Consistently use recommended curl flags `-sfLS`

Could you provide the source for that? It'd also be advantageous to capture such established conventions.

> Add .nvmrc to indicate the node version we are using

Does this imply that nvm is required to develop/build Upstream?

proxy/coco/src/git_helper.rs
FintanH: I believe `&PathBuf` is equivalent to `&Path`. Analogous to `&Vec<T>` and `&[T]`

geigerzaehler:
> > Add .nvmrc to indicate the node version we are using
> 
> Does this imply that nvm is required to develop/build Upstream?

No, the file (much like `.node-version`) is completely optional to have so I’d be ok with leaving it out. But it serves as a good indicator what version of node is required.

proxy/coco/src/git_helper.rs
geigerzaehler: Yes, clippy now complains about this. (Can’t find the lint name atm.)

proxy/coco/src/project/create/validation.rs
geigerzaehler: Clippy [complains](https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_drop) because the `git2::Refernce` implements `Drop`. Alternatives for fixing this are explicit variable names
~~~
let _default_branch_ref = Self::existing_branch(&repo, &default_branch)?;
~~~
explicit drop
~~~
drop(Self::existing_branch(&repo, &default_branch)?);
~~~
or disabling this lint.

I have no preference either way and would leave it up to you.

geigerzaehler:
> > Consistently use recommended curl flags `-sfLS`
> 
> Could you provide the source for that?

Couldn’t actually find a comprehensive source. But this is what you usually see in scripts and docker files. Here’s a quick rundown of the options and why we should use them
* Without `-f` a 404 will not produce an error and the command will output the error message associated with the 404 which would result in confusion.
* Without `-s` will see the progress status which is not really helpful for automated script and often messes up logs.
* We need `-S` to show errors when `-s` is set
* Without `-L` any redirects made on the server side will not be respected and the command will output the body associated with the redirect. Almost never what you want.

> It'd also be advantageous to capture such established conventions.

Any suggestion where to do this?

proxy/coco/src/project/create/validation.rs
FintanH: Cool, thanks for the link! I think the `_name` would be my preferred convention.

geigerzaehler: > Cool, thanks for the link! I think the `_name` would be my preferred convention.

Well, that doesn’t work quite as expected:

~~~rust
let _remote = Self::existing_remote(&repo, &url)?;
Ok(Self::Existing {
    repo,
    url,
    default_branch,
})
~~~

The borrow checker complains because the lifetime of `_remote` is the same as the lifetime of `&repo`, that is the whole scope. But `repo` is moved in the final statement. I’m leaning towards just ignoring this lint. Alternatively we could create a dedicated scope around the two `Self::existing_remote` statements or use an explicit `drop`.

FintanH: I would probably opt for introducing a scope, personally. Sorry, this is a pain :sweat_smile:

geigerzaehler: No worries. It’s fixed in https://github.com/radicle-dev/radicle-upstream/pull/1548/commits/29b0e54e8a4bc63c482c1ba1d155fff6a314beb6

FintanH (approved):
Noice :+1:

xla (approved):
✋ 💳 📡 🌐

geigerzaehler:
Closed in e3897d5c44942b4a04c61ff6383ba599fc724203

PR #1550, opened 2021-01-25
feat(ui): implement app updates

geigerzaehler:
Based on #1526. Only the last commit is relevant.

We add functionality that checks for updates to the app and notifies the user if they enable it.

This does not implement the design fully: The update indicator on the settings icon in the sidebar is not implemented yet. I’ll create a follow-up task


brandonhaslegs:
> * We are not moving the “Version” section to the top of the settings when a new version is available. This is very hard to achieve with Svelte and how the file is organized.

Can't we just use Flexbox and change the position with `align-self`? Just want to make sure we've tried everything here, since it'll be quite strange to get a notification about an update, click to update, go to settings and not see any immediate call to action (and have to scroll down to the bottom of settings to find the update button). But if it's really not possible with Svelte, then maybe we should just move the Version section to always be at the top?

rudolfs:
> We are not moving the “Version” section to the top of the settings when a new version is available. This is very hard to achieve with Svelte and how the file is organized.

I think the easiest would be just to extract the "Version" section into its own component and then include this component in two places (one at the top and the other where it is now) in the Settings screen:
```html
{if newVersion}
<VersionSection />
{/if}
// ... all the other sections
{if !newVersion}
<VersionSection />
{/if}
```

ui/src/updateChecker.ts
```diff
@@ -0,0 +1,197 @@
+import * as svelteStore from "svelte/store";
+import persistenStore from "svelte-persistent-store/dist";
```
rudolfs, line 2: Is this a typo? `persistenStore` -> `persistentStore`.

brandonhaslegs:
The notification should not automatically disappear, since it has a dismiss button.

ui/src/updateChecker.ts
```diff
@@ -0,0 +1,197 @@
...
+  private currentVersion: svelteStore.Writable<
+    string | undefined
+  > = svelteStore.writable(undefined);
+
+  // Create an `UpdateChecker` and initialize it.
+  //
+  // If notifications have been enabled previously, we start checking
+  // for updates.
+  //
+  // If the user has not configured update checking we aks them to do
```
rudolfs, line 60:
```suggestion
  // If the user has not configured update checking we ask them to do
```

ui/src/updateChecker.ts
```diff
@@ -0,0 +1,197 @@
...
+      lastNotified === null ||
+      now >= lastNotified + VERSION_NOTIFY_SILENCE_INTERVAL
+    ) {
+      notification.primary({
+        message: "There is a new version of Upstream available",
+        actions: [
+          {
+            label: `Check out Version ${version}`,
+            handler: () => {
+              window.open(announcementUrl);
```
rudolfs, line 184: This doesn't work for me on macOS. Have a look at how it's done in the `<BSOD>` component for the [support button](https://github.com/radicle-dev/radicle-upstream/blob/master/ui/Screen/Bsod.svelte#L30).

rudolfs (commented):
- [ ] when I click on the "Check out Version 0.1.6" button in settings or on the notification nothing happens and I see `User tried opening URL with invalid URI scheme: about:blank` in the terminal, expected that my external browser would open a link

- [ ] we should check whether the version is actually greater than, not just equal
Screenshot 2021-01-25 at 14 19 08

geigerzaehler:
> I think the easiest would be just to extract the "Version" section into its own component 

Unfortunately that doesn’t quite work. The version component does not respect the styles for `section`, etc. since they are scoped to the component. We would need to copy the styles.

> Can't we just use Flexbox and change the position with `align-self`?

You’re right! It’s not `align-self` but we can use [`order`](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Ordering_Flex_Items#the_order_property). Fixed in c4b9f94e3a9d4e782f35323accaf073e31b8d592

> The notification should not automatically disappear, since it has a dismiss button.

Fixed in 97b1f3d4d47ab2760e60498e503339ae5f605d4e

ui/src/updateChecker.ts
geigerzaehler: Fixed in 5e26fcf738cf918b39014a3af853fb5be194490c

geigerzaehler:
> we should check whether the version is actually greater than, not just equal

Fixed in fb6c17cf704456fc13e6e6d840a6694539e83d8c

brandonhaslegs (dismissed):

rudolfs:
Wouldn't it make more sense for `latest.json` to be on `https://releases.radicle.xyz` instead of `https://builds.radicle.xyz`?

geigerzaehler:
> Wouldn't it make more sense for `latest.json` to be on `https://releases.radicle.xyz` instead of `https://builds.radicle.xyz`?

Yes! Will change this.

ui/src/updateChecker.ts
geigerzaehler: To make this testable I’ve introduced `ipc.openUrl()` in https://github.com/radicle-dev/radicle-upstream/pull/1550/commits/7b869629a3e7338366a8c350d6d73f4841878ce7

geigerzaehler:
@rudolfs: I’ve addressed all your comments. PTAL

rudolfs:
Something is still off with the version checker, it shows that there is an update even though the version of the running app is greater than the one of the update:

Screenshot 2021-01-27 at 11 44 31

Tested via:
```diff
diff --git a/package.json b/package.json
index ac657dcc..d0f9bdaf 100644
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "radicle-upstream",
   "productName": "Radicle Upstream",
-  "version": "0.1.8",
+  "version": "0.2.0",
```

rudolfs:
Might as well use `semver` for comparing versions as it is already in our transitive deps.

rudolfs:
The "new update available" notification disappears after an app reload or a page refresh without me explicitly dismissing it.

Screenshot 2021-01-27 at 11 56 29

The behaviour is different from how the "automatically check for new versions" notification works, which is shown even after an app reload or page refresh.
Screenshot 2021-01-27 at 11 59 19

Moreover, clicking the `Check out Version 0.1.8` or `Go to settings` buttons hide the notifications, but in one case the notification is shown again after a refresh and in the other not. I think it would be more consistent and intuitive if the notifications stayed visible when clicking `Check out Version 0.1.8` or `Go to settings` and only get hidden when clicking the `Dismiss` button.

Is this intentional? \cc @brandonhaslegs

geigerzaehler:
> I think it would be more consistent and intuitive if the notifications stayed visible when clicking `Check out Version 0.1.8` or `Go to settings` and only get hidden when clicking the `Dismiss` button.

I’d prefer to not show the notification again after the user interacts with it. This means if you click on "Go to settings" the app will not show the notification again. Same for clicking on “Check out Version xxx”. This is to avoid confusion where you click on “Go to settings”, then you see that notifications are disabled by default and do nothing. 

> The behaviour is different from how the "automatically check for new versions" notification works, which is shown even after an app reload or page refresh.

This is not intentional. I also think that this behavior is of little consequence since the notification is persistent and the user will most likely interact with it before restarting the app. And if not, it’s hard to determine what the best behavior would be (weighing nosiness vs notifying). So I’m hesitant to change it.

rudolfs:
Missed a version check somewhere: when I turn on notifications, a notification pops up offering me to downgrade from 2.0.0 to 0.1.8.

Screenshot 2021-01-27 at 12 46 12

geigerzaehler:
> Missed a version check somewhere: when I turn on notifications, a notification pops up offering me to downgrade from 2.0.0 to 0.1.8.

Great catch! It’s fixed in https://github.com/radicle-dev/radicle-upstream/pull/1550/commits/be844c25b75161d4b93cdaabac12409d3481115a

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "radicle-upstream",
   "productName": "Radicle Upstream",
-  "version": "0.1.8",
+  "version": "0.2.7",
```
rudolfs, line 4:
```suggestion
  "version": "0.1.8",
```

rudolfs:
Can we silence these via `rollup.app.js`?
```
Circular dependency: node_modules/semver/classes/comparator.js -> node_modules/semver/classes/range.js -> node_modules/semver/classes/comparator.js
Circular dependency: node_modules/semver/classes/comparator.js -> node_modules/semver/classes/range.js -> /Users/rudolfs/work/radicle-upstream-2/node_modules/semver/classes/comparator.js?commonjs-proxy -> node_modules/semver/classes/comparator.js
```

ui/Screen/Settings.svelte
```diff
@@ -147,186 +175,210 @@
...
+            </a>
+          </div>
         </div>
-        <div class="action">
-          <a class="typo-link" href="https://matrix.radicle.community">
-            matrix.radicle.community
-          </a>
+      </section>
+
+      <section data-cy="version" style={showVersionAtTop ? 'order: -1' : ''}>
```
rudolfs, line 313: 👏

rudolfs (approved):
LGTM 🚀 

I assume showing the upgrade notification on the lock screen is okay?

Screenshot 2021-01-27 at 14 32 09

brandonhaslegs (approved):

geigerzaehler:
Thanks folks! Merged in 52d1bcdb58b6f4604c666447e55ba12663ca2ba0

Issue #1529, opened 2021-01-20
Automate manual networking QA tests


rudolfs:
With our network integration test infrastructure in place, we can now automate the network part of our QA tests.
We should at least cover the _happy-path_.

### Network connectivity indicator
- [x] network connectivity icon reacts to connectivity changes

### Replicating projects
- [x] Spin up 2 nodes and create a project on node A
- [x] Replicate project from node A on node B via the search feature
  - [x] The project shows up in the _Following_ tab of the profile screen in
        the waiting area
  - [x] After a while the project is replicated and moves out of the waiting
        area
  - [x] When going to the Project source screen, the maintainer is selected
        in the peer selector
    - [x] Your identity does not show up in the peer selector

### Adding/Removing remotes
- [x] On node A select the created project and add node B as a remote in the remote management modal
  - [x] After a brief period the remote changes state to "Following"
- [x] Add another non-existing remote to the project
  - [x] After a timeout the remote should still be in the "Waiting" state
  - [x] Remove the non-existing remote with the "Unfollow" button

### Forking/Checking out changes
- [x] Fork the replicated project to a local directory on node B
  - [x] The forked project should now appear in the _Projects_ tab of the
        profile screen
  - [x] The identity of node B shows up in the peer selector and is selected by
        default instead of the identity of node A (the maintainer)
  - [x] The button is now called _Checkout_ instead of _Fork_

### Replicating changes
- [x] On node B create a commit and publish your changes via `git rad push`
  - [x] The published commit appears in the _Commits_ tab of the project on node B
  - [x] The commit also appears on node A under the project's contributor's peer

Depends on: https://github.com/radicle-dev/radicle-upstream/issues/1475.

PR #1559, opened 2021-01-28
test: project replication

rudolfs:
Working towards automating more of the manual QA steps.
This implements the following sections of #1529:
  - Network connectivity indicator
  - Replicating projects
  - Adding/Removing remotes

https://user-images.githubusercontent.com/158411/106255424-51718d00-621a-11eb-8136-ed49c9658523.mov

### DIFF
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -61,26 +61,9 @@ context("p2p networking", () => {
             nodeManager.asNode(contributorNode);
             cy.log("navigate to the 'Following' tab");
             commands.pick("following-tab").click();
-            commands
-              .pickWithContent(["following-tab-contents"], "Still looking")
-              .should("not.exist");
-
             commands.pick("sidebar", "search").click();
             commands.pick("search-input").type(urn);
             commands.pick("follow-toggle").click();
-
-            cy.log("project is briefly shown in the waitingroom");
-            commands
-              .pickWithContent(["following-tab-contents"], "Still looking")
-              .should("exist");
-            commands
-              .pickWithContent(
-                ["following-tab-contents"],
-                urn.replace("rad:git:", "")
-              )
-              .should("exist");
-
-            // TODO: check that notifications are shown
           });
 
           cy.log("project moved out of the waiting area and is available");
@@ -106,7 +89,6 @@ context("p2p networking", () => {
             .should("not.exist");
 
           cy.log("add contributor remote on maintainer's node");
-
           nodeManager.asNode(maintainerNode);
           commands.pick("project-list-entry-new-fancy-project.xyz").click();
 
@@ -114,23 +96,15 @@ context("p2p networking", () => {
           commands.pick("manage-remotes").click();
 
           commands.pick("followed-peers", "peer-abbey").should("not.exist");
-          cy.contains("Still looking").should("not.exist");
 
           commands.pick("peer-input").type(contributorNode.peerId);
           commands.pick("follow-button").click();
 
           cy.log("the remote shows up in the waiting area");
-          cy.contains("Still looking").should("exist");
           const shortenedPeerId = contributorNode.peerId.slice(0, 7);
-          commands
-            .pickWithContent(["pending-peers"], shortenedPeerId)
-            .should("exist");
-
-          cy.log("the remote moves out of the waiting area");
           commands
             .pickWithContent(["followed-peers", "peer-abbey"], shortenedPeerId)
             .should("exist");
-          cy.contains("Still looking").should("not.exist");
           commands
             .pickWithContent(
               ["followed-peers", "peer-abbey", "follow-toggle"],
@@ -143,22 +117,14 @@ context("p2p networking", () => {
             "hydysjnrpb4gwtsz66ei9xsyzta1afuhc1h6neufmfghozduo4p3ua";
           const shortenedNonExistingRemote = nonExistingRemote.substr(0, 7);
 
-          cy.contains("Still looking").should("not.exist");
-
           commands.pick("peer-input").type(nonExistingRemote);
           commands.pick("follow-button").click();
-
-          cy.log("check that remote stays in waiting area");
-          cy.wait(1000);
-          cy.contains("Still looking").should("exist");
-          commands.pickWithContent(
-            ["pending-peers"],
-            shortenedNonExistingRemote
-          );
+          commands
+            .pickWithContent(["pending-peers"], shortenedNonExistingRemote)
+            .should("exist");
 
           cy.log("remove the pending remote");
           commands.pick("pending-peers", "follow-toggle").click();
-          cy.contains("Still looking").should("not.exist");
           commands.pick("pending-peers").should("not.exist");
         },
         { node1Handle: "rudolfs", node2Handle: "abbey" }
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -104,44 +104,37 @@ context("p2p networking", () => {
             .should("not.exist");
 
           cy.log("add contributor remote on maintainer's node");
-          commands.pick("sidebar", "profile").click();
-          commands.pick("peer-id").then(el => {
-            const peerId = el.attr("title");
-            if (!peerId) {
-              throw new Error("Could not find PeerId");
-            }
-            const shortenedPeerId = peerId.slice(0, 7);
 
-            nodeManager.asNode(maintainerNode);
-            commands.pick("project-list-entry-new-fancy-project.xyz").click();
+          nodeManager.asNode(maintainerNode);
+          commands.pick("project-list-entry-new-fancy-project.xyz").click();
 
-            commands.pick("peer-selector").click();
-            commands.pick("manage-remotes").click();
+          commands.pick("peer-selector").click();
+          commands.pick("manage-remotes").click();
 
-            commands.pick("followed-peers", "peer-abbey").should("not.exist");
-            cy.contains("Still looking").should("not.exist");
+          commands.pick("followed-peers", "peer-abbey").should("not.exist");
+          cy.contains("Still looking").should("not.exist");
 
-            commands.pick("peer-input").type(peerId);
-            commands.pick("follow-button").click();
+          commands.pick("peer-input").type(contributorNode.peerId);
+          commands.pick("follow-button").click();
 
-            cy.log("the remote shows up in the waiting area");
-            cy.contains("Still looking").should("exist");
-            commands
-              .pick("pending-peers")
-              .contains(shortenedPeerId)
-              .should("exist");
+          cy.log("the remote shows up in the waiting area");
+          cy.contains("Still looking").should("exist");
+          const shortenedPeerId = contributorNode.peerId.slice(0, 7);
+          commands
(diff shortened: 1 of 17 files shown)


FintanH (approved):
Nice work :+1:

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+          cy.wait(1000);
+          cy.contains("Still looking").should("exist");
+          commands.pick("pending-peers").contains(shortenedNonExistingRemote);
+
+          cy.log("remove the pending remote");
+          commands.pick("pending-peers", "follow-toggle").click();
+          cy.contains("Still looking").should("not.exist");
+          commands.pick("pending-peers").should("not.exist");
+        },
+        { node1Handle: "rudolfs", node2Handle: "abbey" }
```
geigerzaehler, line 163: I don’t think it’s ideal to have the parameters here at the very end, disconnected from `withTwoOnboardedNodes()`. I propose we address this be replacing the callback pattern with the more common promise chains. That is
```
startTwoOnborardedNodes({ node1Handle: "rudolfs", node2Handle: "abbey" }).then(([node1, node2]) => {
  // ...
})
```

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+          commands.pick("create-project-button").click();
 
-        commands
-          .pick("project-screen", "header")
-          .contains("new-fancy-project")
-          .should("exist");
+          commands
+            .pick("project-screen", "header")
+            .contains("new-fancy-project")
+            .should("exist");
```
geigerzaehler, line 54: Can’t we just use `createProjectWithFixture()`? Project creation is tested thoroughly elsewhere and this would reduce the amount code and make the tests faster.

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+              .contains(shortenedPeerId)
+              .should("exist");
+            cy.contains("Still looking").should("not.exist");
+            commands
+              .pick("followed-peers", "peer-abbey", "follow-toggle")
+              .contains("Following")
+              .should("exist");
+          });
+
+          cy.log("add a random non-existing remote");
```
geigerzaehler, line 143: This should probably be a separate test case.

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+            .should("not.exist");
+
+          cy.log("add contributor remote on maintainer's node");
+          commands.pick("sidebar", "profile").click();
+          commands.pick("peer-id").then(el => {
+            const peerId = el.attr("title");
+            if (!peerId) {
+              throw new Error("Could not find PeerId");
+            }
+            const shortenedPeerId = peerId.slice(0, 7);
```
geigerzaehler, line 110: Instead of getting the peer ID from the UI we could add it to `NodeSession` and return it from the node manager. This would reduce the complexity of the code drastically.

geigerzaehler (commented):
Feels so good to finally see some proper test cases with multiple nodes.

The main concern I have at the moment is that the test case does too much. I’ve added some comments with ideas how to reduce the complexity.

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+              throw new Error("Could not find URN");
+            }
 
-          nodeManager.asNode(node2);
-          commands.pick("sidebar", "search").click();
-          commands.pick("search-input").type(urn || "");
-          commands.pick("follow-toggle").click();
+            nodeManager.asNode(contributorNode);
+            cy.log("navigate to the 'Following' tab");
+            commands.pick("Following").click();
```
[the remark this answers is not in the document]
rudolfs: We were assigning the `data-cy` attribute programmatically from the `title`. Cleaned it up here: https://github.com/radicle-dev/radicle-upstream/pull/1559/commits/2008e24c18efa5d60e6dd51212d115a828ad0879.

cypress/integration/networking.spec.ts
rudolfs: I also thought about this initially, but then decided against it thinking it would be good to simulate exactly what a user is doing. I guess picking the IDs from the UI is already covered elsewhere.

https://github.com/radicle-dev/radicle-upstream/pull/1559/commits/e98a7a1d411eb16112ab59a2beb54a65a1ed8d92

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+            commands
+              .pick("following-tab")
+              .contains("Still looking")
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1559/commits/70a20c2cccd3ec3f02fc985fa0752e5f4f92f24d

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+              .should("not.exist");
+
+            commands.pick("sidebar", "search").click();
+            commands.pick("search-input").type(urn || "");
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1559/commits/70a20c2cccd3ec3f02fc985fa0752e5f4f92f24d

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+            commands.pick("follow-toggle").click();
+
+            cy.log("project is briefly shown in the waitingroom");
```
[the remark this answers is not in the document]
rudolfs: I didn't bump into any flakiness, but maybe I got lucky. Do we test these pending states anywhere else?

Removed all of the pending state tests as per your recommendation: https://github.com/radicle-dev/radicle-upstream/pull/1559/commits/26e963709ee58008632956b9fba3e22a0379e0cf.

cypress/integration/networking.spec.ts
```diff
@@ -3,49 +3,165 @@ import * as path from "path";
...
+            commands
+              .pick("following-tab")
+              .contains("Still looking")
+              .should("exist");
```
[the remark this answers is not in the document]
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1559#discussion_r566968809

cypress/integration/networking.spec.ts
rudolfs: I'm not sure how to do this, because I want to encapsulate the cleanup in the callback wrapper. One way to do it would be using the disposer pattern, but then we still would have to pass around a callback. Another way would be to just flip the arguments around, i.e. options first, then callback. The drawback with this approach would be losing option defaults.

Happy to pair on this if you have a clear idea on how to make this better.

cypress/integration/networking.spec.ts
rudolfs: I think technically we could, but we wouldn't have a way to get to the original working directory of the project, which I was planning on using later on in the test to test commit replication.

cypress/integration/networking.spec.ts
rudolfs: I'm not sure yet how to effectively split a large integration spec up into multiple smaller ones without sacrificing performance because all steps depend on all of the previous steps.

https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion

cypress/integration/networking.spec.ts
geigerzaehler: > I'm not sure how to do this, because I want to encapsulate the cleanup in the callback wrapper.

You’re right. I didn’t realize that the cleanup is now part of `startTwoOnborardedNodes`.


> Another way would be to just flip the arguments around, i.e. options first, then callback. The drawback with this approach would be losing option defaults.

I’d prefer this but will defer to you. We can still have some optionality by making the fields optional and doing something like `startTwoOnboardedNodes({}, () => { ... })`.

cypress/integration/networking.spec.ts
geigerzaehler: Ok. Could you create an issue for this?

rudolfs: It's captured in https://github.com/radicle-dev/radicle-upstream/issues/1529 in these sections:
  - Forking/Checking out changes
  - Replicating changes

cypress/integration/networking.spec.ts
geigerzaehler: This is indeed correct for interdependent steps. But in our case the last part only depends on project creation and not on the rest of the test and doesn’t even need a second node. Splitting it into a separate test case will make it easier to grasp the test suite and it will be easier to isolate failures and fix them. The only disadvantage would be that we have to repeat the project setup. But I guess for the last part using `commands.createProjectWithFixture()` is sufficient.

All in all, I don’t think this is a major issue though, so feel free to keep it as is if your not convinced.

cypress/integration/networking.spec.ts
rudolfs: I'm planning on working on the second half of #1529 in a separate PR because we first need to make checkouts in `--test` mode work before we can test those.

cypress/integration/networking.spec.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1559/commits/e32842d80a5401300abd3a5b71882c3a83197f21

cypress/integration/networking.spec.ts
rudolfs: Ha, I just realised that we have all of this already covered by: https://github.com/radicle-dev/radicle-upstream/blob/master/cypress/integration/project/peer_management.spec.ts#L24. I was just mechanically translating the QA steps into this spec and didn't look around properly.

Thanks for being persistant!

rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1559/commits/3f28102219c1cc1be7391a57043be9627e3b7fb0

geigerzaehler (dismissed):
:chains:

cypress/plugins/nodeManager/shared.ts
```diff
@@ -5,6 +5,7 @@
 
 export interface NodeSession {
   id: number;
+  peerId: string;
```
FintanH, line 8: Right?
```suggestion
  peerId: PeerId;
```

FintanH (approved):
:kissing_closed_eyes::ok_hand:

cypress/plugins/nodeManager/shared.ts
rudolfs: Good point: https://github.com/radicle-dev/radicle-upstream/pull/1559/commits/55b2f7e0e5f703e91955119c04ae4019e7a1451b.

geigerzaehler (approved):

FintanH (approved):

rudolfs:
Merged in via: 1a66efa1310d34c8474e4ea6ce67f7566bc76e0c.

PR #1563, opened 2021-02-01
fix(ui): transaction center not collapsing

NunoAlexandre:
This was an issue experienced by everyone on the Funding demo last Friday.

The issue is that the way we are resolving whether to expand (or to keep it expanded) the transaction centre is by reactively checking whether it is expanded already or if there are ongoing transactions.

While developing against a local node, this wasn't noticeable since their transactions are included in blocks and marked as completed fairly quickly.

The solution proposed here involves "marking" transactions that the user has already had displayed in the expanded tx center as "displayed" so that we can figure out whether or not there are new transactions to be shown (when one is just created) and display it if so.

### DIFF
--- a/ui/DesignSystem/Component/Transaction/Center.svelte
+++ b/ui/DesignSystem/Component/Transaction/Center.svelte
@@ -11,16 +11,26 @@
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let element: any;
   let expand = false;
+  let initialized = false;
+
+  // The set of transaction hashes that have already been displayed
+  // in the expanded transaction stack.
+  let displayedTxs: Set<string> = new Set();
+
+  if (!initialized) {
+    displayedTxs = new Set(transactions.map(tx => tx.hash));
+    initialized = false;
+  }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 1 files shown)


rudolfs (approved):

geigerzaehler (commented):
While your works, I find it a bit hard to follow and understand. The main complexity comes from the fact that the change to `expand` and `displayedTxs` in `toggleStack` in turn force a re-evalutation of line 27.

I would suggest the following alternative
~~~typescript
$: {
  if transactions.some(tx => !displayedTxs.has(tx.hash)) {
    expand = true;
    transactions.forEach(tx => displayedTxs.add(tx.hash))
  }
}
~~~
 and removing lines 24 and 27

NunoAlexandre:
@geigerzaehler thanks for your feedback as a reader and for the suggestion :ok_hand: 

I've applied it with a slight modification to make it read better and be more efficient (though efficiency doesn't make any difference here).

geigerzaehler (approved):
> I've applied it with a slight modification to make it read better and be more efficient (though efficiency doesn't make any difference here).

Even better!

rudolfs (approved):

NunoAlexandre:
merged in 08051efc75fb25cf0100177cf61ed52e181f5276

PR #1564, opened 2021-02-01
fix: handle disconnecting wallet-side

NunoAlexandre:
When we now disconnect wallet-side, the Upstream doesn't register that event and remains as if there was still a connection.

Here, we fix that by subscribing to the `disconnect` event and call `wallet.disconnect` when triggered.

`Note`: This change was introduced alongside changes supporting the feature in testnet. In that context, this fix has the disconnection from the wallet side working all the time, consistently. However, now set aside and running against a local node the disconnection doesn't seem to always work. This is all we can do, though, afaik.

### DIFF
--- a/ui/src/wallet.ts
+++ b/ui/src/wallet.ts
@@ -70,7 +70,7 @@ export function build(): Wallet {
     qrcodeModal: qrCodeModal,
   });
 
-  const signer = new WalletConnectSigner(walletConnect, provider);
+  const signer = new WalletConnectSigner(walletConnect, provider, disconnect);
   const daiTokenContract = contract.daiToken(signer);
 
   window.ethereumDebug = new EthereumDebug(provider);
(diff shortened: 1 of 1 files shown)


ui/src/wallet.ts
```diff
@@ -96,28 +96,29 @@ export function build(): Wallet {
...
+      // We need to reinitialize `WalletConnect` until this issue is fixed:
+      // https://github.com/WalletConnect/walletconnect-monorepo/pull/370
+      walletConnect = new WalletConnect({
+        bridge: "https://bridge.walletconnect.org",
+        qrcodeModal: qrCodeModal,
+      });
+      signer.walletConnect = walletConnect;
+    } catch {
+      // When the user disconnects wallet-side, calling `killSession` app-side trows an error
+      // because the wallet has already closed its socket. Therefore, we simply ignore it.
```
geigerzaehler, line 111: Can we wrap this around `await walletConnect.killSession()` then. Also, is it on purpose, that in this case we don’t reinitialize `walletConnect` and signer.walletConnect`?

NunoAlexandre: Done! Not on purpose, it was an oversight on my part. Everything seemed to work well without it though, but I have addressed that as well :+1: thank you!

ui/src/wallet.ts
```diff
@@ -96,28 +96,30 @@ export function build(): Wallet {
...
-      qrcodeModal: qrCodeModal,
-    });
-    signer.walletConnect = walletConnect;
-    stateStore.set({ status: Status.NotConnected });
+    try {
+      await walletConnect.killSession();
+    } catch {
+      // When the user disconnects wallet-side, calling `killSession` app-side trows an error
+      // because the wallet has already closed its socket. Therefore, we simply ignore it.
+    } finally {
```
geigerzaehler, line 104: I usually try to reserve `finally` for cleanup code. To ignore errors it’s more idiomatic to use

~~~javascript
// When the user disconnects wallet-side, calling `killSession` app-side trows an error
// because the wallet has already closed its socket. Therefore, we simply ignore it.
await walletConnect.killSession().catch(() => {})
~~~

geigerzaehler (approved):

rudolfs (approved):

ui/src/wallet.ts
NunoAlexandre: Thanks for the input. Last time I used `try-catch` was 5 years ago in Java, and I don't remember it being possible to do such thing so I appreciate you providing cleaner ways :ok_hand:

NunoAlexandre: Done.

geigerzaehler (approved):

rudolfs (approved):

NunoAlexandre:
Merged in 1747a8e124f9638688dc484fe92e6fb48f744235

Issue #1560, opened 2021-01-28
Improve error text with `next`-upgrade in mind


juliendonck:
With the `next`-update being a breaking change there are a few experiences where we expect errors and that we could guardrail by linking to some troubleshooting text in the docs

The places we've discovered are:
- when re-adding a project that was managed by your old (pre next) identity, you'll have to remove the remotes before adding the project.
  -> `You're adding a project that is already managed by a rad id, click here to read more about how to fix this.`
- when pasting a project ID in search you got from the old (pre next) seed node, it won't be able to resolve that urn.
  -> `This project ID doesn't exist, is it possible you copied it from an old seed node? See the new seed node.`

Any other places that we might run into problems?

juliendonck:
@cloudhead @FintanH @MeBrei @abbey-titcomb here's the issue we talked about today.

FintanH:
> Any other places that we might run into problems?

Not that I can think of, but I'm sure we'll find out ;)

PR #1567, opened 2021-02-01
fix(proxy): urn warnings for upgrade

FintanH:
Fixes https://github.com/radicle-dev/radicle-upstream/issues/1560

For the existing remote, I realised that the URL won't parse full-stop. So I decided to handle this case specifically by renaming the existing remote, thus preserving the reference data. Then we create a fresh `rad` remote. A user can then copy any of their references over from `temp_rad` and remove it when they're done.

~For the search box, I'm not sure how to detect if it was an old URN and give a specific message in that case. So what I did here was give a small prompt that it might be an old URN. Maybe we could try document the differences in the community post.~

### DIFF
--- a/proxy/coco/src/project/create/validation.rs
+++ b/proxy/coco/src/project/create/validation.rs
@@ -377,10 +377,12 @@ impl Repository {
         match Remote::<LocalUrl>::find(repo, reflike!("rad")) {
             Err(remote::FindError::ParseUrl(_)) => {
                 log::warn!("an old/invalid URL was found when trying to load the `rad` remote");
-                log::warn!("we are going to rename the remote to `temp_rad` and create a new `rad` remote");
+                log::warn!(
+                    "we are going to rename the remote to `temp_rad` and create a new `rad` remote"
+                );
                 repo.remote_rename("rad", "temp_rad")?;
                 Ok(None)
-            }
+            },
             Err(err) => Err(err.into()),
             Ok(Some(remote)) if remote.url != *url => Err(Error::UrlMismatch {
                 expected: url.to_string(),
--- a/proxy/coco/src/project/create/validation.rs
+++ b/proxy/coco/src/project/create/validation.rs
@@ -378,9 +378,9 @@ impl Repository {
             Err(remote::FindError::ParseUrl(_)) => {
                 log::warn!("an old/invalid URL was found when trying to load the `rad` remote");
                 log::warn!(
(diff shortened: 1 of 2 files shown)


ui/src/search.ts
```diff
@@ -47,7 +47,7 @@ const VALID_URN_MATCH = /^rad:git:[1-9A-HJ-NP-Za-km-z]{37}/;
 const urnConstraints = {
   format: {
     pattern: VALID_URN_MATCH,
-    message: `Not a valid project URN`,
+    message: `Not a valid project URN, perhaps you're using an old URN?`,
```
cloudhead, line 50: I'm not sure this is an improvement, as-is. Users will have no idea what that means.

proxy/coco/src/project/create/validation.rs
```diff
@@ -375,6 +375,14 @@ impl Repository {
         url: &LocalUrl,
     ) -> Result<Option<Remote<LocalUrl>>, Error> {
         match Remote::<LocalUrl>::find(repo, reflike!("rad")) {
+            Err(remote::FindError::ParseUrl(_)) => {
+                log::warn!("an old/invalid URL was found when trying to load the `rad` remote");
+                log::warn!(
+                    "we are going to rename the remote to `temp_rad` and create a new `rad` remote"
+                );
+                repo.remote_rename("rad", "temp_rad")?;
```
cloudhead, line 383: It's not temporary is it? Perhaps `rad_old`?

ui/src/search.ts
FintanH: > Users will have no idea what that means.

Not until they read the post ;P

How could this be improved? A link to the post (possibly to a specific URN section)?

proxy/coco/src/project/create/validation.rs
FintanH: It is if they remove it later :smile: `rad_old` works too :+1:

ui/src/search.ts
cloudhead: Hmm, I wonder if it's worthwhile to have this (to then have to eventually remove it). I think just leaving it as it was might be fine -- we can explain in the post that old URNs aren't compatible, at least they will have the context there.

FintanH: Works for me. We can also add a section to the troubleshooting docs

cloudhead (approved):

PR #1573, opened 2021-02-02
Nuno/overlap free footer

### LINKED ISSUE #1568: Notifications and Transaction Center overlap
In last week's Funding demo on the testnet, one of the most experienced issues was related to not being able to close the Transaction Center. That bug had two parts: one caused by the TS code, fixed in 08051efc75fb25cf0100177cf61ed52e181f5276; the other caused by an overlap between the notification faucet and the transaction center as you can see here:

overlap

There are a couple of solutions we can consider:

1. Have the Transaction Center's `z-index` higher than the notification faucet
    That solves one issue and creates another, _unless_ if we set a minimum width for the whole app.

2. Set the width of the notifications faucet to be  `100vw - var(--sidebar-width) - var(--transaction-center-width) - var(--transaction-center-right-margin)` (pseudo-code).
   That would mean that the transactions would sit between the sidebar and the transaction center.
    However, for users who don't use the funding feature, the transactions would look off-place.

3. We move the transaction centre somewhere else

4. We gather the two elements in a common `div` and have them share the space appropriately.
    This seems to be the best approach. However, setting a minimum width for the whole app would
    still be relevant as having both elements in a small viewport would not work anyways. That can
    be treated as a separate issue, though.

NunoAlexandre:
Fixes #1568 following option 4.

This also improves things by:

+ having `NotificationFaucet` and `TransactionCenter` under the same parent grants equal margins, instead of duplicating them
+ `NotificationFaucet` takes as much space as there is available, instead of us calculating it which doesn't work for both when there is and there isn't a `TransactionCenter`
+ Seamless responsiveness

The only caveat I found for this option was to hide the transaction centre when a modal is being shown, which is not evasive at all imo.

### DIFF
--- a/ui/App/Footer.svelte
+++ b/ui/App/Footer.svelte
@@ -1,0 +1,19 @@
+<style>
+  #footer {
+    display: flex;
+    align-items: flex-end;
+    justify-content: flex-end;
+    gap: var(--content-padding);
+
+    position: fixed;
+    bottom: var(--content-padding);
+    left: var(--sidebar-width);
+    width: calc(100vw - var(--sidebar-width));
+    padding: 0 var(--content-padding);
+    z-index: 1001;
+  }
+</style>
+
+<div id="footer">
+  <slot />
+</div>
--- a/ui/App/TransactionCenter.svelte
+++ b/ui/App/TransactionCenter.svelte
@@ -15,9 +15,6 @@
 
 <style>
   .transaction-center {
-    position: fixed;
-    bottom: 2rem;
-    right: 2rem;
     z-index: 50;
   }
 </style>
--- a/ui/App/TransactionCenter.svelte
+++ b/ui/App/TransactionCenter.svelte
@@ -13,14 +13,6 @@
   };
 </script>
 
-<style>
-  .transaction-center {
-    z-index: 50;
-  }
-</style>
-
 {#if $transactions.length > 0}
-  <div class="transaction-center">
-    <Center on:select={onSelect} transactions={$transactions} />
-  </div>
+  <Center on:select={onSelect} transactions={$transactions} />
 {/if}
--- a/ui/DesignSystem/Component/NotificationFaucet.svelte
+++ b/ui/DesignSystem/Component/NotificationFaucet.svelte
@@ -9,7 +9,7 @@
 </script>
 
 <style>
-  .wrapper {
+  .notification-faucet {
     display: flex;
     flex-direction: column;
     align-items: center;
@@ -19,7 +19,7 @@
   }
 </style>
 
-<div class="wrapper" {style}>
+<div class="notification-faucet" {style}>
   {#each $store as notification (notification.id)}
     <div
       data-cy="notification"
--- a/ui/DesignSystem/Component/NotificationFaucet.svelte
+++ b/ui/DesignSystem/Component/NotificationFaucet.svelte
@@ -13,11 +13,9 @@
     display: flex;
     flex-direction: column;
     align-items: center;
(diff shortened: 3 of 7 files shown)


ui/DesignSystem/Component/NotificationFaucet.svelte
```diff
@@ -9,19 +9,17 @@
...
-  .wrapper {
+  .notification-faucet {
     display: flex;
     flex-direction: column;
     align-items: center;
-    position: fixed;
-    bottom: 2rem;
+    flex: 1;
+
     z-index: 1001;
```
geigerzaehler, line 18: Since this is now controlled by the footer it can be removed.

ui/App/Footer.svelte
```diff
@@ -0,0 +1,19 @@
+<style>
+  #footer {
+    display: flex;
+    align-items: flex-end;
+    justify-content: flex-end;
+    gap: var(--content-padding);
```
geigerzaehler, line 6: TIL

ui/App/TransactionCenter.svelte
```diff
@@ -15,9 +15,6 @@
 
 <style>
   .transaction-center {
-    position: fixed;
-    bottom: 2rem;
-    right: 2rem;
     z-index: 50;
```
geigerzaehler, line 18: Same here, we can remove this because the footer takes care of it.

geigerzaehler (commented):
This looks good from implementation standpoint. But, as mentioned in #1568, I don’t think this is the right solution. The problems are
* With the transaction center visible notifications are not centered in the screen throwing of the visual balance
* The transaction center vanishes when a modal is shown
* We don’t address the problem where the notification faucet covers other parts of the app

I’d like to here from @brandonhaslegs  and @juliendonck on these points

NunoAlexandre:
> * With the transaction center visible notifications are not centered in the screen throwing of the visual balance

I don't feel you on this one. Everything is centred within its context, just like anywhere else in the app :)

> * The transaction center vanishes when a modal is shown
 
That is a non-problem, though, in the sense that - sure it is a workaround - we don't want to show the transaction center in modals anyways, so having them in the bg but not accessible or not showing them makes no difference for users.

> * We don’t address the problem where the notification faucet covers other parts of the app

That is right, but that is not what this PR seeks to address. As for what this PR seeks to resolve, it does so.
It's an improvement that fixes the concrete issues we have.

> I’d like to here from @brandonhaslegs and @juliendonck on these points

Me too! :) I'd rather have that discussion happen in #1568, where I have already pinged for their feedback.

ui/App/TransactionCenter.svelte
NunoAlexandre: oops, forgot that, my bad.

geigerzaehler (approved):

rudolfs:
I think this workaround is not ideal as it is:
  - the position of the notification isn't centered
  - the position of the notification changes when switching between modal and normal view
  - the transaction center is visible on the BSOD screen

https://user-images.githubusercontent.com/158411/106613291-52802280-656a-11eb-84b2-baf6b1f0c6c2.mov

Screenshot 2021-02-02 at 15 34 49

NunoAlexandre:
@rudolfs @geigerzaehler thanks for your input. I am closing this given the fussy trade-off and hoping for a definitive solution in #1568.

rudolfs:
@NunoAlexandre was worth a try! Thanks for putting in the effort. My personal hunch is that we should get rid of the transaction center and have a proper screen for it. 👹

PR #1584, opened 2021-02-04
fix(proxy): update reset script

FintanH:
The radicle directory has now moved to radicle-link.

### DIFF
--- a/scripts/reset-state.sh
+++ b/scripts/reset-state.sh
@@ -7,7 +7,7 @@ if [ "$platform" == "Darwin" ]; then
     dirs=(
         "$HOME/.radicle"
         "$HOME/Library/Application Support/Radicle Upstream"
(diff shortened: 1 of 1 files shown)


rudolfs (commented):
Line 10 also needs to change to:
```
        "$HOME/Library/Application Support/xyz.radicle.radicle-link"
```

MeBrei (changes requested):
👍 
Can you also update the directory for other platforms? It should be `xyz.radicle.radicle-link` in the following line: https://github.com/radicle-dev/radicle-upstream/blob/0b672cb8918c422a6d3a9082b1930003659a1a67/scripts/reset-state.sh#L10

rudolfs (approved):

MeBrei (approved):

PR #1589, opened 2021-02-05
test: commit replication

rudolfs:
This tests bi-directional commit propagation between two nodes.
The following sections of #1529 are now automated:
  - Forking/Checking out changes
  - Replicating changes

Fixes race condition in Manage Remotes modal: https://github.com/radicle-dev/radicle-upstream/pull/1589/commits/abd07624d0780f659cd67a46d6e73f1e57cb2418.

Closes https://github.com/radicle-dev/radicle-upstream/issues/1529.

https://user-images.githubusercontent.com/158411/107740988-96162180-6d0c-11eb-8d58-85873afd6e84.mov

### DIFF
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -134,12 +134,11 @@ context("p2p networking", () => {
             "Commit replication from maintainer to contributor";
 
           cy.exec(
-            `function commands() {cd ${projctPath} && ` +
-              `touch README.md && ` +
-              `git add . && ` +
-              `git commit -m "${maintainerCommitSubject}" && ` +
-              `git -c credential.helper=${credentialHelper} push rad; }; ` +
-              `PATH=$PWD/proxy/target/release:$PATH commands`,
+            `set -euo pipefail
+             export PATH=$PWD/proxy/target/release:$PATH
+             cd ${projctPath} &&
+             git commit --allow-empty -m "${maintainerCommitSubject}" &&
+             git -c credential.helper=${credentialHelper} push rad`,
             {
               env: {
                 RAD_HOME: maintainerNode.storagePath,
@@ -197,13 +196,13 @@ context("p2p networking", () => {
             contributorNodeWorkingDir,
             projectName
           );
+
           cy.exec(
-            `function commands () {cd ${forkedProjectPath} && ` +
-              `touch CONTRIBUTOR.md && ` +
-              `git add . && ` +
-              `git commit -m "${contributorCommitSubject}" && ` +
-              `git -c credential.helper=${contributorCredentialHelper} push rad;}; ` +
-              `PATH=$PWD/proxy/target/release:$PATH commands`,
+            `set -euo pipefail
+             export PATH=$PWD/proxy/target/release:$PATH
+              cd ${forkedProjectPath} &&
+              git commit --allow-empty -m "${contributorCommitSubject}" &&
+              git -c credential.helper=${contributorCredentialHelper} push rad`,
             {
               env: {
                 RAD_HOME: contributorNode.storagePath,
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -129,22 +129,15 @@ context("p2p networking", () => {
 
           cy.log("add a new commit to the project from maintainer's node");
           const projctPath = path.join(maintainerNodeWorkingDir, projectName);
-          const credentialHelper = `'!f() { test "$1" = get && echo "password=${node1User.passphrase}"; }; f'`;
           const maintainerCommitSubject =
             "Commit replication from maintainer to contributor";
 
-          cy.exec(
-            `set -euo pipefail
-             export PATH=$PWD/proxy/target/release:$PATH
-             cd ${projctPath}
-             git commit --allow-empty -m "${maintainerCommitSubject}"
-             git -c credential.helper=${credentialHelper} push rad`,
-            {
-              env: {
-                RAD_HOME: maintainerNode.storagePath,
-              },
-            }
-          );
+          nodeManager.createCommit({
+            repositoryPath: projctPath,
+            monorepoPath: maintainerNode.storagePath,
+            subject: maintainerCommitSubject,
+            passphrase: node1User.passphrase,
+          });
 
           cy.log("refresh the UI, because new commits don't show up otherwise");
           cy.get("body").type("{esc}");
@@ -188,27 +181,19 @@ context("p2p networking", () => {
             .should("exist");
 
           cy.log("add a new commit to the project from contributor's node");
-          const contributorCredentialHelper = `'!f() { test "$1" = get && echo "password=${node2User.passphrase}"; }; f'`;
           const contributorCommitSubject =
             "Commit replication from contributor to maintainer";
-
           const forkedProjectPath = path.join(
             contributorNodeWorkingDir,
             projectName
           );
 
-          cy.exec(
-            `set -euo pipefail
-             export PATH=$PWD/proxy/target/release:$PATH
-             cd ${forkedProjectPath}
-             git commit --allow-empty -m "${contributorCommitSubject}"
-             git -c credential.helper=${contributorCredentialHelper} push rad`,
-            {
-              env: {
-                RAD_HOME: contributorNode.storagePath,
-              },
-            }
-          );
+          nodeManager.createCommit({
+            repositoryPath: forkedProjectPath,
+            monorepoPath: contributorNode.storagePath,
+            subject: contributorCommitSubject,
+            passphrase: node2User.passphrase,
+          });
 
           cy.log("refresh the UI, because new commits don't show up otherwise");
           cy.get("body").type("{esc}");
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -195,7 +195,9 @@ context("p2p networking", () => {
           commands.pick("checkout-button").click();
 
           // Wait for checkout to write the files to disk.
-          cy.wait(1000);
+          commands
+            .pickWithContent(["notification"], `${projectName} checked out to`)
+            .should("exist");
 
           cy.log("add a new commit to the project from contributor's node");
           const contributorCredentialHelper = `'!f() { test "$1" = get && echo "password=${node2User.passphrase}"; }; f'`;
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -25,12 +25,12 @@ context("p2p networking", () => {
   });
 
   it("replicates a project from one node to another", () => {
-    const node1User = {
+    const maintainer = {
       handle: "rudolfs",
       fullName: "Rūdolfs Ošiņš",
       passphrase: "1111",
     };
-    const node2User = {
+    const contributor = {
       handle: "abbey",
       fullName: "Abbey Titcomb",
       passphrase: "2222",
@@ -39,8 +39,8 @@ context("p2p networking", () => {
     commands.withTempDir(tempDirPath => {
       nodeManager.withTwoOnboardedNodes(
         {
-          node1User,
-          node2User,
+          maintainer,
+          contributor,
         },
         (maintainerNode, contributorNode) => {
           nodeManager.connectTwoNodes(maintainerNode, contributorNode);
@@ -144,8 +144,8 @@ context("p2p networking", () => {
             repositoryPath: projctPath,
             monorepoPath: maintainerNode.storagePath,
             subject: maintainerCommitSubject,
-            passphrase: node1User.passphrase,
-            name: node1User.fullName,
+            passphrase: maintainer.passphrase,
+            name: maintainer.fullName,
           });
 
           cy.log("refresh the UI, because new commits don't show up otherwise");
@@ -201,8 +201,8 @@ context("p2p networking", () => {
             repositoryPath: forkedProjectPath,
             monorepoPath: contributorNode.storagePath,
             subject: contributorCommitSubject,
-            passphrase: node2User.passphrase,
-            name: node2User.fullName,
+            passphrase: contributor.passphrase,
+            name: contributor.fullName,
           });
 
           cy.log("refresh the UI, because new commits don't show up otherwise");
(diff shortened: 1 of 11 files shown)


cypress/plugins/nodeManager/plugin.ts
```diff
@@ -108,25 +111,34 @@ class Node {
...
       [
         "--test",
+        "--test-use-rad-home",
         "--http-listen",
         `${HOST}:${this.httpPort}`,
         "--peer-listen",
         `${HOST}:${this.peerPort}`,
       ],
-      {}
+      { env: envCopy }
```
cloudhead, line 136: So how come both the env and the option are needed?

rudolfs: Because it's still WIP. Will clean it up once I have everything working.

proxy/api/src/service.rs
```diff
@@ -39,27 +46,56 @@ impl Environment {
...
+                    None => {
+                        println!("RAD_HOME must be set when running with --test-use-rad-home");
+                        std::process::exit(1);
+                    },
+                    Some(root) => path::Path::new(&root).to_path_buf(),
+                };
+                let store_path = root_path.join("store");
+                let coco_profile =
+                    coco::profile::Profile::from_root(path::Path::new(&root_path), None)?;
+                let keystore = Arc::new(coco::keystore::file(coco_profile.paths().clone()));
```
geigerzaehler, line 62: This is exactly the behavior when the `--test` flag is not provided. I think we don’t need the additional `--test-use-rad-home` flag and can just run the test node with `RAD_HOME` set.

cypress/plugins/nodeManager/plugin.ts
```diff
@@ -108,25 +111,34 @@ class Node {
     this.httpPort = options.id;
     this.peerPort = options.id;
     this.proxyBinaryPath = path.join(ROOT_PATH, options.proxyBinaryPath);
+    this.storagePath = path.join(CYPRESS_WORKSPACE_PATH, uuid.v4());
   }
 
   async start() {
     this.logger.log("starting node");
 
+    fs.mkdirsSync(this.storagePath);
```
geigerzaehler, line 120: We should avoid using blocking code
```suggestion
    await fs.promises.mkdir(this.storagePath);
```

cypress/plugins/nodeManager/plugin.ts
geigerzaehler, line 136: There’s no need to copy all environment variables. The child process inherits all variables that are not explicitly set.
```suggestion
      { env: { RAD_HOME: this.storagePath } }
```

cypress/plugins/nodeManager/plugin.ts
```diff
@@ -217,6 +229,11 @@ class Node {
       this.logger.log("ignoring stop node command, node wasn't running");
     }
   }
+
+  private cleanup(): void {
+    this.logger.log("cleaning up state");
+    fs.removeSync(this.storagePath);
```
geigerzaehler, line 235: See comment about blocking code above

cypress/integration/networking.spec.ts
```diff
@@ -112,6 +124,127 @@ context("p2p networking", () => {
...
+
+          const forkedProjectPath = path.join(
+            contributorNodeWorkingDir,
+            projectName
+          );
+          cy.exec(
+            `function commands () {cd ${forkedProjectPath} && ` +
+              `touch CONTRIBUTOR.md && ` +
+              `git add . && ` +
+              `git commit -m "${contributorCommitSubject}" && ` +
```
geigerzaehler, line 204: Instead of touching and adding a file you can use `git commit --allow-empty`.

cypress/integration/networking.spec.ts
```diff
@@ -112,6 +124,127 @@ context("p2p networking", () => {
...
+          cy.exec(`mkdir -p ${contributorNodeWorkingDir}`);
+
+          ipcStub.getStubs().then(stubs => {
+            stubs.IPC_DIALOG_SHOWOPENDIALOG.returns(contributorNodeWorkingDir);
+          });
+          commands.pick("checkout-modal-toggle").click();
+          commands.pick("choose-path-button").click();
+          // Make sure UI has time to update path value from stub,
+          // this prevents this spec from failing on CI.
+          cy.wait(500);
```
geigerzaehler: As discussed: We don’t need to wait here.

cypress/integration/networking.spec.ts
```diff
@@ -112,6 +124,127 @@ context("p2p networking", () => {
...
+          const contributorCredentialHelper = `'!f() { test "$1" = get && echo "password=${node2User.passphrase}"; }; f'`;
+          const contributorCommitSubject =
+            "Commit replication from contributor to maintainer";
+
+          const forkedProjectPath = path.join(
+            contributorNodeWorkingDir,
+            projectName
+          );
+          cy.exec(
+            `function commands () {cd ${forkedProjectPath} && ` +
```
geigerzaehler, line 201: Is there a reason this is a function and not a simple script?
```js
`set -euo pipefail
export PATH=$PWD/proxy/target/release:$PATH
cd ${forkedProjectPath}
...`
```

rudolfs: Good old `pipefail` saves the day!
ab65566baffc0bb4805c34c4486f23284652f0d6

cypress/integration/networking.spec.ts
rudolfs: ab65566baffc0bb4805c34c4486f23284652f0d6

cypress/plugins/nodeManager/plugin.ts
rudolfs: This is called in the exit hook, which AFAIK should only do `sync` things. Should we then switch to: https://www.npmjs.com/package/async-exit-hook?

cypress/plugins/nodeManager/plugin.ts
rudolfs: I tried this out, it doesn't work for some reason.

```
const envCopy: NodeJS.ProcessEnv = {};
Object.assign(envCopy, global.process.env);
envCopy["RAD_HOME"] = this.storagePath;

const process = childProcess.spawn("env", [], {
  env: envCopy,
});
```
Results in the full env being shown:

```
STDOUT: npm_package_author_email=dev@radicle.xyz
npm_package_dependencies_ethers=^5.0.23
npm_package_devDependencies__types_sinon=^9.0.10
npm_package_scripts_test_unit_watch=jest --watchAll
npm_package_devDependencies__types_twemoji=^12.1.1
npm_config_version_git_sign=
npm_config_ignore_scripts=
npm_config_user_agent=yarn/1.22.5 npm/? node/v14.5.0 darwin x64
npm
...
```

But:
```
const process = childProcess.spawn("env", [], {
  env: { RAD_HOME: "mystoragepath" },
});
```

Results in only the one variable being set:
```
 STDOUT: RAD_HOME=mystoragepath
```

proxy/api/src/service.rs
rudolfs: Makes sense. Will all the control API endpoints work though? We don't need them for this integration test, but maybe for some future ones?

rudolfs: Removed it in: b58fbc64a31d03ad13495d01838d4480f4829915.

cypress/plugins/nodeManager/plugin.ts
rudolfs: For more details see: https://github.com/radicle-dev/radicle-upstream/pull/1589/files#r574634262.

cypress/plugins/nodeManager/plugin.ts
rudolfs: See: https://github.com/radicle-dev/radicle-upstream/pull/1589/files#r574708931.

cypress/integration/networking.spec.ts
rudolfs: 6d50aaabb45f1a3ab92a4737c3c830e8e5d11251

cypress/integration/networking.spec.ts
```diff
@@ -1,4 +1,4 @@
-import * as commands from "../support/commands";
+import * as cmd from "../support/commands";
```
geigerzaehler, line 1: Can we stick with `commands`? We are using using the full word in all other tests and I’d prefer to be consistent and keep the changes to a minimum. (In general I also prefer to use the module name as the qualified import to aid readability at the sacrifice of writing code quicker.)

cypress/plugins/nodeManager/plugin.ts
geigerzaehler: I don’t think your comment applies since we are not calling `start()` from the sync exit hook. The `start` method is already async.

FintanH (approved):
Looks good from my pov

cypress/plugins/nodeManager/plugin.ts
geigerzaehler: From what I can see this is only called in the `exit` handler of a process and not in the `exitHook()` callback. It is ok for the `exit` event handler to be async (and in fact, it already is) and await `cleanup()`.

When it comes to `exitHook()` it turns out that the `exit` event on a the node `process` is never triggered when we kall `stopAllNodes()` in `exitHook()`. I don’t think this is too big of an issue if we occasionally leak these temporary directories.

cypress/support/nodeManager.ts
```diff
@@ -24,6 +29,37 @@ export const connectTwoNodes = (
...
+export GIT_AUTHOR_EMAIL="${options.email || "john@pipefail.com"}"
+export GIT_COMMITTER_NAME=$GIT_AUTHOR_NAME
+export GIT_COMMITTER_EMAIL=$GIT_AUTHOR_EMAIL
+export PATH=$PWD/proxy/target/release:$PATH
+cd ${options.repositoryPath}
+git commit --allow-empty -m "${options.subject}"
+git -c credential.helper=${credentialsHelper(options.passphrase)} push rad`,
+    {
+      env: {
+        RAD_HOME: options.monorepoPath,
```
geigerzaehler, line 57: We are using two ways now to expose environment variables. Via `export`  in the script or adding them here. I’d prefer to just use one.

geigerzaehler: `RAD_HOME` is not exactly the monorepo path. To avoid confusion I’d suggest to just call the property `radHome`, then it’s unambiguous what it is for.

cypress/plugins/nodeManager/plugin.ts
geigerzaehler: Yes you’re right. A more idiomatic and concise way to write this would be
```suggestion
      { env: { ...process.env, RAD_HOME: this.storage_path } }
```

cypress/plugins/nodeManager/plugin.ts
rudolfs: I still think it does.

In the `start` method we register a callback that calls `cleanup` when the process exits: https://github.com/radicle-dev/radicle-upstream/blob/rudolfs/replication-tests/cypress/plugins/nodeManager/plugin.ts#L361

And since we run the `stop` method for each node in the `exitHook` that code will be indirectly invoked.

cypress/plugins/nodeManager/plugin.ts
rudolfs: Good call!

5ec74caf75c4e1e0896f7e4d3123c8e594da991a

cypress/integration/networking.spec.ts
rudolfs: Consistency muss sein: 3318e58b06e141e686e4fdb5853d624815179d1d.
Just wanted to avoid line wrapping in some places.

cypress/support/nodeManager.ts
rudolfs: > We are using two ways now to expose environment variables. Via export in the script or adding them here. I’d prefer to just use one.

I'm not sure I follow, could you please elaborate?

> RAD_HOME is not exactly the monorepo path. To avoid confusion I’d suggest to just call the property radHome, then it’s unambiguous what it is for.

Good thinking. I like how everything now falls into place: 7463b20192904ac8811c12b29a751291c666c948.

cypress/plugins/nodeManager/plugin.ts
rudolfs: Hm, you're right. The process exit callback never fires : |

rudolfs: This seems to fix our problem: 9eec21f550be4b4a033be67f6ecf74650b8accf0.

cypress/plugins/nodeManager/plugin.ts
rudolfs: > From what I can see this is only called in the exit handler of a process and not in the exitHook() callback. It is ok for the exit event handler to be async (and in fact, it already is) and await cleanup().

Let there be `async` then: eda5044f75097aa8ceefffbdd6baa3e7084269a0.

> When it comes to exitHook() it turns out that the exit event on a the node process is never triggered when we kall stopAllNodes() in exitHook(). I don’t think this is too big of an issue if we occasionally leak these temporary directories.

You're right, the child process exit handlers didn't get called because we killed the child process with `.kill`. Sending a shutdown signal should let the child processes clean up after themselves: 9eec21f550be4b4a033be67f6ecf74650b8accf0.

cypress/support/nodeManager.ts
rudolfs: > We are using two ways now to expose environment variables. Via export in the script or adding them here. I’d prefer to just use one.

Thanks for clearing it up, makes sense to do it consistently: 14c09f337d7c54f194c8a4c232093332e41579ab.

cypress/plugins/nodeManager/plugin.ts
rudolfs: > Sending a shutdown signal lets the child processes clean up after themselves: 9eec21f.

Ugh, didn't work the way I thought it would, never mind.

geigerzaehler (approved):

rudolfs:
Merged into https://github.com/radicle-dev/radicle-upstream/tree/xla/1236-thank-you-next via fe72f889986721eaefbad0fb735b0a19f4f8e623.

PR #1582, opened 2021-02-04
feat: add testnet support for funding pool 

NunoAlexandre:
This PR introduces the following changes:

- [x] Support switching between the `Local` and `Ropsten` Ethereum environments
  - [x] Offer toggle in the Settings
  - [x] Extend `WalletConnectSigner` so that it works both when used in both environments

As follow up issues found from running the feature in the Ropsten network:

- [x] Correct the token scaling, from and to the Radicle Contracts
- [x] Use a big-number lib that supports decimals. This was important to:
  - [x] Support setting a decimal value as weekly budget, top-up, and withdraw amounts
  - [x] Preserve precision when converting Radicle Tokens from and to the contracts.
- [x] Show a `Wrong Network` screen when the user switches their connected wallet's network
- [x] Improve the waiting experiences while transactions are to be included in a block
- [x] Documentation ++

PR #1591, opened 2021-02-08
fix(ui): reinitialize walletconnect on dismissed modal

NunoAlexandre:
Currently, if we click on "Connect your wallet", get the QR Code Modal, and dismiss that modal, the button won't work anymore.
The root caused was attempted to be fixed here but it seems that the issue persists.

Until WalletConnect addresses the issue, what we do is detecting when the modal was dismissed (i.e., the user closed it manually) and reinitialize the walletconnect instance when that happens.

Note that for that to be possible, we now allow an optional `onHide` parameter to be specified to a modal being shown.
This approach has no impact on the API of the modal module anywhere else and it offers a nice general-purpose capability, imo.

`Note`: This branch is based off #1582. Will change base when that PR is merged.

### DIFF
--- a/ui/src/modal.ts
+++ b/ui/src/modal.ts
@@ -1,7 +1,12 @@
 import { derived, get, writable } from "svelte/store";
 
+type OnHide = () => void;
+const doNothing = () => {
+  /**/
+};
+
 type ModalOverlay =
-  | { show: true; route: string }
+  | { show: true; route: string; onHide: OnHide }
   | { show: false; route: null };
 
 const overlayStore = writable<ModalOverlay>({
@@ -10,14 +15,21 @@ const overlayStore = writable<ModalOverlay>({
 });
 export const store = derived(overlayStore, $store => $store);
 
-export const hide = (): void => overlayStore.set({ show: false, route: null });
+export const hide = (): void => {
+  const modal = get(store);
+  if (modal.show) {
+    modal.onHide();
+  }
 
-export const toggle = (route: string): void => {
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-  if (get(store).show && get(store).route === route) {
+  overlayStore.set({ show: false, route: null });
+};
+
+export const toggle = (route: string, onHide: OnHide = doNothing): void => {
+  const modal = get(store);
+  if (modal.show && modal.route === route) {
     hide();
     return;
   }
 
-  overlayStore.set({ show: true, route });
+  overlayStore.set({ show: true, route, onHide });
 };
--- a/ui/src/wallet.ts
+++ b/ui/src/wallet.ts
@@ -65,25 +65,6 @@ function getProvider(
   }
 }
 
-const qrCodeModal = {
-  open: (uri: string, _cb: unknown, _opts?: unknown) => {
-    uriStore.set(uri);
-    modal.toggle(path.walletQRCode());
-  },
-  close: () => {
-    // N.B: this is actually called when the connection is established,
-    // not when the modal is closed per se.
-    modal.hide();
-  },
-};
-
-function newWalletConnect(): WalletConnect {
-  return new WalletConnect({
-    bridge: "https://bridge.walletconnect.org",
-    qrcodeModal: qrCodeModal,
-  });
-}
-
 export function build(
   environment: ethereum.Environment,
   provider: ethers.providers.Provider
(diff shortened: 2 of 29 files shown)


ui/src/wallet.ts
```diff
@@ -91,6 +72,40 @@ export function build(
...
+    walletConnect = newWalletConnect();
+    signer.walletConnect = walletConnect;
+  }
+
+  const qrCodeModal = {
+    open: (uri: string, _cb: unknown, _opts?: unknown) => {
+      uriStore.set(uri);
+      modal.toggle(path.walletQRCode(), onModalHide);
+    },
+    close: async () => {
```
rudolfs, line 101: Is the `async` necessary? We're not `await`ing anything in the body.

ui/src/wallet.ts
```diff
@@ -91,6 +72,40 @@ export function build(
...
+
+  function newWalletConnect(): WalletConnect {
+    return new WalletConnect({
+      bridge: "https://bridge.walletconnect.org",
+      qrcodeModal: qrCodeModal,
+    });
+  }
+
+  // We need to reinitialize `WalletConnect` until this issue is fixed:
+  // https://github.com/WalletConnect/walletconnect-monorepo/pull/370
```
rudolfs, line 90: Looks like the referenced PR got merged, maybe we can upgrade walletconnect and this won't be necessary?

rudolfs (approved):

ui/src/wallet.ts
NunoAlexandre: The PR is merged but the issue isn't fixed :) It's all in the PR, have a look for your interest.

rudolfs: Too bad.

ui/src/modal.ts
```diff
@@ -1,7 +1,12 @@
 import { derived, get, writable } from "svelte/store";
 
+type OnHide = () => void;
+const doNothing = () => {
+  /**/
+};
+
 type ModalOverlay =
-  | { show: true; route: string }
+  | { show: true; route: string; onHide: OnHide }
```
geigerzaehler, line 9: Wouldn’t it be more appropriate to call this `onClose`?

ui/src/wallet.ts
```diff
@@ -91,6 +72,40 @@ export function build(
...
+
+  const qrCodeModal = {
+    open: (uri: string, _cb: unknown, _opts?: unknown) => {
+      uriStore.set(uri);
+      modal.toggle(path.walletQRCode(), onModalHide);
+    },
+    close: () => {
+      // N.B: this is actually called when the connection is established,
+      // not when the modal is closed per se.
+      stateStore.set({ status: Status.Connecting });
```
geigerzaehler, line 104: Setting the status to “connecting” seems to contradict the comment above. Shouldn’t it be `Status.Connected`? And why wasn’t this there before?

ui/src/modal.ts
NunoAlexandre: Also thought so. However, we use the 'hide' terminology elsewhere.

ui/src/wallet.ts
NunoAlexandre: > Setting the status to “connecting” seems to contradict the comment above

Indeed. I think rephrasing the comment to the following is more correct:

> // N.B: this is actually called when the connection is initiated,
> // not when the modal is closed per se.

Meaning, when the user scans the QR Code and the connection gets going, this closure is called.

> Shouldn’t it be `Status.Connected`

Ideally, this closure would be `onConnected` or something alike. But the actual connection establishment happens in `walletConnect.connect` (line 125). When it succeeds, we then call `await initialize();`, line 135.

> And why wasn’t this there before?

Because until the need for this workaround, we didn't need to eagerly set the status to what it factually is but to which is not yet set at this point. i.e, this closure is called before the `walletConnect.connect()` promise resolves. In that time in between, `modal.hide()` is called, which now calls the passed `onHide` parameter. When that happens, we need to know whether the modal was closed as a dismissal (the user just closed it) or if it was closed because the connection is being established. We can only capture the latter, in this very closure. So, setting the status to `Connecting` eagerly here allows us to check it in `onModalHide`; if it's `Connecting`, we don't do anything. If it's `NOT_CONNECTED`, then we know the user has dismissed the modal and that we must, therefore, reinitialize `walletConnect`.

geigerzaehler (approved):

rudolfs (approved):

NunoAlexandre:
Facing github buggy merge conflicts, will open a new PR.

PR #1595, opened 2021-02-09
chore(ci): clean up macOS build environment before each run

rudolfs:
Since the macOS host doesn't have build process isolation like its Linux
counterpart, it can happen that leftovers from a previously cancelled
build are still hanging around. This makes sure that we stop any
lingering radicle-proxy processes before starting a new build.

This is a workaround, we should figure out how to do proper build
isolation on macOS in the future.

Affected builds:
https://buildkite.com/monadic/radicle-upstream/builds/7576#887aa993-12f0-475e-907c-a0e0c9ff515d
https://buildkite.com/monadic/radicle-upstream/builds/7576#3e0c96cc-065c-421e-a5f8-814a1acc8877

### DIFF
--- a/.buildkite/env.sh
+++ b/.buildkite/env.sh
@@ -2,6 +2,9 @@
 set -Eeou pipefail
 
 if [[ "${BUILDKITE_AGENT_META_DATA_PLATFORM:-}" == "macos" ]]; then
+  echo "--- Cleaning up macOS environment"
+  killall radicle-proxy
+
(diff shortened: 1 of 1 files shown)


geigerzaehler (approved):

MeBrei (commented):
⭐

MeBrei (approved):

rudolfs:
Closed via: 848cbb413502979dd2b77a4541d580cdb524da8b.

PR #1608, opened 2021-02-10
fix(infra): check minimum node version

geigerzaehler:
We check the minimum node version used to run the `set-latest-release.ts` to ensure that `fs.promise.rm` is available.

### DIFF
--- a/scripts/set-latest-release.ts
+++ b/scripts/set-latest-release.ts
@@ -11,6 +11,7 @@ import * as path from "path";
 import * as fs from "fs";
 import * as childProcess from "child_process";
 
+import * as semver from "semver";
 import fetch from "node-fetch";
 
 const pkg = require("../package.json");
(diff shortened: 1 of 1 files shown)


rudolfs:
Can we also update the node version in `.tool-versions` and `.nvmrc`?

geigerzaehler:
> Can we also update the node version in `.tool-versions` and `.nvmrc`?

Do you mean setting the version in these two files to v14.14.0 or updating it to the latest released node version? Both require us to rebuild the docker image and I don’t see a benefit at the moment. The current version (v14.15.4) is reasonably up-to-date within the LTS channel and downgrading does not seem like a good idea.

NunoAlexandre (approved):

rudolfs:
> Do you mean setting the version in these two files to v14.14.0 or updating it to the latest released node version? Both require us to rebuild the docker image and I don’t see a benefit at the moment. The current version (v14.15.4) is reasonably up-to-date within the LTS channel and downgrading does not seem like a good idea.

I see. Yeah I wanted to avoid the situation where I try to run `./scripts/set-latest-release.ts` and it errors because my toolchain is not at the right version, especially if we already have a mechanism to prevent that. But I see now why you went with showing an error, I guess we can upgrade the toolchain another time then.

rudolfs (approved):

geigerzaehler:
Merged in 79420abcf662454692ad4b63491bfa4b42384a17

PR #1611, opened 2021-02-11
refactor(ui): better type checking of node manager task api

geigerzaehler:
*Happy to delay this until more tests have been merged if that reduces overhead*

We improve the node manager plugin API to achieve the following

* Introduce a common interface for the plugin API that can be used in the test runner and the background process.
* Eliminate manual calls to `cy.task()`. Instead we have a generic `nodeManagerPlugin` object with the correct types.
* We can now expose the `NodeManager` instance directly through the plugin API.
* The plugin API now returns values from `startNode()` and other calls. This eliminates the need for bookkeeping and ensuring that the implemenations match.


rudolfs (approved):
Neat!

NunoAlexandre (approved):

PR #1619, opened 2021-02-17
fix(ui): more permissive seed validation

### LINKED ISSUE #1618: Seed node address validation regex is too restrictive
[The regex](https://github.com/radicle-dev/radicle-upstream/blob/3f24a7a3d30644563270b76ad36008635966207e/ui/src/session.ts#L218) that validates a seed address doesn't allow domain names such as for example `foo-example.com` in the node's address.

geigerzaehler:
We allow all valid DNS names in the seed node address validation.

Note that the validation is in-fact too permissive since it allows `-`, `-.-`, `123` and other invalid names. Take all edge cases into account would increase the complexit of the regexp though.

Fixes #1618.

### DIFF
--- a/ui/src/session.test.ts
+++ b/ui/src/session.test.ts
@@ -1,0 +1,31 @@
+import { __test__ as session } from "./session";
+
+describe("VALID_SEED_MATCH", () => {
+  const peer_id = new Array(54).fill("a").join("");
+
+  it("matches valid values", () => {
+    const values = [
+      `${peer_id}@radicle.xyz:1`,
+      `${peer_id}@radicle.xyz:12345`,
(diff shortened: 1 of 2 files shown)


MeBrei (approved):

NunoAlexandre (approved):
Would be nice to add tests :angel:

geigerzaehler:
> Would be nice to add tests

Done in https://github.com/radicle-dev/radicle-upstream/pull/1619/commits/7da0424b49d04cc0d2dee66c1e0283d331572cfa

NunoAlexandre (approved):
:heavy_check_mark:

MeBrei (dismissed):

Issue #1445, opened 2020-12-02
Creating a new project does not use branch name specified in git config


Nemo157:
I use a custom default branch name for my repositories:

```console
> git config --get init.defaultBranch
prīmum
```

I would expect any tool creating a git repository for me to follow this config, but it appears that radicle-upstream uses a hardcoded `main` branch name instead.

cloudhead:
So one way this could work is:
* If I haven't set my config, upstream defaults to `main`
* If I have set my config, upstream defaults to whatever I set

```
> git config --get init.defaultBranch            # default to 'main'
> git config --global --add init.defaultBranch master   # default to 'master'                                                              
> git config --get init.defaultBranch
master
>
```

NunoAlexandre:
@Nemo157 thanks for raising this! It has merged in https://github.com/radicle-dev/radicle-upstream/commit/1ad2ae34d8ebe97bef201f6fde5b63ee57b9b10e. Keep tuned for the next release :surfing_man: 

![primum](https://user-images.githubusercontent.com/7075260/108542605-3d362280-72e4-11eb-82ef-76393dd3e7fa.png)

PR #1621, opened 2021-02-17
feat(ui): pick user-defined default git branch

NunoAlexandre:
Fixes https://github.com/radicle-dev/radicle-upstream/issues/1445

We look up the user-defined global git default branch and use that, if obtainable, as the default git branch for new repositories.
When that property is not obtainable (which should happen in machines running a git version lower than `v2.28`), we use `main` as the default branch.

We also test both of these possible use cases.
 
Note that we only do this for new projects we are responsible to create, not for imported projects. There, we should look up the repo-specific default branch config property, not the global one. I will address that in a separate PR.


ui/src/config.ts
```diff
@@ -1,5 +1,11 @@
...
+
 export const HIDDEN_BRANCHES = ["rad/contributor", "rad/project"];
 export const UPSTREAM_DEFAULT_BRANCH = "main";
 export const GIT_DEFAULT_BRANCH = "master";
 export const NOTIFICATION_TIMEOUT = 8000; // ms
 export const FADE_DURATION = 200;
+
+export function usersGitDefaultBranch(): Promise<string | undefined> {
+  return ipc.usersGitDefaultBranch();
+}
```
NunoAlexandre, line 11: :information_source: I added this indirection to mean that this is seen as a 'config' property, which happens to be obtained via `ipc`.

native/main.ts
```diff
@@ -163,6 +164,18 @@ ipcMain.handle(RendererMessage.OPEN_URL, (_event, url) => {
   openExternalLink(url);
 });
 
+ipcMain.handle(RendererMessage.USERS_GIT_DEFAULT_BRANCH, () => {
+  try {
+    return childProcess
+      .execSync("git config --global --get init.defaultBranch", {
+        encoding: "utf-8",
+      })
```
geigerzaehler, line 172: We should avoid sync calls so we don’t block the whole VM.

`child_process` doesn’t have great APIs for this so a possible solution could be this
```suggestion
    const stdout = await new Promise((resolve, reject) => 
      childProcess
      .exec("git config --global --get init.defaultBranch", {
        encoding: "utf-8",
      }, (error, stdout, stderr) => {
        if (error) {
          reject(error)
        } else {
          resolve(stdout)
        }
      })
```

native/main.ts
```diff
@@ -163,6 +164,18 @@ ipcMain.handle(RendererMessage.OPEN_URL, (_event, url) => {
...
+      .trim();
+  } catch {
+    return undefined;
```
geigerzaehler, line 175: We should probably at least log a warning if this fails.

cypress/integration/project/creation.spec.ts
```diff
@@ -289,36 +290,53 @@ context("project creation", () => {
     });
 
     context("happy paths", () => {
-      it("creates a new project from an empty directory", () => {
-        withEmptyDirectoryStub(() => {
-          commands.pick("new-project-button").click();
+      context("creates a new project from an empty directory", () => {
+        function go(expectedDefaultBranch: string) {
```
MeBrei, line 294: Can we change the name of the function to something more descriptive, so that it reads better where it is used?

cypress/integration/project/creation.spec.ts
```diff
@@ -289,36 +290,53 @@ context("project creation", () => {
...
+              .pick("profile-screen", "project-list")
+              .contains("My new fancy project");
+          });
+        }
+
+        it("picks the user-defined git default branch", () => {
+          go("trunk");
+        });
+
+        it("picks the Upstream default git branch when can not obtain a user-defined", () => {
```
MeBrei, line 335:
```suggestion
        it("picks the Upstream default git branch when it can not obtain a user-defined", () => {
```

native/ipc-types.ts
```diff
@@ -22,4 +22,5 @@ export enum RendererMessage {
   GET_VERSION = "GET_VERSION",
   OPEN_PATH = "IPC_OPEN_PATH",
   OPEN_URL = "IPC_OPEN_URL",
+  USERS_GIT_DEFAULT_BRANCH = "USERS_GIT_DEFAULT_BRANCH",
```
geigerzaehler, line 25: How about `GET_GIT_GLOBAL_DEFAULT_BRANCH`? It aligns better with `git config --global --get init.defaultBranch` and the fact that all the other variants are verb phrases.

ui/src/config.ts
geigerzaehler: > I added this indirection to mean that this is seen as a 'config' property, which happens to be obtained via `ipc`.

This is something we might want to reconsider. In addition to an unnecessary indirection this also has the potential to create problems down the line. For example, if the `ipc` module has to be configured, we would like to have that value in `config`, but this would create a circular dependency.

I’d also like to point out that, despite its name, this module has not been there for configuration but just served as a place to collect random constants. In this light I’d like to avoid extending this module further and complicating the module dependency graph.

MeBrei (commented):
The changes look good, but I was wondering if using the globally set default branch isn't more a concern of `radicle-link` or at least the coco side of upstream.

native/ipc-types.ts
NunoAlexandre: Good with me!

native/main.ts
NunoAlexandre: Are you referring to any specific type of log? I mean, you are not suggesting to show a notification error for this, is that right?

cypress/integration/project/creation.spec.ts
NunoAlexandre: So the function is defined within its `context`, so for me `go` is as descriptive as it gets without repeating the same thing again. Would `run` be better, to mean "run" the test? Or feel free to shoot other suggestions :)

MeBrei: You are right, since it is within the context, the definition of the function is not far away :)
I was thinking something like: `testNewProjectCreationWith`

NunoAlexandre:
@geigerzaehler @MeBrei all addresses, let me know if I missed anything.

@geigerzaehler regarding your comment https://github.com/radicle-dev/radicle-upstream/pull/1621#discussion_r578256830, I wouldn't know if we should log it as I wouldn't think of it as a 'warning' that the user doesn't have the latest version of git. Fetching that git config only works for users who use the latest version of git, so we just support both scenarios gracefully.

native/main.ts
geigerzaehler: Showing a notification is too intrusive. The logging is mainly for debugging purposes. Since this is the main process and not the renderer process using `console.warn` will print something to `stderr`.

geigerzaehler:
> @geigerzaehler regarding your comment [#1621 (comment)](https://github.com/radicle-dev/radicle-upstream/pull/1621#discussion_r578256830), I wouldn't know if we should log it as I wouldn't think of it as a 'warning' that the user doesn't have the latest version of git. Fetching that git config only works for users who use the latest version of git, so we just support both scenarios gracefully.

Good point. I suggested using `console.warn()` above which would only be visible on `stderr` when the app is run in the terminal. Since we catch _all_ errors I think it would be useful to see this in tests, etc. But I’m also ok if we don’t log if you feel differently.

NunoAlexandre:
> > @geigerzaehler regarding your comment [#1621 (comment)](https://github.com/radicle-dev/radicle-upstream/pull/1621#discussion_r578256830), I wouldn't know if we should log it as I wouldn't think of it as a 'warning' that the user doesn't have the latest version of git. Fetching that git config only works for users who use the latest version of git, so we just support both scenarios gracefully.
> 
> Good point. I suggested using `console.warn()` above which would only be visible on `stderr` when the app is run in the terminal. Since we catch _all_ errors I think it would be useful to see this in tests, etc. But I’m also ok if we don’t log if you feel differently.

I would skip it then. If we feel different once we get using it ourselves we can always add it :+1:

MeBrei:
When no default branch is set in the `.gitconfig` file, there is an error output in the terminal logs. I wouldn't have expected an error, since we have a fallback in upstream, and git uses `master` if `defaultBranch` is not set.
Screenshot 2021-02-19 at 11 41 00

NunoAlexandre:
> When no default branch is set in the `.gitconfig` file, there is an error output in the terminal logs. I wouldn't have expected an error, since we have a fallback in upstream, and git uses `master` if `defaultBranch` is not set.
> Screenshot 2021-02-19 at 11 41 00

I see. This seems to be out of our reach though, looking at the API of [`childProcess.exec`](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback). So I would take it as [this point](https://github.com/radicle-dev/radicle-upstream/pull/1621#issuecomment-781987039) being there out of the box. 

What do you think?

native/main.ts
```diff
@@ -163,6 +164,26 @@ ipcMain.handle(RendererMessage.OPEN_URL, (_event, url) => {
...
+ipcMain.handle(RendererMessage.GET_GIT_GLOBAL_DEFAULT_BRANCH, async () => {
+  return await new Promise((resolve, reject) =>
+    childProcess.exec(
+      "git config --global --get init.defaultBranch",
+      {
+        encoding: "utf-8",
+      },
+      (error, stdout, stderr) => {
+        if (error) {
+          reject(error);
```
geigerzaehler, line 176: This throws now.Either we wrap the whole promise in `try/catch` or we don’t reject here.

NunoAlexandre: So my idea initially was to try to get the git config. If it failed, the only expectable reason would be the running git version not supporting it yet. So we'd either get it or the config is undefined, thus the `string | undefined` return type I had initially. In other terms, `Maybe<string>`.

When we moved to use the async alternative, since it returns a `Promise`, I felt divided between `Promise<string | undefined>` (aka `Promise<Maybe<string>>`) or capturing the error lane as a `reject` and see it more like an async `Either`.

Since we are not interested in the particular error, using a `Maybe`-like approach seems appropriate in the end. Meaning that we could wrap the whole promise in a `try/catch` and return `undefined` when it throws.

wdyt?

geigerzaehler: > Meaning that we could wrap the whole promise in a `try/catch` and return `undefined` when it throws.

Sounds good

NunoAlexandre: Done!

geigerzaehler (approved):

MeBrei (approved):

NunoAlexandre:
Merged in 1ad2ae34d8ebe97bef201f6fde5b63ee57b9b10e.

Issue #1496, opened 2020-12-26
Unable to create an identity


kon3gor:
I was trying to create an identity, but got this after entering passphrase:
```
{
  "code": "create-identity-failed",
  "message": "Could not create identity",
  "details": {
    "handle": "kon3gor"
  },
  "source": {
    "code": "UnknownException",
    "message": "Failed to fetch",
    "stack": "TypeError: Failed to fetch",
    "details": {
      "name": "TypeError"
    }
  }
}
```
Followed by this (different usernames because I tried different passwords and nicknames)
```
{
  "code": "create-identity-failed",
  "message": "Could not create identity",
  "details": {
    "handle": "eshendo"
  },
  "source": {
    "code": "UnknownException",
    "message": "A key already exists",
    "details": {
      "message": "A key already exists",
      "variant": "KEY_EXISTS"
    }
  }
}
```
And after few more attempts it became this
```
{
  "code": "create-identity-failed",
  "message": "Could not create identity",
  "details": {
    "handle": "kon3gor"
  },
  "source": {
    "code": "UnknownException",
    "message": "Incorrect input",
    "stack": "Error: Incorrect input\n    at un (file:///tmp/.mount_radicljQ30uK/resources/app.asar/public/bundle.js:1:27043)\n    at async yn (file:///tmp/.mount_radicljQ30uK/resources/app.asar/public/bundle.js:1:27513)\n    at async u (file:///tmp/.mount_radicljQ30uK/resources/app.asar/public/bundle.js:12:331027)",
    "details": {
      "response": {},
      "name": "Error"
    }
  }
}
```
How can I fix this?

AlexMikhalev:
I had the same issue, are you using master or pre-packaged?

kon3gor:
@AlexMikhalev  I'am using a file from [here](https://radicle.xyz/downloads.html). I believe it's what u called pre-packaged

AlexMikhalev:
@kon3gor two things helped me - compiling from git master and checking if you have any firewalls blocking communications.

imachug:
I have the same issue. I use Linux, Ubuntu 20.04. I built Upstream 0.1.6 from source. If that matters, I use Node 12.18.2, rustc 1.49.0.

I've repeated the following process several times: run `./scripts/reset-state.sh`, run `yarn start`, try to register. The results are `--+--+-`, where `-` means I get an error (the same all the time):

```json
{
  "code": "create-identity-failed",
[...]

PR #1623, opened 2021-02-20
Windows Support

bltavares:
- [x] Yarn/node scripts working
- [x] Electron starts but fails to start `radicle-proxy`
- [x] Update `radicle-proxy` to latest `radicle-link` commits
- [x] Update `coco` to latest `radicle-link` commits

Pending updates to `radicle-proxy` and `coco`:

-  Merged in #1463 [needs another attempt to update]
-  (likely needs) radicle-dev/radicle-link#515

Patches needed upstream:
- https://github.com/radicle-dev/radicle-link/pull/468
- https://github.com/radicle-dev/twemoji-svg-assets/pull/3


bltavares:
I've back-ported some of the changes from `radicle-link` on top of the current version used on `radicle-upstream`, and I was able to get things to compile, even tho they don't work properly yet.

image

I should be able to open a PR with only the electron changes needed to copile windows ([branch windows-backport](https://github.com/radicle-dev/radicle-upstream/compare/master...bltavares:backport-windows)) - but it would not yet be usable until `radicle-link` gets updated.

bltavares:
After deleting the keystore, I was able to recreate my credentials and have it running 🥳 
image

bltavares:
now that radicle-link has been updated (congrats!) we can try this again

bltavares:
`yarn run start:dev:win` should run it now ;)

bltavares:
Even tho it runs, it still has windows errors, likely with fixes needed on `radicle-link`

```
 Cloning failed with: fetcher error: invalid path for filesystem 'C:/Users/bltav/AppData/Roaming/radicle/radicle-link/data/17561e66-70b2-429c-8cbe-4d0a978d25bc/git/refs/namespaces/hnrkehankrg4owmaohjxr5gug3u3fpdnd3hxo/refs/remotes/hynkyndc6w3p8urucakobzna7sxwgcqny7xxtw88dtx3pkf7m3nrzc/rad/ids/hnrkqdpm9ub19oc8dccx44echy76hzfsezyio': The filename or extension is too long.
```

Should we merge this or wait for more fixes to make it fully useable?

bltavares:
I was able to circumvent windows long-path limits and get Radicle Upstream to clone projects from the sprout server

```pwsh
$env:RAD_HOME="D:\rad"
yarn run start:dev:win
```

image

I had a few issues starting the project - something about #1496 race condition. It looks like the `POST /v1/identities` has a race condition, and when I added a `console.error` on the onboarding screen, between creating the keystore and fetching the identity, I was able to continue.

I think we are good to go with the `RAD_HOME` pointing to a root folder :)

bltavares:
If someone wants to try [on a Windows VM](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/) close to what others might have (i.e. no developer tools) you [can test this binary](https://github.com/bltavares/radicle-upstream/releases/tag/pr-1623)

rudolfs (changes requested):
Really awesome work, Bruno! 🎉 
Can't wait for this to land :)

Left some remarks and questions inline. 

> Should we merge this or wait for more fixes to make it fully usable?

I think if we clearly document it as being experimental, we could merge it in. This way we don't have to put in more work keeping this branch up-to-date.

Here's a couple of things I'd like to clarify before moving on:
  - did you try running the unit and integration tests on windows, if so, how much breakage was there?
  - I think it would be also beneficial to document the next steps to make the windows user experience on-par with what we already have for Linux/macOS (e.g. running tests on a Windows CI, listing all the little things that need to be fixed, etc.)

And lastly, we require all commits to be verified, see [here](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification) for how to set that up.

package.json
```diff
@@ -133,16 +143,20 @@
     "webpack-cli": "^4.6.0"
   },
   "scripts": {
-    "start": "RADICLE_UPSTREAM_PROXY_PATH=./target/release/radicle-proxy yarn _private:start",
-    "start:dev": "RADICLE_UPSTREAM_PROXY_PATH=./target/debug/radicle-proxy yarn _private:start:dev",
-    "start:test": "RADICLE_UPSTREAM_PROXY_PATH=./target/release/radicle-proxy RADICLE_UPSTREAM_PROXY_ARGS='--test' yarn _private:start",
+    "start": "cross-env RADICLE_UPSTREAM_PROXY_PATH=./target/release/radicle-proxy yarn _private:start",
+    "start:win": "cross-env RADICLE_UPSTREAM_PROXY_PATH=./target/release/radicle-proxy.exe yarn _private:start",
+    "start:dev": "cross-env RADICLE_UPSTREAM_PROXY_PATH=./target/debug/radicle-proxy yarn _private:start:dev",
+    "start:dev:win": "cross-env RADICLE_UPSTREAM_PROXY_PATH=./target/debug/radicle-proxy.exe yarn _private:start:dev",
```
geigerzaehler, line 149: I’m wondering whether the proxy path should be detected by the app. We would only need one script then.

package.json
```diff
@@ -133,16 +143,20 @@
...
+    "start:test": "cross-env RADICLE_UPSTREAM_PROXY_PATH=./target/release/radicle-proxy RADICLE_UPSTREAM_PROXY_ARGS='--test' yarn _private:start",
+    "start:test:win": "cross-env RADICLE_UPSTREAM_PROXY_PATH=./target/release/radicle-proxy.exe RADICLE_UPSTREAM_PROXY_ARGS='--test' yarn _private:start",
     "ethereum:start": "./scripts/ethereum-dev-node.sh",
-    "test": "TZ='UTC' yarn test:unit && TZ='UTC' yarn test:integration",
-    "test:integration": "TZ='UTC' run-p --race _private:proxy:start:test _private:test:integration",
-    "test:integration:debug": "TZ='UTC' run-p --race _private:webpack:ui:watch _private:proxy:start:test:watch _private:test:integration:debug",
+    "test": "cross-env TZ='UTC' yarn test:unit && TZ='UTC' yarn test:integration",
```
geigerzaehler, line 153: I’m curious why we don’t use `cross-env` for the second command `TZ=UTC yarn test:integration`.

proxy/coco/src/state.rs
```diff
@@ -1149,6 +1149,7 @@ pub mod test {
         );
 
         assert!(!project_names.contains(&fakie.subject().name.to_string()));
+        drop(tmp_dir); // ensure repo lives until end of test
```
geigerzaehler, line 1152: Could you explain why you added this? Without the `drop()` Rust should keep `tmp_dir` around until it goes out of scope, i.e. here. Maybe you were running into this error (#1378)?

```
test can_ask_and_clone_project ... fatal: '.' does not appear to be a git repository
```

geigerzaehler (commented):
Great effort! Looking forward to having this merged.

geigerzaehler:
I’ve added issue #1847 for Windows CI builds.

package.json
bltavares: I missed it hehe

proxy/coco/src/state.rs
bltavares: In this case we can remove it. I had a few tests on previously where the tmpdir was moved into a method, and was dropped before finishing the test.

You are right, the method should drop by the end of it. I've added the `drop` as a guardrail to prevent moving the tmpdir and dropping it earlier with a compiler error.

package.json
bltavares: Yes, I've considered this also, but it was more effort

ui/src/profile.ts
```diff
@@ -3,6 +3,7 @@ import { derived, Readable } from "svelte/store";
 import * as error from "./error";
 import * as localPeer from "./localPeer";
 import * as project from "./project";
+import { Status } from "./remote";
```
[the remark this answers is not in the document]
bltavares: On VSCode it was complaining about it

ui/Screen/Onboarding.svelte
```diff
@@ -44,7 +44,14 @@
...
       const identity = await withRetry(
-        () => createIdentity({ handle }),
+        async () => {
+          try {
+            return await createIdentity({ handle });
+          } catch (_) {
+            // The proxy might fail the first request after store creation
+            throw new Error("Failed to fetch");
+          }
+        },
```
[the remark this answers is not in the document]
bltavares: The `withRetry` only fails if it gets a `Failed to fetch` message, meaning the connection didn't work. Here, the proxy is returning a `403` and it's not retrying as it has a different error message.

I'm wrapping it to trigger the retry on 403 as well. There is likely a more elegant way to allow retries on different status code but I mostly wanted to get it working.

proxy/api/src/config.rs
```diff
@@ -45,7 +45,10 @@ pub fn store_dir(profile_id: &str) -> path::PathBuf {
 ///
 ///   * Could not get the user home path from the HOME env variable
 pub fn bin_dir() -> Result<path::PathBuf, Error> {
-    let home_dir = std::env::var("HOME")?;
+    let home_dir = std::env::var("HOME").or_else(|_| {
+        // Windows don't have HOME by default
+        std::env::var("USERPROFILE")
+    })?;
```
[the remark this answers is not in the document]
bltavares: Windows users might have set HOME previously - specially on dev envs. It could be done with `cfg` tho.

package.json
```diff
@@ -210,5 +224,8 @@
       "prettier --write",
       "eslint --fix --max-warnings=0"
     ]
+  },
+  "volta": {
+    "node": "14.15.5"
```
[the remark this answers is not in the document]
bltavares: I can remove it - no need to keep all those tools on the repo. Volta only runs if this entry exists, so it's a side-effect from my env.

bltavares:
> did you try running the unit and integration tests on windows, if so, how much breakage was there?
Most of the tests are passing. There are a few integration tests failing, likely due to difference between windows/linux concurrent file access and tmpdir location and deletion.

> And lastly, we require all commits to be verified
Do they need to be GPG signed, not only with `signed off`?

package.json
geigerzaehler: :ok_hand: Would you mind creating an issue for this after the PR is merged?

rudolfs:
> Do they need to be GPG signed, not only with signed off?

Correct, we also [require GPG signing](https://github.com/radicle-dev/radicle-upstream/blob/master/DEVELOPMENT.md#how-do-we-work) in this repo.

ui/Screen/Onboarding.svelte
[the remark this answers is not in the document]
geigerzaehler: This is curious: `createIdentity` should never return a 403 :thinking:. If it does, there is something wrong and we don’t want to retry.  Can you reproduce this @bltavares? If so, we should open a separate issue.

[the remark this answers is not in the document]
bltavares: It happens frequently, but sometimes it works. There seem to be a race condition on the proxy, where the `session.createKeystore()` respond before the session token is stored, and then `createIdentity` fails to authenticate. #1496 seems to showcase it's happening on other OSes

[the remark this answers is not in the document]
geigerzaehler: To me it seems like #1496 describes an issue where identity creation fails permanently and requires manual interaction to resolve. If we keep retrying this would just delay showing an error. If it fails once and then recovers we also loose valuable information about existing errors if we keep re-trying. This is why I think we should not include this code and find a solution to the underlying issue.

Could you add error information to #1496. You can get it by clicking “copy to clipboard” when an error shows. Our error reporting infrastructure has improved since the issue was opened.

proxy/coco/src/state.rs
geigerzaehler: Not relevant anymore. The file is gone.

package.json
```diff
@@ -35,6 +35,14 @@
...
         "to": "./"
       },
+      {
+        "from": "target/release/radicle-proxy.exe",
+        "to": "./"
+      },
+      {
+        "from": "target/release/git-remote-rad.exe",
+        "to": "./"
+      },
```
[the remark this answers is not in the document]
rudolfs: I just tried the globs, didn't work.

ui/Screen/Onboarding.svelte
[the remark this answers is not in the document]
rudolfs: I removed this in https://github.com/radicle-dev/radicle-upstream/pull/2375.

package.json
rudolfs: Done in https://github.com/radicle-dev/radicle-upstream/pull/2375.

package.json
rudolfs: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/2375.

rudolfs:
Closing this in favour of: https://github.com/radicle-dev/radicle-upstream/pull/2375.

Issue #1612, opened 2021-02-12
Make project title clickable as a way to go back to the readme view


juliendonck:
Currently the only way to "reset" the project view is by clicking on "Files" in the tabs. That's not always clear as it's already selected. @cloudhead suggested we used the project title to also do this:
<img alt="Screen Shot 2021-02-10 at 16 47 03" width="611" src="https://user-images.githubusercontent.com/2326909/107533836-a371c580-6bbf-11eb-8df5-b6c5e51e11ef.png">

NunoAlexandre:
This is low-hanging fruit, I will tackle it.

PR #1630, opened 2021-02-23
feat(ui): click on project name to go to root view

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-upstream/issues/1612

~~Note: will add a test spec-ing this.~~ :heavy_check_mark:

### DIFF
--- a/cypress/integration/project/source_browsing.spec.ts
+++ b/cypress/integration/project/source_browsing.spec.ts
@@ -191,8 +191,8 @@ context("project source browsing", () => {
               )
               .should("exist");
 
-            // Going to a different path and then switching back to the root path
-            // shows the README again.
+            // Going to a different path and then clicking on 'Files'
+            // sends us back to the project root, showing the README again.
             commands.pick("source-tree").within(() => {
               cy.contains(".i-am-well-hidden").click();
             });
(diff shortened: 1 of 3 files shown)


geigerzaehler (dismissed):

MeBrei (approved):
👍

cypress/integration/project/source_browsing.spec.ts
```diff
@@ -214,6 +214,22 @@ context("project source browsing", () => {
...
+            commands.pick("project-screen", "commits-tab").click();
+            commands
+              .pick("horizontal-menu", "commits-tab")
+              .get("p")
+              .should("have.class", "active");
+            commands.pick("header", "entity-name").click();
+            commands
+              .pick("horizontal-menu", "files-tab")
+              .get("p")
+              .should("have.class", "active");
```
geigerzaehler, line 228: Unfortunately this does not do what you would expect. `.get("p")` will select _all_ `p` elements and just find one that has the `active` class. In general we should also try to avoid using classes and element names to control the tests as they are prone to break.

My suggestion for this test would be to drop the assertion here since it is effectively checked by the assertion below that we are back at the project “home” screen. As for the assertion above I’d suggest to use `commands.pick("project-screen", "file-view").should("not.exist")`. We don’t want to check any specifics about the commits tab, just that we go back to the project home if we were not on it.

geigerzaehler (dismissed):

MeBrei (dismissed):

NunoAlexandre:
Merged in eb0d62bd71ac47a3d0492000b72afa683ffcfc7c

PR #1633, opened 2021-02-24
refactor(ui): use idiomatic async code instead of Promise

geigerzaehler:
We rewrite some code that uses the `Promise` constructor and callbacks to more idiomatic async/await code and log the error that was swallowed previously.

### DIFF
--- a/ui/src/source.ts
+++ b/ui/src/source.ts
@@ -1,6 +1,7 @@
 import { format } from "timeago.js";
 
 import * as api from "./api";
+import * as error from "./error";
 import type { PeerId } from "./identity";
 import type { Urn } from "./urn";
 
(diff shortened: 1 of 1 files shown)


xla:
What is the signal of that error besides that the project doesn't have a readme? Which is already surfaced in the ui.

geigerzaehler:
> What is the signal of that error besides that the project doesn't have a readme? Which is already surfaced in the ui.

If the project does not have a readme then the function returns early because `path` is `null`. So if `fetchBlob` throws then we have a readme but we failed to fetch it for some reason. In this case we should show an error.

xla (approved):
🐝 🚭 🈚️ 🏘

NunoAlexandre (approved):

geigerzaehler:
Merged in 4a6788500722f1144d61e2e0de39e0afc8dc82bb

PR #1638, opened 2021-02-24
test: commit replication (#1589)

### LINKED ISSUE #1529: Automate manual networking QA tests

geigerzaehler:
Backport of https://github.com/radicle-dev/radicle-upstream/pull/1589 to `master`.

Original description

---

* Automate remaining networking QA steps
* Use pasteInto to speed up specs
* Run integration tests in --release mode to speed them up
* Rename storagePath to radHome
* Don't block the plugin thread
* Fix race condition in remote management modal

The remote list in the modal sometimes wouldn't update correctly showing
a newly added remote in a "still looking" state forever. This happened
consistently on CI because of subtle timing differences.

The cause of this bug was that clicking the "Add" remote button would
trigger a backend request to get the updated list, at this moment the
list would still contain the remote in a "still looking" state. Meanwhile
the backend would find the remote and fire a refresh event that the UI
would pick up and update the store to show the found remote, this all
happens before the first request completes. At this moment the remote
list shows the correct state, but now the first request finishes and
re-updates the list with the old value showing the remote again in a
"still looking" state.

To prevent the race condition we additionally add the locking mechanism
to the fetchPeers function.

Closes #1529

### DIFF
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -9,37 +9,56 @@ context("p2p networking", () => {
       "reacts to network state changes",
       { defaultCommandTimeout: 8000 },
       () => {
-        nodeManager.withTwoOnboardedNodes({}, (node1, node2) => {
-          nodeManager.asNode(node1);
-          commands.pick("connection-status-offline").should("exist");
-          nodeManager.connectTwoNodes(node1, node2);
-          commands.pick("connection-status-online").should("exist");
-          nodeManager.asNode(node2);
-          commands.pick("connection-status-online").should("exist");
-        });
+        nodeManager.withTwoOnboardedNodes(
+          { node1User: {}, node2User: {} },
+          (node1, node2) => {
+            nodeManager.asNode(node1);
+            commands.pick("connection-status-offline").should("exist");
+            nodeManager.connectTwoNodes(node1, node2);
+            commands.pick("connection-status-online").should("exist");
+            nodeManager.asNode(node2);
+            commands.pick("connection-status-online").should("exist");
+          }
+        );
       }
     );
   });
 
   it("replicates a project from one node to another", () => {
+    const maintainer = {
+      handle: "rudolfs",
+      fullName: "Rūdolfs Ošiņš",
+      passphrase: "1111",
+    };
+    const contributor = {
+      handle: "abbey",
+      fullName: "Abbey Titcomb",
+      passphrase: "2222",
+    };
+
     commands.withTempDir(tempDirPath => {
       nodeManager.withTwoOnboardedNodes(
-        { node1Handle: "rudolfs", node2Handle: "abbey" },
+        {
+          node1User: maintainer,
+          node2User: contributor,
+        },
         (maintainerNode, contributorNode) => {
           nodeManager.connectTwoNodes(maintainerNode, contributorNode);
           nodeManager.asNode(maintainerNode);
-          const newProjectPath = path.join(
-            tempDirPath,
-            "maintainerNode/new-project"
-          );
 
-          cy.exec(`mkdir -p ${newProjectPath}`);
+          const maintainerNodeWorkingDir = path.join(
+            tempDirPath,
+            "maintainerNode"
+          );
+          cy.exec(`mkdir -p ${maintainerNodeWorkingDir}`);
 
           ipcStub.getStubs().then(stubs => {
-            stubs.IPC_DIALOG_SHOWOPENDIALOG.returns(newProjectPath);
+            stubs.IPC_DIALOG_SHOWOPENDIALOG.returns(maintainerNodeWorkingDir);
           });
+          const projectName = "new-fancy-project.xyz";
+
           commands.pick("new-project-button").click();
-          commands.pick("name").type("new-fancy-project.xyz");
+          commands.pasteInto(["name"], projectName);
 
           commands.pick("new-project").click();
           commands.pick("new-project", "choose-path-button").click();
@@ -60,10 +79,11 @@ context("p2p networking", () => {
             }
 
             nodeManager.asNode(contributorNode);
+
             cy.log("navigate to the 'Following' tab");
             commands.pick("following-tab").click();
             commands.pick("sidebar", "search").click();
-            commands.pick("search-input").type(urn);
+            commands.pasteInto(["search-input"], urn);
             commands.pick("follow-toggle").click();
           });
 
@@ -75,7 +95,7 @@ context("p2p networking", () => {
             )
             .click();
 
-          cy.log("the maintainer shows up in the peer selector");
+          cy.log("maintainer shows up in the peer selector");
           commands
             .pickWithContent(["peer-selector"], "rudolfs")
             .should("exist");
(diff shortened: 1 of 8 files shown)


cypress/integration/networking.spec.ts
```diff
@@ -112,6 +132,111 @@ context("p2p networking", () => {
...
+            subject: maintainerCommitSubject,
+            passphrase: maintainer.passphrase,
+            name: maintainer.fullName,
+          });
+
+          cy.log("refresh the UI for the new commit to show up");
+          cy.get("body").type("{esc}");
+          commands.pick("sidebar", "profile").click();
+
+          cy.log("make sure new commit shows up in the maintainer's UI");
```
NunoAlexandre, line 155: I like these `cy.log` :ok_hand:

cypress/plugins/nodeManager/shared.ts
```diff
@@ -19,3 +23,10 @@ export enum Commands {
   GetOnboardedNodes = "getOnboardedNodes",
   ConnectNodes = "connectNodes",
 }
+
+// A directory that can be used for temporary test data.
+//
+// It is located within this repository so that there is no extra setup
+// necessary when using it locally or on CI. To avoid committing any left-over
+// temp data this directory ignored via .gitignore.
+export const CYPRESS_WORKSPACE_PATH = path.join(__dirname, "../../workspace");
```
NunoAlexandre, line 32:
```suggestion
export const CYPRESS_WORKSPACE_PATH = path.join(__dirname, "../../workspace");

```

cypress/support/nodeManager.ts
```diff
@@ -24,6 +29,37 @@ export const connectTwoNodes = (
...
+  cy.exec(
+    `set -euo pipefail
+export PATH=$PWD/target/release:$PATH
+cd ${options.repositoryPath}
+git commit --allow-empty -m "${options.subject}"
+git -c credential.helper=${credentialsHelper(options.passphrase)} push rad`,
+    {
+      env: {
+        RAD_HOME: options.radHome,
+        GIT_AUTHOR_NAME: options.name || "John McPipefail",
```
NunoAlexandre, line 54: On "John McPipefail"

Does that mean that John Doe married and change his last name?

package.json
```diff
@@ -154,8 +154,8 @@
     "_private:proxy:build": "cargo build --all-features --all-targets",
     "_private:proxy:build:release": "cargo build --release",
     "_private:proxy:clean": "cargo clean",
-    "_private:proxy:start:test": "cargo build --bin git-remote-rad && cargo run -- --test",
-    "_private:proxy:start:test:watch": "cargo build --bin git-remote-rad && cargo watch -x 'run -- --test'",
+    "_private:proxy:start:test": "cargo build --release --bin git-remote-rad && cargo run --release -- --test",
```
NunoAlexandre, line 157: Won't run this with `--release` slow things down considerably?

cypress/integration/networking.spec.ts
```diff
@@ -9,37 +9,56 @@ context("p2p networking", () => {
...
 
           ipcStub.getStubs().then(stubs => {
-            stubs.IPC_DIALOG_SHOWOPENDIALOG.returns(newProjectPath);
+            stubs.IPC_DIALOG_SHOWOPENDIALOG.returns(maintainerNodeWorkingDir);
           });
+          const projectName = "new-fancy-project.xyz";
+
           commands.pick("new-project-button").click();
-          commands.pick("name").type("new-fancy-project.xyz");
+          commands.pasteInto(["name"], projectName);
```
NunoAlexandre, line 61: I would suggest us undoing these changes from `pick(_).type(_)` to `pastInto(_, _)` for the sake of consistency and avoiding introducing another way of doing things that doesn't (seem to) offer any benefit.

NunoAlexandre: Ok, I see the note in the description saying that `pasteInto` speeds things up. I would still consider if that marginal speed gain is worth the introduction of inconsistency.

geigerzaehler:
Thanks for the review @NunoAlexandre. This is simply a backport #1592  from `next` to `master` and has gone through review already. It does not make sense to change the PR because that would create conflicts with #1463.

NunoAlexandre:
> Thanks for the review @NunoAlexandre. This is simply a backport #1592 from `next` to `master` and has gone through review already. It does not make sense to change the PR because that would create conflicts with #1463.

Ok! the linked PR as an empty diff so I couldn't tell if anything was added differently :)

geigerzaehler:
> Ok! the linked PR as an empty diff so I couldn't tell if anything was added differently :)

Sorry. I linked the wrong PR. The right one is https://github.com/radicle-dev/radicle-upstream/pull/1589

package.json
FintanH: It speeds up signing things with the keystore

cypress/integration/networking.spec.ts
```diff
@@ -112,6 +132,111 @@ context("p2p networking", () => {
               "Following"
             )
             .should("exist");
+
+          cy.log("test commit replication from maintainer to contributor");
+
+          cy.log("add a new commit to the maintainer's project working dir");
+          const projctPath = path.join(maintainerNodeWorkingDir, projectName);
```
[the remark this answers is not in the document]
FintanH: I think typos are worth fixing even in a backport @geigerzaehler  :smile:

FintanH (approved):

package.json
NunoAlexandre: Papoy!

cypress/integration/networking.spec.ts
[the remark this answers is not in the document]
geigerzaehler: I’ll follow up on this but I’d like to avoid merge conflicts at the moment.

NunoAlexandre (approved):

geigerzaehler:
Merged in b50c5d7787210427fd73c67334a4334e6883a2f3

PR #1639, opened 2021-02-24
refactor(ui): electron start path improvements

geigerzaehler:
A two changes that improve the path handling when the app starts

* `RADICLE_UPSTREAM_PROXY_PATH` is now relative to the working directory instead of the source file. Now the user can set this properly without knowing the location of the file.
* We create the electron storage directory recursively so that it works even if `RAD_HOME` does not exist yet.

### DIFF
--- a/native/main.ts
+++ b/native/main.ts
@@ -19,7 +19,7 @@ let proxyArgs: string[] = [];
 
 if (isDev) {
   if (process.env.RADICLE_UPSTREAM_PROXY_PATH) {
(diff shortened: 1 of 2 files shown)


NunoAlexandre (commented):
:bee: :honey_pot: :bear:

NunoAlexandre (approved):
:bee: :honey_pot: :bear:

MeBrei (approved):

PR #1649, opened 2021-03-03
fix(build): Don’t minify rollup bundles for dev builds

geigerzaehler:
We only minify the frontend rollup bundle when we are creating a release build. This decreases the bundling time by roughly 40% for non-watch builds. It has no effect on watched builds though.


NunoAlexandre (approved):

rudolfs (dismissed):
Looks good!

Could you provide an example how this can be used, please?

geigerzaehler:
> Could you provide an example how this can be used, please?

If I don’t need to watch the source files and I’m required to restart electron I do the following
~~~bash
yarn run rollup -c
RADICLE_UPSTREAM_PROXY_PATH=target/release/radicle-proxy NODE_ENV=development npx electron  .
~~~
I can easily restart electron without rebuilding all the source files.

PR #1650, opened 2021-03-04
tracking memes: Announce after clone

xla:
This is stacked on top of tyn for review pleasure of one @FintanH 

### DIFF
--- a/proxy/coco/src/peer/subroutines.rs
+++ b/proxy/coco/src/peer/subroutines.rs
@@ -421,9 +421,13 @@ async fn clone(
     match state::clone_project(&peer, urn.clone(), remote_peer, None, None).await {
         Ok(_urn) => {
             sender
-                .send(Input::Request(input::Request::Cloned(urn, remote_peer)))
+                .send(Input::Request(input::Request::Cloned(
+                    urn.clone(),
+                    remote_peer,
+                )))
                 .await
                 .ok();
+            gossip::announce(&peer, &urn, None);
         },
         Err(err) => {
             log::warn!(
--- a/proxy/coco/tests/gossip.rs
+++ b/proxy/coco/tests/gossip.rs
@@ -1,5 +1,6 @@
 use std::time::{Duration, SystemTime};
 
+use assert_matches::assert_matches;
 use futures::{future, StreamExt as _};
 use tokio::time::timeout;
 
@@ -131,6 +132,7 @@ async fn can_ask_and_clone_project() -> Result<(), Box<dyn std::error::Error>> {
     let alice_peer = build_peer(&alice_tmp_dir, RunConfig::default()).await?;
     let alice_peer_id = alice_peer.peer.peer_id();
     let alice = state::init_owner(&alice_peer.peer, "alice".to_string()).await?;
+    let mut alice_events = alice_peer.subscribe();
 
     let (alice_peer, alice_addrs) = {
         let peer = alice_peer.peer.clone();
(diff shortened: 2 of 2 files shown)


proxy/coco/tests/gossip.rs
```diff
@@ -174,9 +177,24 @@ async fn can_ask_and_clone_project() -> Result<(), Box<dyn std::error::Error>> {
 
     requested(query_listener, &urn).await?;
     assert_cloned(clone_listener, &urn.clone(), alice_peer_id).await?;
-    // TODO(finto): List projects
-    let project = state::get_project(&bob_peer, urn).await;
+    let project = state::get_project(&bob_peer, urn.clone()).await;
     assert!(project.is_ok());
 
+    let announced = async_stream::stream! { loop { yield alice_events.recv().await } }
```
FintanH, line 183: It might make more sense for `alice` to `track` `bob` before he requests, to line up with the case we were testing for. And then also assert that `alice` has `bob` as a remote of the project.

proxy/coco/tests/gossip.rs
```diff
@@ -174,9 +177,24 @@ async fn can_ask_and_clone_project() -> Result<(), Box<dyn std::error::Error>> {
...
+        .filter_map(|res| match res.unwrap() {
+            coco::PeerEvent::GossipFetched {
+                gossip, provider, ..
+            } if provider.peer_id == bob_peer_id && gossip.urn.id == urn.id => {
```
FintanH, line 187: So in my other comment, I noticed we weren't using the local `PeerId`. I have there are two `GossipFetched` events appearing, one with `alice`'s `PeerId` and one with `bob`'s. But I'm not 100% on this :thinking: Nonetheless, I'm a little surprised this worked :smile:

proxy/coco/tests/gossip.rs
xla: But that's not really what we test here. The thing that is asserted is that there will be a gossip message after a project has been successfully cloned through the waiting room.

FintanH: Then I would say it would be a nice _addition_ because that's the behaviour we were trying to test out and confirm.

proxy/coco/src/peer/subroutines.rs
```diff
@@ -421,9 +421,13 @@ async fn clone(
...
         Ok(_urn) => {
             sender
-                .send(Input::Request(input::Request::Cloned(urn, remote_peer)))
+                .send(Input::Request(input::Request::Cloned(
+                    urn.clone(),
+                    remote_peer,
+                )))
                 .await
                 .ok();
+            gossip::announce(&peer, &urn, None);
```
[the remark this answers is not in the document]
FintanH: Nevermind, reading things wrong :upside_down_face:

proxy/coco/tests/gossip.rs
```diff
@@ -165,18 +169,60 @@ async fn can_ask_and_clone_project() -> Result<(), Box<dyn std::error::Error>> {
...
-            .urn()
+            .urn();
+
+        urn
     };
 
+    // Alice will track Bob in anticipation of upcoming contributions.
+    state::track(&alice_peer, urn.clone(), bob_peer_id).await?;
+
+    // Make sure Bob is NotReplicaed.
```
FintanH, line 182:
```suggestion
    // Make sure Bob is NotReplicated.
```

FintanH (approved):
Sick :call_me_hand:

PR #1651, opened 2021-03-04
chore: spawn peer run explicitly in tests

### LINKED ISSUE #1648: Tests leak resources due to `into_running()`
The proxy tests commonly use
~~~rust
tokio::task::spawn(peer.into_running())
~~~
in conjunction with a temporary directory that the peer reads and writes.

Since we don’t store the task handle and abort it, the peer will run as long as the tokio executor is running. This leaks resources and might interfer with the rest of the environment. One particular issue is that the temporary directory created by `tempfile::tempdir()` is dropped and the peer still tries to write to it which leads to errors and potentially failing tests.

To address this issue we need to rethink the following pattern
~~~rust
let coco_peer = coco::bootstrap(/* uses tmpdir */).await?;
let peer = coco_peer.peer.clone();
let peer_control = coco_peer.control();
tokio::spawn(coco_peer.into_running());
return (peer_control, peer)
~~~
The goal would be to return a handle that stops the peer when it is dropped.

xla:
Fixes #1648

### DIFF
--- a/proxy/api/src/context.rs
+++ b/proxy/api/src/context.rs
@@ -198,13 +198,15 @@ impl Unsealed {
     /// * coco key creation fails
     /// * creation of the [`kv::Store`] fails
     #[cfg(test)]
-    pub async fn tmp(tmp_dir: &tempfile::TempDir) -> Result<Self, crate::error::Error> {
+    pub async fn tmp(
+        tmp_dir: &tempfile::TempDir,
+    ) -> Result<(Self, impl std::future::Future<Output = ()>), crate::error::Error> {
         let store = kv::Store::new(kv::Config::new(tmp_dir.path().join("store")))?;
 
         let key = coco::keys::SecretKey::new();
         let signer = signer::BoxedSigner::from(signer::SomeSigner { signer: key });
 
-        let (peer_control, peer) = {
+        let (peer_control, peer, run_handle) = {
             let config = coco::config::default(signer.clone(), tmp_dir.path())?;
             let disco = coco::config::static_seed_discovery(&[]);
             let coco_peer =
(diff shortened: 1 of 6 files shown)


FintanH (commented):
I really think a new `struct` using `Deref` would be a cleaner approach over using a tuple. At least then we know the handle is always being dropped when the context is finished being used, whereas with the tuple we have to guess when to unpack and `abort`.

xla:
> I really think a new `struct` using `Deref` would be a cleaner approach over using a tuple. At least then we know the handle is always being dropped when the context is finished being used, whereas with the tuple we have to guess when to unpack and `abort`.

We already have a illustrates amount of "structs": Context -> coco::Peer -> net::Peer. We seem to just move the issue. with the extremely small amount of tests depending on a running peer it seems unnecessary to introduce more infrastructure to enable that use-case.

xla:
What would the `Deref` deref to?

FintanH:
> What would the Deref deref to?

It would deref to `Unsealed`, so something like:
```rust
#[cfg(test)]
pub struct TestCtx {
  ctx: Unsealed,
  run_handle: Handle, // or whatever the type is
}

impl Deref for TestCtx {
  type Target = Unsealed;

  fn deref(&self) -> &Self::Target
}
```

> We already have a illustrates amount of "structs": Context -> coco::Peer -> net::Peer. We seem to just move the issue. with the extremely small amount of tests depending on a running peer it seems unnecessary to introduce more infrastructure to enable that use-case.

I'm seeing the tuple as ad-hoc infrastructure. I think it would make it more manageable in the tests to use the struct. If you think is the better route, then I'm happy to go with it.

geigerzaehler:
I like the idea of having a `TestCtx`. One thing this would help with is that we can spawn the peer task when `TestCtx` and abort it when `TestCtx` is dropped. This way we could eliminate the the additional calls to `tokio::spawn()` and `handle.abort()` in the tests. This would also allow us to take ownership of `tempdir` and ensure that it is dropped after the peer is aborted. But I’m also fine with the current implementation which is already an improvement.

On `Deref`: Using `Deref` for polymorphism is mostly considered an [antipattern](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html). The standard library documentation also says that [“Deref should only be implemented for smart pointers to avoid confusion.”](https://doc.rust-lang.org/std/ops/trait.Deref.html). This is why it’d try to avoid it and use the `ctx` field explicitly.

Issue #1512, opened 2021-01-09
feat: deep-linking


abbey-titcomb:
[From radicle.community](https://radicle.community/t/http-links-to-open-in-the-client/463)

I have one feature suggestion. It would be cool to have a button to copy the URL that opens the Radicle client app, similarly to how https://t.me/telegram is opened by Telegram, or https://us04web.zoom.us/j/XXX?pwd=XXX in Zoom.

It’d be very convenient to share such links on the internet in general, but my goal is to make a badge in the repo readme with the link to open my project in Radicle. Does it sound like a good idea?

the URL could be something like: https://open.radicle.xyz/<project_id>

Adding to [Feedback Hub](https://github.com/orgs/radicle-dev/projects/19?add_cards_query=is%3Aopen)

dragon788:
Related to #1498 as webapps (PWA/SPA/etc) can also be registered as system protocol handlers in some cases.

abbey-titcomb:
@cloudhead said we could do this with xdg-open for Linux, @rudolfs - any idea what we could do for Mac?

rudolfs:
I think the sanest cross-platform way to get Upstream to open if a user clicks a special link somewhere in a browser would be registering a custom protocol. Basically what @dragon788 mentioned above and @xla [suggested back in the day](https://github.com/radicle-dev/radicle-upstream/issues/840#issuecomment-703491425).

rudolfs:
> the URL could be something like: https://open.radicle.xyz/<project_id>

I'm not entirely sure what's the benefit of this additional layer of wrapping. Could you elaborate on that?

cloudhead:
The additional layer is definitely not needed for now.

abbey-titcomb:
Saw another tweet that references this today - https://twitter.com/grin_io/status/1352630590173016066?s=20

Would be cool to try to build this in soon !

NunoAlexandre:
@abbey-titcomb I'd be interested in picking this up. For that, I would like us to scope this and come up with a v1 concept. My idea is for this fist iteration is to support deep linking for projects only.

Now, if users contain the linked project locally, we would simply open the screen with that project's overview.
If the user does \**not\** contain a local copy of that project, they'd first need to follow it. For that, I'd say it would be nice to show a modal informing the user that the project they requested to open is not found locally and asking the user whether they want to follow it. Alternatively, we could just fill in the value in the search box, but that would provide no context for unfamiliar users.

[...]

PR #1652, opened 2021-03-05
feat(ui): open upstream via a custom protocol

rudolfs:
This PR implements the simplest possible deep-linking support. It's based on this [community discussion](https://radicle.community/t/opening-upstream-via-links-on-the-web/1856). Implementing a parser/router for the [proposed URL scheme](https://github.com/radicle-dev/radicle-decisions/pull/10) is left as a follow-up for when we all agree on the scheme. For now the parser just looks for the first `Radicle ID` in the URI and shows the search screen with the input field pre-filled with that URI.

Refs https://github.com/radicle-dev/radicle-upstream/issues/1512
Refs https://github.com/radicle-dev/radicle-decisions/pull/10
Refs https://github.com/radicle-dev/radicle-docs/pull/126

- [x] make it work on macOS
- [x] make it work on Linux
- [x] add section in docs.radicle.xyz about how to set up a `.desktop` file by hand for Linux users: https://github.com/radicle-dev/radicle-docs/pull/126
- [x] tests
- [x] implement parser/handler according to spec

Test this from a browser:
```html
<a href="radicle://link/v0/rad:git:hnrkjm5z3rwae9g3n6jhyo6kzh9eup5ku5odo">radicle://link/v0/rad:git:hnrkjm5z3rwae9g3n6jhyo6kzh9eup5ku5odo</a>
```

Or from the terminal:
```sh
// Linux
xdg-open "radicle://link/v0/rad:git:hnrkjm5z3rwae9g3n6jhyo6kzh9eup5ku5odo"

// macOS
open "radicle://link/v0/rad:git:hnrkjm5z3rwae9g3n6jhyo6kzh9eup5ku5odo"
```

### DIFF
--- a/native/index.ts
+++ b/native/index.ts
@@ -16,6 +16,7 @@ import {
   MainProcess,
   mainProcessMethods,
 } from "./ipc-types";
+import { handleCustomProtocolInvocation } from "./nativeCustomProtocolHandler";
 
 const isDev = process.env.NODE_ENV === "development";
 
@@ -140,6 +141,18 @@ class WindowManager {
 
     this.window = window;
   }
+
+  focus() {
+    if (!this.window) {
+      return;
+    }
+
+    if (this.window.isMinimized()) {
+      this.window.restore();
+    }
+
+    this.window.focus();
+  }
 }
 
 const windowManager = new WindowManager();
@@ -230,27 +243,61 @@ app.on("will-quit", () => {
   proxyProcessManager.kill();
 });
 
-// This method will be called when Electron has finished
-// initialization and is ready to create browser windows.
-// Some APIs can only be used after this event occurs.
-app.on("ready", () => {
-  proxyProcessManager.run().then(({ status, signal, output }) => {
+// Handle custom protocol on macOS
+app.on("open-url", (event, url) => {
+  event.preventDefault();
+  handleCustomProtocolInvocation(url, url => {
     windowManager.sendMessage({
-      kind: MainMessageKind.PROXY_ERROR,
-      data: {
-        status,
-        signal,
-        output,
-      },
+      kind: MainMessageKind.CUSTOM_PROTOCOL_INVOCATION,
+      data: { url },
+    });
+  });
+});
+
+if (app.requestSingleInstanceLock()) {
+  // Handle custom protocol on Linux when Upstream is already running
+  app.on("second-instance", (_event, argv, _workingDirectory) => {
+    handleCustomProtocolInvocation(argv[1], url => {
+      windowManager.focus();
+      windowManager.sendMessage({
+        kind: MainMessageKind.CUSTOM_PROTOCOL_INVOCATION,
+        data: { url },
+      });
     });
   });
 
-  if (isDev) {
-    setupWatcher();
-  }
+  // Handle custom protocol on Linux when Upstream is not running
+  handleCustomProtocolInvocation(process.argv[1], url => {
+    windowManager.sendMessage({
+      kind: MainMessageKind.CUSTOM_PROTOCOL_INVOCATION,
+      data: { url },
+    });
+  });
 
-  windowManager.open();
-});
+  // This method will be called when Electron has finished
+  // initialization and is ready to create browser windows.
+  // Some APIs can only be used after this event occurs.
+  app.on("ready", () => {
+    proxyProcessManager.run().then(({ status, signal, output }) => {
+      windowManager.sendMessage({
+        kind: MainMessageKind.PROXY_ERROR,
+        data: {
+          status,
+          signal,
+          output,
+        },
+      });
+    });
+
+    if (isDev) {
+      setupWatcher();
+    }
+
+    windowManager.open();
+  });
+} else {
+  app.quit();
+}
 
 // Quit when all windows are closed.
 app.on("window-all-closed", () => {
--- a/native/nativeCustomProtocolHandler.ts
+++ b/native/nativeCustomProtocolHandler.ts
@@ -1,0 +1,23 @@
+import lodash from "lodash";
+
+const THROTTLE_TIMEOUT = 1000; // 1 second
+
+export const handleCustomProtocolInvocation: (
+  url: string,
+  callback: (url: string) => void
+) => void = lodash.throttle(
+  (url, callback) => {
+    if (
+      typeof url !== "string" ||
+      url.length === 0 ||
+      Buffer.byteLength(url, "utf8") > 1024 ||
+      !url.toLowerCase().match(/^radicle:\/\//)
+    ) {
+      return;
+    }
+
+    callback(url);
+  },
+  THROTTLE_TIMEOUT,
+  { trailing: false }
+);
--- a/package.json
+++ b/package.json
@@ -40,6 +40,14 @@
         "to": "assets"
       }
     ],
+    "protocols": [
+      {
+        "name": "radicle",
+        "schemes": [
+          "radicle"
+        ]
+      }
+    ],
     "linux": {
       "target": [
         "Appimage"
--- a/ui/App.svelte
+++ b/ui/App.svelte
@@ -7,6 +7,7 @@
   import * as path from "./src/path.ts";
   import * as remote from "./src/remote.ts";
   import * as error from "./src/error.ts";
+  import * as customProtocolHandler from "./src/customProtocolHandler.ts";
   import { fetch, session as store, Status } from "./src/session.ts";
 
   import {
@@ -103,6 +104,8 @@
   $: sessionIsUnsealed =
     $store.status === remote.Status.Success &&
     $store.data.status === Status.UnsealedSession;
+
+  customProtocolHandler.register();
 </script>
 
 <style>
--- a/ui/src/customProtocolHandler.ts
+++ b/ui/src/customProtocolHandler.ts
@@ -1,0 +1,64 @@
+import * as ipc from "./ipc";
+import * as session from "./session";
+import * as modal from "./modal";
+import * as path from "./path";
+import * as error from "./error";
+import { inputStore } from "./search";
+
+export const register = (): void => {
+  ipc.listenCustomProtocolInvocation(async message => {
+    await session.waitUnsealed();
+    handleMessage(message);
+  });
+};
+
+const handleMessage = (message: ipc.CustomProtocolInvocation): void => {
+  const match = message.url.match(
+    /^radicle:\/\/(\w+)\/v([\d+])\/?(rad:git:[1-9A-HJ-NP-Za-km-z]{37})?/
+  );
+
+  if (!match) {
+    error.show({
+      code: error.Code.CustomProtocolParseError,
+      message: "Could not parse the provided URL",
+      details: { url: message.url },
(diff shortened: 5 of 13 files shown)


ui/src/modal.ts
```diff
@@ -1,9 +1,8 @@
 import { derived, get, writable } from "svelte/store";
 
 type OnHide = () => void;
-const doNothing = () => {
-  /**/
```
rudolfs, line 5: This was messing up my syntax highlighting.

package.json
```diff
@@ -38,6 +38,14 @@
         "to": "assets"
       }
     ],
+    "protocols": [
+      {
+        "name": "upstream",
+        "schemes": [
+          "upstream"
```
NunoAlexandre, line 45: From a product perspective, I would say the scheme would be something like `rad://<>`, which the `Radicle Upstream` client becomes responsible to handle.

ui/src/modal.ts
NunoAlexandre: Which syntax highlighting do you use? Might be good to use the same

ui/src/search.ts
```diff
@@ -6,6 +8,8 @@ import * as validation from "./validation";
 import type * as waitingRoom from "./waitingRoom";
 
 // STATE
+export const projectSearchInputStore: Writable<string> = writable("");
```
NunoAlexandre, line 11: Now we only support searching for projects but we may support other entities in the future, so I'd suggest renaming this to `inputStore`; shorter and its namespace already. Up to you :skateboard:

ui/src/customProtocolHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+      handleMessage(message);
+    } else {
+      messageQueue.push(message);
+    }
+  });
+
+  session.waitUnsealed().then(() => {
+    messageQueue.forEach(message => {
+      handleMessage(message);
+    });
```
NunoAlexandre, line 23:
```suggestion
    messageQueue.forEach(handleMessage);
```

NunoAlexandre (commented):
Looking good overall :raised_hands: great you could get it working on macOS!

Left some thoughts, but in sum:

- Are we use the custom protocol is `upstream` and not `rad`/`radicle`? Upstream is just a client that handles our custom protocol. Say we have a new client in the future, handling `upstream://` URLs will not make sense.

- I wonder if using a queue to digest incoming requests is a good idea, since it _could_ lead the UI to go nuts with a batch of requests coming in. Maybe that's not something we have to worry about though. I'd be happy to give it a go and see.

ui/src/modal.ts
rudolfs: I believe it's the Neovim built-in typescript syntax highlighting.

[`/usr/local/Cellar/neovim/0.4.4/share/nvim/runtime/syntax/typescript.vim`](https://github.com/neovim/neovim/blob/v0.4.4/runtime/syntax/typescript.vim)

Screenshot 2021-03-09 at 10 43 50

ui/src/search.ts
rudolfs: 86ff1adecb2a5cc6bb68d3813165ae9cd498b553

ui/src/customProtocolHandler.ts
rudolfs: 64034d48a8ad7b634c304238d3c398f987c40bd2

package.json
rudolfs: I didn't put a lot of thought into this, because this PR is just the proof-of-concept.

One thing to keep in mind is that radicle-link already uses an URN scheme which starts with `rad:`, which might be weird if we mixed it with the Upstream scheme. If we went with `rad`, it'd have to look something like this:

```
rad://v1/search/rad:git:hwd1yred6ptyar9opufxegtdibjbpe87obg1p9a4euyj1tuxpp7rugcoc9c
```

Note that  this: `rad://rad:git:hwd1yred6ptyar9opufxegtdibjbpe87obg1p9a4euyj1tuxpp7rugcoc9c` is not a valid URL and won't work, there has to be at least one extra namespace after the custom protocol, before you can use `:` again.

I think we should move this discussion to the community [post](https://radicle.community/t/opening-upstream-via-links-on-the-web/1856). Could you perhaps outline your thoughts on the scheme there?

native/main.ts
```diff
@@ -221,27 +234,42 @@ app.on("open-url", (event, url) => {
...
-    });
+if (app.requestSingleInstanceLock()) {
+  // Handle custom protocol on Linux
+  app.on("second-instance", (_event, commandLine, _workingDirectory) => {
+    windowManager.focus();
+    if (commandLine[1] && commandLine[1].toLowerCase().match(/^upstream/)) {
+      windowManager.sendMessage({
+        kind: MainMessageKind.CUSTOM_PROTOCOL_INVOCATION,
+        data: { url: commandLine[1] },
+      });
```
NunoAlexandre, line 245: interesting! didn't find this hack. did you get it working?

rudolfs:
> Are we use the custom protocol is upstream and not rad/radicle? Upstream is just a client that handles our custom protocol. Say we have a new client in the future, handling upstream:// URLs will not make sense.

Good point, I switched it to `radicle://` because `rad://` is already used by radicle-link.

> I wonder if using a queue to digest incoming requests is a good idea, since it could lead the UI to go nuts with a batch of requests coming in. Maybe that's not something we have to worry about though. I'd be happy to give it a go and see.

Also a good question. Having a queue was the easiest way to implement, I'd prefer to leave it as is and see if we run into any problems.

NunoAlexandre (commented):
An early review suggestion would be to dry the id of the custom protocol. Now we duplicate and hardcode `radicle` in several places.

ui/src/customProtocolHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as path from "./path";
+import * as error from "./error";
+import { inputStore } from "./search";
+
+export const register = (): void => {
+  ipc.listenCustomProtocolInvocation(message => {
+    session.waitUnsealed().then(() => {
+      handleMessage(message);
+    });
+  });
```
geigerzaehler, line 13:
```suggestion
  ipc.listenCustomProtocolInvocation(async message => {
    await session.waitUnsealed();
    handleMessage(message);
  });
```

ui/App.svelte
```diff
@@ -103,6 +104,8 @@
   $: sessionIsUnsealed =
     $store.status === remote.Status.Success &&
     $store.data.status === Status.UnsealedSession;
+
+  customProtocolHandler.register();
```
geigerzaehler, line 108: Is there a reason we are doing this explicitly instead of doing it as part of a module side effect?

native/index.ts
```diff
@@ -230,27 +242,60 @@ app.on("will-quit", () => {
...
+if (app.requestSingleInstanceLock()) {
+  // Handle custom protocol on Linux when Upstream is already running
+  app.on("second-instance", (_event, commandLine, _workingDirectory) => {
+    windowManager.focus();
+    if (commandLine[1] && commandLine[1].toLowerCase().match(/^radicle:\/\//)) {
+      windowManager.sendMessage({
+        kind: MainMessageKind.CUSTOM_PROTOCOL_INVOCATION,
+        data: { url: commandLine[1] },
+      });
+    }
```
geigerzaehler, line 264: This block is duplicated

ui/src/customProtocolHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+    error.show({
+      code: error.Code.CustomProtocolParseError,
+      message: "Could not extract project Radicle ID from the provided URL",
+      details: { url: message.url },
+    });
+  }
+};
+
+const parseUrl = (url: string): string | null => {
+  const match = url.match(/(rad:git:[\w]{59})/);
```
geigerzaehler, line 32: We should probably be more strict about the parsing, document, this and write some unit tests.

native/index.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1652/commits/fcc8543f16bc5a902b0b82ddc05824d4b98d2c9d

ui/src/customProtocolHandler.ts
rudolfs: Agreed. I'm waiting for the [protocol discussion](https://github.com/radicle-dev/radicle-decisions/pull/10) to be settled before implementing the proper parser logic and tests.

ui/App.svelte
rudolfs: Just a personal preference to have things explicit.

Are you suggesting to wrap the contents of `customProtocolHandler.ts` into a class and then instantiate a singleton of that class upon export?

Would this not lead to an unused import TS warning?

geigerzaehler: > Are you suggesting to wrap the contents of `customProtocolHandler.ts` into a class and then instantiate a singleton of that class upon export?

No, I think this would be overkill. I was suggesting executing the function body of `register` in the `customProtocolHandler` module itself and get rid of the `register` function. But there’s no clean advantage to this so let’s leave it like it is. I was just curious.

rudolfs:
> An early review suggestion would be to dry the id of the custom protocol. Now we duplicate and hardcode radicle in several places.

I got rid of some repetition here: https://github.com/radicle-dev/radicle-upstream/pull/1652/commits/fcc8543f16bc5a902b0b82ddc05824d4b98d2c9d.

Another thing we could do is import `package.json` into `config.ts` and use the `build.protocols.schemes` array, but I think that's overkill at this point, I rather keep it simple.

ui/src/customProtocolHandler.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1652/commits/097bed52b267da37f9df942e94969af1d4219a66

native/nativeCustomProtocolHandler.ts
```diff
@@ -0,0 +1,20 @@
+import lodash from "lodash";
+
+const THROTTLE_TIMEOUT = 1000; // 1 second
+
+export const handleCustomProtocolInvocation = lodash.throttle(
```
geigerzaehler, line 5: Not sure why eslint doesn’t complain, but this should have some type annotations.

native/index.ts
```diff
@@ -230,27 +243,61 @@ app.on("will-quit", () => {
...
+    });
+  });
 
-  windowManager.open();
-});
+  // Handle custom protocol on Linux when Upstream is not running
+  handleCustomProtocolInvocation(process.argv[1], () => {
+    windowManager.sendMessage({
+      kind: MainMessageKind.CUSTOM_PROTOCOL_INVOCATION,
+      data: { url: process.argv[1] },
```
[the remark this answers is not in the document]
rudolfs: 7b09d658480ebc3caa9522451efb1e1ad6f7d6ec

native/nativeCustomProtocolHandler.ts
rudolfs: d4b0d75aad8ebefb53837237b34ecbd1c434b098

geigerzaehler (dismissed):

PR #1653, opened 2021-03-05
Fix Upstream not working in release mode

rudolfs:
Fixes https://github.com/radicle-dev/radicle-upstream/pull/1463#pullrequestreview-603005282.

### DIFF
--- a/builder/entitlements.mac.plist
+++ b/builder/entitlements.mac.plist
@@ -4,5 +4,7 @@
   <dict>
     <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
     <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
   </dict>
 </plist>


xla (dismissed):
utACK

📵 🍪 🛡 👸

builder/entitlements.mac.plist
```diff
@@ -4,5 +4,7 @@
   <dict>
     <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
     <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
```
xla, line 7: What are the implications of this configuration key?

rudolfs: As far as I can tell, this prevents macOS from running binaries with unsigned linked libraries packaged within a DMG.
Setting this config key to `false` disables such checks.

[Here's some more info on this config key](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-library-validation).

For some reason this error only pops up in the `xla/1236-thank-you-next` branch, I assume that `openssl` wasn't linked in `proxy` before?

On my local dev machine as well as on CI `openssl` was installed via brew:
```
~ » brew info openssl
openssl@1.1: stable 1.1.1i (bottled) [keg-only]
Cryptography and SSL/TLS Toolkit
...
For compilers to find openssl@1.1 you may need to set:
  export LDFLAGS="-L/usr/local/opt/openssl@1.1/lib"
  export CPPFLAGS="-I/usr/local/opt/openssl@1.1/include"
```

And it is not signed by anyone.
```
~ » codesign -dv --verbose=4  /usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib
/usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib: code object is not signed at all
```

FintanH:
My biggest concern here is that we're disabling _all_ security checks. That just seems overkill. Since we're `brew`ing `openssl` we can at least validate that _we_ trust this binary and sign it on behalf of _our_ application. I think this could be part of the release process.

What do you think?

rudolfs:
> My biggest concern here is that we're disabling all security checks. That just seems overkill. Since we're brewing openssl we can at least validate that we trust this binary and sign it on behalf of our application. I think this could be part of the release process.

> What do you think?

I see where you're coming from, I'm not opposed to making things more secure.

To me having `com.apple.security.cs.disable-library-validation` set to `true` doesn't make Upstream any more secure, because I would not know how to check that a library which I'll have to manually sign isn't malicious. So leaving this security feature turned "on" just complicates our already lengthy CI/release setup without making the process any more secure.

I'll leave it up to you on solving this however you see fit.

rudolfs:
This was solved more elegantly by banning `openssl` from our dependency graph.

PR #1660, opened 2021-03-09
fix(proxy): announce heads and tags with correct path

geigerzaehler:
We properly announce changed refs. Instead of announcing the urn `rad:git:<project-id>:foo` if the `foo` branch changed we announce `rad:git:<project-id>:refs/heads/foo`. This is what `radicle-link` expects. In addition we also announce tags.

I’ve also changed the visibility of methods in the module that are unused externally.

I don’t have tests for this yet but will implement them in a separate PR. I’ll also backport this to `master`.


proxy/coco/src/peer/announcement.rs
```diff
@@ -71,7 +70,20 @@ pub async fn build(peer: &Peer<BoxedSigner>) -> Result<Updates, Error> {
                     for (head, hash) in &refs.heads {
```
FintanH, line 70: We could do `refs.iter_categories` so that we announce all the refs in the future

FintanH (changes requested):

proxy/coco/src/peer/announcement.rs
FintanH: Yes, I would say so. We haven't seen this happening because updates of `Person` or `Project` haven't been implemented in Upstream, but I would think that the network should see these so followers can get updates and maintainers can resolve updates :)

xla (approved):
🏺 🌔 👠 🍘

PR #1665, opened 2021-03-10
feat: sync periodically instead of on startup

xla:
While getting rid of the sync on startup behaviour, this change
introduces a periodic tick like done for announcements and stats to
issue a sync on an interval with connected peers.


proxy/coco/src/peer/run_state/config.rs
```diff
@@ -7,15 +7,12 @@ pub(super) const DEFAULT_ANNOUNCE_INTERVAL: Duration = std::time::Duration::from
 
 pub(super) const DEFAULT_STATS_INTERVAL: Duration = Duration::from_millis(100);
 
+pub(super) const DEFAULT_SYNC_INTERVAL: Duration = std::time::Duration::from_secs(1);
+
 /// Default number of peers a full sync is attempting with up on startup.
 /// TODO(xla): Revise number.
 pub(super) const DEFAULT_SYNC_MAX_PEERS: usize = 5;
```
FintanH, line 14: We can remove this right?

FintanH (approved):
Looks good

PR #1667, opened 2021-03-11
update app icon to new seedling design

brandonhaslegs:
Now that we've redesigned the seedling icon everywhere, it didn't make sense to keep the pixel art seedling on the app icon.

icon


rudolfs (approved):
Looks good.

Tested on macOS.

Screenshot 2021-03-11 at 13 34 28
Screenshot 2021-03-11 at 13 34 20
Screenshot 2021-03-11 at 13 33 56
Screenshot 2021-03-11 at 13 33 13
Screenshot 2021-03-11 at 13 33 08


We should probably also change the logo on the welcome screen, but that can be done in a follow up.

Screenshot 2021-03-11 at 13 32 00

brandonhaslegs:
> We should probably also change the logo on the welcome screen, but that can be done in a follow up.

Addressed here: #1669

PR #1672, opened 2021-03-15
refactor(test): make node data directories discoverable

geigerzaehler:
Before, each node had its own temporary directory `cypress/workspace/<uuid>` where `uuid` was different for each node. This made debugging the node state (in particular their monorepos) difficult.

With this change node data is located in the per-test temporary directory created by `withTempDir`.

### DIFF
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -9,17 +9,19 @@ context("p2p networking", () => {
       "reacts to network state changes",
       { defaultCommandTimeout: 8000 },
       () => {
-        nodeManager.withTwoOnboardedNodes(
-          { node1User: {}, node2User: {} },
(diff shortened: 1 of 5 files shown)


NunoAlexandre:
@geigerzaehler could you expand on how this change makes it easier for you to debug? thanks!

geigerzaehler:
Before this change each node would have their own temporary directory `cypress/workspace/<uuid>` and `withTempDir` would create another such directory. After this change only `withTempDir` creates a directory and the node data directories are created in that directory and are named `node-<id>`. This makes it easier to find the directory belonging to a node. In addition, it’s easier to disable the automatic removal of the temporary directory in code to inspect node data after a test run.

NunoAlexandre (approved):

NunoAlexandre (approved):

PR #1674, opened 2021-03-15
fix(tests): cypress tests wait until bundle is build

geigerzaehler:
Before it was possible to run cypress tests before the bundle was build which resulted in the app not working.


NunoAlexandre (approved):
Could you amend the typo in the commit message when merging? :star:

NunoAlexandre (approved):

PR #1675, opened 2021-03-15
refactor(ui): make ipc interface generic and type-safe

geigerzaehler:
We make the IPC interface between the electron main process and the renderer generic and type-safe. This reduces boilerplate and increases robustness.


NunoAlexandre (dismissed):
This is nice

NunoAlexandre (approved):

PR #1681, opened 2021-03-17
chore: put the peer in a loop

xla:
To not implode the entire networking machinery on an accept error, we
follow suit with the example of the link seed code and put the bind and
accept in a loop. Only when an error on bind is encountered it's
terminate.

### DIFF
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -7,16 +7,17 @@ use std::{
     net::SocketAddr,
     pin::Pin,
     task::{Context, Poll},
+    vec,
 };
 
 use futures::{future::FutureExt as _, stream::StreamExt as _};
 use tokio::{
-    sync::{broadcast, mpsc},
+    sync::{broadcast, mpsc, watch},
     task::{JoinError, JoinHandle},
 };
 
-use crate::{seed::Seed, state};
-use librad::{net, net::protocol, signer::BoxedSigner};
+use crate::state;
+use librad::{net, signer::BoxedSigner};
 
 mod announcement;
 pub use announcement::Announcement;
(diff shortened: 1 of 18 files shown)


proxy/coco/src/peer.rs
```diff
@@ -236,7 +230,7 @@ impl Future for Running {
         let err = match self.protocol.poll_unpin(cx) {
             Poll::Ready(val) => match val {
                 Err(e) => Some(Error::Join(e)),
-                Ok(res) => Some(Error::Protocol(res.err().expect("the impossible has happened, someone changed the `!` type in `Running::protocol`, if only there was some sort of function that could show this at compile time"))),
+                Ok(res) => Some(res.err().expect("the impossible has happened, someone changed the `!` type in `Running::protocol`, if only there was some sort of function that could show this at compile time")),
```
kim, line 233: > if only there was some sort of function that could show this at compile time

Well, it's not too hard to define [into_ok](https://doc.rust-lang.org/std/result/enum.Result.html#method.into_ok) with the type parameters swapped:

```rust
fn into_err<T, E>(r: Result<T, E>) -> E
where
    T: Into<!>
{
    match r {
        Ok(x) => x.into(),
        Err(e) => e,
    }
}
```

kim (approved):
lgtm

proxy/coco/src/peer.rs
```diff
@@ -191,19 +153,51 @@ where
...
+                                },
+                                Err(e) => {
+                                    return Err(Error::Io(e));
+                                },
+                            }
+
+                            if let Err(e) =
+                                net::protocol::accept(bound, disco.clone().discover()).await
+                            {
+                                log::error!("acept error: {}", e);
```
FintanH, line 190:
```suggestion
                                log::error!("accept error: {}", e);
```

proxy/coco/src/peer.rs
FintanH: Ooh that's nice

FintanH (approved):
Looks good, minor fix :+1:

PR #1682, opened 2021-03-18
feat(ui): Merge requests

juliendonck:
*Opening this to keep track of things to fix while working on it myself.*

##Todo
- [ ] navigating to commits from merge request detail page fails (except commits that exist on default branch)

### DIFF
--- a/cypress/support/nodeManager.ts
+++ b/cypress/support/nodeManager.ts
@@ -52,6 +52,16 @@ interface createCommitOptions {
   email?: string;
 }
 
+interface createMergeRequestOptions {
+  repositoryPath: string;
+  radHome: string;
+  tag: string;
+  subject: string;
+  passphrase: string;
+  name?: string;
+  email?: string;
+}
+
 const credentialsHelper = (passphrase: string) =>
   `'!f() { test "$1" = get && echo "password=${passphrase}"; }; f'`;
 
@@ -62,6 +72,33 @@ export PATH=$PWD/target/release:$PATH
 cd ${options.repositoryPath}
 git commit --allow-empty -m "${options.subject}"
 git -c credential.helper=${credentialsHelper(options.passphrase)} push rad`,
+    {
+      env: {
+        HOME: options.radHome,
+        RAD_HOME: options.radHome,
+        GIT_AUTHOR_NAME: options.name || "John McPipefail",
+        GIT_AUTHOR_EMAIL: options.email || "john@mcpipefail.com",
+        GIT_COMMITTER_NAME: options.name || "John McPipefail",
+        GIT_COMMITTER_EMAIL: options.email || "john@mcpipefail.com",
+      },
+    }
+  );
+};
+
+export const createMergeRequest = (
+  options: createMergeRequestOptions
+): void => {
+  createCommit(options);
+  cy.exec(
+    `set -euo pipefail
+export PATH=$PWD/target/release:$PATH
+cd ${options.repositoryPath}
+git tag -a --message "This is an awesome feature" merge-request/${
+      options.tag
+    } HEAD
+git -c credential.helper=${credentialsHelper(
+      options.passphrase
+    )} push --tag rad`,
     {
       env: {
         HOME: options.radHome,
--- a/ui/Screen/Project/Source.svelte
+++ b/ui/Screen/Project/Source.svelte
@@ -24,10 +24,13 @@
   import RevisionSelector from "../../DesignSystem/Component/SourceBrowser/RevisionSelector.svelte";
 
   import CheckoutButton from "./Source/CheckoutButton.svelte";
+  import MergeRequestButton from "./Source/MergeRequestButton.svelte";
 
   import Code from "./Source/Code.svelte";
   import Commit from "./Source/Commit.svelte";
   import Commits from "./Source/Commits.svelte";
+  import MergeRequests from "./Source/MergeRequests.svelte";
+  import MergeRequest from "./Source/MergeRequest.svelte";
 
   export let project: Project;
   export let selectedPeer: User;
@@ -37,7 +40,14 @@
     "/projects/:urn/source/code": Code,
     "/projects/:urn/source/commit/:hash": Commit,
     "/projects/:urn/source/commits": Commits,
+    "/projects/:urn/source/merge_requests": MergeRequests,
+    "/projects/:urn/source/merge_request": MergeRequest,
   };
+  let mergeRequestsTab = path.active(
+    path.projectSourceMergeRequests(project.urn),
+    get(location),
+    true
+  );
 
   const onCheckout = async (
     { detail: { checkoutPath } }: { detail: { checkoutPath: string } },
@@ -81,6 +91,7 @@
     ) {
       selectPath("");
     }
+    mergeRequestsTab = item.title === "Merge Requests";
   };
   const onSelectRevision = ({ detail: revision }: { detail: Branch | Tag }) => {
     selectRevision(revision);
@@ -101,22 +112,28 @@
   <ActionBar>
     <div slot="left">
       <div style="display: flex">
-        <div class="revision-selector-wrapper">
-          <RevisionSelector
-            loading={selectedRevision.request !== null}
-            on:select={onSelectRevision}
-            selected={selectedRevision.selected}
-            defaultBranch={project.metadata.defaultBranch}
-            {revisions} />
-        </div>
+        {#if !mergeRequestsTab}
+          <div class="revision-selector-wrapper">
+            <RevisionSelector
+              loading={selectedRevision.request !== null}
+              on:select={onSelectRevision}
+              selected={selectedRevision.selected}
+              defaultBranch={project.metadata.defaultBranch}
+              {revisions} />
+          </div>
+        {/if}
 
         <HorizontalMenu items={menuItems} on:select={onMenuSelect} />
       </div>
     </div>
     <div slot="right">
-      <CheckoutButton
-        fork={!isContributor}
-        on:checkout={ev => onCheckout(ev, project, selectedPeer)} />
+      {#if mergeRequestsTab}
+        <MergeRequestButton />
+      {:else}
+        <CheckoutButton
+          fork={!isContributor}
+          on:checkout={ev => onCheckout(ev, project, selectedPeer)} />
+      {/if}
     </div>
   </ActionBar>
 
--- a/ui/src/path.ts
+++ b/ui/src/path.ts
@@ -1,7 +1,8 @@
-import { parse, ParsedQs } from "qs";
+import { parse, ParsedQs, stringify } from "qs";
 import regexparam from "regexparam";
 
 import type { Urn } from "./urn";
+import type { MergeRequest } from "./source";
 
 export const blank = (): string => "/";
 export const settings = (): string => "/settings";
@@ -25,6 +26,15 @@ export const projectSourceCommit = (urn: Urn, hash: string): string =>
   `/projects/${urn}/source/commit/${hash}`;
 export const projectSourceCommits = (urn: Urn): string =>
   `/projects/${urn}/source/commits`;
+export const projectSourceMergeRequests = (urn: Urn): string =>
+  `/projects/${urn}/source/merge_requests`;
+export const projectSourceMergeRequest = (
+  urn: Urn,
+  mergeRequest: MergeRequest
+): string =>
+  `/projects/${urn}/source/merge_request?${stringify({
+    mergeRequest,
+  })}`;
 export const project = projectSourceFiles;
 
 export const parseQueryString = (querystring: string): ParsedQs => {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1233,7 +1233,7 @@ dependencies = [
 [[package]]
 name = "librad"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-link.git?rev=07b4bb364e8d268127b7b887cc94224676652afc#07b4bb364e8d268127b7b887cc94224676652afc"
+source = "git+https://github.com/radicle-dev/radicle-link.git?rev=b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8#b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8"
 dependencies = [
  "async-stream",
  "async-trait",
@@ -1925,7 +1925,7 @@ dependencies = [
 [[package]]
 name = "radicle-git-ext"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-link.git?rev=07b4bb364e8d268127b7b887cc94224676652afc#07b4bb364e8d268127b7b887cc94224676652afc"
+source = "git+https://github.com/radicle-dev/radicle-link.git?rev=b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8#b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8"
 dependencies = [
  "git2",
  "minicbor",
@@ -1940,7 +1940,7 @@ dependencies = [
 [[package]]
 name = "radicle-git-helpers"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-link.git?rev=07b4bb364e8d268127b7b887cc94224676652afc#07b4bb364e8d268127b7b887cc94224676652afc"
+source = "git+https://github.com/radicle-dev/radicle-link.git?rev=b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8#b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8"
 dependencies = [
  "anyhow",
  "git2",
@@ -1973,7 +1973,7 @@ dependencies = [
 [[package]]
 name = "radicle-macros"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-link.git?rev=07b4bb364e8d268127b7b887cc94224676652afc#07b4bb364e8d268127b7b887cc94224676652afc"
+source = "git+https://github.com/radicle-dev/radicle-link.git?rev=b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8#b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8"
 dependencies = [
  "quote",
  "radicle-git-ext",
@@ -1983,7 +1983,7 @@ dependencies = [
 [[package]]
 name = "radicle-std-ext"
 version = "0.1.0"
-source = "git+https://github.com/radicle-dev/radicle-link.git?rev=07b4bb364e8d268127b7b887cc94224676652afc#07b4bb364e8d268127b7b887cc94224676652afc"
+source = "git+https://github.com/radicle-dev/radicle-link.git?rev=b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8#b6f3ffcdfd7f4ade57f6c1aab2032faa96f808c8"
 
 [[package]]
 name = "radicle-surf"
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -241,4 +241,178 @@ context("p2p networking", () => {
       );
     });
   });
+
+  it("replicates a merge request from contributor to maintainer", () => {
+    const maintainer = {
+      handle: "rudolfs",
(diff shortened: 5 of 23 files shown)


juliendonck:
### Usability review
by @brandonhaslegs & @juliendonck 

- [ ] patch amount should be "open patches" not all patches 
Screen Shot 2021-04-12 at 13 20 53

- [ ] When the maintainer merges my patch, I as a contributor want to be able get that latest state in my project (as a contributor when my patch is merged into the original project, I want to be able to pull those changes in and have that reflected in my project view)
- [ ] nothing updates automatically (new patch) I always have to refresh the app
- [ ] contributors remote disconnected and is in "searching for" again. Had to unfollow and re-add the remote to get the latest patch
- [ ] as a contributor when I checkout the project, the maintainer is no longer in the peer selector
- [ ] back button on patch detail page goes to "open patches" tab, not the tab it was last on
- [ ] closed patches have all commits, not just new commits
Screen Shot 2021-04-12 at 13 34 05

juliendonck:
@geigerzaehler @MeBrei check comment

juliendonck (commented):
### Usability review
by @brandonhaslegs & @juliendonck 

- [ ] When the maintainer merges my patch, I as a contributor want to be able get that latest state in my project (as a contributor when my patch is merged into the original project, I want to be able to pull those changes in and have that reflected in my project view)
- [ ] back button on patch detail page goes to "open patches" tab, not the tab it was last on
- [ ] merge button only gives half of the commands that you need to do. Doesn't tell you you need to fetch it as well.
- [ ] patch amount should be "open patches" not all patches 
Screen Shot 2021-04-12 at 13 20 53

**Will not fix**
- [ ] nothing updates automatically (new patch) I always have to refresh the app
- [ ] closed patches have all commits, not just new commits
Screen Shot 2021-04-12 at 13 34 05

**Replication bugs**
- [ ] contributors remote disconnected and is in "searching for" again. Had to unfollow and re-add the remote to get the latest patch
- [ ] as a contributor when I checkout the project, the maintainer is no longer in the peer selector

juliendonck (commented):
`CheckoutMergeRequestButton` `AcceptMergeRequestButton` `MergeRequestButton` are all basically the same. They could be one component

geigerzaehler:
>  contributors remote disconnected and is in "searching for" again. Had to unfollow and re-add the remote to get the latest patch

This is a bug unrelated to patches. 

>  as a contributor when I checkout the project, the maintainer is no longer in the peer selector

Also seems to be an unrelated bug.

> closed patches have all commits, not just new commits

Yes, this is a known issue. With the current implementation there’s no way of knowing which commits belong to a patch. So we can either show all commits or no commits and a message that the patch has been merged.

> nothing updates automatically (new patch) I always have to refresh the app

This is also a known issue. It would be good to know if this is a must have for a release.

juliendonck:
> > contributors remote disconnected and is in "searching for" again. Had to unfollow and re-add the remote to get the latest patch
> 
> This is a bug unrelated to patches.

Apparently fixed in master.

> 
> > as a contributor when I checkout the project, the maintainer is no longer in the peer selector
> 
> Also seems to be an unrelated bug.

Apparently fixed in master.

> 
> > closed patches have all commits, not just new commits
> 
> Yes, this is a known issue. With the current implementation there’s no way of knowing which commits belong to a patch. So we can either show all commits or no commits and a message that the patch has been merged.

OK

> 
> > nothing updates automatically (new patch) I always have to refresh the app
> 
> This is also a known issue. It would be good to know if this is a must have for a release.

OK

geigerzaehler:
> When the maintainer merges my patch, I as a contributor want to be able get that latest state in my project (as a contributor when my patch is merged into the original project, I want to be able to pull those changes in and have that reflected in my project view)

Let’s go through this together and come up with a solution

> back button on patch detail page goes to "open patches" tab, not the tab it was last on

Good catch. Should be an easy fix

> merge button only gives half of the commands that you need to do. Doesn't tell you you need to fetch it as well.

To clarify this we could either include the `git fetch` command or reference it. But I’m not a hundred percent convinced that we should have the merge button anyways. It’s up to you to make call on this, @juliendonck.

> patch amount should be "open patches" not all patches

Should be an easy fix

geigerzaehler:
> `CheckoutMergeRequestButton` `AcceptMergeRequestButton` `MergeRequestButton` are all basically the same. They could be one component

I had a look at this and the similarities are quite hard to extract. The duplication is mainly in the styles and in the toggling logic. To improve the latter we would need to replace the `Overlay` component. Since it is widely used this would create a lot of churn for minimal gain. The replacement component would also need to take care of the styles and it would require some effort to figure out how this should behave in the design system. Given all this I think we can justify the duplication.

If you feel that we will have a need for a styled `Popover` component in the future feel free to open an issue. While working on this I gained some insight that could help in implementing it.

MeBrei:
While testing out the current state of this PR, I found the following small ui issues:

- On the patch detail view, the horizontal menu bar switches to the file-view-version of the project, ie it shows the branch selector and the checkout project button, but should not show the selector and should show the checkout tag button. 
Screenshot 2021-04-20 at 12 05 10

- On the commit detail view for the patch, it shows `undefined` where normally the `branch` of the commit would be shown
Screenshot 2021-04-20 at 11 59 30

- When a patch is "closed" the `PatchCard` and detail view say "Closed by <patch-creator-identity>" which might be confusing, as the patch is closed by the maintainer, not the person who created the patch.
Screenshot 2021-04-20 at 12 13 41

geigerzaehler:
> On the commit detail view for the patch, it shows `undefined` where normally the `branch` of the commit would be shown

We already have an issue for this: https://github.com/radicle-dev/radicle-upstream/issues/1793

MeBrei:
> * When a patch is "closed" the `PatchCard` and detail view say "Closed by " which might be confusing, as the patch is closed by the maintainer, not the person who created the patch.
> 
> Screenshot 2021-04-20 at 12 13 41

@brandonhaslegs @juliendonck What are your thoughts on this? 
I can think of the following two options:
1) rearranging the Opened/Closed indicator and the author, so they are not read in a sentence.
2) showing `Closed by <maintainer-name>`. I'm not sure if we can/if it's feasible to find the actual person who merged the patch, for now we can assume it is the one maintainer and add it in the ui, but this might change in the future. Another problem with this is that it would be harder to find your own closed patches as they wouldn't show your identity anymore

brandonhaslegs:
@MeBrei Can we maybe just keep it the same as the open Patch design. So keep the text "Opened by 😀 **juliendonck**" and just the only change is the icon being red? Especially if the merger is not important for now, the opener of the Patch does seem to be important to show still.

ui/Screen/Project/Source.svelte
```diff
@@ -102,22 +115,28 @@
...
       <div style="display: flex">
-        <div class="revision-selector-wrapper">
-          <RevisionSelector
-            loading={selectedRevision.request !== null}
-            on:select={onSelectRevision}
-            selected={selectedRevision.selected}
-            defaultBranch={project.metadata.defaultBranch}
-            {revisions} />
-        </div>
+        {#if !patchesTabSelected}
```
rudolfs, line 118: This might be a UI/UX follow up, but it feels a bit weird to me that the tab menu bar shifts to the left when the branch selector is gone. 🤷
 
What do you think @juliendonck and @brandonhaslegs?

https://user-images.githubusercontent.com/158411/117034660-d382bd80-ad03-11eb-8557-fb7132c90d1a.mov

ui/Screen/Project/Source/CheckoutPatchButton.svelte
```diff
@@ -0,0 +1,81 @@
...
+  export let myPeerId: string;
+
+  let expanded = false;
+  let copyable: Copyable;
+  const hide = () => (expanded = false);
+  const toggleDropdown = () => {
+    expanded = !expanded;
+  };
+  const copy = () => {
+    copyable.copy();
```
rudolfs, line 16: TIL that you can call functions of components via their exported props. 🤯

ui/src/path.ts
```diff
@@ -22,6 +27,58 @@ export const projectSourceCommit = (urn: Urn, hash: string): string =>
...
+  urn: Urn,
+  filter: PatchFilter
+): string => `${projectSourcePatches(urn)}?${qs.stringify({ filter })}`;
+
+export const projectSourcePatches = (urn: Urn): string =>
+  `/projects/${urn}/source/patches`;
+
+// Holds the query paramters for the patches route if that route is
+// active. Otherwise holds `undefined`
+export const projectSourcePatchesQuery = (): svelteStore.Readable<
```
rudolfs, line 57: The way filtering is implemented here and in the respective component is quite elegant! 👍

ui/src/project/patch.ts
```diff
@@ -0,0 +1,80 @@
...
+import * as proxy from "../proxy";
+
+export type { Patch } from "../proxy";
+
+export interface PatchDetails {
+  patch: proxy.Patch;
+  commits: source.GroupedCommitsHistory;
+}
+
+export const tagPrefix = "radicle-patch/";
```
rudolfs, line 13: In which cases do we write constants camelCase and in which ALL_CAPS, like here 👇 ? https://github.com/radicle-dev/radicle-upstream/blob/a86273751c4a1299617c0be7a59e5a48df28e84c/ui/src/config.ts#L4

rudolfs (commented):
Really excited about this feature!! 👏 🚀 

Since we're adding a bunch of new files, this would be a good opportunity to rewrite all the imports to absolute ones. I'll leave it up to you though, definitely not a blocker.

ui/Screen/Project/Source.svelte
juliendonck: I agree, but the only way around this would be to move it to the header. Which is something we discussed but didn't want to do.

cypress/integration/project/patches.spec.ts
```diff
@@ -0,0 +1,362 @@
...
+            commands.pick("patch-card-my-patch").click();
+            commands
+              .pickWithContent(["patch-page", "patch-title"], patchId)
+              .should("exist");
+          });
+      });
+    });
+  });
+
+  it("replicates a patch from contributor to maintainer", () => {
```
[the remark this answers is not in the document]
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/82b4a6e091ea888026ff690128beda1a22900d3c

cypress/integration/project/patches.spec.ts
```diff
@@ -0,0 +1,362 @@
...
+            cy.log("verify that only the single patch commit is displayed");
+            commands
+              .pick("patch-page", "history", "commit-group")
+              .should("have.length", 1);
+            commands
+              .pick("patch-page", "history", "commit-group", "commit")
+              .should("have.length", 1);
+            commands
+              .pick("patch-page", "history", "commit-group", "commit")
+              .contains(commitMessage);
```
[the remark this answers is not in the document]
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/82b4a6e091ea888026ff690128beda1a22900d3c

cypress/support/commands.ts
```diff
@@ -108,6 +108,49 @@ export const createEmptyProject = (
     }),
   });
 
+export const followProject = (
+  urn: string,
+  localhost: number = 17246
```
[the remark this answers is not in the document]
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/82b4a6e091ea888026ff690128beda1a22900d3c

cypress/support/nodeManager.ts
```diff
@@ -20,7 +20,7 @@ const startAndOnboardNode = (
   });
 };
 
-const withNodeManager = (callback: () => void): void => {
+export const withNodeManager = (callback: () => void): void => {
```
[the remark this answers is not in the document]
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/82b4a6e091ea888026ff690128beda1a22900d3c

proxy/coco/src/patch.rs
```diff
@@ -0,0 +1,169 @@
...
+
+const TAG_PREFIX: &str = "radicle-patch/";
+
+/// A patch is a change set that a user wants the maintainer to merge into a projects default
+/// branch.
+///
+/// A patch is represented by an annotated tag, prefixed with `radicle-patch/`.
+#[derive(Debug, Clone)]
+pub struct Patch {
+    /// ID of a patch. This is the portion of the tag name followin the `radicle-patch/` prefix.
```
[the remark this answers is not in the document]
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/82b4a6e091ea888026ff690128beda1a22900d3c

proxy/coco/src/patch.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub id: String,
+    /// Peer that the patch originated from
+    pub peer: project::Peer<peer::Status<Person>>,
+    /// Message attached to the patch. This is the message of the annotated tag.
+    pub message: Option<String>,
+    /// Head commit that this
```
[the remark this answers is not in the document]
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/82b4a6e091ea888026ff690128beda1a22900d3c

proxy/coco/src/patch.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub commit: git2::Oid,
+    /// The merge base of [`Patch::commit`] and the head commit of the first maintainers default
+    /// branch.
```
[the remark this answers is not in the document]
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/82b4a6e091ea888026ff690128beda1a22900d3c

proxy/coco/src/patch.rs
```diff
@@ -0,0 +1,169 @@
...
+                },
+            }
+        }
+    }
+    Ok(patches)
+}
+
+#[derive(thiserror::Error, Debug)]
+enum PatchExtractError {
+    #[error("cannot peel reference to tag")]
```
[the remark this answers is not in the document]
geigerzaehler: “Peeling” is a git specific term. It’s explained [here](https://stackoverflow.com/a/26492602/12221117)

ui/src/project/patch.ts
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/82b4a6e091ea888026ff690128beda1a22900d3c

cypress/support/nodeManager.ts
```diff
@@ -6,7 +6,7 @@ import {
 
 const nodeManagerPlugin = createNodeManagerPlugin();
 
-const startAndOnboardNode = (
+export const startAndOnboardNode = (
```
rudolfs, line 9: Do we still need to export this?

geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1682/commits/a73c2ff28bfaf69978ad91eeb60a6f440bd464bc

rudolfs (approved):
The UI ts/svelte side solid. 👍 
Would be great if somebody could also have a look at the rust side of things.

Additionally we could use a test-case that includes a `/` in the tag name.

From the UX perspective I think we could improve on:
1. Adding additional information in the "Checkout" modal that you first need to fork the project to have a local copy, otherwise the given command doesn't make sense.

2. The command in the "Checkout" modal only fetches the change, but it doesn't actually check out the respective tag, so a user is not able to see the commit that is in the patch. It would make sense to add more instructions on how to do that, something like:
```
To inspect the changeset do:
git checkout tags/radicle-patch/rudolfs/rudolfs/add-readme
```

proxy/api/src/patch.rs
```diff
@@ -0,0 +1,144 @@
...
+        crate::browser::using(peer, branch, {
+            let patches = &mut patches;
+            move |browser| {
+                let tags = browser.list_tags(ref_scope)?;
+                for tag in tags {
+                    match tag {
+                        radicle_source::surf::git::Tag::Light { .. } => {
+                            continue;
+                        },
+                        radicle_source::surf::git::Tag::Annotated {
```
FintanH, line 111: Using the suggestion of splitting `Patch` data, you could define a smart constructor `fn patch(tag: surf::Tag) -> Option<Patch>` and use a `filter_map` here.

proxy/api/src/patch.rs
```diff
@@ -0,0 +1,144 @@
...
+
+const TAG_PREFIX: &str = "radicle-patch/";
+
+/// A patch is a change set that a user wants the maintainer to merge into a projects default
+/// branch.
+///
+/// A patch is represented by an annotated tag, prefixed with `radicle-patch/`.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Patch {
```
FintanH, line 20: Take this suggestion with a grain of salt :) There seems to be a mixture of data here. There's the `Patch` and where it came from. On top of that, `Identity` contains the `PeerId` so you could abstract over where it came from using an enum. So maybe something like this:

```rust
pub struct Patch {
  pub id: String,
  pub message: Option<String>,
  pub commit: Oid,
  pub merge_base: Option<Oid>,
}

pub enum Origin {
  Peer(PeerId),
  Identity(identity::Identity),
}

pub struct UserPatch {
  pub patch: Patch,
  pub origin: Origin,
}
```

FintanH (commented):
Just a few coding org suggestions that might be of interest :) Functionality makes sense to moi

proxy/api/src/patch.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// ID of a patch. This is the portion of the tag name following the `radicle-patch/` prefix.
+    pub id: String,
```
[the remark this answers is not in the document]
geigerzaehler: Neither the API code nor `radicle-source` use `RefLike` so I’m a bit hesitant to include it.

proxy/api/src/patch.rs
geigerzaehler: Alternatively, how do you feel about replacing `peer_id` and `identity` with
```rust
peer: project::Peer<peer::Status<Person>>
```

proxy/api/src/patch.rs
geigerzaehler: I’ve tried it but because of errors and access to `browser` this is call-site code I came up with to replace the for-loop.
```rust
Ok(patches.extend(
    browser
        .list_tags(ref_scope)?
        .into_iter()
        .filter_map(|tag| tag_to_patch(browser, tag).transpose())
        .collect::<Result<Vec<Patch>, _>>()?,
))
```
And `tag_to_patch` would be pretty much the body of the for-loop. This looks much harder to understand and is probably less performant because it allocates. So iteration seems fine to me.

proxy/api/src/patch.rs
FintanH: I thought about this but wasn't sure if it was the correct semantics. Not all variants of a `Status` have a `PeerId`, however, maybe it doesn't make sense for a `Patch` to exist if it's a non-replicated peer?

proxy/api/src/patch.rs
FintanH: Whatever you found best works :)

proxy/api/src/patch.rs
geigerzaehler: > Not all variants of a `Status` have a `PeerId`,

But the outer `project::Peer` always has a peer ID so this works. I’ve updated the PR.

FintanH: Nice, looks good!

geigerzaehler:
A heads-up on the API code. That code is localized to one `Patch` struct, one `patch::list` function and one `GET patches` endpoint. It only uses publicly available APIs from `radicle-surf` and `radicle-daemon`. With regards to using the upcoming daemon API (https://github.com/radicle-dev/radicle-link/pull/673) that will move the `api` crate code into link we have to options.

1. We’ll move the code that assembles the patches in `patch::list` to the Upstream. This should be easy enough given that all internal APIs are not present on `radicle-surf`.
2. If Upstream maintains a shim API process we should be able to reuse the `patch::list` code and just have it talk to the daemon library.

rudolfs (dismissed):
Looks good from my side.
We should test it once before we merge though.

alexjg (dismissed):

rudolfs (dismissed):

PR #1684, opened 2021-03-18
refactor(proxy): avoid spawning tasks when running a peer

geigerzaehler:
Instead of spawning tasks when `peer.into_running()` is called we construct a future. This simplifies the code while retaining the behavior that when the returned object is dropped we stop polling the protocol and subroutines.

We also replace the custom `futures::select` implementation.

### DIFF
--- a/proxy/coco/src/peer.rs
+++ b/proxy/coco/src/peer.rs
@@ -1,19 +1,12 @@
 //! Machinery to advance the underlying network protocol and manage auxiliary tasks ensuring
 //! prorper state updates.
 
-use std::{
-    future::Future,
-    io,
-    net::SocketAddr,
-    pin::Pin,
-    task::{Context, Poll},
-    vec,
-};
+use std::{io, net::SocketAddr, vec};
 
-use futures::{future::FutureExt as _, stream::StreamExt as _};
+use futures::prelude::*;
 use tokio::{
     sync::{broadcast, mpsc, watch},
-    task::{JoinError, JoinHandle},
+    task::JoinError,
 };
 
 use crate::state;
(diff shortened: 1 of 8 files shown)


FintanH (approved):
Looks like Rust to me :+1:

proxy/coco/src/peer.rs
```diff
@@ -1,19 +1,12 @@
...
-    io,
-    net::SocketAddr,
-    pin::Pin,
-    task::{Context, Poll},
-    vec,
-};
+use std::{io, net::SocketAddr, vec};
 
-use futures::{future::FutureExt as _, stream::StreamExt as _};
+use futures::prelude::*;
```
xla, line 6: Is the wildcard import for the entire prelude necessary?

proxy/coco/src/peer.rs
```diff
@@ -152,113 +145,52 @@ where
...
-}
-
-fn into_err<T, E>(r: Result<T, E>) -> E
-where
-    T: Into<!>,
-{
-    match r {
-        Ok(x) => x.into(),
-        Err(e) => e,
+        Ok(())
```
xla, line 194: Can this ever be hit given that the output of the protocol future is `Result<!, Error>`?

xla (commented):
This is looks like a worthwhile simplification.

proxy/coco/src/peer.rs
geigerzaehler: Yes. Like before, we return when the subroutine future is ready.

proxy/coco/src/peer.rs
geigerzaehler: Not really. If you want I can import the traits explicitly.

xla: It seems preferrable to be explicit esp. with pulling in the traits.

xla (approved):
🔱 👺 🔉 🔟

geigerzaehler:
Merged in 2d06db51e4181ef7eb501677bedfcbd3f5884d0c

PR #1685, opened 2021-03-18
Igor/ethereum attestation

CodeSandwich:
The first step of adding Ethereum attestations to the backend. Allows setting a claim when creating a new identity and returns it when queried. There's no update feature yet, that's to be built on top of this PR when it's ok.

The RFC has been reviewed before, so it should be fine as it is, it's only updated for the updated `Claims` contract. Putting it here feels right, 98% of its implementation resides here, but I'm fine with moving it somewhere else.

### DIFF
--- a/docs/ethereum_attestation.md
+++ b/docs/ethereum_attestation.md
@@ -1,0 +1,137 @@
+# RFC: Ethereum attestation
+
+* Author: @CodeSandwich
+* Date: 2021-03-05
+* Status: draft
+* Community discussion: n/a
+
+## Motivation
+
+The attestation between Radicle Link and Ethereum is a valuable building block for a user identity.
+It brings the Radicle Link reputation coming from projects and contributions to the
+Ethereum world of DAOs and donations, where it's important to know who's behind an address.
+On the other hand, it lends Ethereum account reputation with its assets and undeniable history
+to Radicle Link to build user's trust in the identity.
+
+## Overview
+
+This RFC is built on top of the [identities spec][identities].
+It introduces support for Ethereum address claims on Radicle Link
+and a smart contract on Ethereum to make Radicle Link identity claims on Ethereum.
+
+## Radicle Link identity JSON extension
+
+The identity doc `payload` structure supports a new key: `https://radicle.xyz/ethereum/claim/v1`.
+It can be used only in person identities.
+Under this key, an Ethereum address claim is stored in this format:
+
+- `address` - the claimed Ethereum address, encoded according to [EIP-55][eip-55],
+e.g. using [ethers.js][ethers-addr]
+- `expiration` - the claim expiration timestamp, encoded as a [JavaScript Date][date].
+The claim is valid only before that timestamp.
+
+Example:
+```json
+{
+    "payload": {
+        "https://radicle.xyz/ethereum/claim/v1": {
+            "address": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
+            "expiration": "2021-03-19T23:15:30.000Z"
+        }
+        ...
+    }
+    ...
+}
+```
+
+## Ethereum smart contract
+
+A new Ethereum smart contract is deployed to the network,
+which lets users claim their Radicle Link identities:
+
+```solidity
+contract Claims {
+    event Claimed(address indexed addr);
+    function claim(uint256 format, bytes calldata payload);
+}
+```
+
+To claim an identity, call `claim` using your Ethereum account
+and pass your Radicle Link identity root.
+It will emit an event `Claimed`, which later can be queried to discover your attestation.
+The claims have no expiration date and don't need to be renewed.
+
+Every new claim invalidates previous ones made with the same account.
+To revoke a claim without creating a new one, claim root hash `0`,
+which is guaranteed to not match any existing identity.
+
+Currently supported `format` values:
+- `1` - `payload` is exactly 20 bytes and contains an SHA-1 Radicle Identity root hash
+- `2` - `payload` is exactly 32 bytes and contains an SHA-256 Radicle Identity root hash
+
+We need to deploy an official instance of the `Claims` smart contract and
+it must be used by all the users.
+If anybody makes a claim using a different instance, it won't be recognized by others.
+
+## Creation of an attestation
+
+You need to perform 2 actions in any order:
+- Add or update an `https://radicle.xyz/ethereum/claim/v1` entry in your identity doc.
+The entry's `address` must be your Ethereum address.
+It's highly recommended to set a short expiration date as Ethereum claims don't expire.
+- Call `claim` in the `Claims` smart contract. The `root` must point to your link identity.
+
+## Discovery
+
+### From an Ethereum address
+
+When you know an Ethereum address, you can find the claimed link identity using an Ethereum client.
+The example calls are based on the standard [client JSON RPC API][rpc] and should be exposed
+by your favourite Ethereum client library.
+It's important that the client must be trusted not to hide the events.
+
+- Use [getLogs][rpc-logs] to get the newest `Claimed` event filtered for the given ethereum address
+- Get the event's `transactionHash` field and use it to fetch the transaction which emitted it with
+[getTransactionByHash][rpc-tx]
+- Validate that the transaction signature matches its data and the ethereum address.
+For reference the signature payload content is listed [here][rpc-sign].
+- Read the Radicle Link identity root hash from the transaction data
+- Verify that the Radicle Link identity doc claims back the Ethereum address,
+see [Discovery from a Radicle Link identity](#from-a-radicle-link-identity)
+
+### From a Radicle Link identity
+
+When you know a Radicle Link identity, you can find the claimed Ethereum address.
+Obtain the tip of its identity chain and read the Ethereum address from the identity doc
+`address` field in section `https://radicle.xyz/ethereum/claim/v1`, unless it's expired.
+You need to verify that the given Ethereum address claims back the Radicle Link identity root,
+see [Discovery from an Ethereum address](#from-an-ethereum-address).
+
+## Revocation of an attestation
+
+When your attestation for whatever reason is no longer valid,
+you should revoke it as soon as possible.
+Only one claim needs to be revoked to break the attestation,
+but to improve security you should revoke both sides if you can.
+
+To revoke a claim on Radicle Link, update the identity doc and publish it in a new revision.
+You can change the claimed Ethereum address or remove
+the `https://radicle.xyz/ethereum/claim/v1` section altogether.
+The other Radicle Link nodes will not notice this update until they fetch the new revision.
+
+To revoke a claim on Ethereum, call the `claim` function in `Claims` contract
+to can claim a different Radicle Link identity root
+or the `0` root to revoke any claim you may have.
+The other users will notice this update almost immediately if they
+are subscribed to the `Claimed` events for your Ethereum address in their Ethereum client.
+
+---
+
+[identities]: ../spec/sections/002-identities/index.md
+[eip-55]: https://eips.ethereum.org/EIPS/eip-55
+[date]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON
+[ethers-addr]: https://docs.ethers.io/v5/api/utils/address/
+[rpc]: https://eth.wiki/json-rpc/API
+[rpc-logs]: https://eth.wiki/json-rpc/API#eth_getlogs
+[rpc-tx]: https://eth.wiki/json-rpc/API#eth_gettransactionbyhash
+[rpc-sign]: https://eth.wiki/json-rpc/API#eth_signtransaction
--- a/proxy/api/src/ethereum/address.rs
+++ b/proxy/api/src/ethereum/address.rs
@@ -1,0 +1,46 @@
+use serde::{Deserialize, Serialize};
+use std::{
+    convert::{TryFrom, TryInto},
+    str::FromStr,
+};
+
+/// An [EIP-55](https://eips.ethereum.org/EIPS/eip-55) Ethereum address.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[serde(try_from = "String", into = "String")]
+pub struct Address(String);
+
+#[derive(Clone, Debug, thiserror::Error)]
+#[error("invalid EIP-55 address")]
+pub struct InvalidEIP55AddressError;
+
+impl TryFrom<String> for Address {
+    type Error = InvalidEIP55AddressError;
+
+    fn try_from(address: String) -> Result<Self, Self::Error> {
+        if eip55::validate_address(&address) {
+            Ok(Self(address))
+        } else {
+            Err(InvalidEIP55AddressError)
+        }
+    }
+}
+
+impl FromStr for Address {
+    type Err = InvalidEIP55AddressError;
+
+    fn from_str(address: &str) -> Result<Self, Self::Err> {
+        address.to_string().try_into()
+    }
+}
+
+impl From<Address> for String {
+    fn from(address: Address) -> Self {
+        address.0
+    }
+}
+
+impl AsRef<str> for Address {
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
--- a/proxy/api/src/http/identity.rs
+++ b/proxy/api/src/http/identity.rs
@@ -1,6 +1,5 @@
 //! Manage the state and stateful interactions with the underlying peer API of librad.
 
-use serde::{Deserialize, Serialize};
 use warp::{filters::BoxedFilter, path, Filter, Rejection, Reply};
 
 use crate::{context, http};
@@ -54,7 +53,7 @@ mod handler {
     /// Create a new [`identity::Identity`].
     pub async fn create(
         ctx: context::Unsealed,
-        input: super::CreateInput,
+        metadata: identity::Metadata,
     ) -> Result<impl Reply, Rejection> {
         if let Some(session) = session::get_current(&ctx.store)? {
             return Err(Rejection::from(error::Error::SessionInUse(
@@ -62,7 +61,7 @@ mod handler {
             )));
         }
 
-        let id = identity::create(&ctx.peer, &input.handle).await?;
+        let id = identity::create(&ctx.peer, metadata).await?;
 
         session::initialize(&ctx.store, id.clone(), &ctx.default_seeds)?;
 
@@ -82,21 +81,12 @@ mod handler {
     }
 }
 
-// TODO(xla): Implement Deserialize on identity::Metadata and drop this type entirely, this will
-// help to avoid duplicate efforts for documentation.
-/// Bundled input data for identity creation.
-#[derive(Debug, Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
-pub struct CreateInput {
-    /// Handle the user wants to go by.
-    handle: String,
-}
-
 #[allow(clippy::non_ascii_literal, clippy::unwrap_used)]
 #[cfg(test)]
 mod test {
     use pretty_assertions::assert_eq;
     use serde_json::{json, Value};
+    use std::convert::TryInto;
     use warp::{http::StatusCode, test::request};
 
     use radicle_avatar as avatar;
@@ -112,8 +102,15 @@ mod test {
         let res = request()
             .method("POST")
             .path("/")
-            .json(&super::CreateInput {
+            .json(&identity::Metadata {
                 handle: "cloudhead".into(),
+                ethereum: Some(identity::Ethereum {
+                    address: "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B"
+                        .to_string()
+                        .try_into()
+                        .expect("Invalid address"),
+                    expiration: "2021-03-19T23:15:30.001Z".parse().expect("Invalid date"),
+                }),
             })
             .reply(&api)
             .await;
@@ -152,6 +149,10 @@ mod test {
                     "urn": urn,
                     "metadata": {
                         "handle": "cloudhead",
+                        "ethereum": {
+                            "address": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
+                            "expiration": "2021-03-19T23:15:30.001Z",
+                        }
                     },
                     "shareableEntityIdentifier": &shareable_entity_identifier,
                 })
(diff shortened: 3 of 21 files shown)


proxy/coco/src/ethereum_claim_ext.rs
```diff
@@ -0,0 +1,32 @@
...
+use url::Url;
+
+/// The user identity doc extension for Ethereum addresses claims.
+/// Meaningful only if confirmed on Ethereum, see [the RFC](docs/ethereum_attestation.md).
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct V1 {
+    /// The Ethereum address claimed by the user, encoded according to
+    /// [EIP-55](https://eips.ethereum.org/EIPS/eip-55).
+    pub address: String,
```
FintanH, line 16: I'd suggest a newtype and parser to make sure the construction is always compliant with EIP-55

proxy/coco/src/ethereum_claim_ext.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// The timestamp before which the address claim is valid, encoded according to
+    /// [JavaScript Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON)
+    pub expiration: String,
```
FintanH, line 19: Why a `String` rather than a time type?

proxy/coco/src/state.rs
```diff
@@ -90,23 +90,27 @@ where
 ///   * Fails to verify `User`.
 ///   * Fails to set the default `rad/self` for this `PeerApi`.
 #[allow(clippy::single_match_else)]
-pub async fn init_owner(peer: &Peer<BoxedSigner>, name: String) -> Result<LocalIdentity, Error> {
+pub async fn init_owner(
+    peer: &Peer<BoxedSigner>,
+    name: String,
+    ethereum: Option<EthereumClaimExtV1>,
```
FintanH, line 96: We should change this to just take `PersonPayload` and remove `name` and `ethereum`. Or even take `impl Into<PersonPayload>` so that the parameter can also easily be an `api` type.

FintanH (commented):
Nice start :)

One thing I'd like to call out is, once thank you next gets merged we'll be ripping out `coco` and putting it into `radicle-link`. So this ethereum addition should live in upstream specific code. I think that's easily achievable if we move the ethereum claim code into `api` and switch to taking a `PersonPayload` (I recommended this in one of my comments). 

I _think_ there's also plans to take the `api` code over too, but I'll need @xla to weigh in on this point -- I think he has a clearer idea of this migration.

proxy/coco/src/ethereum_claim_ext.rs
CodeSandwich: I don't have a strong opinion and this may be a wrong approach, but: we're just passing around the value, we don't care about its content. The time type would be parsing this string only to format it, which would introduce new bug opportunities and new dependencies. The app would still needs to parse and validate the expiration date received from the API.

FintanH: Are you not going to check the expiration value at any point? At which point you'll need to compare times, not strings.

CodeSandwich: I think that it's the App's responsibility to interpret the expiration value (e.g. to tell that it's already in the past) and only then comparisons will be needed. The only thing that api and coco can check is if the format is correct.

Anyway I thought about it and I'll add the validation. At the very least it's worth guarding the identity doc from storing junk. It's also easier to reason about the successfully parsed time than about "a string, which should follow this format, but it may not and it's still fine despite being wrong".

FintanH: Ya, not accepting junk is very valuable ;)

proxy/coco/src/state.rs
CodeSandwich: Done and moved payload extension to API

proxy/coco/src/ethereum_claim_ext.rs
CodeSandwich: Fixed

proxy/coco/src/ethereum_claim_ext.rs
CodeSandwich: Fixed

proxy/api/src/test_utils.rs
```diff
@@ -0,0 +1,22 @@
...
+    identities_payload::{Person, PersonPayload},
+    net::peer::Peer,
+    signer::BoxedSigner,
+    state::{init_owner, Error},
+    LocalIdentity,
+};
+
+/// A facade for [`init_owner`](init_owner)
+/// for creating basic identities without person payload extensions.
+pub async fn init_owner_named(
```
xla, line 13: Why is was this function introduced, what is the benefit over the normal owner intialisation?

xla:
> I _think_ there's also plans to take the `api` code over too, but I'll need @xla to weigh in on this point -- I think he has a clearer idea of this migration.

There are rough plans indeed. Current trajectory is to have little verticals that alongside the library offering like `coco` also offer pluggable http endpoints. How that manifests in Upstream is undecided as of yet. Generally I'd also advise to compartmentalise Upstream specific concerns from coco and largely api. The ideal outcome of the developments after next is to remove a lot of the indirection that shrouds the link domain and has been developed over time to marry the "lower-level" protocol domain with the product focused Upstream use-cases. Evidently this lead us to the place we are today where every non-trivial change in the protocol necessitates large migration/adoption phases up the stack, which hurts velocity but also quality of the final product.

proxy/api/src/test_utils.rs
CodeSandwich: It's a test utility. The benefit is that the content of this function would need to be pasted in multiple tests introducing noise.

xla: I don't understand how it is different from what `State::init_owner` does under the hood.

CodeSandwich:
Good to know, thanks! Does it afffect this PR? The current stack doesn't have plugins yet and `api` seems tailored for high level `app` use cases.

proxy/api/src/test_utils.rs
CodeSandwich: It's not different, it's a facade, which wraps `state::init_owner`. It hides some features (payload extensions), but offers a cleaner and simpler API, which is good enough for most tests, but requires less clutter.

proxy/api/src/ethereum_address.rs
```diff
@@ -0,0 +1,35 @@
+use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+
+/// An [EIP-55](https://eips.ethereum.org/EIPS/eip-55) Ethereum address.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[serde(try_from = "String", into = "String")]
+pub struct EthereumAddress(String);
```
FintanH, line 7: A FromStr impl would be useful too.

proxy/api/src/ethereum_address.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[derive(Clone, Debug, thiserror::Error)]
+#[error("Invalid EIP-55 address")]
```
FintanH, line 10: Someone forgot their own proposal for lowercase errors ;)

proxy/coco/src/lib.rs
```diff
@@ -25,7 +25,7 @@
 
 pub use librad::{
     git::{self, identities::local::LocalIdentity, include, local::url::LocalUrl, Urn},
-    identities::{Person, Project},
+    identities::{payload as identities_payload, Person, Project},
```
FintanH, line 28: I think `self` would be "nicer" :) Makes me think why not just re-export `librad` though really :thinking:

proxy/api/src/test_utils.rs
FintanH: The part I'm confused about is that it's repeated in 3 places now

proxy/api/src/identity.rs
```diff
@@ -95,11 +97,18 @@ impl From<Ethereum> for EthereumClaimExtV1 {
...
-    let handle = metadata.handle.clone();
-    let ethereum = metadata.ethereum.clone().map(Into::into);
-    let user = coco::state::init_owner(peer, handle, ethereum).await?;
+    let person = identities_payload::Person {
+        name: metadata.handle.into(),
+    };
+    let mut payload = identities_payload::PersonPayload::new(person);
+    if let Some(ethereum) = metadata.ethereum {
+        payload
+            .set_ext(EthereumClaimExtV1::from(ethereum))
```
FintanH, line 108: Ahhh I didn't realise `set_ext` returned an error. My thinking for `impl Into<PersonPayload>` in `state.rs` was so that `Metadata` could implement `From` and then you'd just pass it into `init_owner`. I'm wondering if that's an improvement now :thinking: 

I think a `TryFrom` for `Metadata` might be cleaner here though, what do you think?

docs/ethereum_attestation.md
```diff
@@ -0,0 +1,133 @@
+# RFC: Ethereum attestation
+
+* Author: @CodeSandwich
+* Date: 2021-03-05
+* Status: draft
+* Community discussion: n/a
+
+## Motivation
+
+The attestation between Link and Ethereum is a valuable building block for a user identity.
```
NunoAlexandre, line 10:
```suggestion
The attestation between Radicle Link and Ethereum is a valuable building block for a user identity.
```

docs/ethereum_attestation.md
```diff
@@ -0,0 +1,133 @@
...
+## Overview
+
+This RFC is built on top of the [identities spec][identities].
+It introduces support for Ethereum address claims on Link
+and a smart contract on Ethereum to make Link identity claims on Ethereum.
+
+## Link identity JSON extension
+
+The identity doc `payload` structure supports a new key: `https://radicle.xyz/ethereum/claim/v1`.
+It can be used only in person identities.
```
NunoAlexandre, line 25:
```suggestion
It can be used only in `Person` identities.
```

docs/ethereum_attestation.md
```diff
@@ -0,0 +1,133 @@
...
+Under this key there is stored an Ethereum address claim in this format:
```
NunoAlexandre, line 26: Might be good to rephrase to transmit that it _can_ be stored an Ethereum address.

```suggestion
Under this key, an Ethereum address claim can be stored with this format:
```

docs/ethereum_attestation.md
```diff
@@ -0,0 +1,133 @@
...
+
+- `address` - the claimed Ethereum address, encoded according to [EIP-55][eip-55],
+e.g. using [ethers.js][ethers-addr]
+- `expiration` - the claim expiration timestamp, encoded as a [JavaScript Date][date].
```
NunoAlexandre, line 30: > `expiration` - the claim expiration timestamp, encoded as a [JavaScript Date][date].

Since we have changed the type of the `expiration` date, I just want to make sure this is still correct as I am afraid it's not.

docs/ethereum_attestation.md
```diff
@@ -0,0 +1,133 @@
...
+- Verify that the Link identity doc claims back the Ethereum address,
+see [Discovery from a Link identity](#discovery-from-a-link-identity)
+
+## Discovery from a Link identity
+
+When you know a Link identity, you can find the claimed Ethereum address.
+Obtain the tip of its identity chain and read the Ethereum address from the identity doc
+`address` field in section `https://radicle.xyz/ethereum/claim/v1`, unless it's expired.
+You need to verify that the given Ethereum address claims back the Link identity root,
+see [Discovery from an Ethereum address](#discovery-from-an-ethereum-address).
```
NunoAlexandre, line 105: suggestion: Have a high-level "Discovery" section with two sub-sections "From an Ethereum address" and "From a Radicle Link Identity"

docs/ethereum_attestation.md
```diff
@@ -0,0 +1,133 @@
...
+Only one claim needs to be revoked to break the attestation,
+but to improve security you should revoke both sides if you can.
+
+To revoke a claim on Link, update the identity doc and publish it in a new revision.
+You can change the claimed Ethereum address or remove
+the `https://radicle.xyz/ethereum/claim/v1` section altogether.
+The other Link nodes will not notice this update until they fetch the new revision.
+
+To revoke a claim on Ethereum, call the `claim` function in `Claims` contract.
+You can claim a different Link identity root or the `0` root to revoke any claim you may have.
```
NunoAlexandre, line 120: These two sentences seem disconnected. Reading the first seem that just calling `claim` will be enough to revoke, whereas in the seconds it starts by talking about claiming another identity.

proxy/api/src/ethereum_address.rs
```diff
@@ -0,0 +1,35 @@
+use serde::{Deserialize, Serialize};
```
NunoAlexandre, line 1: I would prefer if we would group these `ethereum_` modules in an `ethereum` directory. So we would have

`ethereum/`
  - `address.rs`
  - `claim_ext.rs`

proxy/api/src/ethereum_claim_ext.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use crate::ethereum_address::EthereumAddress;
+use chrono::{DateTime, Utc};
+use coco::identities_payload::HasNamespace;
+use lazy_static::lazy_static;
+use serde::{Deserialize, Serialize};
+use url::Url;
+
+/// The user identity doc extension for Ethereum addresses claims.
+/// Meaningful only if confirmed on Ethereum, see [the RFC](docs/ethereum_attestation.md).
```
NunoAlexandre, line 12:
```suggestion
/// Meaningful only if confirmed on Ethereum. See [the RFC](docs/ethereum_attestation.md).
```

proxy/coco/src/lib.rs
NunoAlexandre, line 28: do we need this alias? I'd prefer to refer to it as `identities::payload` instead.

NunoAlexandre (commented):
Looking good :rocket:

proxy/api/src/test_utils.rs
CodeSandwich: Me too :( We have 3 different test scopes:
1. Coco tests, they run on coco compiled with `cfg(test)`, so they see tools in `mod tests`
2. Api tests, they run on coco compiled as production code, so they can see only the production API. This should be possible to be mitigated with some features and dev-dependencies trickery, but AFAIR you've tried it before and there were some issues. I don't want to bloat the scope of this PR either, I'm fine with adding a `tests` feature and DRY-ing it later.
3. Coco integration tests, they run on coco compiled as production code, so they can see only the production API. AFAIK they have no sane method of accessing the coco test API: https://github.com/rust-lang/cargo/issues/2911

I haven't found any method of DRY-ing this tool without putting it in the production API of coco, where it's unneeded.

proxy/api/src/identity.rs
CodeSandwich: > `TryFrom` for `Metadata`

Why haven't I thought about it :facepalm:

proxy/coco/src/lib.rs
CodeSandwich: Fixed:
```
identities::{
        payload::{self as identities_payload},
        Person, Project,
    },
```
Is this what you meant? IMO the old one was a little nicer, but I'm fine with this change :)

proxy/api/src/ethereum_address.rs
CodeSandwich: You got me, fixed :laughing: Some errors in `proxy/api/src/error.rs` are uppercase too, should I fix them now or in a separate PR to avoid noise?

proxy/api/src/ethereum_address.rs
CodeSandwich: Fixed

proxy/coco/src/lib.rs
FintanH: Whoops, should have been clearer. I was saying that we should just re-export `identities`, and as @NunoAlexandre commented somewhere else, we would use it as `coco::identities::payload` :)

proxy/api/src/ethereum_address.rs
FintanH: Ya, let's not add noise but thanks for calling that out!

proxy/api/src/identity.rs
CodeSandwich: Fixed, now it turned out that we don't need any extra test tools, the production API is IMO good enough

proxy/coco/src/lib.rs
CodeSandwich: Fixed. I hope that we haven't used up a precious name :grimacing:

proxy/api/src/test_utils.rs
CodeSandwich: The problem solved itself, the new production API doesn't need any test tools

proxy/coco/src/state.rs
```diff
@@ -90,23 +90,21 @@ where
...
+    payload: impl Into<payload::PersonPayload> + Send,
+) -> Result<LocalIdentity, Error> {
     match peer
         .using_storage(move |store| local::default(store))
         .await??
     {
         Some(owner) => Ok(owner),
         None => {
             let pk = keys::PublicKey::from(peer.signer().public_key());
+            let payload = payload.into();
```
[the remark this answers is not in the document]
CodeSandwich: That was my initial solution. Unfortunately that would require an unconverted `payload` to be passed across a thread boundary, so it can be executed in a task. This is possible, but requires an unnecessary additional `Send` type constraint on the generic `payload` parameter, which is much noisier than this awkward conversion.

proxy/coco/src/lib.rs
CodeSandwich: Fixed

proxy/api/src/http/identity.rs
```diff
@@ -112,8 +102,15 @@ mod test {
...
             .path("/")
-            .json(&super::CreateInput {
+            .json(&identity::Metadata {
                 handle: "cloudhead".into(),
+                ethereum: Some(identity::Ethereum {
+                    address: "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B"
+                        .to_string()
+                        .try_into()
+                        .expect("Invalid address"),
+                    expiration: "2021-03-19T23:15:30.000Z".parse().expect("Invalid date"),
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

proxy/api/src/ethereum_claim_ext.rs
CodeSandwich: Fixed

proxy/api/src/ethereum_address.rs
CodeSandwich: Fixed

proxy/coco/src/state/error.rs
```diff
@@ -113,6 +113,10 @@ pub enum Error {
         /// The reference that we looked for in the `Storage`.
         reference: Reference<One>,
     },
+
+    /// A document payload extension was malformed
+    #[error(transparent)]
+    PayloadExt(#[from] librad::identities::payload::ExtError),
```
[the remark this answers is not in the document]
CodeSandwich: Fixed

docs/ethereum_attestation.md
CodeSandwich: Is it right? `person` is an identity kind as a high level concept, `` `Person` `` looks more like a name of a specific Rust type

docs/ethereum_attestation.md
CodeSandwich: I think that it's still valid, we've just added verification of that format.

docs/ethereum_attestation.md
CodeSandwich: Hmm, the first sentence describes the action and the second its details, `claim` can't be called without any parameters. Do you have an idea for a clearer expression of that?

docs/ethereum_attestation.md
CodeSandwich: Fixed

docs/ethereum_attestation.md
CodeSandwich: It seems more ambiguous, the claim MUST be stored and the format MUST be obeyed. I'm afraid of losing that clarity.

docs/ethereum_attestation.md
CodeSandwich: Fixed

docs/ethereum_attestation.md
NunoAlexandre: hm. when I read the current sentence, I read that a claim must exist, whereas claiming an Ethereum address is an optional extension. That's what my suggestion tries to clarify.

docs/ethereum_attestation.md
NunoAlexandre: It's good now :ok_hand:

docs/ethereum_attestation.md
NunoAlexandre: Ok, what about:

> Under this key, an Ethereum address claim is stored in this format:

CodeSandwich: Perfect, fixed

NunoAlexandre (approved):
:speedboat:

proxy/api/src/ethereum/address.rs
```diff
@@ -0,0 +1,46 @@
...
+            Err(InvalidEIP55AddressError)
+        }
+    }
+}
+
+impl FromStr for Address {
+    type Err = InvalidEIP55AddressError;
+
+    fn from_str(address: &str) -> Result<Self, Self::Err> {
+        address.to_string().try_into()
```
FintanH, line 32: nit: since `validate_address` takes a `&str` it would indicate that `FromStr` should be the implementation and `TryFrom<String>` can use the `FromStr` impl. That means we're not sticking a `&str` on the heap every time we parse.

proxy/api/src/identity.rs
```diff
@@ -28,14 +38,22 @@ impl From<(coco::PeerId, coco::Person)> for Identity {
     fn from((peer_id, user): (coco::PeerId, coco::Person)) -> Self {
         let urn = user.urn();
         let handle = user.subject().name.to_string();
+        // Silently ignore the extension payload if malformed, the identity itself is still valid
```
FintanH, line 41: I have a feeling that silently ignoring will lead us to wonder why some error happened down the line :D Can we maybe log as a warning?

FintanH (approved):
This is looking good :) Just a couple of nits

proxy/api/src/ethereum/address.rs
CodeSandwich: Nitpick time? Nitpick time! I was thinking about 2 cases, constructing `Address` from `&str` and from `String`.

In case of `&str` we need to copy it onto heap as a `String` at some point anyway. In the current implementation it's a wasted effort only if an error occurs.

In case of `String` we don't need to copy anything, just wrap it in an `Address`. If we push a `String` through `from_str`, it's effectively cloned and wasting resources every time.

FintanH: Ah good point! Nitpick time over!

proxy/api/src/identity.rs
CodeSandwich: Fixed, added logging

PR #1686, opened 2021-03-19
ui: display only the new commits of a merge request

NunoAlexandre:


NunoAlexandre:
@geigerzaehler do you have a project URN I could test with from the network? (I just created a merge request on a project of mine).

xla:
Holistically we want to support [gitrevisions](https://git-scm.com/docs/gitrevisions) proper on an endpoint like diff. We ignored a lot of the complexity stemming from potentially supporting revisions, but this sounds like the right point to address them and provide the utility so features like this can be built. What we get already is the infrastructure (types, UI, etc.) to pass around and display diffs, so that's a win.

NunoAlexandre:
> Holistically we want to support [gitrevisions](https://git-scm.com/docs/gitrevisions) proper on an endpoint like diff. We ignored a lot of the complexity stemming from potentially supporting revisions, but this sounds like the right point to address them and provide the utility so features like this can be built. What we get already is the infrastructure (types, UI, etc.) to pass around and display diffs, so that's a win.

Ya I agree, thus having kept this in draft. Do you have or could point me to a detailed set of requirements for such endpoint, _if_ we have that? It's the first time I am looking into this part of the code so all info available is much appreciated.

geigerzaehler:
I agree, an endpoint that gets a commit list from a general git revision parameter would be useful not only here but also in general. But we also have to keep an eye on the effort that implementing this endpoint requires. I’d like to avoid a situation where this blocks the merge requests. A separate issue for this would be a good place to drive forward the discussion

NunoAlexandre:
@geigerzaehler considering your comment above, I have set this as ready for review. The solution proposed here is far from ideal but it's non-invasive and it provides the UX value we want. Will follow up with an issue that takes xla's comment into account and move the filtering to be offered by a dedicated endpoint.

geigerzaehler:
I’ve fixed a couple of things on the base branch. Could you rebase, @NunoAlexandre?

NunoAlexandre:
@geigerzaehler done.

ui/DesignSystem/Component/SourceBrowser/History.svelte
```diff
@@ -45,21 +46,23 @@
...
-    <header>
-      <p>{group.time}</p>
-    </header>
-    <ul>
-      {#each group.commits as commit (commit.sha1)}
-        <li class="commit" on:click={() => onSelect(commit)}>
-          <CommitTeaser
-            {commit}
-            style="background: none; --commit-message-color:
+<div class="history" data-cy="history">
```
NunoAlexandre, line 49: :information_source: the only change here is wrapping the list of commit groups in a `div.history` element.

ui/src/screen/project/source.ts
```diff
@@ -394,20 +407,40 @@ export const fetchMergeRequestCommits = async (
...
+        : 0;
+      const newCommits = mrCommits.history.slice(
+        0,
+        nothingNewIdx === -1 ? 0 : nothingNewIdx
+      );
+
+      mergeRequestCommitsStore.success({
+        history: newCommits,
+        stats: { ...mrCommits.stats, commits: newCommits.length },
+      });
```
geigerzaehler, line 443: I believe this `source.CommitHistory` object needs to be a `CommitHistory` object. Otherwise the app does not work. This also requires an adjustment of the type of the `mergeRequestCommits` store.

ui/src/screen/project/source.ts
```diff
@@ -61,6 +61,17 @@ interface Screen {
   tree: Writable<source.Tree>;
 }
 
+export interface CommitsHistory {
+  history: CommitGroup[];
+  stats: source.Stats;
+}
```
geigerzaehler, line 67: To avoid having two different types with the same name I’d suggest to call this `GroupedCommitHistory`.

ui/DesignSystem/Component/SourceBrowser/History.svelte
NunoAlexandre: My bad, the previous rebase & merge-resolution didn't go well. Just rebased again and now it's all in good shape.

So, answering your question, see https://github.com/radicle-dev/radicle-upstream/pull/1686/files#diff-c1135bec4b86c0f227a805cabfd0d9933c669329ede40a2ef36a2c031c066264R423. The list of commit groups wasn't wrapped in an element so testing the number of commits would be unreliable as if we had more group-commits displayed anywhere else in the page things wouldn't add up.

ui/src/screen/project/source.ts
NunoAlexandre: Done!

ui/src/screen/project/source.ts
NunoAlexandre: Obsolete, result of a bad rebase, fixed now but please double check.

ui/DesignSystem/Component/SourceBrowser/History.svelte
NunoAlexandre: :heavy_check_mark:

geigerzaehler (approved):

Issue #1687, opened 2021-03-22
Upgrade to Yarn v2


geigerzaehler:
We should upgrade to Yarn v2 since v1 has been in maintenance mode for over a year. The main driver for this is that I run into a Yarn [bug](https://github.com/yarnpkg/yarn/issues/4563) when trying to setup Github Actions.

PR #1688, opened 2021-03-22
refactor(ui): download twemoji assets directly

geigerzaehler:
Instead of using our own `twemoji-svg-assets` package we download the assets directly as part of the `postinstall` scripts.

This has two advantages: First, we don’t need to maintain a separate package. Second, this removes a blocker of the [upgrade to Yarn v2][yv2]. Our `twemoji-svg-asstest` package copied the assets as part of *its* [post-install script][pis]. This only works under certain assumptions about the directory layout which don’t hold for Yarn v2.

[pis]: https://github.com/radicle-dev/twemoji-svg-assets/blob/master/package.json#L10
[yv2]: https://github.com/radicle-dev/radicle-upstream/issues/1687

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,7 +157,7 @@
     "_private:rollup:build": "rollup -c --failAfterWarnings",
     "_private:rollup:build:release": "ROLLUP_RELEASE=true rollup -c --failAfterWarnings",
     "_private:rollup:watch": "rollup -c -w",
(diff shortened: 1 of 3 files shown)


xla (approved):
🌴 ☀️ 🗓 🔑

xla (approved):
🆗 🍤 ♨️ 💭

PR #1689, opened 2021-03-22
refactor(build): don’t clean artifacts that are overwritten

geigerzaehler:
We remove those steps in the build script that remove old build artifacts from disk. This step is not necessary since those files will be overwritten by output of subsequent build steps.

The cleanup resulted in [CI failures][ci-failure] if the artifacts were not present before running `yarn:test`. (In fact, these artifacts were never present before running the scripts. It just did not error because of specific shell setting for glob matching). Moreover, `cargo clean` lead to unnecessary long build times when running `yarn dist` without providing a significant advantage.

[ci-failure]: https://buildkite.com/monadic/radicle-upstream/builds/8042#f449736c-2c9a-4602-9572-5a7ee15f10d8/317-336


NunoAlexandre:
> The cleanup resulted in CI failures if the artifacts were not present before running yarn:test

Does this mean that the tests might run on outdated build outputs?

geigerzaehler:
> Does this mean that the tests might run on outdated build outputs?

No, the build agent [removes all untracked files](https://buildkite.com/monadic/radicle-upstream/builds/7989#438156db-0579-4a67-a749-7f0851abe5c1/12-19) from the work space before the build starts.

NunoAlexandre (dismissed):
:hourglass:

rudolfs:
What if we run `yarn dist` locally?

geigerzaehler:
> What if we run `yarn dist` locally?

`yarn dist` only succeeds if the artifacts are correctly created. This means that we can never bundle outdated artifacts.

PR #1690, opened 2021-03-22
fix(ci): run-p --race kills process

geigerzaehler:
Before `run-p --race` did not kill processes on CI and we had to to this manually. The reason for this behavior was that the `ps` binary was not installed. We fix that by installing the `procps` package. Having `ps` in the image is also useful for debugging.

We also take the opportunity to clean up the Dockerfile a bit.
* We move the `build-essential` package to the top and stop listing its dependencies (like gcc) explicitly.
* We install `yarn` via `npm` instead of a debian package


xla (approved):
🏭 ➗ 🚊 📺

PR #1692, opened 2021-03-22
eth: bump radicle-contracts & add claims contract

NunoAlexandre:
We bump the radicle contracts dependency, which involves complying with the latest API of the Erc20Pool contract.
We also add the Claims contract.


package.json
```diff
@@ -203,4 +203,4 @@
       "eslint --fix --max-warnings=0"
     ]
   }
-}
+}
```
geigerzaehler, line 206: Could you make sure there’s a new line here so this doesn’t get changed unnecessarily?

geigerzaehler (dismissed):

package.json
NunoAlexandre, line 206:
```suggestion
}

```

CodeSandwich (dismissed):

geigerzaehler (dismissed):

NunoAlexandre:
merged in 3897f1492713274b59fa1ddb9b470b1a313d307d

PR #1697, opened 2021-03-23
refactor(dev): don't bundle electron in dev mode

geigerzaehler:
We stop bundling the entrypoint of the electron main process in development. Instead we load typescript dynamically. This decreases the bundling time in development and will allow us to switch to a different bundler more easily.


xla (dismissed):
♑️ 🗑 🏕 ⚒

PR #1699, opened 2021-03-23
Disable sync tests

CodeSandwich:
Since the last commit these tests started failing


FintanH:
I've fixed them, so you can close this :)

CodeSandwich:
Cool! On which branch?

PR #1700, opened 2021-03-23
fix(build): fix bundling of electron main files

geigerzaehler:
We fix the bundling of the electron main process file and it to the tests.


NunoAlexandre (approved):

xla (approved):
☕️ 🗒 🈳 🌵

PR #1707, opened 2021-03-25
fix(ci): automatically clean cache when it is almost full

geigerzaehler:
Instead of having to clear caches manually on the build agent we do this as part of the build script.

This code was adapted from the [registry][].

[registry]: https://github.com/radicle-dev/radicle-registry/blob/00436dc9b2fad0ad96c2e562f76873d0aa21a2c0/ci/run#L27-L36


xla (approved):
ʅ（◞‿◟）ʃ

NunoAlexandre (approved):
:broom:

PR #1710, opened 2021-03-25
refactor(build): replace rollup with webpack to make build faster

geigerzaehler:
We replace `rollup` with `webpack` to make our builds faster. While webpack is not faster without a cache it supports persistent caches in the file system which makes it significantly faster in development. Webpack is also significantly faster in watch mode.

On my machine a rollup build and a webpack build without a cache take about 60 seconds. If a cache is present webpack builds the project in 8-9 seconds. When updating a file in watch mode rollup takes 7-8 seconds to rebuild. Webpack rebuilds the project in under a second.

Another thing to note is that in the past I spent multiple hours investigating issues with rollup and tweaking the config. None of this was necessary for webpack. We also gain more functionality, for example we can now leverage templating in the HTML document.

Some implementation notes:
* We patch `package.json` of `svelte-persistent-store` because of [this issue][sps-issue]. This only worked before because rollup itself was buggy.
* We add some dependencies that are used in our source code but were not explicitly specified.
* We use `html-webpack-plugin` to dynamically build `public/index.html` and configure it for different environments.
* We remove `bundle.css`. CSS is now injected during runtime via JS.

Follup-up:
* We still need to leverage the webpack cache on CI

[sps-issue]: https://github.com/andsala/svelte-persistent-store/issues/15


webpack.config.ts
```diff
@@ -0,0 +1,112 @@
...
+      compilerOptions: {
+        noEmit: false,
+        module: "es6",
+      },
+    },
+  },
+};
+
+function electronMain(_env: unknown, argv: Argv): webpack.Configuration {
+  const mode = argv.mode || "development";
```
NunoAlexandre, line 25: `suggestion`

Have a function to dry the mode fallback and call it directly when setting the `mode` field:

```
  mode: getMode(argv),
```

NunoAlexandre (approved):
Awesome :sunflower:

xla (approved):
utACK

🔐 🐂 👴 🍿

rudolfs (approved):

Issue #965, opened 2020-09-25
feat: Support eth address two-way attestation


NunoAlexandre:
**Depends on:** https://github.com/orgs/radicle-dev/projects/17 the integration of radicle-link's `next` branch.

This feature has an impact the whole Funding experience we are building by linking user's Radicle identities and their Ethereum accounts, leveraging discoverability (users will be able to tag each other in the Funding features by name instead of by addresses) and trust (users can trust that an address is associated with a given Radicle user and vice-versa).

### What needs to be done

- [ ] Add a "ethAddress" field to the Radicle Identity document (metadata?) or have a dedicated document for this purpose
- [ ]  Create the necessary infrastructure in `proxy` to set and get this field (http endpoint, etc)
- [ ] Wire up the attention flow in the frontend (see below)

#### Attestation flow

1. As a Radicle user, I have no registered eth address
2. I want to access the Radicle Funding features, therefore I go to "Funding"
3. In "Funding", I see a prompt to connect my wallet using the Eth account I want to use to register in the Radicle network with
4. I connect my wallet app using the QR code provided by Radicle
5. Now on Radicle, I see the correct account address displayed
6. I confirm I want to associate that eth account with my Radicle identity (attestation step 1/2)
6.2 I confirm a transaction storing that association on-chain (attestation step 2/2), which will prompt my wallet app for submission
7. I confirm the registration in my wallet app
9. the app confirms the registration is successful. Now I have access to all funding features.

10. Next time I connect using another account, I get a warning informing me that I am using an account that's not the one I have registered

You can see the stubbed flow from step 5 until 9:

![Peek 2020-12-03 16-59](https://user-images.githubusercontent.com/7075260/101055336-180a0180-358a-11eb-8ffc-5876ce41b208.gif)

juliendonck:
@NunoAlexandre is this still relevant with the latest thinking?

NunoAlexandre:
The issue is still a relevant one but out of scope for the Funding Pool MVP.

xla:
@NunoAlexandre Would it be possible to capture this work somewhere else or do you think it has value to keep the issue around?

NunoAlexandre:
@xla @juliendonck I have updated the description and scope of this issue to fully capture what it entails.



> @NunoAlexandre Would it be possible to capture this work somewhere else or do you think it has value to keep the issue around?

[...]

PR #1712, opened 2021-03-27
feat(proxy): add PUT identity metadata endpoint

CodeSandwich:
Part of https://github.com/radicle-dev/radicle-upstream/issues/965. Adds an API endpoint for updating the user payload. This allows updating the ethereum claims and as a byproduct also changing the user handle.

### DIFF
--- a/proxy/api/src/http/identity.rs
+++ b/proxy/api/src/http/identity.rs
@@ -8,6 +8,7 @@ use crate::{context, http};
 pub fn filters(ctx: context::Context) -> BoxedFilter<(impl Reply,)> {
     get_filter(ctx.clone())
         .or(create_filter(ctx.clone()))
+        .or(update_filter(ctx.clone()))
         .or(list_filter(ctx))
         .boxed()
 }
@@ -21,6 +22,17 @@ fn create_filter(
         .and(http::with_context_unsealed(ctx))
         .and(warp::body::json())
         .and_then(handler::create)
+}
+
+/// `PUT /`
+fn update_filter(
+    ctx: context::Context,
+) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    path::end()
+        .and(warp::put())
+        .and(http::with_context_unsealed(ctx))
+        .and(warp::body::json())
+        .and_then(handler::update)
 }
 
 /// `GET /<id>`
@@ -48,7 +60,7 @@ fn list_filter(
 mod handler {
     use warp::{http::StatusCode, reply, Rejection, Reply};
 
-    use crate::{context, error, identity, session};
+    use crate::{context, error, http, identity, session};
 
     /// Create a new [`identity::Identity`].
     pub async fn create(
@@ -66,6 +78,18 @@ mod handler {
         session::initialize(&ctx.store, id.clone(), &ctx.default_seeds)?;
 
         Ok(reply::with_status(reply::json(&id), StatusCode::CREATED))
+    }
+
+    /// Update the [`identity::Identity`] metadata.
+    pub async fn update(
+        ctx: context::Unsealed,
+        metadata: identity::Metadata,
+    ) -> Result<impl Reply, Rejection> {
+        session::get_current(&ctx.store)?.ok_or(http::error::Routing::NoSession)?;
+        let id = identity::update(&ctx.peer, metadata).await?;
+        session::update_identity(&ctx.store, id.clone())?;
+
+        Ok(reply::with_status(reply::json(&id), StatusCode::OK))
     }
 
     /// Get the [`identity::Identity`] for the given `id`.
@@ -153,6 +177,163 @@ mod test {
                             "address": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
                             "expiration": "2021-03-19T23:15:30.001Z",
                         }
+                    },
+                    "shareableEntityIdentifier": &shareable_entity_identifier,
+                })
+            );
+        });
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn update() -> Result<(), error::Error> {
+        let tmp_dir = tempfile::tempdir()?;
+        let (ctx, _) = context::Unsealed::tmp(&tmp_dir)?;
+        let api = super::filters(ctx.clone().into());
+
+        let res = request()
+            .method("POST")
+            .path("/")
+            .json(&identity::Metadata {
+                handle: "cloudhead".into(),
+                ethereum: None,
+            })
+            .reply(&api)
+            .await;
+        http::test::assert_response(&res, StatusCode::CREATED, |_| ());
+
+        let res = request()
+            .method("PUT")
+            .path("/")
+            .json(&identity::Metadata {
+                handle: "cloudhead_next".into(),
+                ethereum: Some(identity::Ethereum {
+                    address: "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B"
+                        .to_string()
+                        .try_into()
+                        .expect("Invalid address"),
+                    expiration: "2021-03-19T23:15:30.001Z".parse().expect("Invalid date"),
+                }),
+            })
+            .reply(&api)
+            .await;
+
+        let urn = {
+            let session = session::get_current(&ctx.store)?.expect("no session exists");
+            session.identity.urn
+        };
+
+        let peer_id = ctx.peer.peer_id();
+
+        // Assert that we set the default owner and it's the same one as the session
+        {
+            assert_eq!(
+                coco::state::default_owner(&ctx.peer)
+                    .await?
+                    .unwrap()
+                    .into_inner()
+                    .into_inner(),
+                coco::state::get_user(&ctx.peer, urn.clone())
+                    .await?
+                    .unwrap()
+                    .into_inner()
+                    .into_inner()
+            );
+        }
+
+        http::test::assert_response(&res, StatusCode::OK, |have| {
+            let avatar = avatar::Avatar::from(&urn.to_string(), avatar::Usage::Identity);
+            let shareable_entity_identifier = format!("cloudhead_next@{}", peer_id);
+            assert_eq!(
+                have,
+                json!({
+                    "peerId": peer_id,
+                    "avatarFallback": avatar,
+                    "urn": urn,
+                    "metadata": {
+                        "handle": "cloudhead_next",
+                        "ethereum": {
+                            "address": "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
+                            "expiration": "2021-03-19T23:15:30.001Z",
+                        }
+                    },
+                    "shareableEntityIdentifier": &shareable_entity_identifier,
+                })
+            );
+        });
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn update_remove_ethereum_claim() -> Result<(), error::Error> {
+        let tmp_dir = tempfile::tempdir()?;
+        let (ctx, _) = context::Unsealed::tmp(&tmp_dir)?;
+        let api = super::filters(ctx.clone().into());
+
+        let res = request()
+            .method("POST")
+            .path("/")
+            .json(&identity::Metadata {
+                handle: "cloudhead".into(),
+                ethereum: Some(identity::Ethereum {
+                    address: "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B"
+                        .to_string()
+                        .try_into()
+                        .expect("Invalid address"),
+                    expiration: "2021-03-19T23:15:30.001Z".parse().expect("Invalid date"),
+                }),
+            })
+            .reply(&api)
+            .await;
+        http::test::assert_response(&res, StatusCode::CREATED, |_| ());
+
+        let res = request()
+            .method("PUT")
+            .path("/")
+            .json(&identity::Metadata {
+                handle: "cloudhead".into(),
+                ethereum: None,
+            })
+            .reply(&api)
+            .await;
+
+        let urn = {
+            let session = session::get_current(&ctx.store)?.expect("no session exists");
+            session.identity.urn
+        };
+
+        let peer_id = ctx.peer.peer_id();
+
+        // Assert that we set the default owner and it's the same one as the session
+        {
+            assert_eq!(
+                coco::state::default_owner(&ctx.peer)
+                    .await?
+                    .unwrap()
+                    .into_inner()
+                    .into_inner(),
+                coco::state::get_user(&ctx.peer, urn.clone())
+                    .await?
+                    .unwrap()
+                    .into_inner()
+                    .into_inner()
+            );
+        }
+
+        http::test::assert_response(&res, StatusCode::OK, |have| {
+            let avatar = avatar::Avatar::from(&urn.to_string(), avatar::Usage::Identity);
+            let shareable_entity_identifier = format!("cloudhead@{}", peer_id);
+            assert_eq!(
+                have,
+                json!({
+                    "peerId": peer_id,
+                    "avatarFallback": avatar,
+                    "urn": urn,
+                    "metadata": {
+                        "handle": "cloudhead",
+                        "ethereum": null
                     },
                     "shareableEntityIdentifier": &shareable_entity_identifier,
                 })
--- a/proxy/api/src/identity.rs
+++ b/proxy/api/src/identity.rs
@@ -70,6 +70,18 @@ pub struct Metadata {
     pub ethereum: Option<Ethereum>,
 }
 
+// Update the given payload using the properties from the given metadata.
+pub fn update_payload(
+    mut payload: PersonPayload,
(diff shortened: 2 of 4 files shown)


proxy/api/src/error.rs
```diff
@@ -23,6 +23,9 @@ pub enum Error {
     #[error("the current session is in use by `{0}`")]
     SessionInUse(coco::Urn),
 
+    #[error("no session in use")]
```
NunoAlexandre, line 26: There is already a `NoSession` error in `http::error::Routing::NoSession`, will replace.

proxy/api/src/http/identity.rs
```diff
@@ -23,6 +24,17 @@ fn create_filter(
         .and_then(handler::create)
 }
 
+/// `PUT /`
```
NunoAlexandre, line 27: Since this only updates the `Metadata`, I'd prefer to namespace it accordingly.

NunoAlexandre:
@FintanH it may be good for you to review this one and having me addressing it.

proxy/coco/src/state.rs
```diff
@@ -103,40 +97,54 @@ where
...
+{
+    let urn = default_owner(peer).await?.ok_or(Error::MissingOwner)?.urn();
+    let payload = payload.try_into()?;
+    peer.using_storage(move |store| person::update(store, &urn, None, payload, None))
+        .await??;
+    // Update the default user info in the config with the new payload data.
+    // This works, because the payload update doesn't change the owner's URN,
+    // so the old config can be used to load the owner with the updated payload.
+    let owner = default_owner(peer).await?.ok_or(Error::MissingOwner)?;
+    set_default_owner(peer, owner.clone()).await?;
```
FintanH, line 146: I'm not sure this will always work in the future. If the `payload` is updating the document with a new key (or new keys), then the `Person` won't be verified since it won't necessarily have a quorum of signatures.

We're getting away with it now because there's only one key in the document and the signing of that key is happening in `update`, I think.

proxy/api/src/identity.rs
```diff
@@ -69,6 +69,18 @@ pub struct Metadata {
     pub ethereum: Option<Ethereum>,
 }
 
+impl Metadata {
+    pub fn update_payload(self, mut payload: PersonPayload) -> Result<PersonPayload, ExtError> {
```
FintanH, line 73: Something I'm thinking about is whether a `Metadata` that has `Option` fields might be more interesting for updates, i.e.
```rust
pub struct MetadataUpdate {
  name: Option<String>,
  ethereum: Option<Eth>,
}
```

Unfortunately this duplicates the structure (higher-kinded data anyone?!) but I think it's nice for targeting fields that you want to update (a "sensible" default is `None` on all fields). This could also allow the endpoint to have all those fields as optional so if they're missing then it just takes them as none and can set query parameters it cares about updating.

Let me know what you think.

proxy/api/src/http/identity.rs
FintanH: What else could it `PUT` though? :thinking:

NunoAlexandre: I get your question but note how all the other endpoints aim at the whole `Identity` type, whereas this only aims at the identity Metadata:

- `POST /` create an `Identity`
- `GET  /<id>` gets an `Identity`
- `GET /` gets a list of `Identity`

Thus, seeing `PUT /` for me would mean "update an `Identity`", not a sub-part of an `Identity`. So I'd rather have an explicit, clear, and namespaced endpoint which would still make sense if, in the future, we introduce some other updatable field in `Identity`. :man_shrugging:

proxy/coco/src/state.rs
NunoAlexandre: Could you point me in a direction where this could be implemented to be future-proof regarding your point?

proxy/api/src/http/identity.rs
FintanH: Ya, makes sense :+1:

proxy/api/src/identity.rs
CodeSandwich: I've designed the proposed API with these problems in mind:
1. We need to be able to **remove** the Ethereum claim. How to express that, on JSON side and on Rust DTO side? `Option<Option<Eth>>`? Special meaning of a missing vs `null` JSON value? A list of removed values? They all are much trickier, especially considering communication between different JSON interpreters and different languages.
2. This structure is handled and interpreted in an identical way as the one passed to create an identity. It's easier to understand an API like this and requires less code to support, both on the frontend an the backend. 

There may be a better design and I'm not going to fight very hard for this one, but to me it seems reasonable.

proxy/api/src/http/identity.rs
CodeSandwich: This `PUT` can update everything there is to update about an `Identity`. Look at `create`, it too accepts only a `Metadata` to describe a whole new identity. I'm fine with moving it into a separate namespace, but IMO it makes sense to keep it here as `update an Identity`.

proxy/api/src/identity.rs
NunoAlexandre: Right, that's a good point regarding being able to _remove_ the claim :thinking:

proxy/api/src/http/identity.rs
NunoAlexandre: We can also do that, sure, but then I'd expect the endpoint to operate on `Identity` and not `Metadata`.

proxy/api/src/identity.rs
FintanH: That's a great point, sorry I missed that :)

I guess the correct structure would be the same as `Option` except it would convert `Update::Replace(Option<Eth>)` to `eth=null` and back again and the omission of `eth=` would be `Update::Dont`. But anyway, you're right this straight-forward and makes sense :)

_MAYBE_ what I want is lenses :P

NunoAlexandre: Lenses :heart:

proxy/coco/src/state.rs
FintanH: So [these tests](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/identities/git/tests/person.rs) give us an idea of what's going on here. The concept being that if I'm working on a Radicle from my desktop and my laptop I'll have two separate keys for them, but I'd like to use the one `Person` document for `fintohaps` alter-ego.

As soon as I [`update`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/identities/person.rs#L123-L148) my document, on my desktop, with my eth address to (GET PAID :money_mouth_face:) I'll have a document that is signed but not verified. This change needs to be propagated to my laptop and I [`merge`](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/identities/person.rs#L150-L172) the changes if I'm happy that it was me. I think the desktop side will also have to `merge` for _it_ to be in a verified state.

I think but I'm not sure that, running `set_default_owner` here _without_ verifying will mean that it fails because `LocalIdentity` is a wrapper around `VerifiedPerson`. If it's the case that it doesn't fail my bet is that it won't have the update you think it'll have and instead it will be at the previously verified state (sans eth address) -- [this comment](https://github.com/radicle-dev/radicle-link/blob/master/librad/src/git/identities/person.rs#L57-L62) alludes to that.

So this updating of documents across devices needs be incorporated into the plans, in my opinion. I've mentioned it several times that we need to work out this flow if we want to open up the new features `radicle-link` now provides. It'll be the same for adding a maintainer to a `Project` :grin: I'm unsure of what thoughts @radicle-dev/product has had around this flow of cross-device verification, but might be worth thinking about as well.

As for this specific function, I would think that we would need to figure out when a document needs verification from other keys (in the case of `Person`) -- I'm not sure `radicle-link` provides this, but maybe @kim can chime in on that point.

proxy/api/src/http/identity.rs
CodeSandwich: That's a good point, we're operating on an `Identity` and then suddenly on a `Metadata`, that's messy :thinking:

On the other hand we simply can't have the same JSON for getting an identity and creating/updating it, they transport different sets of parameters. To introduce segregation based on the JSONs, we'd need to put getters from setters in separate namespaces despite them working on the same thing.

I don't have strong feeling about it, I'm just nitpicking now and I believe that your solution will be fine :)

NunoAlexandre: Let's leave it as is and change when/if needed

proxy/coco/src/state.rs
kim: > As for this specific function, I would think that we would need to figure out when a document needs verification from other keys (in the case of Person) -- I'm not sure radicle-link provides this, but maybe @kim can chime in on that point.

The identity history permits commits near the tip of the `rad/id` branch which don't pass the quorum rules. This means that, if the oid the tip is pointing to is not equal to the `content_id` of the verified identity, some action is required. This can be determined using `get` vs. `verify` (we could also add some sugar to call `verify` on the result of `get`, but haven't).

Afaics, updating the "default owner" doesn't make much sense here indeed -- the default it just the URN stored in the git config, and since that is read at the beginning, it is presumably still the same. That is, `person::update` is enough for what you want to do here, I think.

FintanH: I'm thinking it would make sense to carry the information in the result of `verify`. At the point we know that it's not current tip so it could be something like: `verify -> Urn -> Verifying<Person>`, where:
```rust
enum Verifying<A> {
  Latest(A),
  Behind(A),
}
```

kim: That doesn’t seem like a huge improvement over a Boolean. If you’d want to rework the fold so it keeps accumulating non-quorum updates after it found a quorum one (similar to how the validation applicative works), I would certainly meet that with appreciation.

CodeSandwich: > Afaics, updating the "default owner" doesn't make much sense here indeed -- the default it just the URN stored in the git config, and since that is read at the beginning, it is presumably still the same.

The update of the default owner updates the name and the email of the git user too. It's not very useful for our short-term use case (update of the Ethereum claim), but at some point it probably should be a part of the update of the metadata. I've added it only for the sake of completeness and to make the API behavior consistent. We can remove it and document that the the metadata update mustn't update the user name.

FintanH: > The update of the default owner updates the name and the email of the git user too

Eh? Where does it do that?

kim: It updates `user.name`, that’s correct. `user.email` is the peer id, so doesn’t change.

Both are only used for commits which are made by `librad`, and are not normally user-visible. They are set only because git expects _something_ in the committer/author fields, we never actually inspect them.

FintanH: > It's not very useful for our short-term use case [..] We can remove it and document that the the metadata update mustn't update the user name.

I'm not advocating we remove this :) I'm thinking about the _long-term_ of these updates since it would benefit _all_ cases of updating `Person` documents -- and also `Project` documents.

proxy/api/src/http/identity.rs
```diff
@@ -68,6 +80,18 @@ mod handler {
...
     }
 
+    /// Update the [`identity::Identity`] metadata.
+    pub async fn update(
+        ctx: context::Unsealed,
+        metadata: identity::Metadata,
+    ) -> Result<impl Reply, Rejection> {
+        session::get_current(&ctx.store)?.ok_or(http::error::Routing::NoSession)?;
+        let id = identity::update(&ctx.peer, metadata).await?;
+        session::update_identity(&ctx.store, id);
```
NunoAlexandre, line 90: FYI @FintanH 

So I've added this line calling the - also added - `session::update_identity` method because I started building the call from the UI calling this endpoint and noticed the session would never update.

proxy/api/src/identity.rs
NunoAlexandre: Getting back to this discussion, more specifically the design goal to be able to **remove** the Ethereum claim, note that this PR does _not_ fulfill that design goal. 

In [these lines below][https://github.com/radicle-dev/radicle-upstream/pull/1712/files#diff-ab07879e13206d820c0c01e08b337b8378c01698c2af0565710bbef5a21d81abR78-R80], we only set the `Ethereum` claim if `Metadata.ethereum` is `Some`.

I have added this PR to `radicle-link` to make it possible for us to remove an extension: https://github.com/radicle-dev/radicle-link/pull/582

Once that's in, and the latest revision of link is integrated in the Upstream, lines 78-80 should become:

```
match self.ethereum {
    Some(ethereum) => payload.set_ext(EthereumClaimExtV1::from(ethereum))?,
    None => payload.remove_ext::<EthereumClaimExtV1>()?,
}
```

NunoAlexandre:
Having started the work integrating this endpoint, I have two things to add here:

1. We need to support the removal of an Ethereum claim
  See https://github.com/radicle-dev/radicle-upstream/pull/1712#discussion_r606136503. Without this, development in the upstream is pretty tedious as we have to keep resetting the state to test the attestation flow.

2. I'd advise to replace `impl Metadata {  pub fn update_payload(self, mut payload: PersonPayload) ... }` with a simple, standalone function:

    ``` rust
    pub fn update_payload(mut payload: PersonPayload, metadata: Metadata) -> Result<PersonPayload, ExtError> {
        payload.subject = Person {
            name: metadata.handle.into(),
        };
        match metadata.ethereum {
            Some(ethereum) => payload.set_ext(EthereumClaimExtV1::from(ethereum))?,
            None => payload.remove_ext(EthereumClaimExtV1::namespace())?,
        }
        Ok(payload)
        }
    }
    ```

   I don't see an added value of having this as an implementation detail of `Metadata`, and by doing (IMHO) the flow gets super confusing, making it read as if we are updating the payload of the metadata using an old payload.

    **now**:

      ``` rust
      let new_payload = metadata
        .update_payload(old_payload)
        .map_err(coco::state::Error::from)?;
     ```

    **after**:

      ``` rust
      let new_payload = update_payload(payload, metadata)
        .map_err(coco::state::Error::from)?;
     ```


@CodeSandwich as I am entering paternity leave, will hand this over back to you :baby: :yellow_heart:

proxy/api/src/identity.rs
CodeSandwich: I think that you've already fixed that issue in `b3423759b2c298a7249811cb0aeebaf3e7da81e4`: `Support removing Ethereum claim ext` :thinking: I've added a test verifying that behavior.

proxy/coco/src/state.rs
CodeSandwich: IIUC:
1. The app doesn't support multi-device, quorum signatures and won't know what to do with the knowledge that an update isn't validated yet
2. Updating git's `user.name` and `user.email` is pointless, nobody will ever see it, especially in the app

Maybe let's stick to an MVP, keep this PR as it is but remove the `set_default_owner` part?

CodeSandwich: Thank you for your reactions @xla and @FintanH, but could you elaborate? Do you have a better plan and what are the steps to implement it?

FintanH: Sure, was composing my thoughts. It's a bit disappointing to punt on helping the implementation of the protocol to both improve the protocol _and_ the upstream client. Is there a rush to get this in with the current semantics?

As I pointed out, it would be useful to bring to the surface when identities are out of date and need interaction. Kim pointed to some possible ways of looking at how to do this. A plan would be to evaluate these methods and implement them.

CodeSandwich: Unfortunately yes, there is a rush to push this out in some form to unblock the rest of the team 🙁 The mechanism for coping with out of date identities looks like a good material for a separate PR. IMO this feature won't be deeply affected when the HTTP API will be extended with information about freshness of an identity (which will be useful only when we add logic to app to handle it and will probably affect multiple HTTP endpoints) or a mechanism for resolution is added to coco.

FintanH: Ja, alles klar. `coco` is on its way over to `radicle-link` now anyway, so it'll be easier to add these things in.

proxy/coco/src/state.rs
```diff
@@ -103,40 +97,54 @@ where
...
+) -> Result<LocalIdentity, Error>
+where
+    P: TryInto<PersonPayload> + Send,
+    Error: From<P::Error>,
+{
+    let urn = default_owner(peer).await?.ok_or(Error::MissingOwner)?.urn();
+    let payload = payload.try_into()?;
+    peer.using_storage(move |store| person::update(store, &urn, None, payload, None))
+        .await??;
+    // Update the default user info in the config with the new payload data.
```
FintanH, line 142: Let's remove the below and then this looks good to go :+1:

FintanH (changes requested):

proxy/coco/src/state.rs
CodeSandwich: Fixed

FintanH (approved):

FintanH (approved):

FintanH (approved):

PR #1713, opened 2021-03-29
fix(ci): don’t clean free cache on mac agent

geigerzaehler:
As a follow-up to #1707 we don’t clean the cache folder on the macos agent. The agent does not provide the GNU version of the `df` utility and does not support the `--output` option.

### DIFF
--- a/.buildkite/run.sh
+++ b/.buildkite/run.sh
@@ -14,14 +14,16 @@ declare -r rust_target_cache="$CACHE_FOLDER/proxy-target"
 mkdir -p "$rust_target_cache"
 ln -s "${rust_target_cache}" ./target
 
-free_cache_space_kb=$(df --output=avail /cache | sed -n 2p)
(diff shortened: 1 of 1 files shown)


NunoAlexandre (approved):

NunoAlexandre:
Does this mean that if mac&cheese gets full that we still have to flush it manually?

geigerzaehler:
> Does this mean that if mac&cheese gets full that we still have to flush it manually?

I guess so. @rudolfs knows more probably.

rudolfs:
> Does this mean that if mac&cheese gets full that we still have to flush it manually?

Correct. 

```
buildkite@mac-n-cheese buildkite-cache % du -hs
 15G

buildkite@mac-n-cheese ~ % df -h
Filesystem      Size   Used  Avail Capacity iused      ifree %iused  Mounted on
/dev/disk1s5   233Gi   10Gi  142Gi     7%  488262 2447613058    0%   /
```

PR #1715, opened 2021-03-29
feat(ui): design improvements from merge request work

geigerzaehler:
This is a collection of design improvements and additions that are backported from the merge request work.

* Add `BranchBox` and `CompareBranches` components
* More specific selectors in the `Header/Back` component. This avoids changing the `fill` property of other SVGs
* Export `copy` from the `Copyable` component so it can be accessed from the outside.
* Drop `overflow: hidden` from the `List` component. This allows popups in the List component to flow outside. This requires us to set the border radius for the list items.
* Add a border to `SegmentedControl`.
* Make border color of transparent button transparent.

### DIFF
--- a/ui/DesignSystem/Component/BranchBox.svelte
+++ b/ui/DesignSystem/Component/BranchBox.svelte
@@ -1,0 +1,24 @@
+<script lang="typescript">
+  import { Icon } from "../Primitive";
+
+  export let name: string;
+  export let style: string = "";
+</script>
+
+<style>
+  .branch {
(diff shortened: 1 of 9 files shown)


NunoAlexandre:
Sorry, merged master thinking this was my PR.

NunoAlexandre (approved):

geigerzaehler:
> This makes sense, this is all I did, you didn't change anything that I can see would break it, right?

Yes, indeed. It only includes the fixes.

PR #1717, opened 2021-03-29
refactor(proxy): remove `branch` from commit info

geigerzaehler:
We remove the `branch` field from the `Commit` object returned by the `source/commit` endpoint. This field was unused and led to panics when a commit was not part of a branch. This case was triggered by the merge request work.

### DIFF
--- a/proxy/coco/src/source.rs
+++ b/proxy/coco/src/source.rs
@@ -157,8 +157,6 @@ pub struct Commit {
     pub stats: CommitStats,
     /// The changeset introduced by this commit.
     pub diff: diff::Diff,
-    /// The branch this commit belongs to.
-    pub branch: Branch,
 }
 
 impl Serialize for Commit {
@@ -170,7 +168,6 @@ impl Serialize for Commit {
         changeset.serialize_field("header", &self.header)?;
         changeset.serialize_field("stats", &self.stats)?;
         changeset.serialize_field("diff", &self.diff)?;
-        changeset.serialize_field("branch", &self.branch)?;
         changeset.end()
     }
 }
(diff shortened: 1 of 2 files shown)


NunoAlexandre (commented):
:broom:

FintanH:
Hmmm, would it make sense to return the `Object` that this commit _is_ associated with?

geigerzaehler:
> Hmmm, would it make sense to return the `Object` that this commit _is_ associated with?

Which `Object` do you mean?

FintanH:
> Which Object do you mean?

Sorry, should have elaborated more :) So a `Commit` can lie in the history of a `Branch`, a `Tag`, or both, right? When I said `Object` I was thinking https://docs.rs/git2/0.13.17/git2/struct.Object.html, but that includes a wider net of git objects.

I'm mostly posing the question: Do we want to delete this just because it's not used in Upstream or should we try make it more correct (for some defintion of more correct)? :)

geigerzaehler:
> So a `Commit` can lie in the history of a `Branch`, a `Tag`, or both, right?

Yes. It can also lie in the history of multiple branches and tags. It may also not have any associated branch or tag. (This may be the case with some special refs like the ones in the `notes` category.)

> I'm mostly posing the question: Do we want to delete this just because it's not used in Upstream or should we try make it more correct (for some defintion of more correct)? :)

It’s difficult to determine what would be correct if we have no need for this at the moment. (My best guess would be to return a list of all the refs the reference this object in their history. But this seems very heavy-handed and might be better suited for a separate endpoint.) I’d prefer to not add anything to keep the code lean and not invest time for something that we might not need or will have to change in the future.

FintanH (dismissed):
Happy with your reasoning. If needs be the multiple object approach can be spec'd out in radicle-surf

PR #1723, opened 2021-03-30
move identifier

FintanH:
I found the identifier module lying around in coco. I don't think it
needs to be there so I'm moving to api for now.

### DIFF
--- a/proxy/api/src/identifier.rs
+++ b/proxy/api/src/identifier.rs
@@ -1,18 +1,18 @@
-//! An `Identifier` is the combination of a user handle and the [`librad::git::Urn`] that
+//! An `Identifier` is the combination of a user handle and the [`coco::PeerId`] that
 //! identifies the user.
 
 use std::{fmt, str::FromStr};
 
 use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
 
-use librad::{identities::Person, peer};
+use coco::{identities::Person, PeerId};
 
 /// Errors captured when parsing a shareable identifier of the form `<handle>@<urn>`.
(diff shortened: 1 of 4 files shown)


xla (changes requested):
doc seems to be broken. Also does the identifier module need to be added to the api crate?

FintanH:
> Also does the identifier module need to be added to the api crate?

Can you elaborate on your question :) Not sure if you're saying if we should:
a. evaluate somewhere else it should go
b. it should stay in the coco realm
c. something else I'm not seeing

xla:
> Can you elaborate on your question :) Not sure if you're saying if we should:
> a. evaluate somewhere else it should go
> b. it should stay in the coco realm
> c. something else I'm not seeing

ha! None of the above, I was asking if you forgot to `git add` the actual identifier rs files, because they are not part of the change-set. :>

FintanH:
> ha! None of the above, I was asking if you forgot to git add the actual identifier rs files, because they are not part of the change-set. :>

Wait wut :thinking: It's moved here in this diff https://github.com/radicle-dev/radicle-upstream/pull/1723/files#diff-a771980c4efd0a18509d1fd44ae67ce45230afdc1ce081b09716242043288bd1

xla:
> Wait wut thinking It's moved here in this diff https://github.com/radicle-dev/radicle-upstream/pull/1723/files#diff-a771980c4efd0a18509d1fd44ae67ce45230afdc1ce081b09716242043288bd1

Disregard I can't read diffs. :facepalm:

xla (approved):
💱 🐙 🐬 🎼

xla (approved):
🏉 🔹 😃 🕵

PR #1726, opened 2021-03-31
feat(ui): add ClaimsContract

NunoAlexandre:
We add a wrapper class around the Radicle-Contracts/Claims contract, offering a method to `claim` a Radicle Identity root and a method to lookup a Radicle Identity root `claimed` by a given Ethereum address.

Once we have this, we can start wiring it up in the `LinkAddress` modal flow. I have tested that locally already :seedling: 

:information_source: to look up the root claimed by an address, we first need to filter the `Claimed` events produced by the contract that had said address as the tx author. We then can lookup said transaction, decode its input data and finally obtain the claimed root (if any).


ui/src/funding/contract.ts
```diff
@@ -27,31 +32,119 @@ const addresses = {
...
 // Get the address of the Pool Contract for the given environment
-export function poolAddress(environment: ethereum.Environment): string {
+export function daiTokenAddress(environment: ethereum.Environment): string {
   switch (environment) {
     case ethereum.Environment.Local:
-      return addresses.pool.local;
+      return addresses.dai.local;
     case ethereum.Environment.Ropsten:
-      return addresses.pool.ropsten;
+      return addresses.dai.ropsten;
```
NunoAlexandre, line 40: These changes are just coded being moved around to have it grouped by the type of contract.

ui/src/funding/contract.ts
```diff
@@ -27,31 +32,119 @@ const addresses = {
...
+
+    if (last === undefined) {
+      return undefined;
+    }
+
+    const tx = await last.getTransaction();
+    const decoder = new InputDataDecoder(this.abi);
+    const decoded = decoder.decodeData(tx.data);
+    const payload = decoded.inputs[1];
+    const root = ethereum.bytesToString(payload);
```
geigerzaehler, line 130: You could try the following
```suggestion
    const claimsFactory = new ClaimsFactory()
    const inputs = claimsFactory.interface.decodeFunctionData("claim", tx.data)
    const root = ethereum.bytesToString(inputs.payload);

```
This code may need some tweaking but it would eliminate the need to add the abi manually and eliminate the InputDataDecoder dependency.

NunoAlexandre: I did try that actually but `decodeFunctionData` does not return the inputs decoded and I am not sure how to proceed from therein. Not needing the abi manually, or being able to retrieve it programatically would absolutely be best.

NunoAlexandre: :eyes: I think I found a way

ui/src/ethereum.ts
```diff
@@ -71,3 +71,11 @@ export function toBaseUnit(n: ethers.BigNumber | Big): Big {
 export function fromBaseUnit(n: Big): ethers.BigNumber {
   return ethers.BigNumber.from(n.mul(TOKEN_DECIMALS).round().toString());
 }
+
+export function bytesToString(array: ethers.Bytes): string {
+  let string = "";
+  for (let i = 0; i < array.length; ++i) {
+    string += String.fromCharCode(array[i]);
```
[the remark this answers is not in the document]
NunoAlexandre: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1726/commits/aa834fcb3040434add01a9e8d4c12c8c67daaf0a.

ui/src/funding/contract.ts
NunoAlexandre: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1726/commits/009d3a42654c1527f2828ee544c242807781f8e1

package.json
```diff
@@ -75,6 +75,7 @@
     "@rollup/plugin-node-resolve": "^10.0.0",
     "@rollup/plugin-typescript": "^8.2.0",
     "@tsconfig/svelte": "^1.0.10",
+    "@typechain/ethers-v5": "^6.0.0",
```
geigerzaehler, line 78: This appears to be unused :thinking:

NunoAlexandre: weird :thinking: I first couldn't access the filters / events API of the contract and Igor pointed out that this dependency was necessary, and then I could get it working. But now I dropped it and it works without it. I must have confused something along the way. Dropped in https://github.com/radicle-dev/radicle-upstream/pull/1726/commits/fe2fd54091086cae7f2a1d0fe4ff519f12b01e2c.

geigerzaehler (approved):
:rocket:

NunoAlexandre:
Merged in c2f7ed2ccf5263c8ae9f58a52e2e2b00d78ce4df

PR #1729, opened 2021-03-31
fix(ui): don't leak store subscriptions

geigerzaehler:
Whenever we call `Store.subscribe` in a component we ensure that we unsubscribe when the component is destroyed. In `Screen/Project` this created particular issues where the current project page refreshed even if only a previously visited received an update.

### DIFF
--- a/ui/Screen/Funding/Pool/Outgoing/Support.svelte
+++ b/ui/Screen/Funding/Pool/Outgoing/Support.svelte
@@ -1,4 +1,5 @@
-<script lang="ts">
+<script lang="typescript">
+  import { onMount } from "svelte";
   import { Button, Icon, Input } from "../../../../DesignSystem/Primitive";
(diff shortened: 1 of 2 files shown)


NunoAlexandre (dismissed):
TIL on `svelte.onMount`

geigerzaehler:
> Instead of keeping these concerns in the component why not back the screen with a ts file as done in other places (project, source, etc.) and therefore not rely on component lifecycle?

This approach makes sense in some circumstances, but here I see some downsides. First, this would mean that we refresh the project screen store even if the project screen is not rendered, for example if we are on the settings page. Second, it embeds side-effects into the module which I try to avoid if possible. (While module side effects don’t create a problem in this instance they may be the source of problems, for example, when creating unit tests.)

On the other hand, I don’t see a reason to avoid component lifecycle hooks if they are don’t create too much complexity.

PR #1731, opened 2021-03-31
chore: use Metadata as payload to create identity

NunoAlexandre:
The REST endpoint that handles this request expects a `Metadata`, not a
type that contains a passphrase. Therefore, we drop `CreateInput` in
favour of a new `Metadata` struct.

Signed-off-by: Nuno Alexandre <hi@nunoalexandre.com>

### DIFF
--- a/ui/src/identity.ts
+++ b/ui/src/identity.ts
@@ -15,11 +15,13 @@ export interface Avatar {
   emoji: string;
 }
 
+export interface Metadata {
+  handle: string;
+}
+
 export interface Identity {
   avatarFallback: Avatar;
-  metadata: {
-    handle: string;
-  };
+  metadata: Metadata;
   peerId: PeerId;
   shareableEntityIdentifier: string;
   urn: Urn;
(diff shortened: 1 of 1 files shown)


FintanH:
Something I'm wondering at this stage is whether these inputs should just be the `PersonPayload` which can include the extensions such as the ethereum address. 

I think that's eventually what that endpoint will be expecting in the future anyway. 

Food for thought :)

ui/src/identity.ts
```diff
@@ -29,13 +31,8 @@ export interface Identity {
 const creationStore = remote.createStore<Identity>();
 export const store = creationStore.readable;
 
-interface CreateInput {
-  handle: string;
-  passphrase: string;
```
FintanH, line 34: I'm presuming the `passphrase` was un-needed? What the hell is it doing there in the first place?! :joy:

FintanH (approved):

ui/src/identity.ts
NunoAlexandre: that's right. I don't know :shrug: I tested the whole onboarding flow to be super sure. but if you follow the code you see that we first call the endpoint that creates a keystore and then we call the endpoint that creates the identity.

xla: It used to be a single call, was changed in multiple iterations of the session feature.

NunoAlexandre:
Merged in 773ca6b5e3383f90c2d8b35d19e43b487080ce19

PR #1732, opened 2021-03-31
chore(proxy): move coco::control

FintanH:
Moving the control module over to api since it wasn't used in coco
(apart from one set of tests easily refactored).

The other changes involved scoping code to only be included as the test
target and updating the imports.

One issue was that since control needs reflike and refpsec_pattern it
needs to include radicle-git-ext.


xla:
> One issue was that since control needs reflike and refpsec_pattern it
> needs to include radicle-git-ext.

Don't think that is an issue, as long as the proxy lives and we keep peeling away the more api will directly depend on crates in radicle-link.

xla (dismissed):
🚑 🚵 🐉 🉐

xla (dismissed):
✖️ 🆙 🍻 🛬

xla (approved):
🛃 👻 🙂 🈯️

Issue #1409, opened 2020-11-30
Flaky unit tests for process manager


geigerzaehler:
We have seen the tests in `native/process-manager.test.js` [fail](https://buildkite.com/monadic/radicle-upstream/builds/6848#434b4c40-940f-4db3-a3d4-27aa8e33ea59). The cause of this is probably output buffering of the process that is under test. The implementation should guard against this by splitting the process output into lines but this needs to be checked again.

geigerzaehler:
Fixed in #1734

PR #1734, opened 2021-03-31
fix(ui): wait for process streams to end

geigerzaehler:
We try to fix flaky tests (#1409) on MacOS by waiting for the output streams of the managed process to end before we return. This should ensure that all the buffered IO has been accounted for.

### DIFF
--- a/native/proxy-process-manager.ts
+++ b/native/proxy-process-manager.ts
@@ -49,10 +49,14 @@ export class ProxyProcessManager {
 
     this.childProcess = childProcess;
 
-    let stdoutBuf = "";
     // We know that `stdout` is set because of the `stdio` spawn options
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    childProcess.stdout!.on("data", (chunk: Buffer) => {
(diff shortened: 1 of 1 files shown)


rudolfs (dismissed):

native/proxy-process-manager.ts
```diff
@@ -49,10 +49,17 @@ export class ProxyProcessManager {
...
-    let stdoutBuf = "";
     // We know that `stdout` is set because of the `stdio` spawn options
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    childProcess.stdout!.on("data", (chunk: Buffer) => {
+    const stdout = childProcess.stdout!;
+
+    const stdoutEnded = new Promise<void>(resolve => {
+      stdout.on("end", () => resolve());
+      stdout.on("error", () => resolve());
+    });
```
NunoAlexandre, line 59: Since you repeat this whole block twice, you could extract it.

rudolfs (approved):

NunoAlexandre (approved):

PR #1735, opened 2021-03-31
fix(proxy): move git-helper

FintanH:
There was nothing coco specific about the git-helper, in fact it's only
important for the upstream client.


xla (approved):
🌦 💽 🎨 ™️

PR #1738, opened 2021-04-01
feat(ui): read and update `Ethereum` claim

NunoAlexandre:
We read the `Metadata.ethereum` field (defined in proxy already) and add API requests to update this claim, either by saving a new claim or by removing it.


CodeSandwich:
I've reviewed this PR and it seems that all its improvement are already present on `master`

PR #1741, opened 2021-04-02
config: increase the fetch limit for peeks

FintanH:
When peeking for some projects, they were _just_ over the 5KB limit,
i.e. ~6,000 bytes. This would prevent us from replicating projects when
we hit those errors.

We double the ceiling for our use-case.


FintanH:
See https://github.com/radicle-dev/radicle-link/issues/583 for some examples.

NunoAlexandre (approved):
Code looks good, but help me out here. Does this mean that we don't peek projects over 10kb in size? What happens to projects larger than that?

FintanH:
> Code looks good, but help me out here. Does this mean that we don't peek projects over 10kb in side? What happens to projects larger than that?

That's only for `rad/*` branches. They're just a small amount of metadata :) But turns out not 5k small :joy:

NunoAlexandre:
> > Code looks good, but help me out here. Does this mean that we don't peek projects over 10kb in side? What happens to projects larger than that?
> 
> That's only for `rad/*` branches. They're just a small amount of metadata :) But turns out not 5k small

Gotcha! thanks

proxy/coco/src/config.rs
```diff
@@ -63,7 +64,12 @@ where
             listen_addr,
             membership: net::protocol::membership::Params::default(),
             network: net::Network::default(),
-            replication: librad::git::replication::Config::default(),
+            replication: replication::Config {
+                fetch_limit: fetch::Limit {
+                    peek: fetch::FIVE_KB * 2,
```
xla, line 69: Doubling it seems arbitrary, is there some heuristic, analysis to understand the upper bounds for these peeks?

xla (changes requested):

proxy/coco/src/config.rs
FintanH: Ja... Good point. So I did a back of the envelope calculation by throwing all the JSON into a byte counter. It only went up to 1,644bytes :weary: 

Any other ideas of how to approximate this? fwiw, the initial 5KB is _also_ fairly arbitrary. Really there should be options to retry with a higher limit :/

proxy/coco/Cargo.toml
```diff
@@ -32,15 +32,15 @@ features = [ "json-value" ]
 
 [dependencies.librad]
 git = "https://github.com/radicle-dev/radicle-link.git"
-rev = "9467c364fa232f6563caf7ec123c43ade9c23284"
+rev = "211fb0dd7d9d3aee2bd71aa22800a5ed553475d2"
 
 [dependencies.radicle-git-ext]
 git = "https://github.com/radicle-dev/radicle-link.git"
-rev = "9467c364fa232f6563caf7ec123c43ade9c23284"
+rev = "211fb0dd7d9d3aee2bd71aa22800a5ed553475d2"
```
geigerzaehler, line 39: The `api` package also uses `radicle-git-ext`. To avoid having two versions of that crate we should update the `rev` there too.

FintanH: Thanks, noticed that too :facepalm: Artifact of trying to build the car while driving it :upside_down_face:

xla: Open heart surgery.

PR #1753, opened 2021-04-07
chore(proxy): daemon & source

FintanH:
This patch switches the dependency of the `api` crate to use
`radicle-daemon` and `radicle-source`. This means that the `coco` crate
was removed, since all functionality from there lives in one of those
two crates.

One consequence of this was splitting out a way to create a `Browser`
given a `Peer`, and so `browser.rs` was added to fill in this hole.

The `kv` crate was set to a stable version rather than a git URL.

All other changes were simply mechanical changes to updating module
imports and new usage of packages, i.e. `browser::using` and `kv`
removing `into_inner`.


xla (dismissed):
over9000

xla (approved):
⛴ 👺 🕜 🕯

geigerzaehler (approved):

PR #1754, opened 2021-04-07
style(ui): typography update

juliendonck:
- [x] remove unused fonts
- [x] allow users to select the system fonts


brandonhaslegs:
are these supposed to be in snake_case?
image

rudolfs:
> are these supposed to be in snake_case?

yep, in rust variables are snake_case, that's why we have a macro in place which converts them to camelCase for the front-end:

```
```

rudolfs (dismissed):

ui/src/settings.ts
```diff
@@ -31,6 +33,22 @@ interface Option<T> {
   title: string;
 }
 
+export enum Theme {
+  Dark = "dark",
+  Light = "light",
+  H4x0r = "h4x0r",
+}
```
geigerzaehler, line 40: This is already imported from `./proxy/settings.ts`. You can remove it here.

ui/src/settings.ts
```diff
@@ -31,6 +33,22 @@ interface Option<T> {
...
+
+export enum UIFont {
+  Inter = "inter",
+  System = "system",
+}
+
+export enum CodeFont {
+  SourceCode = "sourceCode",
+  System = "system",
+}
```
geigerzaehler, line 50: These definitions should be moved into `proxy/settings.ts` and you need to update the [`Appearance` schema](https://github.com/radicle-dev/radicle-upstream/blob/b8c4ad4afb781342c3446beebf57ee36e3f42b35/ui/src/proxy/settings.ts#L31-L36) to ensure we properly decode this. (Happy to help if you need support with that).

ui/src/settings.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1754/commits/1cfc90ec9a3d1ed94fabbc9431509d5efde31afc

ui/src/settings.ts
rudolfs: https://github.com/radicle-dev/radicle-upstream/pull/1754/commits/1cfc90ec9a3d1ed94fabbc9431509d5efde31afc

ui/src/settings.ts
juliendonck: thx

rudolfs:
Tested on Linux, it works:

https://user-images.githubusercontent.com/158411/114364908-31dfd480-9b7a-11eb-8485-432c4f406679.mov

rudolfs (dismissed):

geigerzaehler (dismissed):

rudolfs (approved):

PR #1756, opened 2021-04-07
refactor(ui): introduce proxy api client

geigerzaehler:
We introduce the `ui/src/proxy` module that provides checked, reusable access to the proxy HTTP. It’s intended to supersede `ui/src/api` in the future.

We gain the following advantages
* We separate stateful code from the stateless proxy API code (e.g. in `ui/src/session`)
* Response payloads are validated against TypeScript types with `zod`.
* We can reuse the proxy client in the cypress test commands.
* Provide an foundation to extend the proxy client with more APIs.

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -1,11 +1,17 @@
 import * as uuid from "uuid";
 import * as path from "path";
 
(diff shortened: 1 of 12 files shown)


rudolfs (dismissed):
Very nice!

PR #1758, opened 2021-04-08
Network diagnostics

alexjg:
Adds a network diagnostics screen - accessible via a `<ctrl>-i` shortcut - which shows the waiting room state and any events that change the state of the waiting room.


ui/Screen/NetworkDiagnostics/ConnectedPeers.svelte
```diff
@@ -0,0 +1,47 @@
...
+    width: 100%;
+    border-collapse: collapse;
+    border: 3px solid purple;
+  }
+
+</style>
+
+<Remote {store} let:data={{ status }}>
+  <div>
+    {#if status.type === StatusType.Online}
```
geigerzaehler, line 21: As mentioned in in https://github.com/radicle-dev/radicle-upstream/issues/1798 `Remote` disables type checking. `status` will be of type any and errors in accessing it’s fields will not be caught. I suggest we use a simple `if` statement. It has the same functionality but does type checking.

```suggestion
{#if $store.status === remote.Status.Success}
  <div>
    {#if $store.data.type === StatusType.Online}
```

ui/Screen/NetworkDiagnostics.svelte
```diff
@@ -0,0 +1,67 @@
+<script lang="typescript">
+  import Router from "svelte-spa-router";
+
+  import { store } from "../src/screen/network";
+  import * as path from "../src/path";
+
+  import ActionBar from "../DesignSystem/Component/ActionBar.svelte";
```
geigerzaehler, line 7: We can use absolute imports now.
```suggestion
  import ActionBar from "ui/DesignSystem/Component/ActionBar.svelte";
```
It would be great if you could make a quick pass and change it everywhere a path goes up a directory with `../`.

webpack.config.ts
```diff
@@ -91,6 +91,10 @@ function ui(_env: unknown, argv: Argv): webpack.Configuration {
           extensions: [".svelte", ".ts", ".js"],
         }),
       ],
+      alias: {
+        svelte: path.resolve("node_modules", "svelte"),
+      },
```
geigerzaehler, line 96: Could you explain why you added this? I don’t think it’s necessary.

ui/src/waitingRoom.ts
```diff
@@ -0,0 +1,88 @@
...
+        zod.literal("available"),
+        zod.literal("inProgress"),
+        zod.literal("failed"),
+      ])
+    ),
+  })
+);
+
+export type Event = zod.infer<typeof eventSchema>;
+export type RoomState = zod.infer<typeof roomStateSchema>;
```
geigerzaehler, line 88: With this approach we don’t declare the `Event` and `RoomState` types explicitly. This makes it harder for developers to understand their shape (we have to look at the schema) and it makes it harder to debug typechecking errors. I’d suggest declaring these types explicitly
```typescript
type Event =
  {
    type: "created"
    urn: string,
 } | {
    // ...
  }
```

To support developers we can use explicit annotations for the schema
```typescript
const eventSchem: zod.Schema<Event> = ...
```

This is also the pattern we’re currently using in the rest of the codebase.

ui/src/path.ts
```diff
@@ -25,3 +25,10 @@ export const projectSourceCommits = (urn: Urn): string =>
 export const project = projectSourceFiles;
 
 export const designSystemGuide = (): string => "/design-system-guide";
+
+export const networkDiagnosticsConnectedPeers = (): string =>
+  "/network-diagnostics/connected-peers";
+export const networkDiagnosticsEvents = (): string =>
+  "/network-diagnostics/events";
```
geigerzaehler, line 32: This path is unmatched in `NetworkDiagnostics.svelte`.

ui/src/screen/network.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const store = derived(
+  [localPeer.status],
+  ([status], set: (screen: remote.Data<Screen>) => void): void => {
+    if (status.status === remote.Status.Success) {
+      set({ status: remote.Status.Success, data: { status: status.data } });
+    }
+  },
+  { status: remote.Status.Loading }
+);
```
geigerzaehler, line 18: Could you explain why we need this derived store? It looks to me like we can use `localPeer.status` in the network screen directly since it contains the exact same information. This would allow us to eliminate this file.

ui/Screen/NetworkDiagnostics/WaitingRoom.svelte
```diff
@@ -0,0 +1,72 @@
...
+    WaitingRoomTransition,
+  } from "../../src/localPeer";
+
+  let lastEvent: WaitingRoomTransition | null = null;
+  let eventsWithoutTicks: WaitingRoomTransition[] = [];
+  const unsub = waitingRoomEventLog.subscribe(e => {
+    lastEvent = e[e.length - 1];
+    eventsWithoutTicks = e.filter(e => e.event.type !== "tick");
+  });
+  onDestroy(unsub);
```
geigerzaehler, line 17: Since we’re dealing with svelte stores we can write this more idiomatically and eliminate the `onDestroy`.
```suggestion
  $: lastEvent = $waitingRoomEventLog.slice(-1)[0] || undefined;
  $: eventsWithoutTicks = $waitingRoomEventLog.filter(
    e => e.event.type !== "tick"
  );
```

ui/Screen/NetworkDiagnostics.svelte
```diff
@@ -0,0 +1,67 @@
...
+  .title {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+  }
+
+</style>
+
+<SidebarLayout>
+  <Remote {store} let:data={{ status }}>
```
geigerzaehler, line 52: See my other comment on `Remote`

geigerzaehler (commented):
Great work! This should be super helpful :mag:

webpack.config.ts
alexjg: It's required because otherwise Webpack attempts to use the pre-bundled svelte components from `svelte-json-tree`, adding this key means that the svelte source is compiled by our svelte compiler.

geigerzaehler: Isn’t that what the `mainFields` key is for? Could you add comment with a reference so this isn’t removed accidentally?

alexjg:
This should be ready to take another look at. The only remaining irritation is that `svelte-json-tree` causes a bunch of warnings when bundling. This will be fixed when https://github.com/tanhauhau/svelte-json-tree/pull/10 lands.

geigerzaehler (approved):
Nice work! :fireworks:

juliendonck:
@brandonhaslegs and I might take a stab at styling this a bit. Should we push to this PR or branch of of this?

geigerzaehler:
> @brandonhaslegs and I might take a stab at styling this a bit. Should we push to this PR or branch of of this?

As discussed, we’ll move forward merging this PR and you can create a separate PR to update the styles.

PR #1762, opened 2021-04-09
Fix: show an error notification when a project is missing peers

rudolfs:
This PR prevents the UI from breaking when no peers are available.

When there is something wrong with replication, it can happen that there are no peers that can be shown in the UI, this triggers an exception in the Svelte renderer and prevents users from further interacting with the app.

Screenshot 2021-04-09 at 08 27 48

Instead we present the user an error notification with the Project ID.

Screenshot 2021-04-09 at 08 30 44

```
{
  "code": "RemoteStoreError",
  "message": "Project rad:git:hnrkdry38qedf63fh4wgj88yyhkn6w4o6x4uo is missing peers",
  "source": {
    "code": "UnknownException",
    "message": "Project rad:git:hnrkdry38qedf63fh4wgj88yyhkn6w4o6x4uo is missing peers",
    "stack": "Error: Project rad:git:hnrkdry38qedf63fh4wgj88yyhkn6w4o6x4uo is missing peers\n    at throwUnlessPeersPresent (webpack-internal:///./ui/src/screen/project.ts:226:15)\n    at eval (webpack-internal:///./ui/src/screen/project.ts:38:9)",
    "details": {
      "name": "Error"
    }
  }
}
```

Signed-off-by: Rūdolfs Ošiņš <rudolfs@osins.org>

### DIFF
--- a/ui/src/screen/project.ts
+++ b/ui/src/screen/project.ts
@@ -32,6 +32,7 @@ export const fetch = (projectUrn: Urn): void => {
     })
     .then(peers => {
       const peerSelection = filterPeers(peers);
+      throwUnlessPeersPresent(peerSelection, projectUrn);
       screenStore.success({
         peers,
         peerSelection,
@@ -63,10 +64,12 @@ export const fetchPeers = (): void => {
     project
       .fetchPeers(current.project.urn, request.signal)
       .then(peers => {
+        const filteredPeers = filterPeers(peers);
+        throwUnlessPeersPresent(filteredPeers, current.project.urn);
         screenStore.success({
           ...current,
           peers,
-          peerSelection: filterPeers(peers),
+          peerSelection: filteredPeers,
           requestInProgress: null,
         });
       })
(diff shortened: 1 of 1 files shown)


xla (approved):
Looks good, would it be possible to show an error state for the project itself instead of/additionally to the notification.

📬 🗨 🕑 🌶

rudolfs:
> Looks good, would it be possible to show an error state for the project itself instead of/additionally to the notification.

You mean like we do in the "Projects" tab, where we validate the projects in proxy and send separate lists of working/broken projects to the front-end and then show the working ones and notifications for the broken ones?

It's possible for sure, and we should do it eventually, but it's a bit more effort than this, because we'd have to move the peer filtering to proxy.

xla:
> You mean like we do in the "Projects" tab, where we validate the projects in proxy and send separate lists of working/broken projects to the front-end and then show the working ones and notifications for the broken ones?

More in this example we have a big empty screen which is the error state of the project view and could be used to contextual information and a back button.

rudolfs:
> More in this example we have a big empty screen which is the error state of the project view and could be used to contextual information and a back button.

That would also be possible. Although if we would re-think how to deal with errors to avoid pop-up notifications, I'd rather have the project/following list visually show the broken ones somehow distinctly and maybe show a tooltip on-hover with the reason of the breakage.

We should put some more thought into this, come up with a new strategy and apply it everywhere in the app at once.

PR #1765, opened 2021-04-09
fix(ui): parse ethereum field of identity in ui

geigerzaehler:
We fix a conflict between #1756 and #1712 and add the `ethereum` field for identities to the UI types and parse it properly.


rudolfs (approved):
💦🔥

PR #1766, opened 2021-04-09
fix(tests): wait for test project to be created

geigerzaehler:
We ensure that cypress waits for a test project to be created before it continues. Otherwise tests may fail.


rudolfs (approved):

xla (approved):
〰️ 🏁 🍵 🙇

PR #1776, opened 2021-04-13
style(ui): UI issue fixes

### LINKED ISSUE #1455: Change text from "README" to "files"
When a project is empty, it says "This project doesn't have a README yet." but it would be slightly more accurate to say "This project doesn't have any files yet."

_User feedback from **aspiers** via Matrix channel_

### LINKED ISSUE #1640: Unclear Following UX
After following the first project on Following Tab, I lost the ability to follow other projects from this Tab.

### LINKED ISSUE #1772: Add a style to disable contextual alternates
Want to add a style that's easy to add to certain elements that just disables contextual alternates. For example Ethereum addresses should have this style, otherwise they will always display as "0×31c8eacbffdd875c74b94b077895bd78cf1e64a3" instead of "0x31c8eacbffdd875c74b94b077895bd78cf1e64a3." Another place where it annoys me is in the new "H4x0r" theme, which displays as "H4×0r."

More context [here](https://twitter.com/brandonhaslegs/status/1377154110730903557).

juliendonck:
- [x] add search action to following list -> closes #1640 
- [x] add a class that disables contextual alternatives for the font-> closes #1772 
- [x] clean up project source empty states -> closes #1455

### DIFF
--- a/public/typography.css
+++ b/public/typography.css
@@ -144,3 +144,8 @@ p,
   text-transform: uppercase;
   letter-spacing: 0.05rem;
 }
+
+/* Disables contextual alternates */
+.typo-disable-calt {
+  font-feature-settings: "calt" 0;
+}
--- a/ui/DesignSystem/Component/SegmentedControl.svelte
+++ b/ui/DesignSystem/Component/SegmentedControl.svelte
@@ -61,7 +61,7 @@
 <div class="segmented-control">
   {#each options as option}
     <button
-      class="typo-semi-bold"
+      class="typo-semi-bold typo-disable-calt"
       class:active={option.value === currentlyActive}
       value={option.value}
       on:click={() => onClick(option)}>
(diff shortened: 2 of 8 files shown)


brandonhaslegs (approved):

public/typography.css
```diff
@@ -144,3 +144,8 @@ p,
   text-transform: uppercase;
   letter-spacing: 0.05rem;
 }
+
+/* Disables contextual alternates */
+.typo-disable-calt {
```
rudolfs, line 149: In which circumstances would we want this contextual alternate representation?
Should we not just disable it globally?

We set other font flags here: https://github.com/radicle-dev/radicle-upstream/blob/master/public/global.css#L13.

brandonhaslegs: Yeah I wish we could disable it globally only for certain characters, but that doesn't seem to be possible. Because the contextual alternates in Inter are really cool, I'd like to keep them around and just target certain strings where we specifically find a problem.

But… maybe we should just disable them globally since we have font-choosing now as a setting, and this style only applies to the Inter font.

🤔

juliendonck: I think this implementation makes sense. Cause Inter is still the default.

rudolfs (approved):

PR #1780, opened 2021-04-13
fix(ui): replace `path.active`

geigerzaehler:
We replace the incorrectly used `patch.active` function. Whenever this function was used it was called with a _rendered_ path like `/projects/rad:git:123` instead of a pattern. The implementation however interpreted that path as a pattern and passed it to `regexparam`.

This approach worked as long as no `?` was contained in the path passed to `path.active`. In my case I added a horizontal menu item where the `href` field looked like `/foo/bar?filter=true`. This `href` field is used as the `path` argument and the `?` is treated as a special character. This resulted in `path.active(href, href)` returning false.

The purpose of detecting whether a certain path is active can be more easily achieved by comparing the rendered path with `$location`. For the “loose” match mode supported by `path.active` we use `$location.startsWith`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,7 +180,6 @@
     "pure-svg-code": "^1.0.6",
     "qs": "^6.10.1",
     "radicle-contracts": "github:radicle-dev/radicle-contracts#commit=9fb23ac74a403b167a22513fc666779d871d292e",
-    "regexparam": "^1.3.0",
     "semver": "^7.3.5",
     "stream-browserify": "^3.0.0",
     "svelte-persistent-store": "^0.1.6",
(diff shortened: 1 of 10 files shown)


rudolfs (approved):
Would be nice to have some tests for this.

geigerzaehler:
> Would be nice to have some tests for this.

Do you mean to test the “active” state for all the menus and the sidebar or are you referring to some sort of unit tests?

rudolfs:
> Do you mean to test the “active” state for all the menus and the sidebar or are you referring to some sort of unit tests?

Yeah, just some cypress tests to check that the "active" state is correct in some places, don't even need to cover all of them.

geigerzaehler:
> Yeah, just some cypress tests to check that the "active" state is correct in some places, don't even need to cover all of them.

I’d prefer to refrain from adding tests for individual cases. Since there is no abstract logic every test can only guarantee that it works for this instance of the check and we would need two cases for every “active” check. This is also non-functional so I don’t think adding tests is worth it considering the added run time and maintenance.

Issue #1771, opened 2021-04-12
Search input is not cleared after a search


rudolfs:
Steps to reproduce:

1. Open the search modal via <kbd>⌘</kbd>+<kbd>p</kbd> or the sidebar "Search" icon.
2. Paste `rad:git:hnrk8mcu5xfwn58ubtcaax3y64fcokhnkihxo` into the search bar
3. Click "Follow" or hit <kbd>enter</kbd>
4. Either the project screen or the "Following" tab will be opened (depending whether you have the project replicated or not)
5. Open the search modal via <kbd>⌘</kbd>+<kbd>p</kbd> or the sidebar "Search" icon.

Expected: the search input to be empty.
Actual: the search input is pre-filled with the previous search request `rad:git:hnrk8mcu5xfwn58ubtcaax3y64fcokhnkihxo`

Regression of: #1652.

PR #1781, opened 2021-04-13
fix(ui): clear search input after every search

rudolfs:
Fixes https://github.com/radicle-dev/radicle-upstream/issues/1771.

### DIFF
--- a/cypress/integration/search.spec.ts
+++ b/cypress/integration/search.spec.ts
@@ -1,0 +1,105 @@
+import * as commands from "../support/commands";
+
+context("search", () => {
+  const projectId = "hnrkfr9g6gxymefc3hto37bgmq3eo86sfckky";
+
+  beforeEach(() => {
+    commands.resetProxyState();
+    commands.onboardUser("cloudhead");
+    commands.createProjectWithFixture();
+    cy.visit("./public/index.html");
+  });
+
+  context("when a project is not yet followed", () => {
+    it("follows the project by clicking the follow button", () => {
+      commands.pick("sidebar", "search").click();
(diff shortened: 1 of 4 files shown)


cypress/integration/search.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      // Make sure the search input is cleared after the search.
+      commands.pick("sidebar", "search").click();
+      commands.pick("search-input").should("have.value", "");
+    });
+
+    it("follows the project by pressing the [enter] hotkey", () => {
+      commands.pick("sidebar", "search").click();
+      commands.pasteInto(["search-input"], `rad:git:${projectId}`);
+      cy.wait(400);
```
geigerzaehler, line 30: Could you add a comment explaining why we need this?

ui/Modal/Search.svelte
```diff
@@ -53,7 +57,9 @@
     }
   };
   const follow = () => {
-    requestProject(value);
+    if ($urnValidation.status === ValidationStatus.Success) {
+      requestProject(value);
```
geigerzaehler, line 61: It’s probably worth adding a test for invalid URNs, either in this PR or as a follow-up

cypress/integration/search.spec.ts
rudolfs: That was actually a debug leftover, removed it here: 1f5137005d4dba1745956e587ce428d6733a3eef.

ui/Modal/Search.svelte
rudolfs: 045faac2fd0ed352b4a2e8d9c0b65f3e12655356

geigerzaehler (approved):
:chains:

PR #1784, opened 2021-04-13
feat(ui): improve error handling

geigerzaehler:
* Make `Error` a class instead of an interface. This means that we get stack traces whenever an error is created. This allows us to trace the sources of errors more easily. Since `Error` is also a subclass of the built-in JS `Error` we can also `throw` instances of `Error` now.

* Introduce `error.fromUnknown` as a replacement for `fromException`. `fromException` was used `try/catch` blocks to turn a built-in JS Error into our error. It is not guaranteed though that the error from a catch block is an instance of the built-in Error so we also need to cover that case.

* We call `event.preventDefault()` on the global handlers to avoid reporting errors to the console twice.

* We log error details and source to the console along with the stack trace.

### DIFF
--- a/ui/DesignSystem/Component/Remote.svelte
+++ b/ui/DesignSystem/Component/Remote.svelte
@@ -24,11 +24,7 @@
     !!noErrorSlotProvided &&
     !disableErrorLogging
   ) {
-    error.show({
-      code: error.Code.RemoteStoreError,
-      message: storeValue.error.message,
-      source: storeValue.error,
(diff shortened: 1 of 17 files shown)


rudolfs (dismissed):

geigerzaehler:
@rudolfs: Could you take another look? I improved on the API and fixed an issue in a file that was not using typescript.

rudolfs (dismissed):

PR #1788, opened 2021-04-14
chore(ui): show version and latest commit sha in dev mode

rudolfs:
Unfortunately [`app.getVersion()`](app.getVersion()
) only returns the app's version when it is bundled and run in production. In dev mode it returns Electron's version.

|before|after|
|------|-----|
| Screenshot 2021-04-14 at 17 39 13 | Screenshot 2021-04-14 at 17 39 39 |


xla (approved):
⚖ 👬 🏚 🔴

PR #1796, opened 2021-04-16
chore: update cypress to v7

### LINKED ISSUE #1752: Update to Cypress v7
We want to update to [Cypress v7](https://docs.cypress.io/guides/references/changelog#7-0-0). Locally, the tests with the new version pass and don’t require any migration. On buildkite however we see that the electron shell crashes with the following error:
> [25213:0406/153645.901261:WARNING:interprocess_frame_pool.cc(80)] Failed to allocate 8294400 bytes.

It looks like the containers don’t provide enough RAM for electron to run, but this needs further investigation.

geigerzaehler:
Since the crashes on Cypress v7 have been fixed (#1752) we can update it. I’ve revied the [changelog][1]. The only change affecting us is that uncaught exceptions will result in test failures. This caused issues when we reset the proxy at the start of a test while the app from the previous test was still loaded in the window. We fix this by unloading the app before we reset the proxy.

Fixes #1752

[1]: https://docs.cypress.io/guides/references/changelog#7-0-0

### DIFF
--- a/cypress/empty.html
+++ b/cypress/empty.html
@@ -1,0 +1,3 @@
+<!DOCTYPE html>
+<!-- we load this file to unload the app -->
+<body />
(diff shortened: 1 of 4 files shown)


rudolfs (approved):
🚀

PR #1799, opened 2021-04-16
refactor(ui): remove duplicate avatar module

geigerzaehler:
We remove `ui/src/avatar`. All items that were exported and used are also provided by `src/proxy/identity`.


rudolfs (dismissed):
🧹

PR #1804, opened 2021-04-19
Increase fetch limit on peeks to 5Mb

### LINKED ISSUE #1795: Reasonable peek limit when cloning
We are currently running into an issue where we are hitting the 10kb fetch limit when we clone a new project. This fetch limit exists to close a possible failure case where a misbehaving peer sends a large amount of data on response to a fetch, exhausting bandwidth. We are hitting this fetch limit on perfectly well behaved peers, which means that we must have misunderstood how much data a well behaved peer will need to send. I've been investigating this to figure out what a good upper bound might be.

When we clone a project we ask the remote to send us all the identities and signed refs it knows about for the project in question. We do this via the git protocol. The first thing that happens in this protocol is that the remote lists all the refs it knows about. In our case that means it will emit something like these four lines for each remote:

```
00ae2f5f1676d8b20f3e9adf94cc5c2f74e36ca597bd refs/namespaces/hnrkgbmhyd5xi7ihhzbw6szpcrfotdrzz4z9y/refs/remotes/hyb7bx5nouu4xatm9u7q7kqb8msnpntkrusfbau1tka78tzmn9bxb6/rad/id
00d5f704f847e451abefa16ac904e6461ba92f714d9d refs/namespaces/hnrkgbmhyd5xi7ihhzbw6szpcrfotdrzz4z9y/refs/remotes/hyb7bx5nouu4xatm9u7q7kqb8msnpntkrusfbau1tka78tzmn9bxb6/rad/ids/hnrk81xjkitazkdqdb5ecj7xkei6o459z1yeo
00b0c0ab52e482acb7d9874370fe22982c0baad9c1dd refs/namespaces/hnrkgbmhyd5xi7ihhzbw6szpcrfotdrzz4z9y/refs/remotes/hyb7bx5nouu4xatm9u7q7kqb8msnpntkrusfbau1tka78tzmn9bxb6/rad/self
00b7cb6ff2e68ce12c14dd9722622862088114df0075 refs/namespaces/hnrkgbmhyd5xi7ihhzbw6szpcrfotdrzz4z9y/refs/remotes/hyb7bx5nouu4xatm9u7q7kqb8msnpntkrusfbau1tka78tzmn9bxb6/rad/signed_refs
```

Each of these lines is around 200 characters so we're looking at ~800 bytes per remote.

Once we've done this we ask the server for all the identities and signed refs. For each identity we get the initial document plus one ref update document. This is the size of the git objects involved:

## Project identity creation:
- Identity doc: 300
- Delegations: 600
- Commit: 500
- tree: 100
- delegations tree: 100

Total: 1600

## Person identity creation:
- identity doc: 300
- commit: 500
- tree: 100

Total: 900

## Sigref update:
- commit: 400
- tree: 50
- updated json: 300

Total: 750

Both projects and persons are immediately updated when they are first created. That means that a project remote will be 2500b and a person remote around 1650 bytes. If we work with the project remote to get an upper bound. That means we need 2500b per remote. Let's imagine a very active project with 200 remotes, then it would have 500kb of git objects to download. This is not that amenable to the delta compression that happens in packfiles because these are all distinct objects, not updates to existing ones. However, git also gzip compresses this stuff, and based on a few experiments I've done in python gzip get's ~60% compression ratios on this data so we're looking at 300 kb of data

Incluing the advertisment lines, for 200 remotes we would be looking at 200 * 200 = 40kb.

In total then we might be looking at around 340kb as a reasonable upper bound.

alexjg:
Fixes #1795

### DIFF
--- a/proxy/coco/src/config.rs
+++ b/proxy/coco/src/config.rs
@@ -59,7 +59,7 @@ where
 {
     let mut repconf = replication::Config::default();
     // This number based on discussions in https://github.com/radicle-dev/radicle-upstream/issues/1795
-    repconf.fetch_limit.peek = 5000000;
+    repconf.fetch_limit.peek = 5_000_000;
     net::peer::Config {
         signer,
         protocol: net::protocol::Config {
--- a/proxy/coco/src/config.rs
+++ b/proxy/coco/src/config.rs
@@ -57,6 +57,9 @@ where
     S: Signer + Clone + Send + Sync + 'static,
     S::Error: std::error::Error + Send + Sync + 'static,
 {
+    let mut repconf = replication::Config::default();
+    // This number based on discussions in https://github.com/radicle-dev/radicle-upstream/issues/1795
+    repconf.fetch_limit.peek = 5000000;
     net::peer::Config {
         signer,
         protocol: net::protocol::Config {
@@ -64,7 +67,7 @@ where
             listen_addr,
             membership: net::protocol::membership::Params::default(),
             network: net::Network::default(),
-            replication: replication::Config::default(),
+            replication: repconf,
         },
         storage_pools: net::peer::PoolSizes::default(),
     }


rudolfs (dismissed):

rudolfs:
Superseded by https://github.com/radicle-dev/radicle-link/pull/638.

Issue #1800, opened 2021-04-16
Replace Svelte store in event handling code


geigerzaehler:
At the moment we are using Svelte stores and in particular `svelteStore.derived` to create objects that allow us to listen to events from the proxy ([example](https://github.com/radicle-dev/radicle-upstream/blob/9ef477dfb8419a66bbf6f4de4332d7f909c2383d/ui/src/localPeer.ts#L143)). Because we use stores we are already seeing the following problems

First, we need to be careful when subscribing to an “event” store. The subscriber callback is called with the last event when the subscription is created—even if that event occurred ages ago. Second, writing filters to select a subset of events requires us to use `null` as a [placeholder](https://github.com/radicle-dev/radicle-upstream/blob/9ef477dfb8419a66bbf6f4de4332d7f909c2383d/ui/src/localPeer.ts#L146-L156) for the absence of events which leads to bloated and brittle code.

The root cause is that a Svelte store does not exhibit the behavior we want. It always has a current value. Instead we want a _stream_.

To address the issue I suggest we use a reactive programming library that provides first class reactive primitives including streams. For TypeScript the most widely used candidates are [`RxJS`](https://github.com/reactivex/rxjs) and [`BaconJS`](https://github.com/baconjs/bacon.js). While BaconJS is used less, it is also smaller, provides a more concise (and less Java-like) API and has better documentation. I suggest we adopt it.

cloudhead:
I'm curious why we use stores that way currently, the example you link (ProjectUpdated) seems like an abuse of the store concept. Instead we should have (in this case) a ProjectState type that can be updated with new state. The app can then always show the latest state. This gets around the problem of "null" events. Basically svelte is already functional-reactive.

Do you have examples of where an event emitter/stream is actually what we want?

cloudhead:
Reflecting on this more after a chat with xla, I guess the place where you envision this fitting is the ??? in:

proxy -> sse events -> ??? -> state stores -> ui

Where ??? is currently the "event store".

Is that right?

geigerzaehler:
> proxy -> sse events -> ??? -> state stores -> ui
> 
> Where ??? is currently the "event store".
> 
> Is that right?

Yes, that’s correct. 

> Instead we should have (in this case) a ProjectState type that can be updated with new state.

If I understand this correctly this would invert the dependencies compared to the current approach. At the moment we have code in `src/Screen/Project.svelte` that listens for project events and updates the store
[...]

PR #1805, opened 2021-04-20
Use baconjs for event handling

geigerzaehler:
A proof of concept that shows how a stream primitive from BaconJS can simplify the event logic.

### DIFF
--- a/ui/Screen/Project.svelte
+++ b/ui/Screen/Project.svelte
@@ -47,10 +47,7 @@
   };
 
   onMount(() => {
-    localPeer.projectEvents.subscribe(event => {
-      if (!event) {
-        return;
-      }
+    localPeer.projectEvents.onValue(event => {
       if (event.urn === urn) {
         refresh();
       }
--- a/package.json
+++ b/package.json
@@ -171,6 +171,7 @@
     "@types/qs": "^6.9.6",
     "@types/uuid": "^8.3.0",
     "@walletconnect/client": "^1.4.1",
+    "baconjs": "^3.0.17",
     "big.js": "^6.0.3",
     "browserify": "^17.0.0",
     "crypto-browserify": "^3.12.0",
--- a/ui/src/bacon.ts
+++ b/ui/src/bacon.ts
@@ -1,0 +1,32 @@
+// Facade for the `baconjs` package with extra functions.
+//
+// Re-exports everything from `baconjs`.
+
+import * as bacon from "baconjs";
+export * from "baconjs";
+
+// Creates a new stream that calls `f` on items on the original stream
+// and emits the result if it is not undefined.
+//
+// This is equivalent to
+// ```
+// stream.filter((x) => f(x) !== undefined).map(f)
+// ```
+// but type checks and calls `f` only once.
+export function filterMap<S, T>(
+  stream: bacon.EventStream<S>,
+  f: (s: S) => T | undefined
+): bacon.EventStream<T> {
+  return stream.transform((event, sink) => {
+    if (bacon.hasValue(event)) {
+      const next = f(event.value);
+      if (next === undefined) {
+        return;
+      } else {
+        return sink(new bacon.Next(next));
+      }
+    } else {
+      return sink((event as unknown) as bacon.Event<T>);
+    }
+  });
+}
--- a/ui/src/localPeer.ts
+++ b/ui/src/localPeer.ts
@@ -1,4 +1,3 @@
-import { derived, writable, Readable } from "svelte/store";
 import { push } from "svelte-spa-router";
 import * as zod from "zod";
 
@@ -10,6 +9,7 @@ import * as remote from "./remote";
 import * as session from "./session";
 import type * as urn from "./urn";
 import * as error from "./error";
+import * as bacon from "./bacon";
 
 // TYPES
 export enum StatusType {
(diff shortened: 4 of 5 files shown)


ui/Screen/Project.svelte
```diff
@@ -45,10 +45,7 @@
...
 
   onMount(() => {
-    localPeer.projectEvents.subscribe(event => {
-      if (!event) {
-        return;
-      }
+    localPeer.projectEvents.onValue(event => {
       if (event.urn === urn) {
         refresh();
       }
```
cloudhead, line 51: I'm still unclear why we need to be reading events directly in UI components. Why can't this screen render content from a store directly (reactively), without having to know about how that store is updated? There's nothing special about this component, it should just render the state of the store :thinking:

geigerzaehler: You’re correct. This code should be moved to `ui/src/screen/project.ts`. As mentioned in #1800 I deliberately excluded this change to focus on the improvements to the event filtering.

cloudhead: Ah gotcha. Then we're on the same page :+1:

ui/Screen/Project.svelte
```diff
@@ -45,10 +45,7 @@
   };
 
   onMount(() => {
-    localPeer.projectEvents.subscribe(event => {
-      if (!event) {
-        return;
-      }
```
xla, line 51: Just for my personal understanding, this new dependency, and the extra infra around it, is pulled in to remove these three lines? The point has been made before that the approach of manually subscribing to stores in components is subpar in general. This screen is already backed by a store which can be equipped to do the subscription to meaningful events and only update the store accordingly.

While the Svelte store might look like an abuse of the mechanism it was modeled with the single original use-case of powering the connectivity indicator. If there is a need for streams than it would look more natural to use the `EventSource` which natively offers dispatch semantics. It can't be exposed raw as it needs parsing and such, but that gives an opportunity to rethink the public API of `localPeer`, which could easily expose a `subscribe` function allowing to pass a list of events that should be emitted. This truly would bring utility to the rest of the codebase imho.

geigerzaehler: > this new dependency, and the extra infra around it, is pulled in to remove these three lines?

It is also pulled in to remove these three lines but not only. The “null” pattern can be eliminated everywhere where events are handled. In addition we remove the “latest event” issue where any subscription to the events immediately calls the subscriber with the last event—even if that event was triggered a long time ago.

> The point has been made before that the approach of manually subscribing to stores in components is subpar in general. This screen is already backed by a store which can be equipped to do the subscription to meaningful events and only update the store accordingly.

As I mentioned in my reply to @cloudhead, this PR is not to fix where we subscribe to events but how to provide better facilities to manage events. I agree with you and @cloudhead that the subscription should happen closer to the store.

> If there is a need for streams than it would look more natural to use the `EventSource` which natively offers dispatch semantics.

Could you elaborate on what you mean by dispatch semantics? I see a couple of problems with `EventSource` that make in unergonomic to use. 1. It does not support combinators that allow for say easy filtering. 2. The `addEventListener`/`removeEventListener` API creates a mismatch with sveltes `subscribe() => unsubscribe` pattern. 3. The EventSource is recreated throughout the life time of the app which makes using it directly cumbersome.

> gives an opportunity to rethink the public API of `localPeer`, which could easily expose a `subscribe` function allowing to pass a list of events that should be emitted.

This is definitely worth exploring. But I could be harder to create properly typed APIs for such a generic `subscribe` function.

geigerzaehler:
@rudolfs, @MeBrei: Could you take a look at this? It’s ready now.

rudolfs (approved):
🚀

xla (commented):
It's unclear to me why the introduction of more technology with their own set of APIs and therefore added complexity is chosen over reworking the `localPeer` semantics. A similar example of that overall tendency is the introduction of a library to give the types used to expect data from the proxy more safety - there I would have expected to go to the source of truth of these types and use them to generate the TS equivalents, which would not be subject to indirection, potential divergence and generally a much higher maintenance cost mid/long-term. The complexities in this codebase should stem from guaranteeing a consistent and polished user experience, yet the amount of moving pieces a contributor has to onboard to in the tooling and inside the actual runtime code is disproportionate imho.

geigerzaehler:
> It's unclear to me why the introduction of more technology with their own set of APIs and therefore added complexity is chosen over reworking the `localPeer` semantics.

This change does rework the semantics of `localPeer`. As outlined in #1800 we’re getting rid of the problem where subscribing to one of the stores immediately calls the subscriber with the last event even if that event was emitted a long time ago. This behavior lead to unintended consequences like unnecessary refreshs. As mentioned in #1800 there are more advantages to replacing `svelte/store` for events.

You’re rightThis PR does not address the organization of the event code. I wanted to keep the PR focused and we will handle this separately.

>  A similar example of that overall tendency is the introduction of a library to give the types used to expect data from the proxy more safety - there I would have expected to go to the source of truth of these types and use them to generate the TS equivalents, which would not be subject to indirection, potential divergence and generally a much higher maintenance cost mid/long-term.

This is a different discussion. Happy to have a PR that generates typescript types from Rust types and eliminates schema checks.

>  yet the amount of moving pieces a contributor has to onboard to in the tooling and inside the actual runtime code is disproportionate imho.

This is a fair concern that I share and one of the reasons I chose a library that is well documented. In this specific case I’d argue that we have eliminated complexity especially for new contributors. When writing event code developers are not required to repeat the `if (!event) return null` pattern and they don’t need to create custom workarounds for old events triggering subscribers. (The latter is especially problematic since the problem only manifests itself under certain conditions in runtime and requires that knowledge.)

PR #1808, opened 2021-04-20
refactor(ui): move onboarding screen to TypeScript

geigerzaehler:


ui/Screen/Onboarding/EnterPassphrase.svelte
```diff
@@ -61,14 +67,16 @@
...
     passphraseValidation = getValidationState("passphrase", validations);
     repeatedPassphraseValidation = getValidationState(
       "repeatedPassphrase",
+      // @ts-expect-error: `validations` is guaranteed to be a validations object
       validations
     );
   };
 
-  $: validate(passphrase, repeatedPassphrase);
+  $: validate();
```
rudolfs, line 79: Is this a typo?

rudolfs: Because if you don't list the variables the reactive statement won't trigger.

geigerzaehler: `validate` does not have any arguments. The original code was an oversight.

rudolfs: I guess we're just abusing the parameters. Not sure it'll work, but maybe we can rewrite it to something like this, to make it more explicit?
```
$: passphrase; repeatedPassphrase; validate();
```

rudolfs: 3bedfee09431dd6318216b48882c2c8a2d681122 -> this also works 👍

PR #1811, opened 2021-04-21
refactor(ui): explicit grouping of commits

geigerzaehler:
In preparation for patches we move the grouping of commits out of `source.fetchCommits` and closer to the UI. This allows us to also obtain ungrouped commits from `fetchCommits` which is needed for patches.

### DIFF
--- a/ui/DesignSystem/Component/SourceBrowser/History.svelte
+++ b/ui/DesignSystem/Component/SourceBrowser/History.svelte
@@ -1,11 +1,14 @@
 <script lang="typescript">
   import { createEventDispatcher } from "svelte";
 
-  import type { CommitHeader, CommitsHistory } from "../../../src/source";
+  import type {
+    CommitHeader,
+    GroupedCommitsHistory,
+  } from "../../../src/source";
 
   import CommitTeaser from "./CommitTeaser.svelte";
 
-  export let history: CommitsHistory;
+  export let history: GroupedCommitsHistory;
 
   const dispatch = createEventDispatcher();
   const onSelect = (commit: CommitHeader) => {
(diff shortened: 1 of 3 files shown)


ui/DesignSystem/Component/SourceBrowser/History.svelte
```diff
@@ -1,11 +1,14 @@
 <script lang="typescript">
   import { createEventDispatcher } from "svelte";
 
-  import type { CommitHeader, CommitsHistory } from "../../../src/source";
+  import type {
+    CommitHeader,
+    GrouppedCommitsHistory,
```
MeBrei, line 6:
```suggestion
    GroupedCommitsHistory,
```

and the same/similar typos in both `source.ts` files

MeBrei (changes requested):
Looks good! Just a typo...

ui/src/screen/project/source.ts
```diff
@@ -92,11 +92,12 @@ export const fetch = async (project: Project, peer: User): Promise<void> => {
       source.fetchCommits(project.urn, peer.peerId, selectedRevision),
       fetchTreeRoot(selectedRevision),
     ]);
+    const grouppedHistory = source.groupCommitHistory(history);
 
     screenStore.success({
       code: writable<Code>(root),
-      history,
-      menuItems: menuItems(project, history),
+      history: grouppedHistory,
```
MeBrei, line 99: Sorry for not making that explicit, but the typo is also in this variable and further down in `grouppedCommits`

geigerzaehler: Forgot to grep for lowercase too :see_no_evil:

MeBrei (approved):

PR #1812, opened 2021-04-21
fix(api): only set eth claim extension when provided

geigerzaehler:
We only set the Ethereum claim extension when it is included in the metadata. Before we would call `payload.set_ext(None)` which would result in a `null` value in the identity document. This in turned failed to parse in `user.payload().get_ext::<EthereumClaimExtV1>()`.

### DIFF
--- a/proxy/api/src/identity.rs
+++ b/proxy/api/src/identity.rs
@@ -78,8 +78,10 @@ pub fn update_payload(
     payload.subject = Person {
         name: metadata.handle.into(),
     };
-    let ethereum_claim = metadata.ethereum.map(EthereumClaimExtV1::from);
-    payload.with_ext(ethereum_claim)
+    if let Some(ethereum) = metadata.ethereum {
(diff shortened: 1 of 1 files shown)


CodeSandwich:
Good catch! But this issue has been solved for good on Link level: https://github.com/radicle-dev/radicle-link/pull/620

geigerzaehler:
Good to know! Do you think it still makes sense to merge this? Without knowing about the extension behavior in detail I feel the code makes it more explicit what we want to achieve.

CodeSandwich:
I have an exactly opposite feeling and I wouldn't merge this PR :D This was a workaround for a hidden behavior, which has been fixed. From proxy perspective the intention of setting an extension to `None` to me seems equivalent to explicitly removing it. Adding special logic for every single extension to avoid using `None` feels like noise. But I don't have strong feeling about it, if you think that it will prevent a future misunderstanding, I'll accept this PR.

PR #1815, opened 2021-04-21
refactor(test): extract empty project creation for networking tests

MeBrei:
In order to speed up the networking tests and limit the ui interactions to the ones we actually want to test, a new empty project creation command is implemented. This will become useful as well when we extend replication tests to other areas of the app, eg #1682

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -79,9 +79,7 @@ export const createProjectWithFixture = (
 
 export const createEmptyProject = (
   name: string = "new-project",
-  description: string,
   path: string,
-  defaultBranch: string = "trunk",
   localhost: number = 17246
 ): Cypress.Chainable<void> =>
   requestOk({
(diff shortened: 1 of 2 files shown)


cypress/support/commands.ts
```diff
@@ -77,6 +77,30 @@ export const createProjectWithFixture = (
   });
 };
 
+export const createEmptyProject = (
+  name: string = "new-project",
+  description: string,
+  path: string,
+  defaultBranch: string = "trunk",
+  localhost: number = 17246
+): Cypress.Chainable<void> =>
```
geigerzaehler, line 86: It’s probably also a good idea to only have parameters that actually make a difference in the test code. We can remove `defaultBranch` and `description` and just hard-code it.

geigerzaehler (approved):

MeBrei:
Closed by 6e195ebe7a59afe381f2f74aae11dfe0a32d2c55

PR #1817, opened 2021-04-22
chore(ci): get rid of warnings about fonts

rudolfs:
Removes the following warnings from the CI console output:
```
[4936:0422/071855.175313:INFO:CONSOLE(0)] "A preload for 'http://localhost:44433/public/fonts/Inter-Bold.otf' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.", source: http://localhost:44433/public/index.html (0)
```

Screenshot 2021-04-22 at 09 52 09

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -7,36 +7,47 @@
     <title>Radicle Upstream</title>
 
     <!-- Prevents font load related flickering on boot -->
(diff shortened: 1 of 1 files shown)


MeBrei (approved):
⭐

PR #1819, opened 2021-04-23
refactor(ui): tsconfig path mapping for absolute imports

### LINKED ISSUE #734: Absolute paths
Wouldn't it be nice if we could type 
```
import * as thing from "src/thing";
```
instead of 
```
import * as thing from "../../../../src/thing";
```
?
[Potential approach](https://dev.to/sjafferi/absolute-paths-in-svelte-488c)

geigerzaehler:
We use TypeScript path mapping to allow absolute imports of the `ui`
folder. I’ve converted some files to the new style to showcase this.

Fixes #734

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     "ts-jest": "^26.5.4",
     "ts-loader": "^8.1.0",
     "ts-node": "^9.1.1",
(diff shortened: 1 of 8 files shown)


rudolfs (approved):
This is awesome!

MeBrei (approved):
⭐

PR #1820, opened 2021-04-23
fix(ui): validate event payload and fix fields

geigerzaehler:
We validate the event payloads from the proxy so that they match the types we provide. This also requires us to rename the `peer` field of `ProjectUpdated` to `provider`.

### DIFF
--- a/ui/src/localPeer.ts
+++ b/ui/src/localPeer.ts
@@ -1,5 +1,6 @@
 import { derived, writable, Readable } from "svelte/store";
 import { push } from "svelte-spa-router";
+import * as zod from "zod";
 
 import type * as identity from "./identity";
 import * as config from "./config";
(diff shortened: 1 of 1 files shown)


MeBrei (approved):

ui/src/localPeer.ts
```diff
@@ -53,7 +74,7 @@ enum EventType {
 
 interface ProjectUpdated {
   type: EventType.ProjectUpdated;
-  peer: identity.PeerId;
+  provider: identity.PeerId;
```
rudolfs, line 77: Was this never used in the code?

geigerzaehler: Yes, it was never used.

rudolfs: Glad we have zod now :)

PR #1821, opened 2021-04-23
Reactive patches list

MeBrei:
We subscribe to the `ProjectUpdated` events to show new patches in the list view without reloading the page. The patches counter in the horizontal menu bar is also updated. Note that the event is currently only triggered for changes coming from other remotes, so updating the list when the user is pushing a patch themselves has to be done in a separate PR.

The following is an example of a `ProjectUpdated` event triggered by a remote pushing a new patch `feature-1`
```
{
    provider: "hyys5xxamaizo7nxbhrxsjkpwh385mabea8uxy3zn1kzmks5jqemj4" 
    type: "projectUpdated" 
    urn: "rad:git:hnrkbpzranjs7kbn45zxfyborhibcugrbyeio/tags/radicle-patch/feature-1"
}
```

### DIFF
--- a/ui/Screen/Project/Source.svelte
+++ b/ui/Screen/Project/Source.svelte
@@ -1,4 +1,5 @@
 <script lang="typescript">
+  import { onDestroy } from "svelte";
   import Router, { location } from "svelte-spa-router";
   import * as router from "svelte-spa-router/wrap";
 
@@ -11,6 +12,7 @@
   import type { Project, User } from "../../src/project";
   import {
     fetch,
+    watchPatchUpdates,
     selectPath,
     selectRevision,
     store,
@@ -100,6 +102,8 @@
     selectRevision(revision);
   };
 
+  const unwatchPatchUpdates = watchPatchUpdates();
+  onDestroy(unwatchPatchUpdates);
   $: fetch(project, selectedPeer);
 </script>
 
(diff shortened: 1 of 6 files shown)


ui/Screen/Project/Source.svelte
```diff
@@ -100,6 +102,7 @@
     selectRevision(revision);
   };
 
+  onDestroy(watchPatchUpdates());
```
rudolfs, line 105: Should this not be registered in `onMount`? And `onDestroy` have the event listener stop listening?

geigerzaehler: > Should this not be registered in `onMount`? And `onDestroy` have the event listener stop listening?

This is what the code does. When it is executed (i.e. on mount) `watchPatchUpdates()` is called. The function it returns is then registered as a callback on destroy.

rudolfs: Hm, got it. Is it possible to make this a bit more explicit? From looking at the code it's not very obvious.

geigerzaehler: Yes, I think it makes sense to be more explicit here: https://github.com/radicle-dev/radicle-upstream/pull/1821/commits/6d7859f1a1f626a2d89f5e272e15e7073f9f017b

rudolfs (dismissed):

PR #1823, opened 2021-04-23
fix(ui): search enter key follow only when fetched

geigerzaehler:
Hitting “enter” in the search dialog only triggers a follow if we know whether the project has been found or not. It doesn’t trigger when we’re still fetching the search result.

We also adapt the tests so that we wait for the search result to be present.


rudolfs (approved):
Works for me! 👌

PR #1824, opened 2021-04-26
refactor(ui): add project proxy client

geigerzaehler:
We add an `Client` to interact with the project API of the proxy and use this throughout the codebase.


rudolfs (approved):
So good! 💯

PR #1825, opened 2021-04-26
fix(tests): eliminate unnecessary waits

geigerzaehler:
We eliminate unnecessary calls to `cy.wait()`. In most places the call was just copied over from other tests and there was never a need for the wait. (The comment attached to `cy.wait()` makes this clear.)

In the only place where it really was required to run an assertion we replace `.then()` with `.should()`. This will rerun the assertion if it failed until it times out. This is more efficient than waiting.


rudolfs (approved):

geigerzaehler:
@rudolfs: I’ve added some more fixes for bugs that were uncovered by removing the `wait`s. Everything passes now

rudolfs (dismissed):

PR #1834, opened 2021-04-28
fix(proxy): qualify announcement

FintanH:
We need to qualify our announcements with the `refs/` prefix so that
they don't get mistintrepeted on the other side of the wire.


rudolfs (approved):

xla (approved):
Strings are hard!

🐙 💵 🎑 🌶

FintanH:
> Strings are hard!

So are types!

PR #1839, opened 2021-04-29
fix(ui): more robust session context

geigerzaehler:
We make acquiring the unsealed session from the context more robust. This was inspired by [comments on #1828][1].

* We stop hardcoding the `session` string into the context
* Instead of using type assertion which don’t fail at runtime we throw an error. If a session is not unsealed this will make it easier to locate the cause.

[1]: https://github.com/radicle-dev/radicle-upstream/pull/1828#discussion_r620346071

### DIFF
--- a/ui/DesignSystem/Component/SidebarLayout.svelte
+++ b/ui/DesignSystem/Component/SidebarLayout.svelte
@@ -1,14 +1,12 @@
 <script lang="typescript">
-  import { getContext } from "svelte";
-
-  import type { UnsealedSession } from "../../src/session";
+  import * as sess from "../../src/session";
 
   import Sidebar from "./Sidebar.svelte";
 
   export let dataCy: string = "";
   export let style: string = "";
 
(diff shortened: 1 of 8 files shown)


rudolfs (approved):
Generally looks good.
Just one question: why do we use a context instead of a svelte store for the session?

geigerzaehler:
> Just one question: why do we use a context instead of a svelte store for the session?

I think both could work but there might be unknown pitfalls. I didn’t replace context with the store to avoid changes to the behavior.

rudolfs:
> I think both could work but there might be unknown pitfalls. I didn’t replace context with the store to avoid changes to the behavior.

Got it, makes sense.

If we touch this code in the future it might be worth thinking about it. The main reason is that contexts are not reactive -- you'd only get the context to propagate a change of its data if you re-render the component where it was set, which in most situations is not very useful. I usually avoid contexts for this reason.

PR #1842, opened 2021-04-29
feat(ui): Peer status indicator

juliendonck:
Indicator to that shows if you are in sync with the maintainer's version and instructions on how to pull that version into your main branch:

Screenshot 2021-05-11 at 09 53 05


ui/DesignSystem/Component/SourceBrowser/FileView/PeerStatus.svelte
```diff
@@ -0,0 +1,28 @@
+<script>
+  import { Avatar, Icon } from "../../../Primitive";
```
geigerzaehler, line 2: We can use absolute imports now.
```suggestion
  import { Avatar, Icon } from "ui/DesignSystem/Primitive";
```

geigerzaehler:
We don’t have an API yet that provides this information. I have  two questions to make sure I understand this correctly.
* Do we show this for every branch? I.e. branches created by the user and branches created by remote peers?
* Do we always compare against the default branch of the maintainer?
I’ll create an issue to implement the API.

juliendonck:
> We don’t have an API yet that provides this information. I have two questions to make sure I understand this correctly.
> 
>     * Do we show this for every branch? I.e. branches created by the user and branches created by remote peers?

I think eventually I'd like to show a dropdown in the overlay where you can select any peer and it shows you the difference. For now I think just showing the maintainer main branch would be enough.

> 
>     * Do we always compare against the default branch of the maintainer?

I think initially it should just be to merge from main -> main. In the end the instructions should teach you how to merge other ones in if you want to.

> 
> 
> I’ll create an issue to implement the API.

geigerzaehler:
It looks like the API described in https://github.com/radicle-dev/radicle-surf/issues/174 should be enough to implement this. It looks to me like the change in `radicle-surf` is a bit more involved because the APIs don’t quite fit this use case but it  should be doable.

juliendonck:
Closed for now until we have bandwidth to work on this.

PR #1843, opened 2021-04-30
fix(dev): implement ethereum dev node in typescript

geigerzaehler:
We implement the script to start the ethereum dev node in typescript with the following improvements
* We wait for the dev node to be ready before deploying the contracts
* We equip the dev account with funds only once
* We eliminate the flaky process management from the shell script
* We provide a better error message when the dev account address cannot be read


rudolfs (dismissed):
utACK

scripts/ethereum-dev-node.ts
```diff
@@ -0,0 +1,81 @@
...
+import { deployAll } from "radicle-contracts";
+import * as ethers from "ethers";
+import waitOn from "wait-on";
+
+const ethAccountFile = "sandbox/.local-eth-account";
+
+main().catch(e => {
+  console.error(e);
+  process.exit(1);
+});
```
rudolfs, line 14: Do we still need to do this?

CodeSandwich: Probably to kill all worker threads? :shrug:

scripts/ethereum-dev-node.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  const provider = new ethers.providers.JsonRpcProvider(
+    "http://localhost:8545"
+  );
+  const signer = provider.getSigner(0);
+  const txCount = await signer.getTransactionCount();
+  if (txCount !== 0) {
+    throw new Error(
+      "Deployer account has non-zero transaction count. You need to reset your chain"
```
CodeSandwich, line 50: Is it possible for Genache to randomly load an old state on startup? If so, maybe we should always reset it first?

CodeSandwich (dismissed):

scripts/ethereum-dev-node.ts
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env -S yarn node -r ts-node/register/transpile-only
```
CodeSandwich, line 1: Cool trick! Is it better than having a `yarn` command?

geigerzaehler: I prefer this because you can also execute the script directly from the terminal. Curiously this is also faster to start up than having a yarn script. Finally, we don’t need to maintain a script entry for script and this avoid breaking things when scripts are moved.

scripts/ethereum-dev-node.ts
geigerzaehler: If we don’t do this an “UnhandledPromiseRejectionWarning” will be emitted which does not kill the process. This is the behavior of Node v14 which we are using in `.tool-version` and CI. Once we adopt Node v15 we can get rid of this.

scripts/ethereum-dev-node.ts
geigerzaehler: Good point. Ganache uses persistent state if the `--db` flag is provided. We don’t do this so we can remove this check.

CodeSandwich (approved):
:+1:

scripts/ethereum-dev-node.ts
```diff
@@ -0,0 +1,75 @@
...
+        ethers.BigNumber.from(10).pow(tokenDecimals)
+      )
+    )
+  ).wait();
+
+  console.log(`\nRad token deployed at ${contracts.rad.address.toLowerCase()}`);
+  console.log(`ENS deployed at ${contracts.ens.address.toLowerCase()}`);
+  console.log(
+    `Eth Pool deployed at ${contracts.ethPool.address.toLowerCase()}`
+  );
```
NunoAlexandre, line 66: We don't use this one anymore so maybe we can just drop it to avoid noise when going through the logs.

NunoAlexandre (approved):

rudolfs (dismissed):

NunoAlexandre (approved):

rudolfs (approved):

PR #1844, opened 2021-04-30
refactor(ui): project requests client

geigerzaehler:
Extract code related to project requests into project API client.


rudolfs (dismissed):
Tested ACK

PR #1848, opened 2021-05-03
feat(ui): improve error reporting for parse failures

geigerzaehler:
We improve the reporting of errors when parsing the proxy response fails.

* We make the `message` and `stack` properties of `Error` enumberable so that they are part of the error payload that is copied to the clipboard.
* We wrap `zod.Schema.parse` so that we can attach information about the requests to the error that is thrown.


rudolfs (dismissed):
utACK

PR #1854, opened 2021-05-04
fix(ui): properly handle failed project

geigerzaehler:
* Fix the schema required by the `projects/failed` endpoint.
* Don’t crash the whole app when fetching the list of failed projects fails.


rudolfs (approved):
Tested ACK

PR #1860, opened 2021-05-05
fix(dev): don’t cache latest release information

geigerzaehler:
We forbid caching the latest release document so that app users are informed immediately about new releases.

We also use `execa` instead of `child_process.spawnSync` to run commands. `execa` will throw if the command fails.


rudolfs (approved):
utACK

PR #1862, opened 2021-05-06
chore(proxy): clean up api crate

xla:
Assortment of cleanup commits in preparation of the great flippening.


rudolfs (approved):
🧹 💨

proxy/api/src/http.rs
```diff
@@ -41,7 +40,6 @@ pub fn api(
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
     let test = ctx.test();
 
-    let avatar_filter = path("avatars").and(avatar::get_filter());
```
rudolfs, line 44: I think I'll need the `avatars` endpoint for https://github.com/radicle-dev/radicle-upstream/pull/1813 after all.

xla: Generally we should reduce the dependence on the proxy going forward, esp. for non-essential features. If we need avatars for ETH addresses there should be readily available solutions employed in wallets like MetaMask.

FintanH: In other words, `radicle-link`/`daemon` won't be maintaining an endpoint for avatars :)

rudolfs: </3

FintanH: :'(

xla: :\

xla: :/

xla: :>

xla: @rudolfs This concern is addressed since radicle-dev/radicle-avatar#5 landed, is that assumption correct?

rudolfs: Yeah, we don't need the avatar endpoint anymore.

rudolfs (approved):

geigerzaehler (approved):

PR #1867, opened 2021-05-07
deps: update radicle-contracts

geigerzaehler:


package.json
```diff
@@ -180,7 +180,7 @@
     "mnemonist": "^0.38.3",
     "pure-svg-code": "^1.0.6",
     "qs": "^6.10.1",
-    "radicle-contracts": "github:radicle-dev/radicle-contracts#commit=9fb23ac74a403b167a22513fc666779d871d292e",
```
rudolfs, line 183: Why was it on that specific commit before?

package.json
```diff
@@ -180,7 +180,7 @@
...
+    "radicle-contracts": "github:radicle-dev/radicle-contracts#head=master",
```
CodeSandwich, line 183: What are the advantages of pinning the dependency to `master`? IIUC this isn't a dependency which needs to be upgraded regularly with `yarn upgrade`, there are no security concerns. Any upgrades will be usually due to changes in the contracts' APIs, which may need adjustments in upstream. `yarn upgrade` may become harder to use if it will be risking fetching breaking changes from `radicle-contracts`. What we could do is create a git tag dependency, it'll be nicer to look at, but functionally equivalent to a hash.

package.json
```diff
@@ -180,7 +180,7 @@
     "mnemonist": "^0.38.3",
     "pure-svg-code": "^1.0.6",
     "qs": "^6.10.1",
-    "radicle-contracts": "github:radicle-dev/radicle-contracts#commit=9fb23ac74a403b167a22513fc666779d871d292e",
+    "radicle-contracts": "github:radicle-dev/radicle-contracts#commit=f401f4f1222cbee48197c5268d81ee2cb92b6421",
```
rudolfs, line 183: Have there been API changes in the contracts since the last bump that we need to integrate in Upstream now?

package.json
geigerzaehler: Excellent point. I changed it back to a commit. I don’t think there’s a need to create a tag. That would just create overhead.

CodeSandwich (approved):

package.json
geigerzaehler: There have been additions but no breaking changes.

rudolfs (approved):

Issue #1778, opened 2021-04-13
Navigation breaks after refresh


juliendonck:
When going to a full screen modal and refreshing the app, the routing breaks and I can't leave the screen by pressing the `X` or by typing the shortcut again. 

Only workaround atm is to go to the settings screen with the shortcut `cmd + ,` and then you can continue using the app

geigerzaehler:
> When going to a full screen modal and refreshing the app,

I guess you’re referring to the design system modal. I was only able to reproduce the issue with that dialog. The problem seems to be caused by an [issue in Electron](https://github.com/electron/electron/issues/16306) which does not retain the History API information after a refresh.

juliendonck:
It's also when you go to a commit detail page and then refresh. The `.pop()` doesn't know where to pop to anymore.

rudolfs:
We'll fix this in https://github.com/radicle-dev/radicle-upstream/pull/1876.

Issue #1522, opened 2021-01-13
Move from using dispatch to pass functions as arguments


NunoAlexandre:
For context: https://github.com/radicle-dev/radicle-upstream/pull/975#issuecomment-759353567 + https://github.com/radicle-dev/radicle-upstream/pull/975#issuecomment-759391371

NunoAlexandre:
Duplicate of https://github.com/radicle-dev/radicle-upstream/issues/1521. Github gave me a 404 when creating the first :shrug:

PR #1876, opened 2021-05-10
refactor(ui): roll our own router

rudolfs:
Using `svelte-spa-router` got more painful over the time because the architecture it imposed on our app was hard to work with.

Working on [orgs](https://github.com/radicle-dev/radicle-upstream/pull/1813) I also realised that I need to have a way to perform data fetching before a routing event, which was possible with `wrap` and route conditions from the spa router, but it didn't work reliably in a sub-route setup.

This PR replaces `svelte-spa-router` with our own router implementation and hopefully addresses all the shortcomings.

Benefits of the new approach:
  - everything is typed which should make it harder to run into runtime bugs
  - props can be passed directly to the component via `push()`
  - we have full control on how we serialize/deserialize and map routes to URLs
  - we have more control on when things get rendered
  - no more warnings about unknown `params` props

TBD:
  - [x] fix pop() breakage after a refresh (in design system guide and commit back button) https://github.com/radicle-dev/radicle-upstream/issues/1778 -> https://github.com/radicle-dev/radicle-upstream/pull/1876/commits/cc5d2afcbe1563515124a48e7f807dc9188d5a95


ui/DesignSystem/Component/HorizontalMenu/MenuItem.svelte
```diff
@@ -1,14 +1,22 @@
...
+  import { createEventDispatcher } from "svelte";
+
+  export let tab: State;
   export let dataCy: string;
   export let icon: typeof SvelteComponent;
   export let title: string;
   export let active: boolean;
   export let counter: number | undefined;
 
+  const dispatch = createEventDispatcher();
```
geigerzaehler, line 14: As mentioned in #1522 it’s preferable to use props instead of events.

ui/src/router.ts
```diff
@@ -0,0 +1,34 @@
...
+const emptyState = { component: null, props: {} };
+
+export const push = (newState: State): void => {
+  const oldHistory = svelteStore.get(writableHistory);
+  writableHistory.set([...oldHistory, newState]);
+};
+
+export const pop = (): void => {
+  const oldHistory = svelteStore.get(writableHistory);
+  writableHistory.set(oldHistory.slice(0, -1));
```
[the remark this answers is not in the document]
rudolfs: 8b09bc1e534c8c112dacf47536f34beb875c66b9

ui/Screen/Project/Source.svelte
```diff
@@ -123,5 +144,13 @@
...
-  <Router {routes} />
+  {#if activeTab === 'files'}
+    <FilesTab />
+  {:else if activeTab === 'commits'}
+    <CommitsTab />
+  {:else if activeTab === 'commit'}
+    {#if commitHash}
+      <CommitTab {commitHash} />
+    {/if}
+  {/if}
```
geigerzaehler, line 155: I think we’re missing some cases here

rudolfs: I think it was because the HorizontalMenu had all the tab types in it previously.
Now each screen has its own tab type which lives in the router module.

ui/Screen/UserProfile/Projects.svelte
```diff
@@ -1,23 +1,27 @@
 <script lang="typescript">
-  import { push } from "svelte-spa-router";
+  import { push } from "ui/src/router";
```
geigerzaehler, line 2: How do you feel about using qualified imports consistently?
```suggestion
  import * as router from "ui/src/router";
```
In the code `router.push()` make’s it more explicit what it is doing.

ui/Screen/Project/Source.svelte
```diff
@@ -121,5 +145,13 @@
...
 
-  <Router {routes} />
+  {#if activeTab === "files"}
+    <FilesTab />
+  {:else if activeTab === "commits"}
+    <CommitsTab />
+  {:else if activeTab === "commit"}
+    {#if commitHash}
+      <CommitTab {commitHash} />
+    {/if}
```
geigerzaehler, line 155: It looks to me like `commit` is not an actual tab. We don’t include it in the tab list and we only show it when a commit hash is present. We could just rely on the presence of `commit` hash to overide the `activeTab`.

Ideally we’d encode this fully in the route type:
```typescript
interface ProjectRoute {
  type: "project"
  urn: string,
  view: 
    { type: "files" }
    | { type: "commits" }
    | { type: "commit", commitHash: string }
}

File: ui/src/router.ts [hunk 6]
```diff
@@ -0,0 +1,59 @@
...
+    }
+  | { type: "settings" };
+
+const persistedState: Route | null = window.history.state;
+
+const writableHistory: svelteStore.Writable<Route[]> = svelteStore.writable(
+  persistedState === null ? [] : [persistedState]
+);
+
+const serializeRoute = (route: Route) => {
```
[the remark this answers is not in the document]
rudolfs: da34a6c48869ac52f93c2eddde0c8ab789dea05e

ui/src/router.ts
```diff
@@ -0,0 +1,59 @@
...
+  ) {
+    subRoute = `/${route.activeTab}`;
+  }
+
+  return `#/${route.type}${subRoute}`;
+};
+
+export const push = (newRoute: Route): void => {
+  writableHistory.update(history => [...history, newRoute]);
+  window.history.pushState(newRoute, "", serializeRoute(newRoute));
```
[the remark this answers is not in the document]
rudolfs: 0eb5833b5bf8b12075c9320641020acfcf04c5f4

ui/App.svelte
```diff
@@ -100,12 +80,27 @@
...
+  {:else if $activeRouteStore.type === "profile"}
+    <Profile activeTab={$activeRouteStore.activeTab} />
+  {:else if $activeRouteStore.type === "project"}
+    <Project
+      activeTab={$activeRouteStore.activeTab}
+      urn={$activeRouteStore.urn}
+      commitHash={$activeRouteStore.commitHash} />
+  {:else if $activeRouteStore.type === "settings"}
+    <Settings />
+  {/if}
```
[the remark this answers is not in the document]
rudolfs: e631f1b7a106d8cb447c73df9bd48ee01143447c

ui/src/router.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const pop = (): void => {
+  writableHistory.update(history => history.slice(0, -1));
+  window.history.back();
+};
+
+export const activeRouteStore: svelteStore.Readable<Route> =
+  svelteStore.derived(writableHistory, state => {
+    if (state.length === 0) {
+      return <Route>{ type: "empty" };
```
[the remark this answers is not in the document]
rudolfs: e631f1b7a106d8cb447c73df9bd48ee01143447c

ui/src/router.ts
```diff
@@ -0,0 +1,59 @@
...
+  | { type: "userProfile"; activeTab: UserProfileTab; urn: string }
+  | {
+      type: "project";
+      activeTab: ProjectTab;
+      urn: string;
+      commitHash: string | null;
+    }
+  | { type: "settings" };
+
+const persistedState: Route | null = window.history.state;
```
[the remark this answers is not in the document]
rudolfs: I believe the History API persists its state only within the current session. If I'd stop the app and start it again, it would start with an empty history. I don't see a problem with how it is at the moment.

ui/src/router.ts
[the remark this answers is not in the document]
rudolfs: I trust MDN: https://developer.mozilla.org/en-US/docs/Web/API/History_API#browser_compatibility.

ui/Screen/Project/Source.svelte
rudolfs: 771b3d0915c64fc52cddad8fe420a27735ae49f7

geigerzaehler (dismissed):
Great work! This will make our lives much easier. :surfing_man:

ui/Screen/Wallet.svelte
```diff
@@ -23,40 +21,43 @@
...
   import { Icon } from "ui/DesignSystem/Primitive";
 
-  const topbarMenuItems = [
-    {
-      icon: Icon.Transactions,
-      title: "Transactions",
-      href: path.walletTransactions(),
-    },
-  ];
+  export let activeTab: router.WalletTab = "transactions";
```
geigerzaehler, line 27: There’s no need for a default value. `activeTab` is always part of the route object.

rudolfs: d6db3de1ce76d9f68c26258d62f72a94a2cf0db9

geigerzaehler (approved):
:ship:

geigerzaehler (approved):

PR #1879, opened 2021-05-10
refactor(ui): add project peers endpoint to proxy client

geigerzaehler:
Add project peers endpoint to proxy client.


geigerzaehler:
I’ve fixed the failures. Could you take another look @rudolfs?

rudolfs (approved):

PR #1888, opened 2021-05-12
fix(deps): update radicle-contracts to have stricter APIs

geigerzaehler:
We update `radicle-contracts` to include https://github.com/radicle-dev/radicle-contracts/pull/141. This makes the TypeScript bindings to the contracts stricter and prevents misuse of those contracts.


rudolfs (approved):

geigerzaehler:
Will be updated by #1872

PR #1894, opened 2021-05-18
chore(deps): update prettier-plugin-svelte

geigerzaehler:
We update `prettier-plugin-svelte` from v1 to v2. To keep the format
changes to a minimum we set two options for the plugin.


rudolfs (approved):
✨

PR #1895, opened 2021-05-19
Point at patched libgit

alexjg:
This is necessary to use a patched version of libgit which respects the force flag on refspecs. Without this we can lose data when replicating with an out of date peer.

I had  to bump the rust-toolchain which triggered a few new lints and changed the name of some older ones so I've had to make some minor changes to the code.


FintanH:
Does the toolchain need to be on par? :thinking: Or has it run out of space?

alexjg:
Looks more like a run-out-of-space error

alexjg:
I believe this is now failing due to the changes in the server sent events which are currently merged into master on link and will be fixed by #1758 , I'll get on to fixing the issues in that PR then we can revisit this.

geigerzaehler (dismissed):

xla (approved):
🎀 🔣 💤 👸

alexjg:
Ugh, code signing, I'll rebase and force push with a signature, at which point I think this will need a bunch more approvals.

geigerzaehler (approved):

geigerzaehler:
Looks like the proxy [panics](https://buildkite.com/monadic/radicle-upstream/builds/9053#96101db2-d72d-4937-a58e-2baacca8192e/6345-6370) on CI.

alexjg:
I _think_ that's not a problem, it's just the `StreamDiscovery` shutting down. It happens for every test but most of them pass. I'm trying to reproduce the problem locally right now.

PR #1897, opened 2021-05-19
refactor: clean up remote store APIs

geigerzaehler:
* Remove unused `chain()`
* Eliminate `filter` argument from `fetch`. This can be more easily accomplished on the caller side.


xla (approved):
⚱ 🌳 ⏰ 💀

PR #1900, opened 2021-05-21
feat(ui): Wallet update

### LINKED ISSUE #1885: Make attestation an optional part of onboarding
Add the attestation steps to the main onboarding flow when a user first opens the app.

juliendonck:
Working on porting over the wallet from the txcenter + profile to it's own tab

image

- closes #1885


ui/App.svelte
```diff
@@ -102,10 +102,6 @@
 <NotificationFaucet />
 <Theme />
 
-{#if isExperimental && sessionIsUnsealed && $location !== path.designSystemGuide()}
-  <TransactionCenter />
```
geigerzaehler, line 106: Could you also remove the corresponding `src/App/TransactionCenter.svelte` file?

ui/Screen/Wallet/Transactions.svelte
```diff
@@ -0,0 +1,64 @@
...
+  import { store as transactions, getShortMonth } from "ui/src/transaction";
+
+  import TxList from "ui/DesignSystem/Component/Wallet/Transactions/TxList.svelte";
+
+  $: pendingTxs = $transactions.filter(
+    tx => tx.status === "Awaiting inclusion"
+  );
+  $: rejectedTxs = $transactions.filter(tx => tx.status === "Rejected");
+
+  $: txMonthSections = $transactions.reduce((res, cur) => {
```
geigerzaehler, line 11: In general, using `reduce` is a good idea. But in this case an imperative loop has some advantages. It’s easier to typecheck, easier to understand, we don’t rely on some implicit ordering of the transactions, and it’s more performant. The latter is particularly important when dealing with lists because they quickly grow in size. I’d propose the following
```typescript
const groupTxs = (txs: Tx[]) => {
  const sections: Array<{ key: string; title: string; items: Tx[] }> = [];
  // Sort from newest to oldest
  txs.sort((a, b) => b.date - a.date);
  for (const tx of txs) {
    const txDate = new Date(tx.date);
    const txMonth = txDate.getMonth();
    const txYear = txDate.getFullYear();
    const key = `${txYear}-${txMonth}`;
    const currentSection = sections[sections.length - 1];
    if (currentSection && currentSection.key === key) {
      currentSection.items.push(tx);
    } else {
      const title = `${getShortMonth(txDate)} ${txYear}`;
      sections.push({
        key,
        title,
        items: [tx],
      });
    }
  }
  return sections;
};

$: txMonthSections = groupTxs($transactions);
```

ui/Screen/Wallet/Pool.svelte
```diff
@@ -1,17 +1,17 @@
...
-  import { TxKind, ongoing } from "../../src/transaction";
-  import { store as txs } from "../../src/transaction";
+  import * as modal from "ui/src/modal";
+  import * as fundingPool from "ui/src/funding/pool";
+  import type { Pool } from "ui/src/funding/pool";
+  import { TxKind, ongoing, store as txs } from "ui/src/transaction";
+  import { store } from "ui/src/wallet";
 
-  export let pool: Pool;
+  let pool: Pool;
```
geigerzaehler, line 14: You can remove this line. `pool` will be defined below.

ui/Screen/Onboarding/Success.svelte
```diff
@@ -50,9 +61,16 @@
...
+    <div class="buttons">
+      <Button
+        variant="transparent"
+        dataCy="go-to-profile-button"
+        on:click={profile}>
+        Go to profile
+      </Button>
+      <Button dataCy="go-to-wallet-button" on:click={wallet}>
+        Set up your wallet
+      </Button>
```
geigerzaehler, line 73: The wallet is only enabled under the dev environment. Shouldn’t we guard it here, too?

ui/DesignSystem/Component/Wallet/Transactions/TxListItem.svelte
```diff
@@ -0,0 +1,62 @@
+<script lang="typescript">
+  import type { Tx } from "ui/src/transaction";
+  import { Icon } from "ui/DesignSystem/Primitive";
+  import { Label } from "ui/DesignSystem/Component";
+  import { getDay, getShortMonth } from "ui/src/transaction";
+
+  export let tx: Tx;
+
+  const price = 0.0043;
```
geigerzaehler, line 9: It will take a bit more effort to get the gas price for transactions. I suggest we remove this for now and add it once we have it. When we merge this we can create a follow-up issue.

ui/src/transaction.ts
```diff
@@ -395,6 +395,29 @@ const monthNames = [
   "December",
 ];
 
+export function getDay(date: Date): string {
+  return `${date.getUTCDate()}`;
+}
```
geigerzaehler, line 400: This is to small of a function to have. I’d prefer to inline it. Plus, this should by `date.getDate` to display it in local time.

juliendonck: Does that make sense since the getShortMonth is right below and both of these are always used in conjunction. It seems to me that either we move both or none.

ui/DesignSystem/Component/Label.svelte
```diff
@@ -1,7 +1,7 @@
-<script>
-  export let title = null;
-  export let color = null;
-  export let style = null;
+<script lang="typescript">
+  export let title: string = "";
+  export let color: string = "";
```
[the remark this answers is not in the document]
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1900/commits/f1d52d43e452a8873acccd5b49b802e7549862a3

ui/Screen/Wallet.svelte
```diff
@@ -0,0 +1,120 @@
+<script lang="ts">
```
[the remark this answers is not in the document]
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1900/commits/ca0a945c7b376947bd43ae265436e00924f17dc0

ui/App.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1900/commits/78d594b8dde49394cc843f6f1a8b69c0faccfe65

ui/Screen/Wallet/Pool.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1900/commits/e4975da1e70e9305b7e16438684e13bef45ef15f

ui/Screen/Onboarding/Success.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1900/commits/b0e1ae3f4abdb9e210ec37314a80be6b9d95ce18

ui/DesignSystem/Component/Wallet/Transactions/TxListItem.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1900/commits/908bc169723b3f22c21ad24edb4e589f7efe6fb2

ui/DesignSystem/Component/Wallet/Transactions/TxListItem.svelte
```diff
@@ -0,0 +1,62 @@
...
+  }
+</style>
+
+<div class="container" on:click>
+  <div class="left">
+    <div class="date">
+      <h5>{getShortMonth(new Date(tx.date))}</h5>
+      <p class="typo-text-bold">{getDay(new Date(tx.date))}</p>
+    </div>
+    <svelte:component this={Icon[txIcon(tx)]} />
```
juliendonck, line 55: @geigerzaehler @rudolfs how do I typescript this `{Icon[txIcon(tx)]}` thing?

geigerzaehler: You can do this by having `txIcon` return the actual icon instead of a string identifying the icon.

```ts
import type { SvelteComponent } from "svelte";

export function txIcon(tx: Tx): typeof SvelteComponent {
  switch (tx.kind) {
    case TxKind.ClaimRadicleIdentity:
      return Icon.Registered;
    // ...
  }
}
```

ui/src/transaction.ts
geigerzaehler: `getShortMonth` is used in two places so it makes sense to have this here. The whole setup is not ideal and we should start to use a library that deals with dates and time at some point. Until we get there we should try to avoid adding more abstraction and indirection. This is why I’d prefer to inline it.

juliendonck: kk

ui/DesignSystem/Component/Wallet/Transactions/TxListItem.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1900/commits/533e08baeb0e6e03c45f213979c8f96ab9bd9ccd

ui/src/transaction.ts
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1900/commits/533e08baeb0e6e03c45f213979c8f96ab9bd9ccd

geigerzaehler:
There are still three typescript errors logged on CI that need fixing. Otherwise this looks good to go.

geigerzaehler (approved):

rudolfs (approved):

PR #1907, opened 2021-05-26
style(ui): add basic padding to network diagnostic screen

juliendonck:
Just added some basic padding to the screen, still left it full width.

@rudolfs is there a way to not show the shortcut?

### DIFF
--- a/ui/Screen/NetworkDiagnostics.svelte
+++ b/ui/Screen/NetworkDiagnostics.svelte
@@ -37,6 +37,12 @@
 </script>
 
 <style>
+  .container {
+    margin: 0 auto;
+    min-width: var(--content-min-width);
+    padding: 0 var(--content-padding);
+  }
+
   .title {
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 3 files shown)


geigerzaehler (approved):

geigerzaehler:
> @rudolfs is there a way to not show the shortcut?

The shortcut for the inspector is only shown in development mode. If you download and run the app it is hidden.

rudolfs:
> The shortcut for the inspector is only shown in development mode. If you download and run the app it is hidden.

That is correct. The shortcut is not shown in the dialog and it is not possible to reach the diagnostics screen in a production app. I wonder how useful this feature is then? If I understood it correctly the intent was to be able to diagnose problems in the field when they actually happen, and not only in dev mode, maybe @alexjg can comment?

alexjg:
Yeah my intention was for it to actually be part of the deployed application so that when things don't work as expected there is a way for users to see at least some of what's going on under the hood without having to understand the librad codebase. In general I'm a fan of exposing the internal state of things to users who are interested.

However, it's also extremely useful just in development as it allows us to use Upstream as a simple debugging tool for network problems, so it's still useful even if only in development mode.

geigerzaehler:
> it is not possible to reach the diagnostics screen in a production app

It is possible. The shortcut also works in the production app. It’s just that the shortcut is not shown.

rudolfs:
> Yeah my intention was for it to actually be part of the deployed application so that when things don't work as expected there is a way for users to see at least some of what's going on under the hood without having to understand the librad codebase.

In this case we have to put it in the publicly accessible shortcut list:
https://github.com/radicle-dev/radicle-upstream/blob/master/ui/src/hotkeys.ts#L36

> @rudolfs is there a way to not show the shortcut?

Currently there's no way to hide the shortcut when it still should be accessible in the production app. We'd have to change the code a little bit.

PR #1911, opened 2021-05-26
refactor: eliminate duplicate code in projects

geigerzaehler:
Before, the functions `refresh` and `fetchPeers` used exactly the same code.

### DIFF
--- a/ui/Screen/Project.svelte
+++ b/ui/Screen/Project.svelte
@@ -1,5 +1,5 @@
 <script lang="typescript">
-  import { onMount } from "svelte";
+  import { onDestroy } from "svelte";
   import { push } from "svelte-spa-router";
 
   import * as localPeer from "../src/localPeer";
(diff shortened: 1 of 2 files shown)


ui/Screen/Project.svelte
```diff
@@ -46,13 +51,13 @@
     selectPeer(peer);
   };
 
-  onMount(() => {
+  onDestroy(
```
rudolfs, line 54: Why did you change this?

geigerzaehler: Because we did not return anything from the callback to `onMount` we never stopped listening to the events. I could have added a return but felt that `onDestroy` is more straight-forward.

rudolfs: Can we make this one more explicit as well?

geigerzaehler: https://github.com/radicle-dev/radicle-upstream/pull/1911/commits/6e2f1b544e8b6c807605520e572978995bd131fa

rudolfs (approved):
utACK

rudolfs (dismissed):

PR #1912, opened 2021-05-26
Add org members list

CodeSandwich:


ui/Screen/Org/Members.svelte
```diff
@@ -19,25 +17,55 @@
...
+    min-width: 0;
   }
 </style>
 
 <div class="container">
   <List items={members} let:item={member} styleHoverState={false}>
     <div class="list-item">
-      <div class="member-name typo-text-bold">{member}</div>
-      <div class="member-address">{ellipsed(member)}</div>
+      {#if member.identity}
```
rudolfs, line 37: We should keep the style like it's in the mockup -- we show the Ethereum address for unattested/unresolved members. And show the Avatar + name for attested ones:

Screenshot 2021-06-16 at 11 47 53

proxy/api/src/http/identity.rs
```diff
@@ -47,6 +48,18 @@ fn get_filter(
         .and_then(handler::get)
 }
 
+/// `GET /remote/<id>`
```
rudolfs, line 51: I wonder if we need a separate endpoint for remote identities. We could probably have a "type" field on `Identity` to distinguish whether it's our own identity or a remote one. But I understand that this was the easiest way, so maybe we can refactor this later in a follow-up.

What do you think @geigerzaehler?

geigerzaehler: It think it makes sense to have separate endpoints because we also have different fields for the two types of identities.

proxy/api/src/identity.rs
```diff
@@ -39,9 +39,38 @@ pub struct Identity {
...
+            metadata: identity.metadata,
+            avatar_fallback: identity.avatar_fallback,
+        }
+    }
+}
+
+/// The remote users personal identifying metadata and keys.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RemoteIdentity {
```
geigerzaehler, line 60: Can we rename this to `Person`? This is the terminology from radicle link. `Identity` is an Upstream-only concept because it includes `peer_id` which is not present in the link data.

proxy/api/src/identity.rs
```diff
@@ -39,9 +39,38 @@ pub struct Identity {
...
+    /// The coco URN.
+    pub urn: Urn,
+    /// Bundle of user provided data.
+    pub metadata: Metadata,
+    /// Generated fallback avatar to be used if actual avatar url is missing or can't be loaded.
+    pub avatar_fallback: avatar::Avatar,
```
geigerzaehler, line 66: Can we drop this for here and use the typescript library to generate this in the client? I’m ok if we do this as a follow-up when we remove avatars alltogether.

ui/src/attestation/contract.ts
```diff
@@ -53,24 +53,45 @@ export class ClaimsContract {
...
+    const filter = this.contract.filters.Claimed(address);
+    const lastEvent = (await this.contract.queryFilter(filter)).pop();
+    if (!lastEvent) {
+      return undefined;
+    }
+    let claimed;
+    try {
+      claimed = await this.getClaimedByTx(lastEvent.transactionHash, address);
+    } catch {
+      return undefined;
```
geigerzaehler, line 68: Is there a reason we swallow errors?

ui/src/attestation/contract.ts
```diff
@@ -53,24 +53,45 @@ export class ClaimsContract {
...
+    }
+    if (claimed.every(hashByte => hashByte === 0)) {
+      return undefined;
```
geigerzaehler, line 71: I don’t understand we why are doing this check. Could you add a comment?

ui/src/org.ts
```diff
@@ -458,3 +489,26 @@ export const openAnchorProjectModal = async (
...
+  const radicleIdBytes = await contract.getClaimed(address);
+  if (!radicleIdBytes) {
+    return undefined;
+  }
+  const urn = identitySha1Urn(radicleIdBytes);
+  let identity;
+  try {
+    identity = await proxy.client.remoteIdentityGet(urn);
+  } catch {
+    return undefined;
```
geigerzaehler, line 506: we should probably check for a 404 here and re-throw all other errors.

CodeSandwich: Fixed

ui/src/attestation/contract.ts
CodeSandwich: Added. This branch could be dropped, but then we would be returning a `0` hash if the identity has been explicitly unclaimed. That would needlessly complicate the function's API.

ui/src/attestation/contract.ts
CodeSandwich: We don't, we return `undefined` to signal that there's no valid claim on the Ethereum side. Added a comment.

proxy/api/src/identity.rs
CodeSandwich: I think that it should be a follow-up, these changes are way outside of this PR scope

proxy/api/src/identity.rs
CodeSandwich: Fixed

ui/Screen/Org/Members.svelte
CodeSandwich: Fixed

geigerzaehler (approved):

rudolfs (approved):
LGTM 🚀 

The replication issues will be fixed in a follow-up: https://github.com/radicle-dev/radicle-upstream/pull/2028.

PR #1913, opened 2021-05-28
improvements to the Graph module

geigerzaehler:
* Use `fetchPolicy` default option
* Throw an error instead of returning an invalid endpoint URI. Throwing the error also results in the UI showing it.
* Fix type annotation from tuple (`[Org]`) to array (`Org[]`)

### DIFF
--- a/ui/src/theGraphApi.ts
+++ b/ui/src/theGraphApi.ts
@@ -7,54 +7,59 @@ import * as ethereum from "ui/src/ethereum";
 import * as error from "ui/src/error";
 import * as urn from "ui/src/urn";
 
-const gnosisClientEndpoint = (): string => {
-  const walletStore = svelteStore.get(wallet.store);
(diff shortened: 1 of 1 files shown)


rudolfs:
> Throw an error instead of returning an invalid endpoint URI. Throwing the error also results in the UI showing it.

When I switch to the "Local" net there is no exception nor a notification.


https://user-images.githubusercontent.com/158411/119958806-ad1f0d80-bfa3-11eb-9693-bc49d15caf2d.mov

geigerzaehler:
> When I switch to the "Local" net there is no exception nor a notification.

This is the same behavior as on the current orgs branch. It seems this is caused by the local eth node not running, the `fetch()` for some data from it failing, and thus the code to fetch data from the graph never being called.

rudolfs (approved):
Thanks for the 🧹 💨!

PR #1915, opened 2021-05-28
Integrate patches with router

geigerzaehler:
Update #1876 with latest master. This means updating patches to use the new router. The diff for this is useless because it also contains all the new patches code.

### DIFF
--- a/ui/Screen/Project/Source/PatchList.svelte
+++ b/ui/Screen/Project/Source/PatchList.svelte
@@ -56,7 +56,6 @@
         filteredPatches = patches;
         break;
       default:
-        console.log(filter);
         router.unreachable(filter);
         break;
     }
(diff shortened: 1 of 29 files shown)


ui/Screen/Project/Source/PatchList.svelte
```diff
@@ -0,0 +1,116 @@
...
+        filteredPatches = patches.filter(patch => !patch.merged);
+        break;
+      case "closed":
+        filteredPatches = patches.filter(patch => patch.merged);
+        break;
+      case "all":
+        filteredPatches = patches;
+        break;
+      default:
+        console.log(filter);
```
rudolfs, line 59: 👀

rudolfs (approved):

ui/Screen/Project/Source/PatchList.svelte
geigerzaehler: So strange that eslint didn’t catch that

geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1915/commits/611d187a74e5057512e9f0220b9cd0622e56658c

PR #1917, opened 2021-05-31
style(ui): Fix border radius on "still looking" projects

### LINKED ISSUE #1890: Fix hover state on "Still looking" element
The inner hover state element is not rounded, so when you hover, the non-roundedness covers the rounded corners. Notice in this screenshot how the corners of the border are cut off. 

image

juliendonck:
- closes #1890 


ui/Screen/Profile/Following.svelte
```diff
@@ -55,6 +55,7 @@
     display: flex;
     align-items: center;
     justify-content: space-between;
+    border-radius: 0.5rem;
```
rudolfs, line 58: Shouldn't we do this by doing `overflow: hidden;` on the container?

juliendonck: 🙇

juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1917/commits/604f8c9d0852cd2c292f4dba04b7498fc2b6ad67

rudolfs (approved):
utACK

PR #1922, opened 2021-05-31
dev: isolate development environment

geigerzaehler:
* Running Uptream in development mode uses `sandbox/rad_home` from the repo as `RAD_HOME`.
* We don’t install `git-remote-rad` in development mode anymore
* We provide a script that makes `git-remote-rad` available in dev mode and sets `RAD_HOME`


rudolfs (approved):
utACK

PR #1924, opened 2021-05-31
fix(tests): unload app when test is successful

geigerzaehler:
We’re running into a [cypress issue][1] where the app from a previous tests still runs while a second test is started. When the cookies are cleared to prepare the second test the app from the first test may error because the authentication token is invalid.

To work around this problem we unload the app after a successful test run.

[1]: https://github.com/cypress-io/cypress/issues/15247


rudolfs (approved):
👀 🤔

PR #1925, opened 2021-05-31
fix: don’t leak ethereum account polls

geigerzaehler:
Before this change we would never clear the interval that polls for account data. Since this interval is created whenever the ethereum environment is changed every change would register another interval. We would end up with one poll request for every environment change.


rudolfs (approved):
Nice catch!

Issue #1521, opened 2021-01-13
Move from using dispatch to passing functions as parameters 


NunoAlexandre:
Currently, we use [Svelte's event system](https://svelte.dev/tutorial/component-events) as the way to configure a (component's behaviour upon some event, e.g., what do when a user clicks a button.  

By reading this [SO thread](https://stackoverflow.com/questions/61569655/svelte-event-forwarding-with-dispatcher-vs-passing-in-handling-function-which) I understand that `dispatch` is the only way to achieve this with native components. However, I find it a bad deal to, therefore, breach it out across the whole codebase for three main reasons:

1. it makes the code highly implicit and difficult to follow
2. the events are dynamically checked, resulting in the whole flow only being checked on runtime
3. no type-checking, one can call the wrong event or pass unfitting arguments and - again - only see it fail on runtime 

My suggestion is to move from using `dispatch` to passing functions as parameters instead, which resolves the 3 issues raised above.

_context: https://github.com/radicle-dev/radicle-upstream/pull/975#issuecomment-759353567 + https://github.com/radicle-dev/radicle-upstream/pull/975#issuecomment-759391371_

geigerzaehler:
@MeBrei, @xla: Do you agree with this approach? If so we don’t want to tackle this all at once but incrementally as we go.

xla:
The alternative needs more realistic scrutiny, as the perceived gains from type-safety needs to be measured against the not insignificant overhead when a component at the bottom of a deep component hierarchy changes its interface. Such a change would necessitate refactoring to all dependent components. Furthermore there should be a proposal for how to capture optionality when binding, something that the current dispatch approach allows. Generally I'm not against improvements which make public APIs more explicit, I'd be sceptical that the implications are clear.

geigerzaehler:
Thanks for the input, @xla. With your comments in mind, I think it’s reasonable to chose a mechanism on a case-by-case basis where we can weigh the pros and cons.

Since we don’t intend to migrate everything to one approach anyway, I suggest we close this issue.

NunoAlexandre:
> The alternative needs more realistic scrutiny, as the perceived gains from type-safety needs to be measured against the not insignificant overhead when a component at the bottom of a deep component hierarchy changes its interface. Such a change would necessitate refactoring to all dependent components.

[...]

PR #1930, opened 2021-06-02
feat(ui): add wallet status to wallet sidebar icon

### LINKED ISSUE #1919: Wallet status on sidebar Wallet icon
Give the user an insight in the status of the wallet by looking at the wallet icon in the sidebar

States:
- Not connected
- Connected
- Connected with pending transactions

Screenshot 2021-05-28 at 13 27 33

juliendonck:
closes #1919 


juliendonck:
@rudolfs I'm not clear on why every time you click on any of the items it seems to rerender the wallet indicator one

ui/DesignSystem/Component/WalletStatusIndicator.svelte
```diff
@@ -0,0 +1,58 @@
...
+    dispatch("walletClick");
+  };
+
+  $: pendingTxs = $transactions.filter(
+    tx => tx.status === "Awaiting inclusion"
+  );
+  $: wallet = $store;
+  // Hack to have Svelte working with checking the $wallet variant
+  // and thus be able to access its appropriate fields.
+  $: w = $wallet;
```
geigerzaehler, line 23: This hack is not necessary anymore. You can just use `$wallet` instead of `w`.

ui/DesignSystem/Component/Sidebar.svelte
```diff
@@ -54,97 +56,46 @@
...
-          class:active={$activeRouteStore.type === "wallet"}
-          on:click|stopPropagation={() =>
-            push({ type: "wallet", activeTab: "transactions" })}>
-          <Icon.Wallet />
-        </div>
-      </Tooltip>
+      <WalletStatusIndicator
+        active={$activeRouteStore.type === "wallet"}
+        on:walletClick={() =>
+          push({ type: "wallet", activeTab: "transactions" })} />
```
geigerzaehler, line 88: We’re trying to [avoid](https://github.com/radicle-dev/radicle-upstream/issues/1521) using events because we can’t type them properly. Instead, we simply use callback props
```suggestion
        onClick={() =>
          push({ type: "wallet", activeTab: "transactions" })} />
```

And in `WalletStatusIndicator`
```typescript
export let onClick: () => void
```

brandonhaslegs:
I'm very hesitant to add a green circle to the icon when it's connected. It draws way too much undeserved attention to the icon as if the connection status of the wallet is the most important thing in the app, when really it's just a "normal" state… just like having the wallet disconnected is a normal state.

My suggestion would just be an open wallet and a closed wallet icon for connected and disconnected. I've added the closed wallet icon to the icon file [here](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=9056%3A556).

juliendonck:
> I'm very hesitant to add a green circle to the icon when it's connected. It draws way too much undeserved attention to the icon as if the connection status of the wallet is the most important thing in the app, when really it's just a "normal" state… just like having the wallet disconnected is a normal state.
> 
> My suggestion would just be an open wallet and a closed wallet icon for connected and disconnected. I've added the closed wallet icon to the icon file [here](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=9056%3A556).

Ok I agree. Do we still use the orange dot when transactions are pending

juliendonck:
```
/build/ui/DesignSystem/Component/WalletStatusIndicator.svelte:26:13
--
Error: This condition will always return 'false' since the types 'Status.Connecting \| Status.NotConnected' and 'Status.Connected' have no overlap. (ts)
--
{:else if $wallet.status === Status.Connected &amp;&amp; pendingTxs.length &gt; 0}
--
```

ui/DesignSystem/Component/Sidebar.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1930/commits/3f87e87b845bdb63021212e7cffd4fe17d2adbca

ui/DesignSystem/Component/WalletStatusIndicator.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1930/commits/3f87e87b845bdb63021212e7cffd4fe17d2adbca

rudolfs:
@juliendonck it's because the first `if` clause already captures the `$wallet.status === Status.Connected` condition and the else if would never be reached in this case. You probably want to do another `if` clause within the first one.

brandonhaslegs (approved):

PR #1931, opened 2021-06-02
Style Update

brandonhaslegs:
Just wanted to change a few styles to make things feel more modern

- Update all border radii to be the same.
- Update all border radii to rems instead of px
- increase roundness of border radius globally


ui/Screen/Project/Source/CheckoutButton.svelte
```diff
@@ -30,9 +30,9 @@
     margin-top: 3rem;
     right: 0;
     position: absolute;
-    border-radius: 0.5rem;
+    border-radius: 1rem;
     background: var(--color-background);
-    border: 1px solid var(--color-foreground-level-3);
+    /* border: 1px solid var(--color-foreground-level-3); */
```
rudolfs, line 35: If this is intentional, we should just delete the line instead of commenting it out.

brandonhaslegs: Ah not intentional. Will do

juliendonck:
This is so nice 😍 
Screenshot 2021-06-02 at 17 11 40

juliendonck (approved):
🤙

rudolfs (approved):
utACK

PR #1932, opened 2021-06-02
fix: don’t destroy wallet when all views unsubscribe

geigerzaehler:
Before, on every page navigation all views would unsubscribe from the wallet store and then subscribe again when the new page was rendered. This resulted in the wallet reconnecting and loading all its data.


rudolfs (approved):
I suspected something like this. Good catch! 👍

PR #1940, opened 2021-06-03
style(ui): Rework project page's actions

### LINKED ISSUE #1936: Introduce Overflow button on Project page
- [x] Change the global Following button back to rendering as icon + text. instead of just an icon in a button
- [x] Move Following button on the project page to a button accessible under a `•••` overflow button (until the following button actually does something, it's totally fine to hide it)
- [x] Change following button under the overflow button to "Unfollow" instead of being a Follow/Unfollow toggle button
- [x] Add other quick actions under the overflow button
   - [x] Copy Radicle ID
   - [ ] Checkout (or Fork… depending on the situation?)
   - [ ] New Patch

image

juliendonck:
Introduces simpler project header
Screenshot 2021-06-03 at 17 17 23

Improves both the PeerSelector as well as adds the actions to the overflow dropdown
Screenshot 2021-06-03 at 17 17 46
Screenshot 2021-06-03 at 17 17 51

closes #1936


ui/DesignSystem/Component/PeerSelector.svelte
```diff
@@ -82,7 +85,9 @@
 
   .peer-dropdown {
     border: 1px solid transparent;
-    border-radius: 0.5rem;
+    border-right: none;
+    border-top-left-radius: 0.5rem;
+    border-bottom-left-radius: 0.5rem;
```
brandonhaslegs, line 90:
```suggestion
    border-bottom-left-radius: 0.5rem;
    border-bottom-right-radius: 0.5rem;
```

brandonhaslegs:
Very pleased with how the overflow menu turned out! Thanks for giving it a go.

ui/DesignSystem/Component/PeerSelector.svelte
juliendonck: Huh, that's wrong. I don't understand

ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
```diff
@@ -18,7 +18,7 @@
     disabled?: boolean;
   }
   export let menuItems: MenuItem[];
-  export let headerTitle: string;
+  export let headerTitle: string = "";
```
rudolfs, line 21: I think this is a bit more idiomatic:
```diff
diff --git a/ui/DesignSystem/Component/AdditionalActionsDropdown.svelte b/ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
index ec32f7c8..3faeb18a 100644
--- a/ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
+++ b/ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
@@ -18,7 +18,7 @@
     disabled?: boolean;
   }
   export let menuItems: MenuItem[];
-  export let headerTitle: string = "";
+  export let headerTitle: string | undefined = undefined;

   export let dataCy = "";
   export let style = "";
@@ -133,7 +133,7 @@
     </button>
     {#if expanded}
       <div out:fade={{ duration: 100 }} class="modal" hidden={!expanded}>
-        {#if headerTitle !== ""}
+        {#if headerTitle}
           <div class="header">
             <StyledCopyable value={headerTitle} />
           </div>
```

ui/DesignSystem/Component/PeerSelector.svelte
brandonhaslegs: It looks weird with the bottom right radius being a corner. Wondering if we should just keep the dropdown fully rounded just to keep it consistent with all dropdowns.

ui/DesignSystem/Component/AdditionalActionsDropdown.svelte
juliendonck: Weird magic, but I like it.

juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/1940/commits/00574b15ac98df5ad530393b890bc2d68b8bd11a

ui/DesignSystem/Component/PeerSelector.svelte
juliendonck: It does? For me it's rounded... Did you run the code? I round the bottom right corner in the <Entry /> to make sure it's only rounded if there's more then 1 item in the list.

juliendonck: Aah, I get it now. It was hard to see in dark mode. I made it dependent on the height of the dropdown, which means that if there is more then 1 item in the list it will be rounded otherwise it's a corner.
https://github.com/radicle-dev/radicle-upstream/pull/1940/commits/f2fd1adb4a2c9774bdca3889aa37badd7f115b3f

brandonhaslegs:
Oh you're missing a hover state on the dropdown and the edit button

brandonhaslegs (dismissed):
Add hover states to dropdown and edit buttons

juliendonck:
> Add hover states to dropdown and edit buttons

@rudolfs do you know why the hover doesn't work? It's properly defined

juliendonck:
> Add hover states to dropdown and edit buttons

Hover states actually work, they are just the same as the background. So I had to make it a shade darker

ui/Screen/Project.svelte
```diff
@@ -31,11 +31,44 @@
 
   export let urn: Urn;
   export let activeView: router.ProjectView = { type: "files" };
+  let hoverstyle: string = "";
+
+  const mouseenter = () => {
+    hoverstyle = "background-color: var(--color-foreground-level-2)";
```
rudolfs, line 37: Ingenious! I wonder if there's a way to put this logic into the button itself?

rudolfs (approved):

ui/Screen/Project.svelte
juliendonck: I mean I'm only doing it here cause I'm overwriting the button hover style cause the background is the same color as the hover background color.

rudolfs: Wonder if we'll also need to override it in other places later? Not a biggie tho.

PR #1944, opened 2021-06-04
refactor(ui): align all modals

### LINKED ISSUE #1939: Make sure all modals are using a standard modal component
List of modals that should use the modal component:
- Add remotes
- New project
- Keyboard shortcuts
- Token Streams onboarding
- Transaction details
…probably forgetting some things here

---

image

Global modal properties:
- 37.5rem width
- 1rem border-radius
- vertically and horizontally centered position
- dark overlay 70% #000
- emoji header (optional)
- headline (h1 size) (optional)
- normal text description (optional)
- Button Navigation anchored to the bottom right (optional)

---
image

While you're at it, would be cool to have a separate "Mini Modal" component that has all the same properties except:
- 18.75rem width
- headline (h3 size) (optional)

[Figma specs](https://www.figma.com/file/owmgsbs6lnUt8R1bixstCA/%F0%9F%8C%B1%F0%9F%86%99-Radicle-Upstream?node-id=7004%3A64488)

juliendonck:
closes #1939 

### DIFF
--- a/ui/DesignSystem/Component/Modal.svelte
+++ b/ui/DesignSystem/Component/Modal.svelte
@@ -1,10 +1,11 @@
 <script lang="typescript">
-  export let dataCy = "";
+  export let dataCy: string = "";
+  export let style: string = "";
 </script>
 
 <style>
   .container {
-    width: 38.5rem;
+    width: 37.5rem;
     background: var(--color-background);
     border-radius: 1rem;
     padding: 2rem;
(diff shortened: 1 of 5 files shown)


juliendonck:
@brandonhaslegs that accounts for the big one. I think until we have a need to use the small modal I don't think I'd add it in the code.

juliendonck:
@rudolfs you think it makes sense to make the icon and title etc fixed optional variables of <Modal /> or we keep the flexibility with it just being an empty canvas?

rudolfs (approved):
Screenshot 2021-06-04 at 16 13 28

Can we add a linebreak so the link is on a separate line?

Issue #1945, opened 2021-06-04
Show testnet name in wallet tooltip


rudolfs:
Like so: `Wallet • Connected to Ropsten`.

For mainnet the tooltip can stay as-is:
<img width="277" alt="Screenshot 2021-06-04 at 12 26 44" src="https://user-images.githubusercontent.com/158411/120795147-68621c00-c539-11eb-82e7-288e75d6688c.png">

PR #1946, opened 2021-06-05
feat(ui): add selectedNetwork to wallet tooltip

sebastinez:
Hi folks,

This PR adds the functionality requested in #1945.
It uses the `environment` property created by `wallet.ts`
https://github.com/radicle-dev/radicle-upstream/blob/eef980fe3976fa28c98517b6c9a61d7e5fb84d8a/ui/src/wallet.ts#L390

Let me know what you think.
Cheers!

### DIFF
--- a/ui/DesignSystem/Component/WalletStatusIndicator.svelte
+++ b/ui/DesignSystem/Component/WalletStatusIndicator.svelte
@@ -32,7 +32,7 @@
         </SidebarItem>
       </Tooltip>
     {:else}
-      <Tooltip value="Wallet · Connected">
+      <Tooltip value={`Wallet · Connected to ${wallet.environment}`}>
         <SidebarItem
           dataCy="wallet"
           indicator


sebastinez:
I realized that the mainnet case will not be taken into account since There is only the local and ropsten option..
Will see how to address this part of the issue.

P.S. The implementation could eventually be more generic to allow deployments on other test networks for example Kovan, Goerli, etc.

rudolfs:
Hey @sebastinez, this should do fine for now.
We require contributors to both [gpg sign](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/signing-commits) commits as well as sign-off (`git commit -s`) on our [DCO](https://github.com/radicle-dev/radicle-upstream/blob/master/DCO). Could you do that and force-push the commit after?

sebastinez:
> Hey @sebastinez, this should do fine for now.
> We require contributors to both [gpg sign](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification/signing-commits) commits as well as sign-off (`git commit -s`) on our [DCO](https://github.com/radicle-dev/radicle-upstream/blob/master/DCO). Could you please do that and force-push the commit after?

Thank you @rudolfs

ui/DesignSystem/Component/WalletStatusIndicator.svelte
```diff
@@ -32,7 +32,7 @@
         </SidebarItem>
       </Tooltip>
     {:else}
-      <Tooltip value="Wallet · Connected">
+      <Tooltip value={`Wallet · Connected to ${wallet.environment}`}></Tooltip>
```
rudolfs, line 35:
```suggestion
      <Tooltip value={`Wallet · Connected to ${wallet.environment}`}>
```

sebastinez: Yes it slipped through in the last commit.. I fixed it in the most recent commit.

rudolfs (approved):

rudolfs:
Merged via 500d63b013eb4de4b6d675ac1bd86e27aa750345.

PR #1951, opened 2021-06-07
refactor: integrate back radicle_daemon::keystore

geigerzaehler:
We move the code from `radicle_daemon::keystore` closer to `radicle-proxy`. `radicle_daemon::keystore` provides an API very specific to the radicle-proxy code and it does not share code with any of the rest of `radicle_daemon`. This was [raised before][1].

[1]: https://github.com/radicle-dev/radicle-link/pull/576#discussion_r608646189


xla (approved):
📄 🆖 🍎 👹

PR #1953, opened 2021-06-07
fix(tests): properly select peer in networking tests

geigerzaehler:
Instead of using the test ID `peer-dropdown-container` to select a peer we use the newly introduced `peer-dropdown-entry`. The former ID would select the container element and clicking it should not result in peers selection. (In practice this worked though because of how cypress translates clicks to coordinates.)

We also inline the `Entry` component for the peer selector dropdown.


rudolfs (approved):
utACK

PR #1969, opened 2021-06-09
feat(dev): fast key store for development

### LINKED ISSUE #1614: Fast file keystore for development
When the proxy and `git-remote-rad` are build without optimization, then decrypting the keys stored in a file takes quite some time (20 seconds on my machine). This slows down development and is a particular problem when running tests since every test case that does not use the in-memory key store requires unlocking which in turn decrypts the key. A workaround is to compile with the `--release` flag and use the release build to test and develop against. However compilation with the release profile takes quite some time too, slowing down development again.

`radicle-keystore` provides a set of [test parameters](https://github.com/radicle-dev/radicle-keystore/blob/e8bfc0052c5cd2d6a68f1c69fcbc74149b9d82ac/src/crypto.rs#L44) for the key derivation function that make the encryption insecure but fast. I suggest we use that for development builds. We would need to ensure that this does not mess with production values.

geigerzaehler:
Fixes #1614.

We add a command line option to the proxy that makes it use weak encryption parameters to encrypt files on disk. This allows reasonably fast encryption even if the proxy is build without code optimization.

The command line option is only available when the proxy is build with the `unsafe-fast-keystore` feature which is not enabled when building releases.

Similarly, we also enable the selection of fast encryption for the remote helper. This is enabled by an environment variable.


geigerzaehler:
@rudolfs Could you test that the git commands work if you follow the instruction from the dev docs? You need to reset your dev state first by removing `sandbox/rad_home`.

xla (approved):
utACK

🕰 ✨ 💎 😏

rudolfs (approved):
Dev setup works for me.
Didn't test the packaged release version though.

geigerzaehler:
I tested the distribution app image. Everything works fine.

PR #1972, opened 2021-06-09
fix: remove pending & rejected transactions from transaction list

### LINKED ISSUE #1970: Pending transactions should show up as pending instead of normal
In the wallet, pending transactions also show up as normal transactions, basically showing up twice

juliendonck:
closes #1970 

### DIFF
--- a/ui/Screen/Wallet/Transactions.svelte
+++ b/ui/Screen/Wallet/Transactions.svelte
@@ -1,13 +1,18 @@
 <script lang="typescript">
-  import { store as transactions, getShortMonth } from "ui/src/transaction";
+  import {
+    store as transactions,
+    getShortMonth,
+    TxStatus,
+  } from "ui/src/transaction";
   import type { Tx } from "ui/src/transaction";
 
   import TxList from "ui/DesignSystem/Component/Wallet/Transactions/TxList.svelte";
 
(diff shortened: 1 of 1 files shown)


brandonhaslegs:
The pending transaction should only show up in one section, not both.
image

brandonhaslegs (dismissed):
image
Pending transaction should only show up in the pending section, not in both.

brandonhaslegs:
Okay I was on the wrong branch. But another thing: I tried to do it again and ended up with two "Claim Radicle Identity" transactions. Even though this is the result of connecting two different wallets. Shouldn't the transactions not carry over when I connect a new wallet?

image

juliendonck:
> Okay I was on the wrong branch. But another thing: I tried to do it again and ended up with two "Claim Radicle Identity" transactions. Even though this is the result of connecting two different wallets. Shouldn't the transactions not carry over when I connect a new wallet?
> 
> image

Yeah this is something about transactions persisting, we know this happens and will fix this in the next cycle when we take some time to clean up the ethereum code.

geigerzaehler (approved):

geigerzaehler:
Merged in 79dd6af895c6b62270411f06133be706dc42089a

PR #1977, opened 2021-06-09
fix(dev): avoid yarn run to properly kill process

geigerzaehler:
We avoid `yarn run` in the ethereum dev node script so that `ganache-cli` is properly killed when an error occurs.


rudolfs (approved):
utACK

PR #1979, opened 2021-06-10
fix(dev): allow empty function lint

geigerzaehler:
We disable the lint that complains about empty functions.

This is achieved by removing the override for typescript files so that extending the override configuration with the rules from `@typescript-eslint/recommended` doesn’t override the top-level rules that already allow empty functions.


rudolfs (approved):
utACK

PR #1980, opened 2021-06-10
fix router issues and reorganize the code

geigerzaehler:
* We split the routing code into two modules. One contains the concrete route definitions and loader function. The other contains the routing logic that is agnostic of the concrete router types.
* We don’t persist `LoadedRoute`. Instead we persist `Route` and load on every push or pop. As a consequence, `activeRouterStore` is not a derived store anymore.
* Consolidate functionality in `setHistory`
* Extract `unreachable`. It’s not specific to the router


geigerzaehler:
A necessary follow up to this is putting `setHistory` behind a lock so that we don’t run into race conditions.

ui/src/router/index.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export const pop = async (): Promise<void> => {
+  const history = svelteStore.get(historyStore);
+  if (history.length > 1) {
+    await setHistory(history.slice(0, -1));
+  }
+};
+
+export const activeRouteStore =
```
rudolfs, line 58: Yeah, having this not be a derived store makes things much more straightforward!

rudolfs (approved):

rudolfs:
Looks like there's still some bug as CI is failing.

rudolfs:
👏

PR #1981, opened 2021-06-10
fix: prevent lock race condition

geigerzaehler:
* We remove `screen.lock()` and `screen.unlock()` and replace it with `screen.withLock()`. This makes it impossible to forget to unlock.
* We fix a race condition when two actions are run with a lock. Before, the first action that would finish would unlock the screen. Now, we unlock only after all actions have finished.


rudolfs (approved):
utACK

PR #1983, opened 2021-06-11
chore(deps): update dependencies

geigerzaehler:
* Includes the following major updates
  * [Electron](https://www.electronjs.org/releases/stable#13.0.0)
  * [jest](https://github.com/facebook/jest/releases/tag/v27.0.0)
  * [svelte-check](https://github.com/sveltejs/language-tools/releases/tag/svelte-check-2.0.0)
  * [sinon](https://github.com/sinonjs/sinon/blob/master/CHANGELOG.md#1100--2021-05-24)

  The only breaking change affecting us is jest changing the default environment to node. We fix this by configuring jsdom as the default environment.

* Includes an update to yarn itself

* We fix the typescript version because type-checking crashes on 4.3.2


rudolfs (approved):
utACK

PR #1985, opened 2021-06-11
refactor: mutex executor to manage concurrent fetches

geigerzaehler:
We introduce `MutexExecutor` that allows us to abstract over behavior were we need to abort previous requests issued by the user. We apply this abstraction in `ui/src/screen/project.ts` where it simplifies the code significantly.

The abstraction can also be used in other places like `ui/src/screen/project/source.ts` and the upcoming changes to the router.

### DIFF
--- a/ui/src/mutexExecutor.test.ts
+++ b/ui/src/mutexExecutor.test.ts
@@ -1,0 +1,83 @@
+import * as mutexExecutor from "./mutexExecutor";
+import { sleep } from "./sleep";
+import * as sinon from "sinon";
+
+test("cancels running task", async () => {
+  const e = mutexExecutor.create();
+
+  const first = e.run(async () => {
+    await sleep(10);
+    return "first";
+  });
+  const second = e.run(async () => {
+    return "second";
+  });
+
+  expect(await first).toBe(undefined);
+  expect(await second).toBe("second");
+
+  const third = e.run(async () => {
+    await sleep(10);
+    return "third";
+  });
+  const fourth = e.run(async () => {
+    return "fourth";
+  });
+
+  expect(await third).toBe(undefined);
+  expect(await fourth).toBe("fourth");
+});
+
+test("cancels multiple tasks", async () => {
+  const e = mutexExecutor.create();
+
+  const canceled1 = e.run(async () => {
+    await sleep(10);
+    return true;
+  });
+  const canceled2 = e.run(async () => {
+    await sleep(10);
+    return true;
+  });
+  const canceled3 = e.run(async () => {
+    await sleep(10);
+    return true;
+  });
+  const last = e.run(async () => {
+    return true;
+  });
+
+  expect(await canceled1).toBe(undefined);
+  expect(await canceled2).toBe(undefined);
+  expect(await canceled3).toBe(undefined);
+  expect(await last).toBe(true);
+});
+
+test("triggers abort signal event", async () => {
+  const e = mutexExecutor.create();
+  const abortListener = sinon.spy();
+
+  e.run(async abort => {
+    abort.addEventListener("abort", abortListener);
+    await sleep(10);
+    return "first";
+  });
+  expect(abortListener.called).toBe(false);
+  e.run(async () => {});
+  expect(abortListener.called).toBe(true);
+});
+
+test("don’t throw error on aborted task", async () => {
+  const e = mutexExecutor.create();
+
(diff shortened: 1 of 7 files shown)


ui/src/mutexExecutor.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.runningTaskId += 1;
+    const taskId = this.runningTaskId;
+
+    if (this.abortController) {
+      this.abortController.abort();
+    }
+    this.abortController = new AbortController();
+    return f(this.abortController.signal).then(
+      data => {
+        if (this.runningTaskId == taskId) {
```
rudolfs, line 47:
```
  if (this.runningTaskId === taskId) {
```

ui/src/mutexExecutor.ts
```diff
@@ -0,0 +1,62 @@
...
+          return data;
+        } else {
+          return undefined;
+        }
+      },
+      err => {
+        if (this.runningTaskId == taskId) {
```
rudolfs, line 54:
```suggestion
        if (this.runningTaskId === taskId) {
```

ui/src/screen/project.ts
```diff
@@ -24,51 +26,43 @@ export const fetch = (projectUrn: Urn): void => {
...
+      const peerSelection = project.userList(peers);
+      throwUnlessPeersPresent(peerSelection, screen.data.project.urn);
+      screenStore.success({
+        ...screen.data,
+        peers,
+        peerSelection,
+      });
+    } catch (err) {
+      screenStore.error(error.fromUnknown(err));
+      return;
```
rudolfs, line 64: Why do we need an explicit `return` here?

rudolfs (commented):
Will these be addressed in a follow-up?

```
ui/src/screen/project/source.ts
53:  requestInProgress: AbortController | null;
154:    const request = new AbortController();
214:    const request = new AbortController();
345:  const request = new AbortController();

ui/src/source.ts
129:  request: AbortController | null;
134:  request: AbortController | null;
```

ui/src/mutexExecutor.ts
geigerzaehler: Shouldn’t this be caught by eslint :thinking:?

ui/src/mutexExecutor.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  e.run(async () => {});
+  expect(abortListener.called).toBe(true);
+});
+
+test("don’t throw error on aborted task", async () => {
+  const e = mutexExecutor.create();
+
+  const first = e.run(async () => {
+    await sleep(10);
+    throw new Error();
```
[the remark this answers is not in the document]
geigerzaehler: You’re making a fair assumption about the behavior. Unfortunately we can’t really abort tasks in JavaScript so the function will continue executing even if the second task is run. This means the function will throw. The test is constructed as a whitebox test to cover the error case in the implementation.

ui/src/mutexExecutor.ts
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1985/commits/384f77f256e8c726e23240e01584e7977039d83f

ui/src/mutexExecutor.ts
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1985/commits/384f77f256e8c726e23240e01584e7977039d83f

ui/src/screen/project.ts
geigerzaehler: We don’t need it. Fixed in https://github.com/radicle-dev/radicle-upstream/pull/1985/commits/e379e2e1e3d7f4bf69fd4293f14ffe076084e96c

geigerzaehler:
> Will these be addressed in a follow-up?

I took a stab at also converting this but it is more complex since the logic is too entangled. If we make the change there we also run the risk of breaking the source browsing. I’d postpone the change until we go back to the source browsing and have to make changes there. Then we would have an immediate benefit and more confidence through testing that we don’t break anything.

rudolfs (approved):
Wanna create a follow-up issue for the remaining refactors?

geigerzaehler:
> Wanna create a follow-up issue for the remaining refactors?

https://github.com/radicle-dev/radicle-upstream/issues/2005

PR #1986, opened 2021-06-11
fix(dev): jest module name mapping for TS paths

geigerzaehler:
We set the `moduleNameMapper` option for Jest to match the `paths` option from `tsconfig.json` so that we can use absolute imports also in modules required by Jest.


rudolfs (approved):
utACK

PR #1989, opened 2021-06-11
chore(ui): add org related icons

rudolfs:
Backport icons needed for org feature to make the [PR](https://github.com/radicle-dev/radicle-upstream/pull/1813) more lean.


ui/DesignSystem/Primitive/Icon/Anchor.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+  import IconWrapper from "./IconWrapper.svelte";
+
+  export let dataCy = "";
+  export let style = "";
```
geigerzaehler, line 5: Another way to write this is
```suggestion
  export let dataCy: string | undefined = undefined;
  export let style: string | undefined = undefined;
```

This has the following advantage: If this Icon is used without specifying the attributes they will be undefined and they will not be set on `IconWrapper`. Down the line, this means we don’t end up with empty attributes like `<div dataCy>` in the DOM. (In this specific case, we’ll need to fix `IconWrapper` though first).

As a general rule I’d prefer to use `undefined` as the default value.

geigerzaehler (approved):
One suggestion. I’m also ok if we keep it like it is.

ui/DesignSystem/Primitive/Icon/Anchor.svelte
rudolfs: I totally agree, in this case I just copied an old component and forgot to update the prop types. Might be worth doing this for all icons in a separate follow-up.

e3c3eacfca73580c324d5d60f54c04e04069f9db

geigerzaehler (approved):

PR #1993, opened 2021-06-15
refactor: enforce curly braces for if/else statments

geigerzaehler:
We enforce curly braces in if/else statments. This makes the code more consistent and arguably [aides readability][1]. It also reduces the number of changes required when a statement is added to a branch.

[1]: https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/


rudolfs (approved):
utACK

PR #1995, opened 2021-06-15
fix(dev): allow more modules to be required in jest tests

geigerzaehler:
We update the `config` and `error` modules so that they can be required (transitively) in tests run by Jest.


rudolfs (approved):
utACK

PR #1996, opened 2021-06-15
Extract presentation logic from screen modules

geigerzaehler:
Two commits that extract presentational logic from screen modules so that they can be reused without relying on the screen modules.

See the commits for details.


rudolfs (approved):
utACK

PR #1997, opened 2021-06-15
fix: commit teaser message overflow

geigerzaehler:
If the message in a commit teaser is too large we make it overflow. This prevents the author and date on the right from being pushed outside of the commit teaser container.

We also use auto margins instead of `justify-content: flex-end` to push the box to the right.

### DIFF
--- a/ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
+++ b/ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
@@ -7,7 +7,7 @@
   import { Icon } from "../../Primitive";
 
   export let commit: CommitHeader;
-  export let style: string = "";
+  export let style: string | undefined = undefined;
 
   const dispatch = createEventDispatcher();
   const onSelect = () => {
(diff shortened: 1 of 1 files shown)


ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
```diff
@@ -29,15 +29,15 @@
 
   .align-left {
     display: flex;
-    flex: 1;
     align-items: center;
-    max-width: 64%;
+    margin-right: 1ex;
```
rudolfs, line 33: Why is this `1ex`?

geigerzaehler: It’s there to have a one character space between the message and the author. Could also be `1em` or `2em`.

rudolfs (approved):

ui/DesignSystem/Component/SourceBrowser/CommitTeaser.svelte
rudolfs: Probably `1rem` would make sense, we use `rem`s everywhere.

PR #1999, opened 2021-06-15
refactor: poll orgs explicitly

geigerzaehler:


ui/src/org.ts
```diff
@@ -62,6 +62,44 @@ const orgFactoryAddress = (network: ethereum.Environment): string => {
...
+// `ORG_POLL_INTERVAL_MS` milliseconds.
+const updateOrgsForever = async (): Promise<never> => {
+  for (;;) {
+    const walletStore = svelteStore.get(wallet.store);
+    const w = svelteStore.get(walletStore);
+
+    if (w.status === wallet.Status.Connected) {
+      await fetchOrgs();
+    }
+    await sleep(ORG_POLL_INTERVAL_MS).catch(err => {
```
rudolfs, line 85: Will this flood the app with error notifications if we're offline?

geigerzaehler: Good point. I’ve updated the code so that we only show the first error. (The `catch` was also in the wrong place).

rudolfs (approved):
utACK

CI still complains tho.

PR #2006, opened 2021-06-16
refactor: make error code optional

geigerzaehler:
We make the `code` property passed to errors optional. This makes it easier to add new errors (which happens frequently) because we don’t need to add it to a global error registry.

Error code’s are beneficial when the cause of the error cannot be deduced from the message or the stack which rarely is the case, or if we want to match on a certain type of error.

### DIFF
--- a/ui/src/error.ts
+++ b/ui/src/error.ts
@@ -7,7 +7,7 @@ import * as ipc from "./ipc";
 
 interface ErrorParams {
   // A unique code for easy identification of the error.
-  code: Code;
+  code?: Code;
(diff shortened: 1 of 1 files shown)


rudolfs (approved):
Would we do the exception clean up in a separate PR?

geigerzaehler:
> Would we do the exception clean up in a separate PR?

You mean removing some of the error codes? Do you think that is necessary?

rudolfs:
> You mean removing some of the error codes? Do you think that is necessary?

Yeah, I meant removing the error codes. I guess we can leave them for now. 🤷

PR #2007, opened 2021-06-16
Wait until wallet is connected before fetching orgs

geigerzaehler:
In the tasks that polls for org updates we wait until the wallet has been connected.  This ensures that we poll as soon as possible after connecting even if the poll interval is quite large.

To achieve this we introduce the `svelteStore.waitUntil` abstraction. This abstraction can also be used to implement `waitUnsealed` more succinctly.


rudolfs (approved):
🤯 🧹

PR #2012, opened 2021-06-17
refactor: introduce svelteStore.waitUntil

geigerzaehler:
We add `svelteStore.waitUntil` that provides some better abstraction over `session.waitUnsealed`. This function will also be reused in the org feature.


rudolfs (approved):
utACK

Issue #2025, opened 2021-06-18
Refactor orgs data layer


rudolfs:
Follow up on: https://github.com/radicle-dev/radicle-upstream/pull/1813#discussion_r649236852.

> On the organization of code in this module and theGraphApi I’d propose the following: Let’s split it into a stateless data fetching layer and the stateful “service” layer.

> The data fetching layer would not contain any stateful code. It would not make use of stores, notifcations, module-level variables, etc. It’s primary purpose would be to export functions that get or update org-related data from the sources. (Maybe we separate data fetchers for the graph, contracts, and gnosis safe.)

> On top of this layer would be the layer that keeps state and interacts with the UI (e.g. through notifications.)

PR #2029, opened 2021-06-18
refactor: extract gnosis safe specific code

geigerzaehler:
We extract gnosis safe specific code into it’s own module. Part of #2025


rudolfs (approved):
utACK 🧹

Issue #2024, opened 2021-06-18
Dependency License Audit


geigerzaehler:
Audit the licenses of all dependencies so that they are compatible with our [licensing model](https://radicle.community/t/radicle-licensing-model/282). This task should be automated and part of CI. [Here](http://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses) is an overview of GPL-compatible licenses

geigerzaehler:
[`license-checker`](https://www.npmjs.com/package/license-checker) is a widely used tool for NPM.

geigerzaehler:
We should also include license header (and again automatically check this) in all source files.

rudolfs:
Can this be closed now?

PR #2036, opened 2021-06-21
feat(dev): check license compliance for NPM packages

geigerzaehler:
We ensure that the licenses of all NPM production dependencies are compatible with GPL-3.0-only. See #2024.


rudolfs (approved):
🦺 ✅

PR #2038, opened 2021-06-21
style(ui): update appearance settings

### LINKED ISSUE #2013: Update appearance settings
### Currently looks like this:
image

### Should look like this:
image

- [ ] **Theme**
Dark mode might be easier on the eyes, but some just want to follow the light. Only true h4x0rs know.

- [ ] **Interface font**
This is the font you’ll read most often around the app. In fact, it’s the one you’re reading right now.

- [ ] **Code font**
This is the font source code is displayed in on repositories. It’s also used for hashes.

- [ ] add lines between the settings

juliendonck:
closes #2013 

Screenshot 2021-06-21 at 11 54 46


brandonhaslegs (approved):

ui/Screen/Settings.svelte
```diff
@@ -197,7 +198,7 @@
           <h3>Devices</h3>
         </header>
         <div class="section-item">
-          <div class="info">
+          <div>
```
geigerzaehler, line 201: I think we can remove all this unnecessary divs that had the `info` class before.

juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/2038/commits/bda1aa5202029dfd406298f6c9913e04d2fbb754

juliendonck: Removed them where I could

PR #2040, opened 2021-06-21
feat(dev): add cargo-deny

geigerzaehler:
We add cargo deny to check our rust dependencies. The deny configuration is copied from `radicle-link`. The main difference is the explicit allow list for licenses.


rudolfs (approved):
utACK

Issue #1803, opened 2021-04-19
Customize the primary UI color to your favorite color ❤️💚💙


brandonhaslegs:
![image](https://user-images.githubusercontent.com/4406983/115257063-0c316d00-a130-11eb-903b-fca1f7061c1f.png)

- [x] offer Blue (#5555FF), Pink (#FF55FF), and Orange (#FF7700) as options for primary colors -> #2380
- [ ] a text field for a hex code or RGB.
- [ ] Support for Standard [HTML color names](https://www.w3schools.com/tags/ref_colornames.asp)
- [ ] Error when user inputs something that isn't a color code

- [x] translate the current "level system" of the primary color (primary-level1, primary-level-2, etc…) into an RGBA solution using opacities

juliendonck:
@brandonhaslegs what happens if I'm in hacker theme? Should it overwrite the orange with blue? The theme and font color kind of collide as separate options, no?

juliendonck:
Also, could you give me the rgba values for the `pink` please

juliendonck:
Also, could you give me the rgba values for the `pink` please

juliendonck:
Also, could you give me the rgba values for the `pink` please

brandonhaslegs:
> Also, could you give me the rgba values for the `pink` please

The pink is `255, 85, 255`

brandonhaslegs:
> @brandonhaslegs what happens if I'm in hacker theme? Should it overwrite the orange with blue?

Yes.

---

> The theme and font color kind of collide as separate options, no?

Don't know what this means.

PR #2041, opened 2021-06-21
style(ui): add primary color levels as opacities

brandonhaslegs:
Change primary color levels to work with opacities in preparation for #1803.


brandonhaslegs:
Can someone check the error page with this change? Just wondering if it still works. I think it should be fine.

rudolfs:
Seems to be working just fine:

Screenshot 2021-06-21 at 14 10 26

Is the halo/border around buttons intentional?

https://user-images.githubusercontent.com/158411/122759909-cb261800-d29a-11eb-9242-27c1b52e70d5.mov

Screenshot 2021-06-21 at 14 11 08

rudolfs (approved):

brandonhaslegs:
The border was introduced last week sometime. There's an issue for it#2030

rudolfs:
Should we convert all the other colors to RGBa as well for consistency?

brandonhaslegs:
Yep I'll make an issue and do that soon.

juliendonck:
And make all the levels opacity based?

rudolfs:
I was only referring to converting `#5555ff` -> `rgba(85, 85, 255, 1);`.
I don't really care whether the other levels are actually opacity based or not.
If it makes sense from a design perspective then yeah, why not.

brandonhaslegs:
This would break some things if we've ever put a lighter level on a darker level.

rudolfs:
> This would break some things if we've ever put a lighter level

Indeed, I was thinking about this as well. From now on we'll have to keep this in mind. With the previous hex-only implementation we were able to mix any color combination without worrying about this issue.

brandonhaslegs:
I think we can live with this restriction though.

PR #2050, opened 2021-06-22
Add license header and checks

geigerzaehler:
As part of #2024 we add license headers to all source files. We also provide a script to automatically add license headers to files and check this. The latter is run on CI.

The PR consists of two commits. The first adds the script, the second updates all files.

### DIFF
--- a/scripts/license-header.ts
+++ b/scripts/license-header.ts
@@ -1,0 +1,149 @@
+#!/usr/bin/env -S yarn node -r ts-node/register/transpile-only
+
+// Copyright © 2021 The Radicle Upstream Contributors
+//
+// This file is part of radicle-upstream, distributed under the GPLv3
+// with Radicle Linking Exception. For full terms see the included
+// LICENSE file.
+
+import globby from "globby";
+import yargs from "yargs";
+import * as fs from "fs/promises";
+import * as Path from "path";
+
+// Error that is shown without a stacktrace to the user
+class UserError extends Error {
+  constructor(public message: string) {
+    super(message);
+  }
+}
+
+async function main() {
+  yargs
+    .command({
+      command: "check",
+      describe: "Check presence of license headers in files",
+      handler: async () => {
+        let failure = false;
+        for (const path of await paths()) {
+          const content = await fs.readFile(path, "utf8");
+          if (!hasLicenseHeader(content)) {
+            failure = true;
+            console.error(`License missing from ${path}`);
+          }
+        }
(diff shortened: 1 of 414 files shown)


scripts/license-header.ts
```diff
@@ -0,0 +1,118 @@
...
+import * as Path from "path";
+
+// Globs for the files we enforce license headers for.
+const fileGlob = [
+  "ui/**/*.(svelte|ts)",
+  "native/**/*.(js|ts)",
+  "cypress/**/*.(js|ts)",
+  "public/**/*.css",
+  "!cypress/workspace",
+];
```
rudolfs, line 15: What about these?

```
builder/notarize.js
ci/*
scripts/*
ui/index.html
svelte.config.js
webpack.config.ts
jest.config.js
```

ui/src/urn.test.ts
```diff
@@ -1,11 +1,10 @@
-/**
- * The jsdom environment lacks `TextEncoder` required by `multibase`,
- * see https://github.com/multiformats/js-multibase/issues/90,
- * will be solved by https://github.com/jsdom/jsdom/issues/2524
```
rudolfs, line 4: Don't we want to keep this explanation?

ui/Theme.svelte
```diff
@@ -1,3 +1,9 @@
+<!-- Copyright © 2021 The Radicle Upstream Contributors
+  --
+  -- This file is part of radicle-upstream, distributed under the GPLv3
+  -- with Radicle Linking Exception. For full terms see the included
+  -- LICENSE file.
+  -->
```
rudolfs, line 6: Do svelte files not have a line break after the licence on purpose? .ts files seem to have a line break.

geigerzaehler: Good spot. Prettier will remove the additional line break so we can’t add it.

scripts/license-header.ts
geigerzaehler: I omitted these at first because I couldn’t make automatic adding of licenses work. But we really only need automatic checking so I added these manually

ui/src/urn.test.ts
geigerzaehler: Fixed.

geigerzaehler:
I’ve made major changes the license check script. I also changed the format of svelte comments so that vim syntax highlighting is not tripped up.

rudolfs (approved):
utACK

PR #2055, opened 2021-06-23
style(ui): restyle buttons and links

### LINKED ISSUE #2031: Fix hover and click states of buttons and links
These have always looked bad. They should just work with opacities or transforms instead of jumping around to different color styles. Just weird overall!

brandonhaslegs:
Restyling buttons and links to not be color/text based, but more tactile and buttony. This is part of a larger animation effort to make the app feel more responsive and alive as you interact with it.

To do:
- [x] fix the edit remotes joint button on the project page (they shouldn't transform alone, they should move together, regardless of where you hover on the button)
- [x] New project modal XXL buttons
- [x] delete (❌) button on the seed address
- [x] close (❌) button on the design system modal
- [x] check out why the color is flashing on hover
- [x] revision selector
- [x] add org button

Closes #2031


brandonhaslegs:
I can't figure out why some of the buttons are flashing on hover. Would like some help!

Jun-23-2021 17-30-40

brandonhaslegs:
It would also be cool if we could put the transition and transform settings in one place so if we ever want to change them or maybe have a setting that turns off motion, it's easy to adjust.

juliendonck:
> I can't figure out why some of the buttons are flashing on hover. Would like some help!
> 
> Jun-23-2021 17-30-40

I gave that one a hover style cause I had to overwrite the default hover. It was the same color as the background

juliendonck:
Let's pair on this tomorrow as well to fix these last things.

brandonhaslegs:
> I gave that one a hover style cause I had to overwrite the default hover. It was the same color as the background:

I think it's happening on more than just this button, but let's check it out tomorrow.

brandonhaslegs:
I figured it out. Was because I had a transition on `all` instead of just on the `transform` property.

brandonhaslegs:
Okay the only thing I couldn't do is make the merge remotes button transition together. That button/dropdown is too complicated for me

public/global.css
```diff
@@ -46,3 +46,21 @@ a {
 .lock-screen {
   cursor: wait;
 }
+
+.transition {
```
geigerzaehler, line 50: `transition` might be to generic as a class name. How about `button-transition`?

ui/DesignSystem/Button.svelte
```diff
@@ -96,33 +88,17 @@
...
+    opacity: 0.25;
     cursor: not-allowed;
   }
 
   .primary.disabled:hover {
-    background-color: var(--color-primary-level-1);
-    box-shadow: 0 0 0 0;
-    border-color: var(--color-primary-level-1);
-    color: var(--color-background);
+    transform: scale(1);
```
geigerzaehler, line 101: Why do we need this here and on the other selectors? For me, everything seems to work without it.

public/global.css
```diff
@@ -46,3 +46,21 @@ a {
...
+  transition: transform 0.2s ease-in-out;
+}
+
+.transition:hover,
+.transition:checked:hover {
```
geigerzaehler, line 55: It’s enough to use one selector
```suggestion
.transition:hover {
```

ui/DesignSystem/Button.svelte
```diff
@@ -77,6 +68,7 @@
     box-shadow: 0 0 0 0;
     border-color: var(--color-foreground-level-1);
     color: var(--color-foreground-level-3);
+    transform: scale(1);
```
geigerzaehler, line 71: We also need to remove `background-color` from this selector (`.vanilla.disabled:hover`).

geigerzaehler (approved):
:rainbow:

brandonhaslegs:
Okay besides the remotes-selector/edit-remotes button, the other thing I'm not sure how to solve is that the disabled buttons have opacity hover states and they should have no hover state.

brandonhaslegs:
Thanks, looks good!

juliendonck (approved):

PR #2056, opened 2021-06-23
chore(dev): bundle licenses to distribute them with our app

geigerzaehler:
We use `license-webpack-plugin` to bundle the licenses from all dependencies and our own license in a file that is distributed alongside the code bundles.

The plugin also takes care of checking licenses and we can remove `license-compliance`.


rudolfs (approved):
utACK

PR #2058, opened 2021-06-23
refactor: add source client for proxy

geigerzaehler:
The low-level proxy client can now handle request for `source/blob`.


rudolfs (approved):
utACK

PR #2060, opened 2021-06-23
refactor: reorganize ethereum modules

geigerzaehler:
This change is one step towards moving all Ethereum infrastructure into the `ui/src/ethereum` folder.

* We extract `WalletConnectSigner` from `ui/src/wallet` and move it under `ui/src/ethereum`.
* We extract the environment types from `ui/src/ethereum` and move it into `ui/src/ethereum/environments`.
* We extract the debug object from `ui/src/wallet` and move it into `ui/src/ethereum/debug`.

No logic has been changed.

As a follow-up we will do the same for `ui/src/wallet` and `ui/src/transaction`


rudolfs (approved):
utACK

PR #2066, opened 2021-06-23
feat(ui): network page

### LINKED ISSUE #1696: Move Network settings to Network page
Task:
Simply move the settings to their own page. Nothing else (for now).

Reasoning:
- When you are connected to a lot of seeds, the list of seeds is very long and hides other settings below and generally just takes up a ton of space
- The network icon in the sidebar looks like a page because it's in the navigation, but is just a status indicator
- If we put it on it's own page now, we can develop the Network page later into something more robust (health indicator, projects list, connected peers, whatever)

juliendonck:
closes #1696 


juliendonck:
@brandonhaslegs can you take a look at this? Do we just leave it as is or do we want to restyle it a bit?

brandonhaslegs:
Yes, love this! You should remove network settings from the settings page though. They shouldn't be in two places.

juliendonck:
Screenshot 2021-06-24 at 14 26 10

brandonhaslegs:
Probably need to update the docs (and inform the whole team who does support) when this change merges. Seed settings moving to their own page is a big change.

juliendonck:
> Need to update the docs (and inform the whole team who does support) when this change merges. Seed settings moving to their own page is a big change.

Yep, as soon as this is approved I'll open a PR on the docs and sync with @viraptor and @abbey-titcomb about making sure the community is up to date

juliendonck:
@geigerzaehler it's crashing on the license headers. Any tips on how to resolve this?

geigerzaehler:
> @geigerzaehler the tests are exiting on the license headers. Any tips on how to resolve this?

You can run `./scripts/license-header.ts add` or manually copy the license header from another svelte file.

ui/Screen/Network.svelte
```diff
@@ -0,0 +1,180 @@
...
+          style="color: var(--color-foreground-level-6); margin-bottom: 2v4px;">
+          Enter seed addresses that you’d like to connect to here.
+          <a
+            style="color: var(--color-foreground-level-5);"
+            class="typo-link"
+            href="https://docs.radicle.xyz/docs/understanding-radicle/glossary#seed"
+            >Learn more</a>
+        </p>
+      </div>
+      <form
```
geigerzaehler, line 145: Looks to me like there is some margin missing between the explanation and the input field.

Bildschirmfoto vom 2021-06-25 10-00-22

ui/Screen/Network.svelte
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  section {
+    align-items: center;
+    margin-bottom: 24px;
+    padding: 0 12px;
+  }
+
+  .seed-entry-form {
+    flex: 1;
```
geigerzaehler, line 64: This property can be removed. The parent container is not a flex box.

ui/Screen/Network.svelte
```diff
@@ -0,0 +1,180 @@
...
+<style>
+  .container {
+    max-width: var(--content-max-width);
+    margin: 64px auto;
+    min-width: var(--content-min-width);
+    padding: 0 var(--content-padding);
+  }
+
+  section {
+    align-items: center;
```
geigerzaehler, line 58: This property can be removed. The element is not a flex box.

ui/Screen/Network.svelte
```diff
@@ -0,0 +1,180 @@
...
+  .seed-entry-field {
+    width: 100%;
+    display: flex;
+    align-items: flex-start;
+  }
+
+  .seeds {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: flex-end;
```
geigerzaehler, line 79: To accomplish a vertical layout I’d suggest
```suggestion
    display: flex;
    flex-direction: column;
```

ui/src/network.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+interface IndicatorState {
+  text: string;
+  cy: string;
+  icon: typeof SvelteComponent;
+  fill: string;
+}
+
+export const indicatorState = (status: Status): IndicatorState | undefined => {
```
geigerzaehler, line 34: We want to omit the `undefined` case from the return type. It should never happen because you covered all status types. Unfortunately Typescript is not smart enough (yet) to figure this out on its own so we need to add the following `else` statement at the end
```typescript
} else {
  return unreachable(status)
}
```
with `import { unreachable } from `ui/src/unreachable`. Then you can remove `undefined` from the return type.

ui/Screen/Network.svelte
```diff
@@ -0,0 +1,180 @@
...
+
+  let statusText: string = "";
+  let statusFill: string = "";
+  $: {
+    const state = indicatorState($status);
+    if (state) {
+      statusText = state.text;
+      statusFill = state.fill;
+    }
+  }
```
geigerzaehler, line 46: There’s a more concise alternative to this. But I’m also ok with keeping it as it it.

```typescript
import * as svelteStore from "svelte/store";

const indicatorStatus = svelteStore.derived(status, indicatorState);
```
Then you can use `$indicatorStatus.text` instead of `statusText` etc.

This depends on fixing the return type of `indicatorState` as mentioned in the comment above.

geigerzaehler (commented):
Very nice!

ui/Screen/Network.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/2066/commits/841cb0998cff274d0adfc34bb2b6b6bb458a8df9

ui/src/network.ts
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/2066/commits/841cb0998cff274d0adfc34bb2b6b6bb458a8df9

ui/Screen/Network.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/2066/commits/841cb0998cff274d0adfc34bb2b6b6bb458a8df9

ui/Screen/Network.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/2066/commits/841cb0998cff274d0adfc34bb2b6b6bb458a8df9

ui/Screen/Network.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/2066/commits/841cb0998cff274d0adfc34bb2b6b6bb458a8df9

ui/Screen/Network.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/2066/commits/841cb0998cff274d0adfc34bb2b6b6bb458a8df9

juliendonck:
@rudolfs @geigerzaehler struggling bit with the tests, could you take a look?

rudolfs:
> @rudolfs @geigerzaehler struggling bit with the tests, could you take a look?

https://github.com/radicle-dev/radicle-upstream/pull/2066/commits/cf9277d084e005847dcedf95b24cf3e0b59c10f5

rudolfs (approved):

PR #2071, opened 2021-06-24
refactor: use dayjs for time formatting

geigerzaehler:
We use `dayjs` to format dates and times instead of our custom formatters. This reduces the amount of code we need to write and also eliminates some bugs already present in our code, like not including leading zeros in minutes and seconds (e.g. “17:3:5” instead of “17:03:05”).


brandonhaslegs:
Nice ⏰

brandonhaslegs:
It's about time!

juliendonck (approved):
🚀

rudolfs (approved):
utACK

PR #2085, opened 2021-06-29
improvement: disable `anchor` button on pending anchor

### LINKED ISSUE #2079: Disable `Anchor project` button if there is a pending anchor
If multiple anchors are pending, only one will be executable. This means that if two people are anchoring a project, the first one to execute will remove the other one that was pending. To avoid confusion, we could disable the `anchor project` button when there is already a pending anchor.

juliendonck:
closes #2079 

### DIFF
--- a/ui/Screen/Org.svelte
+++ b/ui/Screen/Org.svelte
@@ -89,6 +89,14 @@
       },
     ];
   };
+  console.log(activeTab);
+
+  let pendingAnchor: boolean = false;
+  $: {
+    pendingAnchor = activeTab.anchoredProjects.some(
+      proj => proj.anchor.type === "pending"
+    );
+  }
 </script>
 
 <SidebarLayout>
@@ -113,7 +121,8 @@
         <ProjectsMenu
           orgAddress={address}
           {gnosisSafeAddress}
-          disabled={activeTab.projectCount === 0} />
+          disabled={activeTab.projectCount === 0 || pendingAnchor}
+          {pendingAnchor} />
       {:else if activeTab.type === "members"}
         <MembersMenu {gnosisSafeAddress} />
       {:else}
--- a/ui/Screen/Org.svelte
+++ b/ui/Screen/Org.svelte
@@ -89,7 +89,6 @@
       },
     ];
   };
-  console.log(activeTab);
 
   let pendingAnchor: boolean = false;
   $: {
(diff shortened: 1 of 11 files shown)


ui/Screen/Org.svelte
```diff
@@ -89,6 +89,14 @@
       },
     ];
   };
+  console.log(activeTab);
```
rudolfs, line 92: 👀

ui/Screen/Org.svelte
```diff
@@ -113,7 +121,8 @@
         <ProjectsMenu
           orgAddress={address}
           {gnosisSafeAddress}
-          disabled={activeTab.projectCount === 0} />
+          disabled={activeTab.projectCount === 0 || pendingAnchor}
```
rudolfs, line 123: Now that the logic to disable this button is a bit more complex we should just pass in project count to the `ProjectsMenu` component as a prop and do all the logic inside the component.

ui/Screen/Org.svelte
```diff
@@ -89,6 +89,13 @@
       },
     ];
   };
+
+  let pendingAnchor: boolean = false;
+  $: {
+    pendingAnchor = activeTab.anchoredProjects.some(
+      proj => proj.anchor.type === "pending"
```
rudolfs, line 96: We should match against the type, not an arbitrary string here.

```suggestion
      proj => proj.anchor.type === project.PendingAnchor
```

And import the type from project at the top:
```
import type * as project from "ui/src/project"
```

ui/Screen/Org.svelte
juliendonck: https://github.com/radicle-dev/radicle-upstream/pull/2085/commits/8f523c9e166bcdf4b4f99eb94b50bb0580ec17c7

ui/Screen/Org.svelte
rudolfs: Yea, disregard this. It doesn't make sense.

rudolfs:
Pushed a fix that covers pending anchors in resolved projects as well as unresolved: https://github.com/radicle-dev/radicle-upstream/pull/2085/commits/e88e9d9c05b9b142d0006144d1c7f09e3ee20750.

Screenshot 2021-06-29 at 19 18 45

Screenshot 2021-06-29 at 19 26 37

rudolfs (approved):

PR #2088, opened 2021-06-30
improvement: differentiate pending anchors in orgs

### LINKED ISSUE #2078: Differentiate pending anchored projects with more than just icon color
When a user signs an anchor transaction but doesn't execute the transaction yet, the project appears to show up as if it was already anchored. We should make this more clear that it is in fact still pending.

juliendonck:
closes #2078

### DIFF
--- a/ui/Screen/Org/Projects.svelte
+++ b/ui/Screen/Org/Projects.svelte
@@ -10,8 +10,6 @@
 
   import UnresolvedAnchorList from "ui/Screen/Org/UnresolvedAnchorList.svelte";
 
-  import type * as project from "ui/src/project";
-
   import * as router from "ui/src/router";
   import * as sess from "ui/src/session";
   import * as org from "ui/src/org";
@@ -21,8 +19,7 @@
   const session = sess.getUnsealedFromContext();
 
   export let address: string;
-  export let anchoredProjects: project.Project[];
-  export let unresolvedAnchors: project.Anchor[];
+  export let anchors: org.OrgAnchors;
   export let gnosisSafeAddress: string;
   export let disableAnchorCreation = false;
 
@@ -33,22 +30,6 @@
       urn: project.urn,
     });
   };
-
-  let pendingResolved: project.Project[];
-  let confirmedResolved: project.Project[];
-  let pendingUnresolved: project.Anchor[];
-  let confirmedUnresolved: project.Anchor[];
-
-  $: {
-    pendingResolved = anchoredProjects.filter(
-      p => p.anchor && p.anchor.type === "pending"
-    );
-    confirmedResolved = anchoredProjects.filter(
-      p => p.anchor && p.anchor.type === "confirmed"
-    );
-    pendingUnresolved = unresolvedAnchors.filter(a => a.type === "pending");
-    confirmedUnresolved = unresolvedAnchors.filter(a => a.type === "confirmed");
-  }
 </script>
 
 <style>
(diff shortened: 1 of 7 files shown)


ui/Screen/Org/Projects.svelte
```diff
@@ -33,6 +33,22 @@
...
+  $: {
+    pendingResolved = anchoredProjects.filter(
+      p => p.anchor && p.anchor.type === "pending"
+    );
+    confirmedResolved = anchoredProjects.filter(
+      p => p.anchor && p.anchor.type === "confirmed"
+    );
+    pendingUnresolved = unresolvedAnchors.filter(a => a.type === "pending");
+    confirmedUnresolved = unresolvedAnchors.filter(a => a.type === "confirmed");
+  }
```
rudolfs: This should all go into the `org` module: daef4a96da3729437c930eaaf7aa806f65b8dde6.

ui/Screen/Org/Projects.svelte
```diff
@@ -41,20 +57,65 @@
...
-    <UnresolvedAnchorList anchors={unresolvedAnchors} />
+  {#if pendingResolved.length !== 0 || pendingUnresolved.length !== 0}
+    <div class="pending">
+      <div class="header">
+        <p class="typo-text-bold">Pending</p>
+        <p style="margin-left: .5rem; color: var(--color-foreground-level-6);">
+          Not enough of the members have signed this anchor.
+        </p>
+      </div>
+      {#if pendingResolved.length !== 0}
```
rudolfs: We don't have to wrap lists in conditionals. If there are no elements in the array, the list won't be rendered. Fixed in 5518ed210d262ff6090d31f4369114c76e471e3a.

ui/Screen/Org/Projects.svelte
```diff
@@ -41,20 +57,65 @@
...
+          projects={pendingResolved}
+          userUrn={session.identity.urn}
+          on:select={select} />
+      {:else if pendingUnresolved.length !== 0}
+        <UnresolvedAnchorList anchors={pendingUnresolved} />
+      {/if}
+    </div>
+  {/if}
+  {#if confirmedResolved.length !== 0 || confirmedUnresolved.length !== 0}
+    {#if pendingResolved.length !== 0 || pendingUnresolved.length !== 0}
```
rudolfs: There's a logic bug here. If there are no confirmed anchors but there are pending ones, then the pending ones would not be shown. Fixed in 5518ed210d262ff6090d31f4369114c76e471e3a.

ui/Screen/Org/Projects.svelte
```diff
@@ -41,20 +57,65 @@
...
+      <div class="header">
+        <p style="color: var(--color-foreground-level-6);">
+          These anchored projects haven't been found in your network yet, try
+          following them.
+        </p>
+      </div>
+      <UnresolvedAnchorList anchors={confirmedUnresolved} />
+    {/if}
   {:else}
     <EmptyState
```
rudolfs, line 94: This should only be shown if there are no anchors at all. Fixed in 5518ed210d262ff6090d31f4369114c76e471e3a.

Screenshot 2021-07-01 at 09 45 50

rudolfs (commented):
👍 

There's one small visual issue with the popover:

Screenshot 2021-07-01 at 09 04 54

ui/Screen/Org/Projects.svelte
juliendonck: Hmm, I was trying to show the header only when there are pending anchors. So that when there are no pending, there is also no title.

juliendonck:
> There's one small visual issue with the popover:
> 
> Screenshot 2021-07-01 at 09 04 54

Fixed https://github.com/radicle-dev/radicle-upstream/pull/2088/commits/3ce4ed9af18f7705c0af2c9dded89f5717c27ac6

brandonhaslegs:
Instead of "Confirmed" should it say "Anchored"?

rudolfs (approved):

PR #2103, opened 2021-07-07
refactor: isolate WalletConnect instance creation

geigerzaehler:
We lift `newWalletConnect()` to the module-level so that it does not interact with internal `Wallet` state. Before, callbacks for the modal would set `stateStore` and internal state variables. This significantly simplifies the connection flow.


CodeSandwich (approved):
Great! You must've experimented a lot with WalletConnect undocumented behavior to support these subtleties :detective:

PR #2104, opened 2021-07-07
fix: wallet disconnect handling

geigerzaehler:
* We don’t call `disconnect()` multiple times when the user disconnects. The second call to `disconnect()` was triggered by the `disconnect` event.
* We properly register the `disconnect` event even after wallet connect is initialized.


CodeSandwich (approved):
:broom:

geigerzaehler:
Seems like this doesn’t work properly. When I disconnect from upstream and reconnect again I’m not presented with the QR modal again. I’ll close this and tackle this properly in an upcoming PR.

PR #2107, opened 2021-07-08
chore: update cargo dependencies to resolve advisory

geigerzaehler:
We update all cargo dependencies to resolve https://rustsec.org/advisories/RUSTSEC-2021-0072


FintanH (approved):
Very good :clap:

Issue #2002, opened 2021-06-16
Ethereum network in Upstream should be set to the wallet's network


rudolfs:
At the moment we disable certain features if the Ethereum network in Upstream is set to something different than what the user has selected in their connected wallet.

<img width="999" alt="Screenshot 2021-06-16 at 11 08 47" src="https://user-images.githubusercontent.com/158411/122191666-3a62cd00-ce93-11eb-9fba-ae081b7f4379.png">

Instead Upstream's Ethereum network setting should just follow whatever is set in the user's wallet to prevent having a mismatch in the first place.

geigerzaehler:
To implement this change we would need to remove `selectedEnvironment` from `ui/src/ethereum` and use `WalletConnect#chainId` instead. `selectedEnvironment` is only used by `ui/src/wallet`. It reacts to changes and re-builds the wallet based on the environment. The environment is used to instantiate an ethers provider and obtain addresses for all the contracts we interact with.

The main challenge of using `WalletConnect#chainId` instead of the selected environment is that at the moment the environment is used before WalletConnect is instantiated and before we know the chain ID. To address this it might be necessary move the contract address resolution into the contract facade objects and do it dynamically on every call. Similarly, for we would need to change `WalletConnectSigner` so that we can dynamically set the provider instead of passing it on construction.

Finally, we need to keep an option to use a local Ethereum node for development with ganache. A command line flag or environment variable should be enough for that.

I estimate that it takes three days to do this.

geigerzaehler:
After investigating this further I came to the conclusion that it’s not a good idea to have the connected wallet control the network while the app is running. This setup opens the door to a lot of issues caused by switching the network while some request is in progress.

I suggest we keep control of the network within the app. We can however make some improvements to the UX. We could move the network selection to the wallet screen. We could also use the network from the wallet when it is initially connected. To change the network, the user would need to reconnect to the wallet again.

rudolfs:
> After investigating this further I came to the conclusion that it’s not a good idea to have the connected wallet control the network while the app is running. This setup opens the door to a lot of issues caused by switching the network while some request is in progress.


The `ethers` docs mention reloading the UI on a network change as best practice. Maybe we could do that?

> So, the best practice when a network change occurs is to simply refresh the page. This should cause all your UI components to reset to a known-safe state, including any banners and warnings to your users if they are on an unsupported network.

[...]

PR #2111, opened 2021-07-08
test: add instructions for manually testing WalletConnect

geigerzaehler:
Add instructions for manually testing WalletConnect.

### DIFF
--- a/docs/wallet-connect-testing.md
+++ b/docs/wallet-connect-testing.md
@@ -1,0 +1,68 @@
+# WalletConnect Testing
+
+This document describes how to test our WalletConnect integration.
+
+As a wallet you can use the Use [Metamask][] or [Rainbow][] mobile apps or the
+WalletConnect [browser test wallet][test-wallet].
+
+## Connect a wallet
+
+1. Go to wallet screen in the app.
+2. Ensure you are disconnected. If that is not the case, disconnect and refresh
+   the app.
+3. Click “Connect your wallet”.
+4. Scan the QR code or paste the link into the wallet.
+5. Approve the connection in the wallet.
+6. The Wallet screen should show the correct address from the wallet
+
+## Disconnect from the app
+
+1. Ensure that you are connected with a wallet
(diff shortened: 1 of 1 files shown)


geigerzaehler:
- [ ] Add wallet network switch to the test case

docs/wallet-connect-testing.md
```diff
@@ -0,0 +1,68 @@
...
+4. Reconnect with the wallet again.
+
+## Close connection modal
+
+1. Ensure you are disconnected.
+2. Click on “Connect your wallet”.
+3. Close the modal by clicking outside of it.
+4. Click on “Connect your wallet” and reconnect with a wallet again.
+
+## Switch network
```
CodeSandwich, line 42: Do we need a similar test for switching the network in upstream settings and seeing what happens?

CodeSandwich (approved):
Looks good! There's no test for actually signing and sending a transaction, but I guess that other manual tests will cover that?

geigerzaehler:
> There's no test for actually signing and sending a transaction, but I guess that other manual tests will cover that?

Yes, I was focusing on the connection for the tests. We didn’t have a lot of issues with signing and the code is pretty straight-forward.

docs/wallet-connect-testing.md
geigerzaehler: The code that handles upstream network changes does not involve wallet connect. This is something we can test automatically later.

geigerzaehler: > are they valuable in the first place?

It’s indeed dubious whether this is valuable. We want to get rid of this anyways https://github.com/radicle-dev/radicle-upstream/issues/2002

> Maybe mocking WC could allow that thinking

Yes, that is what I’m working on.

rudolfs:
When would we use these? Should they be part of the QA procedure?

geigerzaehler:
> When would we use these? Should they be part of the QA procedure?

I don’t think we necessarily need to do them as part of release QA. I was running through this whenever I changed something related to wallet connect.

PR #2114, opened 2021-07-09
fix: correct radicle-contracts types

geigerzaehler:
We correct the all the radicle-contracts code that failed to type-check after the types were properly included in https://github.com/radicle-dev/radicle-contracts/pull/150.


CodeSandwich (approved):
Nice :ok_hand: We could update to an even fresher radicle-contracts hash to include more improvements.

geigerzaehler:
> We could update to an even fresher radicle-contracts hash to include more improvements.

Done!

PR #2115, opened 2021-07-09
fix: race condition when watching claim updates

geigerzaehler:
We ensure that the `onClaimed` callback is called in the same order as the events that arrive. Before it was possible for `onClaimed` to be called for an event Y that occured _after_ the most recent event X _before_ `onClaimed` was called for `X`. This left the application in a state where the claim from `X` was considered more recent. This behavior was triggered by writing tests against a local Ethereum node.

### DIFF
--- a/ui/src/mutexExecutor.ts
+++ b/ui/src/mutexExecutor.ts
@@ -4,17 +4,19 @@
 // with Radicle Linking Exception. For full terms see the included
 // LICENSE file.
 
+import * as Bacon from "ui/src/bacon";
+
 // A task executor that runs only one task concurrently. If a new task
 // is run, any previously running task is aborted and the promise
 // returned from `run()` will return undefined.
 //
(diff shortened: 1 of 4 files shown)


ui/src/mutexExecutor.ts
```diff
@@ -4,36 +4,60 @@
...
+//
+//     import * as mutexExecutor from "ui/src/mutexExecutor"
+//     const worker = mutexExecutor.createWorker(async (value) => {
+//       await sleep(1000)
+//       return value
+//     })
+//
+//     const firstPromise = worker.output.firstToPromise()
+//     worker.push("first)
+//     assert.equal(await first, "first")
```
CodeSandwich, line 49:
```suggestion
//     assert.equal(await firstPromise, "first")
```

CodeSandwich (approved):

PR #2118, opened 2021-07-12
fix: improved error message from server

geigerzaehler:
Whenever we encounter an unmatched error we respond with a 500 instead of a 400 and also send the error message to better identify the underlying issue in user reports.


FintanH (approved):
Makes sense :+1:

PR #2121, opened 2021-07-12
chore: update rust toolchain to 2021-06-17

geigerzaehler:
Also fixes new clippy lints


proxy/api/src/http/session.rs
```diff
@@ -46,6 +46,7 @@ mod handler {
     use crate::{context, error, http, session};
 
     /// Fetch the [`session::Session`].
+    #[allow(clippy::unused_async)]
```
FintanH, line 49: Any reason to allow these rather than changing the signature?

FintanH (approved):

proxy/api/src/http/session.rs
FintanH: Thanks for explaining :)

PR #2123, opened 2021-07-14
refactor: wallet connect facade

geigerzaehler:
We introduce a `WalletConnect` interface that replaces the WalletConnect `Connector` in `ui/src/wallet.ts`. This interface serves to purposes

1. We move all WalletConnect specific quirks behind its implementation.
2. We can implement this interface to simulate WalletConnect in tests.

### DIFF
--- a/ui/src/ethereum/walletConnect.ts
+++ b/ui/src/ethereum/walletConnect.ts
@@ -1,0 +1,184 @@
+// Copyright © 2021 The Radicle Upstream Contributors
+//
+// This file is part of radicle-upstream, distributed under the GPLv3
+// with Radicle Linking Exception. For full terms see the included
+// LICENSE file.
+
+// This module provides a facade for WalletConnect that makes it easier to use.
+
+import { Mutex } from "async-mutex";
+import Connector from "@walletconnect/client";
+import type { ITxData } from "@walletconnect/types";
+
+import * as Error from "ui/src/error";
+import * as modal from "ui/src/modal";
+import ModalWalletQRCode from "ui/Modal/Wallet/QRCode.svelte";
+import * as Bacon from "ui/src/bacon";
+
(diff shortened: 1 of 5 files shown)


ui/src/ethereum/walletConnect.ts
```diff
@@ -0,0 +1,179 @@
...
+
+// Data provided by a connected wallet
+export interface Connection {
+  accountAddress: string;
+  chainId: number;
+}
+
+export interface WalletConnect {
+  // Emits an event when the wallet is disconnected. The disconnection
+  // can may be triggered by the app or by the wallet.
```
rudolfs, line 26:
```suggestion
  // may be triggered by the app or by the wallet.
```

ui/src/ethereum/walletConnect.ts
```diff
@@ -0,0 +1,179 @@
...
+  sendTransaction(tx: ITxData): Promise<string> {
+    return this.connector.sendTransaction(tx);
+  }
+
+  signTransaction(tx: ITxData): Promise<string> {
+    return this.connector.sendTransaction(tx);
+  }
+
+  // We need to reinitialize `WalletConnect` until this issue is fixed:
+  // https://github.com/WalletConnect/walletconnect-monorepo/pull/370
```
rudolfs, line 121: This PR seems to be merged now, do we still need to re-init now?

geigerzaehler: Yes, I’ve updated the comment. in https://github.com/radicle-dev/radicle-upstream/pull/2123/commits/0e406d57a60c2de4021960f4043e352ddd9eb39a

CodeSandwich (approved):
Nice refactoring :+1:

rudolfs (approved):
utACK

PR #2151, opened 2021-07-20
fix(tests): forward `stdout` of test node to console

geigerzaehler:
We forward the `stdout` of nodes started by the tests to the terminal that cypress runs on. This ensures that it is visible.


rudolfs (approved):
utACK

PR #2152, opened 2021-07-20
refactor: consolidate org route loading

geigerzaehler:
* Instead of having much of the code that loads the org route in the global routing definition we move it to `src/Screen/Org/route.ts`.

* We introduce the org `Owner` type and remove `OrgWithSafe` to better represent owner information and to distinguish between the safe and an externally owned account.


rudolfs (approved):
Nice!

PR #2153, opened 2021-07-20
refactor: consolidate cypress test config

geigerzaehler:
We move detection of the cypress test environment to `ui/src/config` so it can be reused. We also update the name of the similar configuration to better reflect what it does.


rudolfs (approved):
utACK

PR #2162, opened 2021-07-21
fix: reduce infura polling time to once every minute

rudolfs:
Given the current way we do requests this would result to ~2350 req/hour per client.
Before this change it was ~5960req/hour per client.

At our current 100k/day limit that would support 42 clients.

### DIFF
--- a/ui/src/transaction.ts
+++ b/ui/src/transaction.ts
@@ -21,12 +21,6 @@ import { Icon } from "ui/DesignSystem";
 // The store where all managed transactions are stored.
 export const store = persistentStore<Tx[]>("transactions", []);
 
-// Periodically refresh the status of all ongoing transactions.
-const POLL_INTERVAL_MILLIS = 3000;
-setInterval(() => {
-  updateStatuses();
-}, POLL_INTERVAL_MILLIS);
-
 export type Tx = TxData & MetaTx;
 
 // The data shared across all types of transactions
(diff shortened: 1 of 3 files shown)


geigerzaehler (approved):

ui/src/transaction.ts
```diff
@@ -210,11 +204,28 @@ function txData(txc: ContractTransaction, date: number = Date.now()): TxData {
...
+  if (tx.status === TxStatus.AwaitingInclusion) {
+    const provider = svelteStore.get(walletStore).provider;
+    provider.waitForTransaction(tx.hash).then(txReceipt => {
+      store.update((txs: Tx[]) => {
+        return txs.map(tx_ => {
+          if (tx_.hash === tx.hash) {
+            tx_.status = status(txReceipt);
+          }
+          return tx_;
+        });
```
geigerzaehler, line 217: Not that it matters much but `find` and in-place mutation is faster
```suggestion
        const tx_ = txs.find(tx_ => tx_.hash === tx.hash)
        if (tx_) {
          tx_.status = status(txReceipt);
        }
        return txs
```

ui/src/transaction.ts
```diff
@@ -210,11 +204,28 @@ function txData(txc: ContractTransaction, date: number = Date.now()): TxData {
   };
 }
 
+function registerTxUpdateCallback(tx: Tx): void {
+  if (tx.status === TxStatus.AwaitingInclusion) {
+    const provider = svelteStore.get(walletStore).provider;
+    provider.waitForTransaction(tx.hash).then(txReceipt => {
```
geigerzaehler, line 210: We should also catch errors and log them with an appropriate error message.

rudolfs: 92945be8d8d15e8943902e3ef25c2c64de5d3601

ui/src/transaction.ts
```diff
@@ -207,15 +207,25 @@ function txData(txc: ContractTransaction, date: number = Date.now()): TxData {
...
+            tx_.status = status(txReceipt);
+          }
+          return txs;
+        });
+      })
+      .catch(err => {
+        error.show(
+          new error.Error({
+            message: "Failed to update transaction status",
+            source: error.fromJsError(err),
```
geigerzaehler, line 225: The following is better. It also handles cases were `err` is not a JS `Error`

```suggestion
            source: err,
```

geigerzaehler:
Merged in 669273290057ba28f035e3b4a89b256235b2d842

PR #2180, opened 2021-07-23
feat: Naming orgs

efstajas:
Frame 876

This PR enables the registration of new and editing of existing ENS names, allowing users to link rich identities with org names, URLs, avatars & social profiles to their orgs. Users can also update an existing ENS name, link a name that they already own to an org, and re-link an already named org with another name.

### Usecases
- [ ] Support multi-sig orgs. This requires changes to the `LinkOrgToName` step and the succeeding success step's copy, and maybe also some kind of indicator on the multisig org page that indicates that an org link transaction is currently pending approval.
- [x] Register a new ENS name if the user enters a non-existing name.
- [x] Link a name that already points to another org to an org. This should display a warning.
- [x] If you enter a name that is already linked to the current org, just allow updating name metadata and skip the linking step.
- [x] If you enter a name you already own but that is not linked to another org, skip name creation and then allow setting the name's address value to the current org and linking the org to the name.

### Edge cases
- [x] Display an error if the entered name is already registered by someone else.
- [x] Display a warning if the entered name is currently already pointing to a different org.
- [x] After linking a name to a different org than before, ensure that the other org shows up without a name afterward.
- [x] Fetch registration and allow updating metadata when entering a name that has already been registered by the user previously.
- [x] Skip the organization link step when the organization is already linked to the entered name.
- [ ] It may be a good idea to fetch the image URL entered in the `populateMetadata` step and check against a max allowed filesize in order to prevent users setting massive images as their org avatar.
- [ ] Persist a commitment including name & salt in `localStorage` as soon as submitted in order to be able to resume it should the user lose connection or drop out after spending the RAD fee.

### UX & Design
- [x] Fix incorrect alignment of loading / validation error icons on `TextInput` when a label is set
- [x] Display avatar image on org page & sidebar
- [x] Display URL on org page if set
- [ ] Display social media links if set
- [ ] When an org already has a name, we probably want to show an "Edit or link other ENS name" button in the overflow menu instead of the current "Register ENS name" button.
- [ ] We probably want to show a list of any already registered .radicle.eth ENS names for the current user on the `enterName` step in addition to the text field for creating a new name.
- [ ] This PR introduces the first actual RAD transaction in upstream. Probably, it should be released together with adding support for RAD balance & transactions to the wallet page.
- [ ] When upstream is aware of RAD balance, it should check the available balance before starting the commitment in order to avoid the user spending a gas fee but then running into a failure because they don't have enough RADs to cover the name commitment fee.
- [ ] Introduce a way to retry a transaction / signature request (plus maybe a timeout?) in case WalletConnect acts funny and the registration confirmation prompt doesn't appear in the wallet. Currently the button just gets locked in "Waiting for confirmation..." and it's unclear how to retry.
- [ ] In the beginning, to create a new name, the user first needs to confirm a signature and then submit the commitment transaction. Currently, this is presented as a single step, but the UI should probably give feedback after the signature to ensure the user doesn't miss the second transaction prompt and wonder why nothing's happening.
- [ ] We should probably be upfront with the RAD fee required for name registration at the very beginning of the flow.
- [ ] Review & improve design of the steps. They currently look quite bland.

### Behind the scenes
- [ ] Don't fetch name data for each org as frequently as we do right now, as that causes way too many Infura calls.
- [ ] May want to improve state management on the `registerEnsNameFlow` component, as it's currently a bit all over the place.
- [x] Use auto-generated types and contract factories in `ensRegistrar.ts`.
- [ ] Refactor ensRegistrar.ts to not include any usecase-specific logic, instead only wrapping the registrar contract's functions. Move usecase-specific logic to the components that import registrar contract functions.

### DIFF
--- a/scripts/license-header.ts
+++ b/scripts/license-header.ts
@@ -8,7 +8,7 @@
 
 import globby from "globby";
 import yargs from "yargs";
-import * as fs from "fs/promises";
+import { promises as fs } from "fs"
 import * as Path from "path";
 
 // Error that is shown without a stacktrace to the user
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,11 +4,11 @@
   "exclude": ["native/bundle.js", "node_modules/*", "public/*"],
   "compilerOptions": {
     /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
-    "target": "ES2018",
+    "target": "ES2020",
     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
     "module": "es6",
     /* Specify library files to be included in the compilation. */
-    "lib": ["dom", "es6", "es7"],
+    "lib": ["dom", "es6", "es7", "es2020"],
     /* Allow javascript files to be compiled. */
     "allowJs": true,
     /* Report errors in .js files. */
--- a/ui/App.svelte
+++ b/ui/App.svelte
@@ -134,6 +134,7 @@
     <NetworkDiagnostics activeTab={$activeRouteStore.activeTab} />
   {:else if $activeRouteStore.type === "singleSigOrg"}
     <SingleSigOrg
+      registration={$activeRouteStore.registration}
       address={$activeRouteStore.address}
       owner={$activeRouteStore.owner}
       projectCount={$activeRouteStore.projectCount}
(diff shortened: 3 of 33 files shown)


geigerzaehler:
Really impressive work @efstajas. For us to be able to continue with this could you rebase and squash your commits? Could you also [sign off your commit](https://github.com/radicle-dev/radicle-upstream/blob/master/CONTRIBUTING.md#certificate-of-origin) according to the DCO.

efstajas:
@geigerzaehler Done!

rudolfs:
Closing this in favour of https://github.com/radicle-dev/radicle-upstream/pull/2191.

PR #2185, opened 2021-07-27
ci: bump github actions cache

geigerzaehler:
We bump the version of the yarn cache on Github actions. The old cache was 1.3GiB in size because it held previous versions of Cypress.


juliendonck (approved):
I don't know what it does or how to test, but this code looks sane 🤷‍♂️

PR #2196, opened 2021-08-09
fix: ignore security advisory that we’re unaffected by

geigerzaehler:
The vulnerability is in the `tar` crate that is only used by `radicle-surf` as a build dependency.


rudolfs (approved):
utACK

Issue #1874, opened 2021-05-10
Display user profiles


brandonhaslegs:
Users should be able to view the profiles of other users. At the moment they can only view their own profile.

- Show the user's projects
- Show the projects the user is following
- Copy link to profile
- Copy user's Device ID
- Block a user (choose to not replicate this user's content and profile)

![image](https://user-images.githubusercontent.com/4406983/117641012-f8ed4c80-b185-11eb-9f1c-47370f3543b3.png)

PR #2197, opened 2021-08-09
Enable other user profiles

rudolfs:
Closes: https://github.com/radicle-dev/radicle-upstream/issues/1874.

Screenshot 2021-08-11 at 10 22 23

Screenshot 2021-08-11 at 10 21 06

Screenshot 2021-08-11 at 10 20 44

Screenshot 2021-08-11 at 10 20 28


ui/src/userProfile.ts
```diff
@@ -24,8 +26,22 @@ export const fetchProjects = (urn: string): void => {
...
+    .remoteIdentityGet(urn)
     .then(userStore.success)
     .catch(err => userStore.error(error.fromUnknown(err)));
 };
+
+export const openUserProfile = (
+  userUrn: string,
+  sessionIdentityUrn: string
+): void => {
+  if (userUrn === sessionIdentityUrn) {
```
geigerzaehler, line 39: As an alternative, we could get the session identity URN from the `ui/src/session` module. The advantage would be that we don’t need to obtain it first every time we want to call this function.

ui/Screen/Org/Members.svelte
```diff
@@ -6,10 +6,22 @@
...
+  import type * as project from "ui/src/project";
+  import * as userProfile from "ui/src/userProfile";
+  import * as Session from "ui/src/session";
+
+  import { Avatar, List, StyledCopyable } from "ui/DesignSystem";
 
   export let members: org.Member[];
+  const session = Session.getUnsealedFromContext();
+
+  function onSelect({ detail: member }: { detail: project.User }) {
```
geigerzaehler, line 19: Just a small suggestion: I find it easier to follow if functions are named by what they do and not by what they are called. I’d suggest calling this `openUserProfile` and having `on:select={openUserProfile}`.

geigerzaehler (commented):
Could you add some tests for this?

ui/Screen/Org/Members.svelte
rudolfs: a01379e1474e3189bb7c25994ab2ce5803affedd

proxy/api/src/identity.rs
```diff
@@ -92,6 +105,8 @@ pub struct Metadata {
     pub handle: String,
     /// The user's Ethereum address.
     pub ethereum: Option<Ethereum>,
+    /// The user's PeerIds.
+    pub peer_ids: Option<Vec<PeerId>>,
```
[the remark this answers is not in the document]
rudolfs: I tried it without wrapping it in an `Option` initially. But then the app wouldn't start because the session endpoint expected `peer_ids` to be set in the metadata.

ui/src/userProfile.ts
rudolfs: I thought of this as well. But then I realised that we set the session via a context in a top-level component, however the modal is outside of that component, so when we try to get the session via `getUnsealedFromContext` it fails.

I think if we want to solve this, we'll have to refactor how we pass the session around. It might make sense to just have a global session store instead of the context. This would be a bigger rewrite though.

geigerzaehler (approved):
:desktop_computer:

geigerzaehler (commented):
Wait a second... This is still lacking some tests ;)

rudolfs:
> Wait a second... This is still lacking some tests ;)

Working on it ; P

rudolfs:
@geigerzaehler is this sufficient?

4610a62b0f19c90c320baf63e25dbb953b73f1c7
0f4ab703d70840c73b6f60e336138a78d0a19b32

geigerzaehler (approved):
Nice work

PR #2199, opened 2021-08-09
refactor: reduce dependency on ui code

geigerzaehler:
We move UI logic from the `org` module to the org svelte components.
This allows us to revert changes to the `ThreeDotsMenu` component that
made it more complicated.


rudolfs (approved):
🧹

PR #2203, opened 2021-08-09
fix: don’t use stale wallet

geigerzaehler:
Before, some of the functions in `ensRegistrar` would use a style `Wallet` value because it was set at the module level. Now, we get the most recent wallet when the functions are called.

Since the functions are inherently stateful, we don’t require functions to pass down the `environment` parameter and get it from the wallet instead.


rudolfs (approved):
Nice! I also had this in my todo list.

PR #2204, opened 2021-08-09
refactor: align use of setName for org contract

geigerzaehler:
We rewrite `org.setNameMultisig` and `org.setSingleSigName` to better align with project anchoring. This means we make the code in `ui/src/org/contract` stateless, i.e. independent of the wallet. We consolidate the stateful code in `ui/src/org`. We also hardcode the ENS addresses since they might not always be available from the provider.


rudolfs (approved):
🧹

PR #2206, opened 2021-08-09
feat: add `suffix` property to text input

geigerzaehler:
This will be needed by the ENS registration. Note that we made a small adjustment to the code from #2191.


rudolfs (approved):
utACK

PR #2208, opened 2021-08-10
refactor: remove unused shareable_entity_identifier attribute

rudolfs:
It was never used. On the first public release we opted for sharing project `URN`s and `DeviceId`s.


geigerzaehler (approved):
:broom: :heart:

geigerzaehler (approved):

PR #2211, opened 2021-08-11
fix: fix race conditions in source browsing tests

geigerzaehler:
We fix race conditions in source browsing tests by ensuring that abort signals for superseded requests are properly delivered.


rudolfs (approved):
🙇‍♂️

PR #2214, opened 2021-08-11
fix: tests wait until keystore is sealed

geigerzaehler:
We ensure that the tests wait until the proxy has been properly sealed before continuing.

To make this work we also need to ensure that the HTTP server shuts down properly by terminating any event streams.


rudolfs (approved):
💯

PR #2220, opened 2021-08-11
refactor: remove session from global context

rudolfs:
The session is already available via a global store, passing it in via a
context makes things more complex without providing any benefit.


ui/src/session.ts
```diff
@@ -48,10 +47,16 @@ sessionStore.subscribe(data => {
 });
 
 // Return the unseleased session if the session is unsealed, undefined otherwise.
-export const unsealed = (): UnsealedSession | undefined => {
+export const unsealed = (): UnsealedSession => {
```
geigerzaehler, line 50: Could you update the documentation of this function?

geigerzaehler (commented):
:heart:

ui/src/session.ts
rudolfs: 0fff2cefd7959ff357eb8cc0a3f10eee99aa5532

geigerzaehler (approved):

PR #2224, opened 2021-08-12
fix: silent retries if graph is unavailable

geigerzaehler:
We properly initialize the remaining tries when the graph is
unavailable.


rudolfs (approved):
utACK

PR #2225, opened 2021-08-12
fix: remove WalletConnect signer defaults

geigerzaehler:
We don’t add default values to sign transaction requests for wallet connect anymore. These values we’re only useful to work with the local network.


rudolfs (approved):
utACK

Issue #2222, opened 2021-08-12
Handle RAD tokens in Upstream


rudolfs:
At the moment Upstream only is aware of ETH and DAI tokens.
For [ENS name registration](https://github.com/radicle-dev/radicle-upstream/pull/2191) we'll need to check the RAD balance.

Hide the DAI balance if the token streams feature is not enabled.

<img width="1552" alt="Screenshot 2021-08-12 at 11 27 43" src="https://user-images.githubusercontent.com/158411/129173499-27a511fc-01d3-409d-99d6-485080963ecb.png">

geigerzaehler:
> Hide the DAI balance if the token streams feature is not enabled.

I’d be ok with always hiding the DAI balance since nobody uses the token streams at the moment.

rudolfs:
> I’d be ok with always hiding the DAI balance since nobody uses the token streams at the moment.

To be consistent, I'd rather hide it behind `isDev`, since it belongs to the token streams, which we also hide behind `isDev`. We can remove it entirely if we also remove that feature from the code, otherwise it'll be confusing when somebody will want to test it. 🤷

PR #2228, opened 2021-08-12
Show RAD token balance on wallet screen

rudolfs:
Required by https://github.com/radicle-dev/radicle-upstream/pull/2191.
Closes https://github.com/radicle-dev/radicle-upstream/issues/2222.

Screenshot 2021-08-12 at 14 31 24

### DIFF
--- a/ui/src/wallet/radToken.ts
+++ b/ui/src/wallet/radToken.ts
@@ -1,0 +1,40 @@
+// Copyright © 2021 The Radicle Upstream Contributors
+//
+// This file is part of radicle-upstream, distributed under the GPLv3
+// with Radicle Linking Exception. For full terms see the included
+// LICENSE file.
+
+import type * as ethers from "ethers";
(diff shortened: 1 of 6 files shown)


ui/src/wallet/radToken.ts
```diff
@@ -0,0 +1,40 @@
...
+      return addresses.mainnet;
+  }
+}
+
+export function connect(
+  signerOrProvider: ethers.Signer | ethers.providers.Provider,
+  address: string
+): ERC20 {
+  return Erc20Factory.connect(address, signerOrProvider);
+}
```
geigerzaehler, line 40: What do you think about taking the environment as the argument instead of the address. This way we don’t need to export `radTokenAddress` and it makes the call simpler.

rudolfs: 25da5de6e85916cd281f525d834b11e29d38751d

geigerzaehler (approved):

PR #2236, opened 2021-08-13
Merge enter name and register flows

rudolfs:
All of this should happen within one screen:

Screenshot 2021-08-13 at 19 01 52


ui/Modal/Org/ConfigureEns/RegisterName.svelte
```diff
@@ -0,0 +1,196 @@
...
+    } else {
+      validationStatus = {
+        status: validation.ValidationStatus.Loading,
+      };
+      if (timeoutHandle) {
+        clearTimeout(timeoutHandle);
+      }
+      timeoutHandle = window.setTimeout(() => {
+        validateFormAndQueueNextAction();
+      }, 1000);
```
geigerzaehler, line 81: Although we cancel the timeout there is still a potential race condition here: Assume the user types "x", waits more then a second, then types "y". This will call `validateFormAndQueueNextAction` twice—first, just before typing "y" and then a second later. Unfortunately it’s not guaranteed that the first async call finishes before the second call finishes. `ensRegistrar.isAvailable()` may take more then a second the first time and be very quick the second time around. This results in the UI being updated with the result of the first validation call for "x".

I suggest we use `lodash.debounce()` together with `MutexWorker` to guard against this. (`MutexExecutor` would work, too.) Happy to pair on this if you need support.

rudolfs: 5150e5d595096eec92d61958f4922514b63a4da3

ui/Modal/Org/ConfigureEns/RegisterName.svelte
```diff
@@ -0,0 +1,221 @@
...
+
+      return {
+        validationState: {
+          status: validation.ValidationStatus.Success,
+        },
+        registration: null,
+      };
+    }
+
+    registration = await ensResolver.getRegistration(
```
geigerzaehler, line 125: This mutates the component state variable and leads to a race.
```suggestion
    const registration = await ensResolver.getRegistration(
```

rudolfs: Good catch!

ui/Modal/Org/ConfigureEns.svelte
```diff
@@ -44,56 +42,47 @@
...
-      // TODO handle exception
-      const registration = await ensResolver.getRegistration(domain);
+  async function registrationDone(result: registerName.Result) {
+    let registration: ensResolver.Registration | null;
+
+    if (result.registration) {
+      registration = result.registration;
+    } else {
+      const domain = `${result.name}.${ensResolver.DOMAIN}`;
+      registration = await ensResolver.getRegistration(domain);
```
geigerzaehler, line 58: Could you retain the comment that we need to handle exceptions?

rudolfs: In what normal use-case would we arrive here and not have a registration? Should this not be impossible?

geigerzaehler: You’re right, this should never be `null` and we handle this case. But this is different from `getRegistration` throwing an exception. This can happen if the networking fails. In that case we should ensure that the UI is in a good state and the user is informed accordingly.

rudolfs: d174180b2b32339dd0ff4fd0dddc055267c76130

ui/src/org/ensRegistrar.ts
```diff
@@ -78,15 +78,30 @@ export function formatFee(fee: ethers.BigNumber): string {
   );
 }
 
-export async function commit(
+// Expire one hour from now.
+const deadline = ethers.BigNumber.from(Math.floor(Date.now() / 1000)).add(3600);
```
geigerzaehler, line 82: This needs to be calculated when `commit` is called and not when the module is instantiated. Otherwise this lies in the past at some point.

ui/src/org/ensRegistrar.ts
```diff
@@ -78,15 +78,30 @@ export function formatFee(fee: ethers.BigNumber): string {
...
+
+export async function getPermitSignature(
```
geigerzaehler, line 84: While your at it: Could you remove the `environment` parameter? We can obtain it from `wallet`.

ui/src/org/ensRegistrar.ts
```diff
@@ -78,15 +78,30 @@ export function formatFee(fee: ethers.BigNumber): string {
...
+): Promise<ethers.Signature> {
+  const wallet = svelteStore.get(Wallet.store);
+  const spender = registrarAddress(environment);
+  const token = radToken();
+  return await permitSignature(wallet.signer, token, spender, fee, deadline);
+}
+
+export interface CommitResult {
+  tx: ethers.ContractTransaction;
   minAge: number;
```
geigerzaehler, line 96: Documentation for this would be great. I always have trouble remembering what this is for.

ui/src/org/ensRegistrar.ts
```diff
@@ -78,15 +78,30 @@ export function formatFee(fee: ethers.BigNumber): string {
...
-  name: string,
-  salt: Uint8Array,
   fee: ethers.BigNumber
-): Promise<{
-  receipt: ethers.providers.TransactionReceipt;
+): Promise<ethers.Signature> {
+  const wallet = svelteStore.get(Wallet.store);
+  const spender = registrarAddress(environment);
+  const token = radToken();
+  return await permitSignature(wallet.signer, token, spender, fee, deadline);
```
geigerzaehler, line 91: Could you inline this function? I don’t see a reason to split it and `permitSignature` calls `svelteStore.get(Wallet.store)` again.

ui/Modal/Org/ConfigureEns.svelte
geigerzaehler: I meant to restore the comment above the call to `getRegistration` because this is where the exception can happen. I don’t think we need to handle the case where `registration` is not defined. (This should indeed never happen.)

ui/src/org/ensRegistrar.ts
rudolfs: 23a171bb6a82f75edbe1fc238dbdcaecba4130aa

ui/src/org/ensRegistrar.ts
rudolfs: 23a171bb6a82f75edbe1fc238dbdcaecba4130aa

ui/src/org/ensRegistrar.ts
rudolfs: 69bca8ffcb5ba70807c23f35aab79f9cfb2701dd

ui/src/org/ensRegistrar.ts
rudolfs: e921f99efff9d155af2a9b1db342e5a5ced8b2cb

ui/Modal/Org/ConfigureEns/RegisterName.svelte
```diff
@@ -0,0 +1,289 @@
...
+  $: runAllValidations(nameInputValue);
+
+  function runAllValidations(name: string | undefined): void {
+    if (!userInputStarted) {
+      userInputStarted = true;
+      return;
+    }
+
+    if (!name) {
+      debouncedValidateFormAndSetState.cancel();
```
geigerzaehler, line 77: Unfortunately, this does not prevent a race condition. When we call `cancel()` `validateFormAndSetState()` may already be running and result in updating the validation. Instead, the following should work
```suggestion
      const result = await validateFormExecutor.run(async () => true)
      if (result) {
        validationState = { .. }
      }
```

This will ensure that any validations that we’re triggered before don’t update the state.

ui/Modal/Org/ConfigureEns/RegisterName.svelte
```diff
@@ -0,0 +1,289 @@
...
+
+  $: runAllValidations(nameInputValue);
+
+  function runAllValidations(name: string | undefined): void {
+    if (!userInputStarted) {
+      userInputStarted = true;
+      return;
+    }
+
+    if (!name) {
```
geigerzaehler, line 76: Could you flip the if-else-statement?

ui/Modal/Org/ConfigureEns.svelte
rudolfs: a0b03cb4e4f8bc62505848a3bff061095bd1451f

ui/Modal/Org/ConfigureEns/RegisterName.svelte
rudolfs: be3f7079135357c303d8ba97ba29652324583f54

ui/Modal/Org/ConfigureEns/RegisterName.svelte
rudolfs: Is this any better?
4a934e46fc5a097814b51404d956a8a477554d02

geigerzaehler (approved):

PR #2237, opened 2021-08-16
fix: wait until session is unsealed

geigerzaehler:
*Review commit-by-commit*

When unsealing the session we’ll wait until it is actually unsealed. This fixes a race condition where we fetch the session before the proxy has had time to restart.

This includes a separate commit that extracts the retry functionality in it’s own module and generalizes it.


rudolfs (approved):
utACK

PR #2238, opened 2021-08-16
fix: don’t handle unhandled errors in cypress runner

geigerzaehler:
When the code is executed by the cypress test runner (i.e. the code were tests are defined) we also don’t want to handle errors and show them to the user. Instead they should be unhandled errors that make the tests fail.


rudolfs (approved):
utACK

PR #2246, opened 2021-08-17
chore: use the default walletconnect bridge

rudolfs:
Since WalletConnect 1.6.0 setting `bridge` to a custom value has no effect anymore. They're urging everyone to use their default infrastructure.

### DIFF
--- a/ui/src/ethereum/walletConnect.ts
+++ b/ui/src/ethereum/walletConnect.ts
@@ -173,7 +173,7 @@ function createConnector(): Connector {
   let modalClosedByWalletConnect = false;
 
   const connector = new Connector({
-    bridge: "https://radicle.bridge.walletconnect.org",
+    bridge: "https://bridge.walletconnect.org",
     qrcodeModal: {
       open: (uri: string, onClose, _opts?: unknown) => {
         modal.toggle(


ui/src/ethereum/walletConnect.ts
```diff
@@ -173,7 +173,7 @@ function createConnector(): Connector {
   let modalClosedByWalletConnect = false;
 
   const connector = new Connector({
-    bridge: "https://radicle.bridge.walletconnect.org",
+    bridge: "https://bridge.walletconnect.org",
```
geigerzaehler, line 176: I think we can just remove this option.

rudolfs: It's still mandatory. If I remove it there's an exception:
```
error.ts?4b5c:174 Error: Missing one of the required parameters: bridge / uri / session
    at new Connector (index.js?4853:32)
    at new WalletConnect (index.js?e07a:5)
    at createConnector (walletConnect.ts?135e:175)
    at WalletConnectClient.reinit (walletConnect.ts?135e:133)
    at new WalletConnectClient (walletConnect.ts?135e:71)
    at createWalletConnect (walletConnect.ts?135e:56)
    at eval (wallet.ts?aa61:106)
    at Module../ui/src/wallet.ts (bundle.js:1)
    at __webpack_require__ (bundle.js:1)
    at eval (transaction.ts:35) {details: {…}}
```

geigerzaehler: The types say it’s optional :facepalm:

geigerzaehler (approved):

PR #2248, opened 2021-08-17
refactor: unify UI configuration

geigerzaehler:
We unify the mechanism to configure the UI. This makes it extendable if we want to add more configuration parameters.

### DIFF
--- a/native/index.ts
+++ b/native/index.ts
@@ -15,6 +15,7 @@ import {
 } from "electron";
 import fs from "fs";
 import path from "path";
+import qs from "qs";
 import { ProxyProcessManager } from "./proxy-process-manager";
 import {
   MainMessage,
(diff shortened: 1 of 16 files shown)


native/index.ts
```diff
@@ -352,3 +351,17 @@ function execAsync(cmd: string): Promise<{ stdout: string; stderr: string }> {
     });
   });
 }
+
+function buildConfig(): Partial<Config> {
+  const config: Partial<Config> = {};
+  if (process.env.RADICLE_UPSTREAM_UI_PROXY_ADDRESS) {
+    config.proxyAddress = process.env.RADICLE_UPSTREAM_UI_PROXY_URL;
```
rudolfs, line 358:
```suggestion
    config.proxyAddress = process.env. RADICLE_UPSTREAM_UI_PROXY_ADDRESS;
```

rudolfs (commented):
Build is broken

native/index.ts
geigerzaehler: :facepalm:

rudolfs (approved):
utACK

PR #2249, opened 2021-08-17
fix: don’t swallow transaction error source

geigerzaehler:
We don’t call `error.fromJsError` so we also handle other types of errors. By passing `e` as the `source` parameter directly we leverage `error.fromUnknown` which is called internally.


rudolfs (approved):
utACK

PR #2250, opened 2021-08-17
dev: add test wallet for dev and testing

geigerzaehler:


ETHEREUM.md
```diff
@@ -69,6 +84,8 @@ Read more about Orgs [here][org].
 
 ### Local environment
 
+**⚠️ This sections is outdated and will be rewritten ⚠️**
```
rudolfs, line 87:
```suggestion
**⚠️ This section is outdated and will be rewritten ⚠️**
```

ui/src/ethereum/walletConnect.ts
```diff
@@ -210,6 +218,84 @@ function createConnector(): Connector {
...
+  private provider: ethers.providers.Provider;
+
+  public connection: svelteStore.Readable<Connection | undefined>;
+
+  constructor(mnemonic: string) {
+    this.connection = svelteStore.derived(this._connection, x => x);
+    this.wallet = ethers.Wallet.fromMnemonic(mnemonic);
+    this.provider = ethers.providers.InfuraProvider.getWebSocketProvider(
+      "rinkeby",
+      "de5e2a8780c04964950e73b696d1bfb1"
```
rudolfs, line 237: We're using this in multiple places, can we extract this into `config.ts`?

rudolfs: Why do we need the live provider actually? Shouldn't all networking be stubbed out?

geigerzaehler: I’ve extracted the key.



> Why do we need the live provider actually? Shouldn't all networking be stubbed out?

The idea for this is to test the app with rinkeby and have a wallet that automatically signs all transactions. I’ll follow this up with a PR that expands this so we can run against a local test net.

PR #2254, opened 2021-08-19
Add tx notifications

rudolfs:


ui/Modal/Org/ConfigureEns/UpdateMetadata.svelte
```diff
@@ -43,54 +44,71 @@
...
+          new error.Error({
+            message: err.message,
+            source: err,
+          })
+        );
+      } finally {
+        updateNotification.remove();
       }
 
+      if (tx) {
```
geigerzaehler, line 106: Instead of this `if`, could we have an early `return` in the `catch` block? Then typescript will know that `if` is defined.

rudolfs: 88a6ff9fc0396effafc654ed334f10895cf2c69d

geigerzaehler (approved):

Issue #2252, opened 2021-08-19
Crash on opening Upstream 0.2.10


tmakarios:
I feel like I've reported this before, but I can't find the issue now.  As soon as I typed in my passphrase after opening Upstream, it crashed.  Two error messages popped up on the screen, each providing the following details:
```
{
  "stack": "Error: Failed to load the session\n    at fn (file:///tmp/.mount_ra$
  "message": "Failed to load the session",
  "code": "SessionFetchFailure",
  "source": {
    "stack": "TypeError: Failed to fetch",                                     
    "message": "Failed to fetch",
    "code": "UnknownException",
    "details": {
      "name": "TypeError"
    }
  }
}
```

I also have a copy of the console output.  I've attached the first and last 1024 lines, since the whole file is 40 MiB (though I don't know how it got that big that quickly).
[2021-08-19-radicle-start.txt](https://github.com/radicle-dev/radicle-upstream/files/7011995/2021-08-19-radicle-start.txt)
[2021-08-19-radicle-end.txt](https://github.com/radicle-dev/radicle-upstream/files/7011999/2021-08-19-radicle-end.txt)

geigerzaehler:
Hey @tmakarios. Thanks for posting the logs. It looks like it takes around nine seconds for the background process to unlock the keys as you can see from the two log lines at the start
```plain
Aug 19 17:24:21.821  INFO proxy::http: POST /v1/keystore/unseal 204 elapsed=249.10487ms
Aug 19 17:24:30.677  INFO api::process: starting API
```

This is rather unusual. We’ll increase the timeout but I would also like to better understand what causes this. What CPU does your machine have?

tmakarios:
Hi @geigerzaehler, according to `/proc/cpuinfo` my CPU is a Celeron Dual-Core T3500 @ 2.10GHz.  It doesn't always take that long to unseal; normally it seems to be a bit under 1 second.  I'm not sure what else might have been going on that time; maybe it was tidying up after a recent `git push`?

geigerzaehler:
Good to know that it doesn’t take that long usually. Maybe there was some other CPU intensive process running. In any case, I’m increasing the timeout to 20 seconds in #2255. That should be enough. I’ll close the issue once this is merged.

PR #2255, opened 2021-08-19
fix: increase timeout to unseal

geigerzaehler:
As reported in #2252, it may take more than five seconds to unseal the
keystore. We’re increasing the timeout to 20 seconds.


rudolfs (approved):
utACK

PR #2260, opened 2021-08-20
Make commitment step resumable

rudolfs:
We're saving the commitment salt to localStorage, so that if the user goes out of the flow, they can resume with the name registration without losing the commitment fee.


ui/Modal/Org/ConfigureEns/BlockTimer.svelte
```diff
@@ -22,6 +22,10 @@
   let done: boolean = false;
 
   const onBlock = (currentBlock: number) => {
+    if (!startBlock) {
+      startBlock = currentBlock;
```
geigerzaehler, line 26: Instead of doing this, we should save the transaction hash of the commitment transaction, fetch the receipt for that tx and extract the block number from there.

It would probably be best to do this right after we call `restoreCommitment`. Then, `startBlock` would not need to be `undefined` and we don’t need to handle that case in the UI.

rudolfs: 5df5994c93784b480fc1866be125de317b11e6f8

ui/Modal/Org/ConfigureEns/BlockTimer.svelte
```diff
@@ -35,6 +36,19 @@
...
+
+  onMount(async () => {
+    const tx = await walletStore.provider.getTransaction(txHash);
+    const block = await walletStore.provider.getBlockNumber();
+
+    // If the block has not been mined, `blockNumber` is null.
+    if (tx.blockNumber) {
+      startBlock = tx.blockNumber;
+    } else {
+      startBlock = block;
```
geigerzaehler, line 48: If the transaction has not been included in a block yet it is not correct to use the current block as the start block. The commitment block is after the start block in this case. This leads to the UI thinking the commitment time has already passed. Here’s my suggestion how to fix this: https://github.com/radicle-dev/radicle-upstream/pull/2269

PR #2265, opened 2021-08-23
refactor: introduce <strong> typo modifier

rudolfs:
Required for https://github.com/radicle-dev/radicle-upstream/pull/2191.

### DIFF
--- a/public/typography.css
+++ b/public/typography.css
@@ -123,6 +123,7 @@ p,
   font-weight: 600;
 }
 
+strong,
 .typo-bold {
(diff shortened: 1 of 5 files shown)


public/typography.css
```diff
@@ -123,6 +123,7 @@ p,
   font-weight: 600;
 }
 
+strong,
```
geigerzaehler, line 126: I don’t think it’s a good idea to add styles to elements. You often end up overwriting the rules when you want to use `<strong>` with different style. Instead we should just add the `typo-bold` class to the element.

PR #2275, opened 2021-08-24
chore: check that error in catch block is unknown

geigerzaehler:
We enable an eslint rule that enforces the `unknown` type for errors in `catch` blocks. Before, these errors where typed as `any` which lead to runtime errors.


rudolfs (approved):
utACK

Issue #1964, opened 2021-06-08
Show QR code on empty (not set up) Wallet page


brandonhaslegs:
I do remember discussing this but I don't see a reason why we don't just show the QR code on the empty wallet screen.

So show this:
![image](https://user-images.githubusercontent.com/4406983/121205421-f1e65680-c877-11eb-9902-538fe29b7bb1.png)

instead of just this:
![image](https://user-images.githubusercontent.com/4406983/121205652-24904f00-c878-11eb-8b6c-bb7f49ebf40b.png)
since this button just opens the modal anyway.

brandonhaslegs:
If you disconnect your wallet, then click on the wallet page, there should be a QR code there already, instead of an empty state with just a button.

![Jul-21-2021 07-46-54](https://user-images.githubusercontent.com/4406983/126437492-b3fc2f24-8c1d-46fd-9702-05ac5f4741c2.gif)

PR #2276, opened 2021-08-24
refactor: decouple wallet connect from qr modal

geigerzaehler:
We don’t hardcode displaying the QR modal for wallet connect in the `walletConnect` module anymore. This eliminates the dependency of that module on UI code which also resulted in a circular dependency.


ui/Screen/Wallet.svelte
```diff
@@ -130,7 +159,7 @@
       </div>
     {:else}
       <ConnectWallet
-        onConnect={wallet.connect}
+        onConnect={connectWallet}
         connecting={$wallet.status === Status.Connecting} />
```
rudolfs, line 163: Could we now just show the QR code directly on this page, like @brandonhaslegs [wanted](https://github.com/radicle-dev/radicle-upstream/issues/1964)?

ui/src/ethereum/walletConnect.ts
```diff
@@ -140,7 +154,28 @@ export class WalletConnectClient implements WalletConnect {
   // and
   // https://github.com/WalletConnect/walletconnect-monorepo/pull/370#issuecomment-776038638
   private reinit() {
-    this.connector = createConnector();
+    this.connector = new Connector({
+      bridge: "https://radicle.bridge.walletconnect.org",
```
rudolfs, line 158: I thought that we should be using their default bridge now?

rudolfs (commented):
I tested this with Rainbow and Metamask mobile, the happy path worked. But when I rejected the connection initiation in the wallet, I didn't get a notification about that. After rejecting the connection I also couldn't re-connect anymore unless I reloaded Upstream.

geigerzaehler:
> But when I rejected the connection initiation in the wallet, I didn't get a notification about that.

That’s a known issue. See https://github.com/radicle-dev/radicle-upstream/blob/master/docs/wallet-connect-testing.md#reject-connection-fails

ui/src/ethereum/walletConnect.ts
geigerzaehler: This was a merge conflict. Fixed in https://github.com/radicle-dev/radicle-upstream/pull/2276/commits/a9becf75c0436f9718d51e6dc8df03c5b33bc96f

ui/Screen/Wallet.svelte
geigerzaehler: I requires a bit more effort but in effect we should be able to do this now.

rudolfs: Cool, we can follow-up on that later.

rudolfs (approved):

PR #2277, opened 2021-08-25
ci: better logging on CI

geigerzaehler:
We enable debug logging in the multi-node CI tests and add some debugging statements to the code.


rudolfs (approved):
utACK

PR #2278, opened 2021-08-25
chore: update radicle-link

geigerzaehler:


rudolfs (approved):
utACK

geigerzaehler:
> Running this against sprout leads Upstream to consume CPU at 300% after about 15min uptime.

So no regressions then :smiling_imp:

PR #2290, opened 2021-08-31
refactor: remove presentational logic from transaction module

geigerzaehler:
We remove presentational logic from `ui/src/transaction.ts` and move it into the components that use it. This eliminates circular dependencies involving the `transaction` module.


rudolfs (approved):
utACK

PR #2291, opened 2021-08-31
dev: add tests for transaction modal

### LINKED ISSUE #1921: Test Attestation Flow
Anything regarding the Ethereum integration is currently untested. This poses a big risk for breaking things. As a start I propose we test attestation with the following scenario

1. Go to the wallet screen
2. Assert that no wallet is connected
3. Click on “connect wallet”
4. Copy the displayed WalletConnect URI
5. Use the URI to connect with a test wallet
6. Check the displayed wallet address and balance
7. Click on “link identity”
8. Assert that the transaction for linking the identity exists
9. Click on the linking transaction
10. Check the content of the transaction modal

The main technical challenge for this test is providing an ethereum node and a test wallet. For the former we should be able to adapt `./scripts/ethereum-dev-node` and integrate it in the test infrastructure. For the latter we’ll need to create a test wallet that can be used programatically. We can either stub some code in `ui/src/wallet.ts` or use the wallet connect library to connect to the application via the wallet connect bridge. In both cases, the address of the wallet will need to be equipped with ETH on the local dev network.

geigerzaehler:
Closes #1921

### DIFF
--- a/cypress/integration/attestation.spec.ts
+++ b/cypress/integration/attestation.spec.ts
@@ -21,9 +21,16 @@ describe("attestation", () => {
       .should("exist");
     cy.contains("button", "Link your ID").click();
     commands.pickWithContent(["confirm-button"], "Link your ID").click();
-    cy.get('[data-cy="transactions"]', { timeout: 8000 }).should(
-      "contain",
-      "Claim Radicle Identity"
-    );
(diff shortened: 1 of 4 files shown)


rudolfs (approved):
utACK

PR #2292, opened 2021-08-31
chore: update dependencies

geigerzaehler:
Because of [a bug in svelte-check][1] we also replace single quotes with
a proper right single quote (’).

[1]: https://github.com/sveltejs/language-tools/issues/1113


rudolfs (approved):
utACK

Issue #2169, opened 2021-07-22
Consolidate the graph radicle orgs and gnosis safe indexes


rudolfs:
We rely on our self-deployed Gnosis safe legacy index. We won't be able to use this index on the mainnet version of the Graph, therefore we should merge these two indices and deploy them on the Graph mainnet. This will also reduce a couple of network requests from our clients as we'll be able to query this aggregate index directly.

Legacy mainnet: https://thegraph.com/legacy-explorer/subgraph/radicle-dev/gnosis-safe
Legacy testnet: https://thegraph.com/legacy-explorer/subgraph/radicle-dev/gnosis-safe-rinkeby

https://github.com/radicle-dev/radicle-subgraph
https://github.com/radicle-dev/gnosis-safe-subgraph

PR #2294, opened 2021-08-31
chore: use the safe client api instead of the graph index

rudolfs:
Instead of using two different the Graph indexes we use the API provided by gnosis to query gnosis related data.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/2169.

### DIFF
--- a/ui/src/org.ts
+++ b/ui/src/org.ts
@@ -7,7 +7,7 @@
 import * as lodash from "lodash";
 import { OperationType } from "@gnosis.pm/safe-core-sdk-types";
 
-import type { Org, MemberResponse } from "./org/theGraphApi";
+import type { Org } from "./org/theGraphApi";
 import type * as identity from "ui/src/proxy/identity";
 import type * as project from "ui/src/project";
 
@@ -39,7 +39,7 @@ import * as wallet from "ui/src/wallet";
 import ModalAnchorProject from "ui/Modal/Org/AnchorProject.svelte";
 import ConfigureEns from "ui/Modal/Org/ConfigureEns.svelte";
 
-export type { MemberResponse, Org };
+export type { Org };
 
 const orgPollExecutor = mutexExecutor.create();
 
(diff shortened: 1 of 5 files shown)


ui/src/org/safe.ts
```diff
@@ -28,6 +28,36 @@ export async function getPendingTransactions(
...
+
+export async function getMetadata(
+  ethEnv: Ethereum.Environment,
+  safeAddress: string
+): Promise<MemberResponse> {
+  safeAddress = ethers.utils.getAddress(safeAddress);
+  const safeServiceClient = createSafeServiceClient(ethEnv);
+  const response = await safeServiceClient.getSafeInfo(safeAddress);
+
+  return <MemberResponse>{
```
geigerzaehler, line 44: The return statement also checks without the explicit typecast. I would prefer to remove the explicit typecast because the it accepts excess properties
```typescript
return <MemberResponse> {
    foo: true,
    threshold: response.threshold,
    members: response.owners,
  };
```
Without the type cast the excess property produces an error.

ui/src/org.ts
```diff
@@ -378,7 +385,8 @@ export async function fetchMembers(
   wallet: wallet.Wallet,
   gnosisSafeAddress: string
 ): Promise<OrgMembers> {
-  const response: MemberResponse = await graph.getGnosisSafeMembers(
+  const response: MemberResponse = await Safe.getMetadata(
```
geigerzaehler, line 388: We can omit the explicit type annotation.

ui/src/org/safe.ts
rudolfs: 6f609f6943e714b73f452722231eff6caf14c608

ui/src/org.ts
rudolfs: 6f609f6943e714b73f452722231eff6caf14c608

geigerzaehler (approved):

PR #2303, opened 2021-09-01
refactor: consolidate and check contract addresses

geigerzaehler:
We consolidate all contract addresses in one file. The dev nodes check that they deploy the contracts to the correct addresses when they are run.


rudolfs (approved):
utACK

Issue #2308, opened 2021-09-01
Maintainer view is not updated when new patch is discovered


geigerzaehler:
The patches count in the maintainer project view is not updated when the maintainer discovers a new patch from a peer. This means the following test fails: https://github.com/radicle-dev/radicle-upstream/blob/f8daaeceef22c2b1c179f7d29feaa7a70b4dbaad/cypress/integration/project/patches.spec.ts#L343

rudolfs:
I believe this is fixed.

PR #2309, opened 2021-09-01
fix: disable failing patches test

geigerzaehler:
We disable a test that fails consistently. The bug has been recorded in #2308.


rudolfs (approved):

rudolfs:
Should we do the same with the networking test?

geigerzaehler:
> Should we do the same with the networking test?

The networking tests are just flaky. It’s quite hard to have them fail locally. I’d prefer to keep them. This way we can retry before a release until things are green.

PR #2310, opened 2021-09-01
dev: use predictable path for test data

geigerzaehler:
For each test we use the same path to store data instead of a new patch for every run of the same test. This means that we overwrite data from previous runs. In practice I haven’t found it useful to have data from previous runs. On the contrary: If there are artifacts from multiple runs it is difficult to find those of the current run.


rudolfs (approved):
utACK

Issue #2108, opened 2021-07-08
Cache org data


geigerzaehler:
To get data for displaying orgs we issue a lot of requests which may result and long load times and potentially rate limiting. We can easily cache the following data fetches because the data is immutable`
* `isMultiSig` from `ui/src/org`
* `getOwner` from `ui/src/org`

We should also cache data that is mutable but changes infrequently. Even a max-age of 60 seconds would decrease the number of requests. The only candidate for this is

* `ClaimContract#getClaimed` as used in `getClaimedIdentity` in `ui/src/org`

We should also investigate more candidates for caching.

rudolfs:
> isMultiSig from ui/src/org
> getOwner from ui/src/org

This was already done in https://github.com/radicle-dev/radicle-upstream/pull/2167.

PR #2336, opened 2021-09-07
fix: cache claimed identities for one minute

rudolfs:
We want to reduce the amount of requests we make to Infura.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/2108.

### DIFF
--- a/ui/src/attestation/contract.ts
+++ b/ui/src/attestation/contract.ts
@@ -4,19 +4,20 @@
 // with Radicle Linking Exception. For full terms see the included
 // LICENSE file.
 
-import * as ethers from "ethers";
 import type { Signer } from "ethers";
+import LruCache from "lru-cache";
+import * as ethers from "ethers";
 
 import {
   Claims,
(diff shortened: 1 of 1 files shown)


ui/src/attestation/contract.ts
```diff
@@ -53,22 +63,34 @@ export class ClaimsContract {
   // Fetches the identity claimed by the given Ethereum address.
   // Returns `undefined` if currently there's no valid claim.
   async getClaimed(address: string): Promise<Uint8Array | undefined> {
+    const normalisedAddress = address.toLowerCase();
+    const cached: ClaimsContractCacheEntry | undefined =
```
geigerzaehler, line 67: Any reason why we don’t use `ui/src/memoizeLru` to separate the caching responsibility?

rudolfs: I couldn't figure out how to make it work within a class.

geigerzaehler: Ah, of course.

Issue #2092, opened 2021-07-01
Improve wording around pending anchor execution


rudolfs:
Once an anchor has been signed by the required amount of people, the transaction still needs to be executed on the Gnosis end. We should improve the wording around that so it is more clear. It might otherwise be confusing to see that there are `1 of 1` signatures, but the anchor is still in pending state.

<img width="1352" alt="Screenshot 2021-07-01 at 10 27 09" src="https://user-images.githubusercontent.com/158411/124097345-c8c27b80-da5b-11eb-9d41-d651aca8c767.png">

<img width="1216" alt="Screenshot 2021-07-01 at 11 02 48" src="https://user-images.githubusercontent.com/158411/124097608-02938200-da5c-11eb-900d-aa29f25f7784.png">

PR #2345, opened 2021-09-07
fix: improve pending anchor message

rudolfs:
When the quorum of members have signed a transaction on Gnosis Safe, this transaction still needs to be executed by one of the members manually before it gets submitted to the blockchain. In this case show a different message to the user so they know what needs to be done.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/2092.

Screenshot 2021-09-07 at 18 23 01

### DIFF
--- a/ui/App/OrgScreen/Projects.svelte
+++ b/ui/App/OrgScreen/Projects.svelte
@@ -6,7 +6,7 @@
  LICENSE file.
 -->
 <script lang="typescript">
-  import type { Project } from "ui/src/project";
+  import type * as Project from "ui/src/project";
 
   import * as router from "ui/src/router";
   import * as Session from "ui/src/session";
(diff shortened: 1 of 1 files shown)


rudolfs:
Does the copy sound good here @brandonhaslegs?

brandonhaslegs:
Can there be more than one pending?

brandonhaslegs:
Maybe "Waiting for a member to execute."

rudolfs:
> Can there be more than one pending?

No.

> As a member what does this look like? It's not clear how's to execute if this is the member view.

This is what a member would see. You only see the anchor stuff if you're a member of the org. Maybe there should be a link to the Gnosis Safe transaction, so that you can execute it?

brandonhaslegs:
> Maybe there should be a link to the Gnosis Safe transaction, so that you can execute it?

Yes plz

rudolfs:
Not enough members have signed:

Screenshot 2021-09-08 at 07 41 00

Quorum has signed, but tx is not yet executed:

Screenshot 2021-09-08 at 07 51 23

"Execute transaction" link goes to:

Screenshot 2021-09-08 at 07 53 14

We can't link to the exact transaction yet, because of https://github.com/gnosis/safe-react/issues/970.

brandonhaslegs:
> Not enough members have signed

Change this text to "Not enough members have signed this anchor." (remove "of the")

rudolfs:
Screenshot 2021-09-08 at 08 46 38

rudolfs:
Now I'm thinking, should this also have a link to the transaction?

`Not enough members have signed this anchor. Sign anchor`

brandonhaslegs:
> Now I'm thinking, should this also have a link to the transaction?

Yes

rudolfs:
I also aligned the wording to match the Gnosis Safe webpage:

Screenshot 2021-09-08 at 09 06 45


Screenshot 2021-09-08 at 09 06 38

brandonhaslegs:
In one of the lines we say "this anchor transaction" and in the other we say just "anchor transaction" Should we add "this" to the other one?

rudolfs:
Done.

Screenshot 2021-09-08 at 09 34 44

brandonhaslegs (approved):

Issue #1321, opened 2020-11-24
Make "copied" notifications more specific


brandonhaslegs:
![image](https://user-images.githubusercontent.com/4406983/100091797-69f9ab80-2e55-11eb-9f52-951ed2a40fa1.png)

---

![image](https://user-images.githubusercontent.com/4406983/100091870-84338980-2e55-11eb-833f-c947fd7ba3a8.png)

## Profile
Especially since we don't ever say explicitly on the profile that this is your Device ID (we just show the device icon), we could note in the notification that you've in fact copied your Device ID.

Suggestion: `Device ID copied to your clipboard`

---

![image](https://user-images.githubusercontent.com/4406983/100091770-5a7a6280-2e55-11eb-8e62-6740dd09d275.png)

## Project
Also need to add it here then too, when you copy a Radicle ID.

Suggestion: `Radicle ID copied to your clipboard`

rudolfs:
I believe this was solved by https://github.com/radicle-dev/radicle-upstream/pull/2274. Re-open the issue if not.

brandonhaslegs:
For the most part yes!

![image](https://user-images.githubusercontent.com/4406983/132457235-114c061b-4041-43dd-b419-349e25abd1fb.png)

this one should say "Radicle ID"

---

![image](https://user-images.githubusercontent.com/4406983/132457350-4d1f9b5d-ff0a-4e06-9af4-099b7f4f92fd.png)

This one should say "command"

---

![image](https://user-images.githubusercontent.com/4406983/132457403-e17677f9-d932-443d-becb-3d8dc3ef0a58.png)

Should say "command"

PR #2356, opened 2021-09-09
style: make copy-to-clipboard notification consistent

rudolfs:
Closes: https://github.com/radicle-dev/radicle-upstream/issues/1321.

Screenshot 2021-09-09 at 18 58 01
Screenshot 2021-09-09 at 18 58 11
Screenshot 2021-09-09 at 18 58 55

### DIFF
--- a/ui/App/ProjectScreen/RemoteHelperHint.svelte
+++ b/ui/App/ProjectScreen/RemoteHelperHint.svelte
@@ -55,7 +55,7 @@
     </ExternalLink>
   </p>
   <Hoverable bind:hovering={hover}>
-    <Copyable name="shell configuration">
+    <Copyable name="shell configuration" tooltipStyle="width: fit-content;">
       <p
         class="typo-text-small-mono"
         style="color: var(--color-foreground-level-6)">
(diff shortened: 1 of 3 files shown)


brandonhaslegs (approved):

ui/App/ProjectScreen/RemoteHelperHint.svelte
```diff
@@ -55,7 +55,7 @@
     </ExternalLink>
   </p>
   <Hoverable bind:hovering={hover}>
-    <Copyable name="shell configuration">
+    <Copyable name="shell configuration" tooltipStyle="width: fit-content;">
```
geigerzaehler, line 58: Is the style applicable to all tooltips? Should it be moved into the `Copyable` component.

rudolfs: I think it's 50/50. In some cases we want it to be `fit-content` in others we want it to take the full width. Here I moved this style into the `<Tooltip>` component and I had to make the overrides to `width: 100%` in other places: https://github.com/radicle-dev/radicle-upstream/commit/5420fdf79c298afc3c9365251cfeb4f45198b9e1.

I'm fine with either way, you can pick which one you prefer.

geigerzaehler: Let’s leave it like it is then. I was just curious.

Issue #1347, opened 2020-11-24
Relative links in rendered markdown interact with UI


rudolfs:
Refs https://github.com/radicle-dev/radicle-upstream/issues/1335.

When clicking on relative links in rendered markdown they lead to the profile page.
Clicking on a relative link in rendered markdown should have no effect in the UI for now.

Steps to repro:

1. `git clone https://github.com/BurntSushi/ripgrep.git`
2. In the UI navigate to the project `ripgrep`
3. In the rendered `README.md` click on the "Installation" link

MeBrei:
Actually making relative links work would require some refactoring in the source browsing code, since navigating to a file is not done via changing the url currently.

I did not find a way to disable relative links via `marked` directly. We could match relative `href`s via css and disable any pointer events, but we would have to make sure we catch all variants and a disabled link that I can see with dev tools might be a bit frustrating.

PR #2352, opened 2021-09-08
style: format all external links consistently

rudolfs:
This is only about links that open outside of Upstream.

Closes: https://github.com/radicle-dev/radicle-upstream/issues/1967.

Screenshot 2021-09-09 at 09 15 57

PR #2376, opened 2021-09-15
style: render external links in markdown with appropriate style

rudolfs:
Fixes a regression introduced with https://github.com/radicle-dev/radicle-upstream/pull/2352. Links in rendered markdown had a wrong styling due to the CSS warning that I added. Wasn't a good idea in hindsight.

We also apply a different style to relative links and prevent user interaction.
Closes https://github.com/radicle-dev/radicle-upstream/issues/1347.

Note: making relative links work properly would require a substantial change to the source browsing code.

Regression:
Screenshot 2021-09-15 at 16 29 11

Fix:
Screenshot 2021-09-15 at 16 30 20

Disabled relative links:
Screenshot 2021-09-15 at 16 30 41

We don't apply the styling if the `<a>` tag contains images:
Screenshot 2021-09-15 at 16 48 48

### DIFF
--- a/ui/DesignSystem/Markdown.svelte
+++ b/ui/DesignSystem/Markdown.svelte
@@ -9,6 +9,26 @@
   import marked from "marked";
 
   export let content: string;
+
+  const renderer = {
+    link(href: string, _title: string, text: string) {
+      if (
+        href.toLowerCase().startsWith("http://") ||
+        href.toLowerCase().startsWith("https://")
+      ) {
+        if (text.includes("<img")) {
+          return `<a href="${href}">${text}</a>`;
+        } else {
+          return `<a class="typo-link" style="text-decoration: none;" href="${href}"><span style="text-decoration: underline; margin-right: 0.1rem;">${text}</span><span style="vertical-align: text-top">↗</span></a>`;
+        }
+      } else {
+        // Internal links don't work yet, so we disable user-interaction.
+        return `<span style="text-decoration: underline; text-underline-offset: 0.25rem;">${text}</span>`;
+      }
+    },
+  };
+
+  marked.use({ renderer });
 </script>
 
 <style>
@@ -101,15 +121,6 @@
   .markdown :global(p) {
     margin-top: 0;
     margin-bottom: 0.625rem;
-  }
-
-  .markdown :global(a) {
-    background-color: initial;
-    color: var(--color-primary);
-    text-decoration: none;
-  }
-  .markdown :global(a:hover) {
-    text-decoration: underline;
   }
 
   .markdown :global(strong) {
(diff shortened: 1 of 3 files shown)


brandonhaslegs (approved):

ui/DesignSystem/Markdown.svelte
```diff
@@ -9,6 +9,26 @@
   import marked from "marked";
 
   export let content: string;
+
+  const renderer = {
+    link(href: string, _title: string, text: string) {
```
geigerzaehler, line 14: If I’m not mistaken, this does not handle inline html links properly. E.g. if your markdown looks like this

```markdown
[md link](foo)
<a href="bar">inline html</a>
```
I would prefer to style this with CSS. We can even use CSS attribute selector to disable user interaction.

I’d also like to keep the global link styles and disable them for `markdown` with the `:not` selector

rudolfs: Inline html works just fine:
```md
[inline link](foo)

[inline link 2](https://twitter.com/)
```

Translates to:

Screenshot 2021-09-20 at 15 45 58

I'm not sure we'll be able to easily achieve the external link styling without additional markup in the specific way that @brandonhaslegs wanted.

How would you distinguish between internal and external links with CSS?

geigerzaehler: > Inline html works just fine:

Your example doesn’t look like it’s using line HTML. I’ve tried the following

```markdown
<a href="https://x">HTML HTTPS</a>

<a href="http://x">HTML HTTP</a>

<a href="./">HTML internal</a>

[Markdown HTTPS](https://x)

[Markdown HTTP](http://x)

[Markdown internal](./foo)
```

and it gives this result

Screenshot from 2021-09-20 16-45-30


> How would you distinguish between internal and external links with CSS?

For example with
```css
  .markdown :global(a[href^="http://"]:after),
  .markdown :global(a[href^="https://"]:after)
  {
    content: "↗";
    margin-left: 0.1rem;
    text-decoration: none;
    /* disables text decoration from containing a element */
    display: inline-block;
  }
```

Not underlining `img` tags might be tricky though

rudolfs: Is this what you had in mind https://github.com/radicle-dev/radicle-upstream/commit/258ce958bddeeb1c5eefad0ea737f05ca3d87d70?

```
<a href="https://x">HTML HTTPS</a>

<a href="http://x">HTML HTTP</a>

<a href="./">HTML internal</a>

[Markdown HTTPS](https://x)

[Markdown HTTP](http://x)

[Markdown internal](./foo)
```

Translates to:

Screenshot 2021-09-21 at 11 03 31

I couldn't figure out how to make the `` tags to not show the `↗` though.

Screenshot 2021-09-21 at 11 04 01

Anchors seem to work, so I left them enabled. Only disabled internal relative links:

Screenshot 2021-09-21 at 11 05 14

Issue #1976, opened 2021-06-09
Show anchored state of the project


brandonhaslegs:
Maybe in the peer selector or revision selector on the project page

rudolfs:
I think it would make more sense in the revision selector.

brandonhaslegs:
So the two options would be either showing the anchor next to the anchored branch
![image](https://user-images.githubusercontent.com/4406983/122215510-98031380-ceab-11eb-8fa5-16d045cbd441.png)

…or showing it next to the org that has anchored the project
![image](https://user-images.githubusercontent.com/4406983/122215576-a94c2000-ceab-11eb-9868-8fda2c09e120.png)

I think the first one makes more sense?

rudolfs:
Hm, but the actual object that is being anchored is a commit.

A branch contains many commits and a peer contains many
branches and commits, so to me it doesn't make sense to
show the anchor on a branch or a peer. Besides, a branch always
points to the latest commit of that branch, but an anchored commit
could be somewhere in that branch but earlier than the latest commit.

To me an anchor is similar to a branch or a tag. I think we should
show it in the revision (aka branch) selector as the shortened
commit SHA + anchor like this:

```
<branch icon> main
<branch icon> feature/bla
<anchor icon> f4c7697
<anchor icon> a89834

brandonhaslegs:
![image](https://user-images.githubusercontent.com/4406983/122225175-216b1380-ceb5-11eb-94a2-decb9be4d90c.png)

rudolfs:
Hmm, now after reflecting a bit more on this and talking to @geigerzaehler I think that you were right after all -- orgs belong in the peer selector. 😬 🤷 

If we have the org in the revision selector then we'd still have to figure out which peer to select in the peer selector. And it's not guaranteed that every peer will have this specific anchored project, which might make this selection heuristic complicated.

Now, if we put the org that the anchor belongs to in the peer selector we've solved the peer pre-selection problem. And for that org we'd show the latest anchor in the revision selector -- which might make that list rather short and boring. To remedy that, we could also show all the historical anchors for that org _and_ mark the latest anchor in some special way.

brandonhaslegs:
Yeah I like it. Makes the most sense to me too. So this?

![image](https://user-images.githubusercontent.com/4406983/122231362-aefd3200-ceba-11eb-88e9-f6d56c113780.png)

juliendonck:
[...]

Issue #2245, opened 2021-08-17
Org Profile Hover Card


brandonhaslegs:
On hover of an org avatar, show an org profile card. Design still open for ideas.

![image](https://user-images.githubusercontent.com/4406983/129669219-3cb059ac-c383-46c3-b69a-eed896a4d74b.png)

Kaihuang724:
Closing this as orgs will not be implemented for MVP

PR #2377, opened 2021-09-17
feat: show anchors on project page

rudolfs:
Closes https://github.com/radicle-dev/radicle-upstream/issues/1976.

Screenshot 2021-09-27 at 12 57 32

Screenshot 2021-09-27 at 12 57 59

Screenshot 2021-09-27 at 13 03 15

### DIFF
--- a/ui/App/OrgIdentity.svelte
+++ b/ui/App/OrgIdentity.svelte
@@ -1,0 +1,39 @@
+<!--
+ Copyright © 2021 The Radicle Upstream Contributors
+
+ This file is part of radicle-upstream, distributed under the GPLv3
+ with Radicle Linking Exception. For full terms see the included
+ LICENSE file.
+-->
+<script lang="typescript">
+  import { Avatar, CopyableIdentifier } from "ui/DesignSystem";
+  import type * as ensResolver from "ui/src/org/ensResolver";
+
+  export let orgAddress: string;
+  export let registration: ensResolver.Registration | undefined = undefined;
+</script>
+
+<style>
+  .org-domain {
+    color: var(--color-foreground-level-6);
+    text-overflow: ellipsis;
+    overflow: hidden;
+    white-space: nowrap;
+  }
+</style>
+
+<div style="display: flex;">
+  <Avatar
+    style="margin-right: 0.5rem;"
+    size="small"
+    kind={registration?.avatar
+      ? { type: "orgImage", url: registration.avatar }
+      : { type: "orgEmoji", uniqueIdentifier: orgAddress }} />
+  {#if registration?.domain}
+    <span class="typo-text-bold org-domain">
+      {registration?.domain}
+    </span>
+  {:else}
+    <CopyableIdentifier kind="ethAddress" value={orgAddress} showIcon={false} />
+  {/if}
+</div>
--- a/ui/App/ProjectScreen/ProjectHeader.svelte
+++ b/ui/App/ProjectScreen/ProjectHeader.svelte
@@ -8,6 +8,8 @@
 <script lang="typescript">
   import type { Stats } from "ui/src/project";
 
+  import dayjs from "dayjs";
+
   import { Icon, CopyableIdentifier } from "ui/DesignSystem";
 
   export let name: string;
@@ -15,6 +17,7 @@
   export let description: string | undefined = undefined;
   export let stats: Stats;
   export let onClick: (() => void) | undefined = undefined;
+  export let latestAnchorTimestamp: number | undefined = undefined;
 </script>
 
 <style>
@@ -40,18 +43,20 @@
   .stat-item {
     display: flex;
     color: var(--color-foreground-level-6);
-    margin-right: 1rem;
   }
 
-  .stat-item p {
-    margin-left: 0.5rem;
-    white-space: nowrap;
-  }
-
-  .stat-separator {
+  .stat-item:not(:last-of-type)::after {
+    content: "•";
     display: flex;
     color: var(--color-foreground-level-3);
-    margin-right: 1rem;
+    margin: 0 1rem 0 1rem;
+    font-family: var(--typeface-mono-bold);
+    font-weight: bold;
+  }
+
+  .stat-item span {
+    margin-left: 0.5rem;
+    white-space: nowrap;
   }
 
   .clickable {
@@ -81,22 +86,29 @@
     {#if stats.branches > 0}
       <div class="stat-item">
         <Icon.Branch />
-        <p>
+        <span>
           {stats.branches === 1 ? `1 Branch` : `${stats.branches} Branches`}
-        </p>
+        </span>
       </div>
     {/if}
-    {#if stats.branches > 0 && stats.contributors > 0}
-      <span class="typo-mono-bold stat-separator">•</span>
-    {/if}
+
     {#if stats.contributors > 0}
       <div class="stat-item">
         <Icon.User />
-        <p>
+        <span>
           {stats.contributors === 1
             ? `1 Contributor`
             : `${stats.contributors} Contributors`}
-        </p>
+        </span>
+      </div>
+    {/if}
+
+    {#if latestAnchorTimestamp}
+      <div class="stat-item">
+        <Icon.Anchor />
+        <span>
+          Last anchored {dayjs().to(dayjs.unix(latestAnchorTimestamp))}
+        </span>
       </div>
     {/if}
   </div>
--- a/ui/App/ProjectScreen/Source/Anchors.svelte
+++ b/ui/App/ProjectScreen/Source/Anchors.svelte
@@ -1,0 +1,160 @@
+<!--
+ Copyright © 2021 The Radicle Upstream Contributors
+
+ This file is part of radicle-upstream, distributed under the GPLv3
+ with Radicle Linking Exception. For full terms see the included
+ LICENSE file.
+-->
+<script lang="typescript">
+  import type * as project from "ui/src/project";
+
+  import dayjs from "dayjs";
+  import * as zod from "zod";
+
+  import * as browserStore from "ui/src/browserStore";
+  import * as org from "ui/src/org";
+  import * as router from "ui/src/router";
+
+  import { Icon, IdentifierLink, format } from "ui/DesignSystem";
+  import AnchorCard from "./AnchorCard.svelte";
+  import EmptyState from "ui/App/ScreenLayout/EmptyState.svelte";
+  import OrgIdentity from "ui/App/OrgIdentity.svelte";
+
+  export let anchors: project.ConfirmedAnchor[];
+
+  $: latest = anchors.slice(-1)[0];
+  $: oldAnchors = anchors.slice(0, -1);
+
+  const openCommit = (commitHash: string, projectId: string) => {
+    router.push({
+      type: "project",
+      params: {
+        activeView: { type: "commit", commitHash },
+        urn: projectId,
+      },
+    });
+  };
+
+  const isAnchorHintVisible = browserStore.create<boolean>(
+    "radicle.isAnchorHintVisible",
+    true,
+    zod.boolean()
+  );
+</script>
+
+<style>
+  .container {
+    margin: 0 auto 6rem;
+    max-width: var(--content-max-width);
+    min-width: var(--content-min-width);
+    padding: 2rem var(--content-padding) 0;
+  }
+
+  .banner {
+    display: flex;
+    padding: 0 1rem 0 1rem;
+    align-items: center;
+    border-radius: 0.5rem;
+    background-color: var(--color-primary-level-1);
+    color: var(--color-primary);
+    height: 3rem;
+    margin-bottom: 1rem;
+  }
+
+  .dismiss {
+    margin-left: auto;
+  }
+
+  .timestamp {
+    color: var(--color-foreground-level-5);
+    margin-left: auto;
+  }
+
+  .anchor-list {
+    border: 1px solid var(--color-foreground-level-2);
+    border-radius: 0.5rem;
+    margin-top: 1rem;
+  }
+
+  .anchor {
+    padding: 1rem;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+  }
+
+  .anchor:not(:last-of-type) {
+    border-bottom: 1px solid var(--color-foreground-level-2);
+  }
+
+  .left {
+    display: flex;
+    gap: 1rem;
+  }
+</style>
+
+<div class="container">
+  {#if $isAnchorHintVisible}
+    <div class="banner">
+      <Icon.InfoCircle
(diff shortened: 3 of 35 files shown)


geigerzaehler:
On adding the `ui/src/ethereum/transcode` module: I prefered the organization of the code before.

The encoding `encodeAnchorData` function and the corresponding decoding are specific to the anchoring contract. The encoding is only used in the contract and it’s not a standard encoding. So I think it’s preferable to keep it close to the contract code to make the contract code easier to understand.

On the other hand, extracting the SHA1 from the URN (and vice versa) is something that is determined by the protocol. The encoding functionality is used by the org and attestation contracts so we need to put it into a shared module. Instead of organizing it under `ui/src/ethereum` it makes more sense to put it into `ui/src/urn`.

rudolfs:
@brandonhaslegs I think this feature is done, could you have a look at the visuals? I'd like to handle the [linking to orgs](https://github.com/radicle-dev/radicle-upstream/issues/2245) in a separate issue.

brandonhaslegs:
Can we add a button to anchor the project you're viewing when you see this empty state? I'd expect to be able to anchor a project from that project page instead of having to go to the org page to do it.
image

brandonhaslegs:
These two lines should be swapped, now that the second line will always be longer than the first line.

rudolfs:
> Can we add a button to anchor the project you're viewing when you see this empty state? I'd expect to be able to anchor a project from that project page instead of having to go to the org page to do it.

A single button wouldn't work with the current implementation. How would you chose in which org the project should be anchored? Currently the anchor modal is in the context of an org that you have previously selected:

Screenshot 2021-09-27 at 15 09 45

brandonhaslegs:
I will make an issue to add an org selection modal here then. Is that all we'd need? Some org-selection modal and then we can direct to the anchoring flow on that org's page?

rudolfs:
> These two lines should be swapped, now that the second line will always be longer than the first line.

Does it really make sense to change the order in the header, given that there can be an optional description and in all other cases (user profiles and orgs) we always keep the title and identifier close by?

Screenshot 2021-09-27 at 15 14 41
Screenshot 2021-09-27 at 15 14 44
Screenshot 2021-09-27 at 15 15 40

rudolfs:
> I will make an issue to add an org selection modal here then. Is that all we'd need? Some org-selection modal and then we can direct to the anchoring flow on that org's page?

Maybe it makes more sense to change the anchoring modal to have a dropdown to select the org in which the commit should be anchored. This would make the modal usable from anywhere in the app. And when you're already on an org, it would just pre-select that org.

ui/App/OrgIdentity.svelte
```diff
@@ -0,0 +1,38 @@
...
+ with Radicle Linking Exception. For full terms see the included
+ LICENSE file.
+-->
+<script lang="typescript">
+  import { Avatar, CopyableIdentifier } from "ui/DesignSystem";
+  import type * as ensResolver from "ui/src/org/ensResolver";
+
+  export let orgAddress: string;
+  export let registration: ensResolver.Registration | undefined = undefined;
+  export let style: string | undefined = undefined;
```
geigerzaehler, line 14: We always include `margin-right: 0.5rem` in the style to ensure an appropriate spacing between the avatar and the domain or org address. It should not be the responsibility of the caller to ensure the spacing and we should hard-code this here.

Except for this the style is only used to add a left margin to the `Avatar`. This can also be accomplished by adding margins to the previous element at the call site. This is preferable to adding the style here and would allow us to eliminate the `style` property altogether.

ui/App/ProjectScreen/Source/Anchors.svelte
```diff
@@ -0,0 +1,223 @@
...
+  import * as router from "ui/src/router";
+
+  import { Icon, IdentifierLink } from "ui/DesignSystem";
+  import EmptyState from "ui/App/ScreenLayout/EmptyState.svelte";
+  import OrgIdentity from "ui/App/OrgIdentity.svelte";
+
+  export let anchors: project.ConfirmedAnchor[];
+
+  const latest = anchors.slice(-1)[0];
+  const oldAnchors = anchors.slice(0, -1);
```
geigerzaehler, line 25: To ensure this is reactive

```suggestion
  $: latest = anchors.slice(-1)[0];
  $: oldAnchors = anchors.slice(0, -1);
```

ui/App/ProjectScreen/Source/Anchors.svelte
```diff
@@ -0,0 +1,223 @@
...
+          <div class="timestamp">
+            {ethereum.formatAnchorTransactionTimestamp(anchor.timestamp)}
+          </div>
+        </div>
+      {/each}
+    </div>
+  {:else}
+    <EmptyState
+      emoji="🚣"
+      text="This project doesn't have any anchors yet. To anchor it, go to one of your orgs and click “Anchor project”." />
```
geigerzaehler, line 221: We should only show this if `latest` is `undefined`.

ui/App/ProjectScreen/Source/Commit.svelte
```diff
@@ -153,6 +192,38 @@
...
+    {#each anchors as anchor}
+      <div class="anchor">
+        <div class="header">
+          <div style="display: flex;">
+            <Icon.Anchor />
+            <p
+              class="typo-text-bold"
+              style="color: var(--color-foreground-level-6); margin: 0 0.5rem 0 0.5rem;">
+              Anchored by
+            </p>
```
geigerzaehler, line 204: See my other comment about making this a `<span>`.

ui/App/ProjectScreen/Source/Commit.svelte
```diff
@@ -72,6 +79,38 @@
...
+    justify-content: space-between;
+    align-items: center;
+    padding: 0 1rem 0 1rem;
+  }
+
+  .timestamp {
+    color: var(--color-foreground-level-5);
+  }
+
+  .metadata {
```
geigerzaehler, line 104: This class is already applied to another element in the slot for the `BackButton` component. Is this intentional? I feel that `anchor-transaction-hash` would be more descriptive.

ui/App/ProjectScreen/Source/Commit.svelte
```diff
@@ -72,6 +79,38 @@
...
+    padding: 1rem 1rem 0 1rem;
+    display: flex;
+    gap: 0.5rem;
+    color: var(--color-foreground-level-6);
+  }
+
+  .metadata:last-of-type {
+    padding-bottom: 1rem;
+  }
```
geigerzaehler, line 113: From what I can see every element of `.metadata` is the last of its type. I think we can set this on every `.metadata` element.

ui/App/ProjectScreen/Source/Commit.svelte
```diff
@@ -153,6 +192,38 @@
       </div>
     </div>
 
+    {#each anchors as anchor}
+      <div class="anchor">
+        <div class="header">
```
geigerzaehler, line 197: Can we call the class `anchor-header` to make disambiguate it from other headers in this component?

ui/App/ProjectScreen/Source/Commit.svelte
```diff
@@ -72,6 +79,38 @@
...
+    border: 1px solid var(--color-foreground-level-2);
+    margin-top: 1rem;
+    margin-bottom: 1rem;
+  }
+
+  .header {
+    border-bottom: 1px solid var(--color-foreground-level-2);
+    min-height: 3.5rem;
+    display: flex;
+    justify-content: space-between;
```
geigerzaehler, line 95: An alternative way to ensure that the timestamp is right-aligned is to put `margin-left: auto` on `.timestamp`. I prefer this because it also works if `.header` has more than two children.

ui/App/ProjectScreen/Source/Commit.svelte
```diff
@@ -72,6 +79,38 @@
...
+    background-color: var(--color-foreground-level-1);
+    border-radius: 0.5rem;
+    border: 1px solid var(--color-foreground-level-2);
+    margin-top: 1rem;
+    margin-bottom: 1rem;
+  }
+
+  .header {
+    border-bottom: 1px solid var(--color-foreground-level-2);
+    min-height: 3.5rem;
```
geigerzaehler, line 93: We could replace this with `padding: 1rem`. This would make it consistent with how padding is used for `metadata`.

brandonhaslegs:
> Maybe it makes more sense to change the anchoring modal to have a dropdown to select the org in which the commit should be anchored. This would make the modal usable from anywhere in the app. And when you're already on an org, it would just pre-select that org.

thought about this. makes sense. will make the issue.

ui/src/ethereum/index.ts
```diff
@@ -73,6 +74,16 @@ export function formatTokenAmount(amount: ethers.BigNumberish): string {
   });
 }
 
+export function formatTransactionTimestamp(
+  timestamp: number | dayjs.Dayjs
+): string {
+  return dayjs(timestamp).format("D MMM YYYY [at] HH:mm");
+}
```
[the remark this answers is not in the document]
rudolfs: Good call!

32885af19de1d6433873ea8ef2ce7d5c49913f75

ui/src/org/theGraphApi.ts
```diff
@@ -147,6 +137,51 @@ export async function getOrgProjectAnchors(
   );
 }
 
+export async function getProjectAnchors(
+  projectId: string
+): Promise<project.ConfirmedAnchor[]> {
+  const response = (
+    await orgsSubgraphClient().query({
```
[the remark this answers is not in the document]
rudolfs: Yep. Was an oversight.

456fa6ec5cb218e2094336b855dee13f7a7ac7c1

ui/App/ProjectScreen/ProjectHeader.svelte
```diff
@@ -40,20 +43,25 @@
...
+  .stat-item::after {
+    content: "•";
+    display: flex;
+    color: var(--color-foreground-level-3);
+    margin: 0 1rem 0 1rem;
+  }
+
+  .stat-item:last-of-type::after {
+    content: "";
+    margin: 0;
```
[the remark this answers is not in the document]
rudolfs: 13c8715cda2b6dcfcb94643df963cc14f5bbf180

ui/App/ProjectScreen/Source/Anchors.svelte
rudolfs: 1066991c037263586762180e95e64f9e59b89cda

ui/App/OrgIdentity.svelte
```diff
@@ -0,0 +1,38 @@
...
+</style>
+
+<Avatar
+  {style}
+  size="small"
+  kind={registration?.avatar
+    ? { type: "orgImage", url: registration.avatar }
+    : { type: "orgEmoji", uniqueIdentifier: orgAddress }} />
+{#if registration?.domain}
+  <p class="typo-text-bold org-domain">
```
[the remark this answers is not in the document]
rudolfs: 15487c135165f2a826a5ff4a1337d7ebfc5f2f8f

ui/App/ProjectScreen/Source/Commit.svelte
rudolfs: 15487c135165f2a826a5ff4a1337d7ebfc5f2f8f

ui/App/ProjectScreen/Source/Anchors.svelte
rudolfs: de06ecaac0475b76208c99c9455fda9e60121a91

ui/App/ProjectScreen/route.ts
```diff
@@ -0,0 +1,61 @@
...
+  let anchors: project.ConfirmedAnchor[] = [];
+
+  if (wallet.isConnected()) {
+    anchors = await theGraphApi.getProjectAnchors(params.urn);
+
+    if (
+      params.activeView.type === "anchors" ||
+      params.activeView.type === "commit"
+    ) {
+      await Promise.allSettled(
```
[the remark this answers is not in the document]
rudolfs: This was also an oversight.
3d911656ced3c4d5499df7d4e54aa319692764c9

ui/App/ProjectScreen/Source/Commit.svelte
rudolfs: aa9bc1f7422719b8afef06ffe2faeb48cffa0f75

ui/App/ProjectScreen/Source/Commit.svelte
rudolfs: aa9bc1f7422719b8afef06ffe2faeb48cffa0f75

ui/App/ProjectScreen/Source/Commit.svelte
rudolfs: Wasn't intentional. Seems like the previous `metadata` class wasn't used.

aa9bc1f7422719b8afef06ffe2faeb48cffa0f75

ui/App/ProjectScreen/Source/Commit.svelte
rudolfs: aa9bc1f7422719b8afef06ffe2faeb48cffa0f75

ui/App/ProjectScreen/Source/Commit.svelte
rudolfs: aa9bc1f7422719b8afef06ffe2faeb48cffa0f75

ui/App/OrgIdentity.svelte
rudolfs: I disagree with this point.

I think it's best to leave positioning and spacing up to the enclosing elements. By abiding this principle throughout the codebase it makes it easier to reason about how something will look without having to look at the component itself or having to preview a render. Knowing that components won't touch their outer positioning makes the mental model easier to deal with for me. I think there's also a benefit in terms of component composition.

So far we've been following this principle and I don't see a good reason to break it now.
To avoid always setting the margin from outside we might use some flexbox styling, e.g. `gap`.

rudolfs: Never mind, this was actually an internal spacing 🤦.

rudolfs: 0ea5679cb704a4cce5f9e9c845b1888ea8377b21

brandonhaslegs:
> Does it really make sense to change the order in the header, given that there can be an optional description and in all other cases (user profiles and orgs) we always keep the title and identifier close by?

fine leave it as it is

geigerzaehler (approved):
:clap:

rudolfs:
I extracted the Anchor card into a separate component, there was some duplication. \cc @brandonhaslegs 

Screenshot 2021-09-28 at 13 09 07
Screenshot 2021-09-28 at 13 09 16

geigerzaehler (approved):

PR #2383, opened 2021-09-21
Fix markdown HTML links

geigerzaehler:


rudolfs (approved):
For posterity: we decided that formatting links that contain images like this is okay for now.

Screenshot 2021-09-21 at 13 39 50

Issue #1197, opened 2020-11-06
Resolve cross-Site Cookie warning


MeBrei:
Upstream uses cookies to authenticate with the proxy and all requests to the proxy are cross-origin. Electron may enable stricter cookie handling in the future (https://github.com/electron/electron/issues/24225). This would change the default cookie policy from `SameSite=None` to `SameSite=Lax` and would require `Secure` when using `SameSite=None`. This would mean that setting cross-origin cookies on unencrypted connections is disallowed and this would break our app. Here’s a [good article](https://web.dev/samesite-cookies-explained/) explaining what is going on. 

The only way for us to address this is to not use cookies or to serve the Upstream code from the proxy.

*edited by @geigerzaehler*

geigerzaehler:
@MeBrei can you find out how pressing this is? I.e. when this will break?

MeBrei:
The feature has been already rolled out on the Chrome side: [Chrome Same Site Update](https://www.chromium.org/updates/same-site)
The Electron v11.1.1 we are using seems to be using chromium v87 but the feature is not enabled yet. There is an open issue in https://github.com/electron/electron/issues/24225

geigerzaehler:
It should be possible to set the cookie with the `Secure` attribute because that attribute [is not considered](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies) for request to localhost domains. However electron does not respect that yet.

PR #2388, opened 2021-09-22
chore: update dependencies

geigerzaehler:
Includes an update of marked from `v2` to `v3` ([changelog][1]). The
breaking changes don’t affect us.

Includes a [major update of Electron][2]. This update does not have any
breaking changes except for a major update in Chromium

[1]: https://github.com/markedjs/marked/releases/tag/v3.0.0
[2]: https://www.electronjs.org/releases/stable#15.0.0


rudolfs (approved):
utACK

Do you remember why you pinned `"multibase": "=4.0.4"`?

geigerzaehler:
> Do you remember why you pinned `"multibase": "=4.0.4"`?

The package of v4.0.5 was broken but v4.0.6 seems to work again. I’ll fix it.

geigerzaehler:
Turns out that a couple of things were broken by the Electron and Cypress upgrade. I added another commit that fixes these issues. Here are the commit details:

After the upgrades to Electron and Cypress we need to fix our cookie
handling.

* Electron seems to have changed their cookie handling without notice.  We now have to set the `SameSite=None` attribute for cookies which  also requires the `Secure` attribute. Fixes #1197.

* Cypress now has stricter cookie parsing when using `cy.setCookie()`.  This means we have to properly extract the cookie value and ignore the  attributes.

* The session endpoint was not guarded properly with the authentication  token. We fixed that.

rudolfs:
There are still some type errors in the build.

geigerzaehler:
Reverted the check for the session endopint. It doesn’t work. We need to fix this as part of the architecture migration.

rudolfs (approved):
utACK

Issue #2232, opened 2021-08-12
Flaky networking test on CI


geigerzaehler:
The networking and patches tests sometimes [fail](https://buildkite.com/monadic/radicle-upstream/builds/10197#fd36e975-02b5-4f15-85ab-c8b215f3030f) on CI. The failure is due to data not replicated between the nodes.

geigerzaehler:
In [builds](https://github.com/radicle-dev/radicle-upstream/actions/runs/1139254366) Github Actions we’re seeing the following warning
```plain
recv git error err=Rere(MissingSignedRefs(Urn { id: Oid(c7c303d71b4617696f8eca12b64780c267ad79a8), path: None }))
```
It’s repeated multiple times with the same URN.

geigerzaehler:
Increasing the timeout for tests doesn’t help.

geigerzaehler:
Here’s [another failure](https://buildkite.com/monadic/radicle-upstream/builds/10249#b02a5c84-e5f1-4e2b-b9d2-3e488f7538dc) that doesn’t seem to be caused by a `.lock` file for a reference to exist https://github.com/radicle-dev/radicle-link/issues/758.

geigerzaehler:
Since https://github.com/radicle-dev/radicle-link/pull/777 we’re seeing the following error
```
recv git error err=Rere(Fetcher(Concurrent { urn: Urn { id: Oid(d8fd76b7f3eaecf62a11ead632bd580cca852b30), path: None }, remote_peer: hydkthp9j3ufuzt69nepfzu5x8wcmqmy9ckoi7hythguitfmoee5dh }))
```

https://buildkite.com/monadic/radicle-upstream/builds/10688#f13a8574-97d3-4309-8390-e2c7a4cb350e

rudolfs:
This was fixed, right?

geigerzaehler:
We’ve added some workarounds in #2391 but we should keep this issue open to remind us that we need to fix the underlying cause.

rudolfs:
How about now?

geigerzaehler:
I think we can close this. The tests are still flaky, but the migration to playwright should fix this.

PR #2391, opened 2021-09-22
fix: try to avoid CI race conditions in networking tests

geigerzaehler:
We’re trying to fix flaky networking tests on CI by allowing retries and by adding sleeps to prevent race conditions between the P2P nodes.

See https://github.com/radicle-dev/radicle-upstream/issues/2232

### DIFF
--- a/cypress/integration/networking.spec.ts
+++ b/cypress/integration/networking.spec.ts
@@ -38,7 +38,14 @@ context("p2p networking", () => {
 
   it(
     "replicates a project from one node to another",
-    { defaultCommandTimeout: 8000 },
+    {
+      defaultCommandTimeout: 8000,
+      // Tests involving two nodes are extremely flaky
+      retries: {
+        runMode: 3,
+        openMode: 1,
+      },
+    },
     () => {
       const maintainer = {
         handle: "rudolfs",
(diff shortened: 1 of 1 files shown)


cypress/integration/networking.spec.ts
```diff
@@ -92,6 +99,9 @@ context("p2p networking", () => {
                 throw new Error("Could not find URN");
               }
 
+              // We’re giving the P2P node some space to avoid race
+              // conditions.
+              cy.sleep(2000);
```
rudolfs, line 104: I think the sleeps are redundant, because we already set `defaultCommandTimeout: 8000` to 8s. Maybe we should double that to 16s?

geigerzaehler: The `sleep` calls are different from the timeout. The timeout controls how long commands are retried. But the first try is issued immediately. The purpose of `sleep()` is to delay the next command (like pushing changes with git) so that the nodes have time to settle down.

rudolfs: Ah, that makes sense.

rudolfs (approved):
May blesseth be thine build ✨ 
 🍏 ✅ 📗 🟢 🥗

rudolfs (approved):

PR #2395, opened 2021-09-22
refactor: eliminate use of radicle_daemon::MaybeFrom

geigerzaehler:
We eliminate the reference to `radicle_daemon::MaybeFrom` to reduce the reliance on `radicle_daemon` code.


rudolfs (approved):
utACK

PR #2397, opened 2021-09-23
Custom primary color in settings

### LINKED ISSUE #1803: Customize the primary UI color to your favorite color ❤️💚💙

juliendonck:
closes #1803


ui/App/SettingsScreen.svelte
```diff
@@ -95,6 +118,40 @@
...
+    );
+    root.style.setProperty(
+      "--color-primary-level-2",
+      `rgba(${colorRgb.r},${colorRgb.g},${colorRgb.b},0.37`
+    );
+    root.style.setProperty(
+      "--color-primary-level-6",
+      `rgba(${colorRgb.r},${colorRgb.g},${colorRgb.b},0.87`
+    );
+  }
```
juliendonck, line 154: @rudolfs if here I could target `[data-primarycolor="hex"]` instead of `root` it would work imo

geigerzaehler: I’ll take a look at this. Maybe we can do something with the StyleSheet DOM APIs

geigerzaehler: This should be possible with the [CSSStyleSheet](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule) API.

brandonhaslegs:
I'm so happy about this. What colorpicker is this? Can we find a nicer looking one?

ui/App/SettingsScreen.svelte
```diff
@@ -95,6 +118,45 @@
...
+    styles += `--color-primary: rgba(${colorRgb.r},${colorRgb.g},${colorRgb.b},0.17`;
+    styles += `--color-primary: rgba(${colorRgb.r},${colorRgb.g},${colorRgb.b},0.37`;
+    styles += `--color-primary: rgba(${colorRgb.r},${colorRgb.g},${colorRgb.b},0.87`;
+    styles += "}";
+
+    // Add the first CSS rule to the stylesheet
+    sheet.insertRule(styles, 0);
+  }
+
+  $: console.log($primaryColorStore);
```
juliendonck, line 159: @geigerzaehler @rudolfs trying it this approach and it's not working. I don't know how I can check if the new stylesheet even exists. Also I don't want to pull in master cause that's gonna break everything.

rudolfs (approved):
🎨 ✨ 🌈

brandonhaslegs (changes requested):
Hey when I click on "Custom" the default color is pure black. Would be cool if it didn't change the color when you clicked on custom.

I also think you can keep the square color swatch there even if you choose one of the default color options. It's weird that it appears and disappears when you click Custom.

brandonhaslegs:
By the way, if possible it would be cool if the last color you had selected became the color that was selected when you click "Custom". If this is too difficult, then defaulting back to blue instead of black is fine.

juliendonck:
Screenshot 2021-09-29 at 10 20 50

I had chosen a yellow, but can make it blue if that's better. It's just that it's gonna show that color always since we don't hide it. Also now you can actually change that color in the picker but it won't switch to custom, which I feel it should. So I'm gonna work on that

juliendonck:
this is the hex of the yellow: `  "#FFFC67`

juliendonck:
@brandonhaslegs I did what you asked here but I think it's strange that I see the color picker, I edit the color but it doesn't update the UI. I feel that when you interact with it, the `SegmentedControl` should switch over to `Custom`.

brandonhaslegs:
> @brandonhaslegs I did what you asked here but I think it's strange that I see the color picker, I edit the color but it doesn't update the UI. I feel that when you interact with it, the `SegmentedControl` should switch over to `Custom`.

The square should always be the color that is selected. So if you have Pink, Blue, or Orange selected, the square should be that color.

brandonhaslegs (changes requested):
Looks like the `color-level-1` isn't set up right when you choose a custom color.
image

image

brandonhaslegs (changes requested):
The square isn't aligned correctly.

image

juliendonck:
@brandonhaslegs check out this intermediate state.

brandonhaslegs (approved):
Looks great!

rudolfs:
The custom animation now causes the sidebar to show up on the right side of the screen when we switch between different pages.

https://user-images.githubusercontent.com/158411/135488826-d86e6e9c-b985-4650-a45e-a6422775a012.mov

ui/App/DesignSystemGuideModal.svelte
```diff
@@ -195,11 +197,16 @@
...
-            on:select={ev => primaryColor.set(ev.detail)} />
+          <div style="display: flex;">
+            <SegmentedControl
+              style="background-color: var(--color-background); margin-right: 1rem;"
+              active={$primaryColor}
+              options={primaryColorOptions}
+              on:select={ev => primaryColor.set(ev.detail)} />
+            <ColorInput
+              style="margin-right: 2rem;"
+              bind:value={$primaryColorHex} />
```
rudolfs, line 208: Can we replace this with `<SegmentedColorPicker>` as well?

juliendonck: yep

PR #2400, opened 2021-09-23
refactor: cleanup identity endpoints

geigerzaehler:
* We follow the pattern outlined for other endpoints and move all the methods into a separate identity client class.
* We rename `remoteIdentityGet` to `personGet` as this is terminology used by the identities CLI.
* We eliminate some unused code and indirections.


rudolfs (approved):
👏  utACK

```
  cypress/support/commands.ts(170,23): error TS2339: Property 'identityCreate' does not exist on type 'Client'.
```

Issue #1849, opened 2021-05-03
Move upstream specific data into local storage


geigerzaehler:
At the moment Upstream specific data like theme settings are stored by the proxy and exposed via an API. This setup makes it harder to change the this kind of data. We need to update that proxy code and take care of backwards compatibility with previously stored data (#1845). Since the data is upstream specific there is also no benefit to storing it in the proxy since the data doesn’t need to be accessed by other applications. Browser local storage is a suitable alternative for storing this data.

If we set this up we should also ensure that we handle old data properly when making changes to the schema. This could be accomplished by adding tests with current data that cannot be modified when the schema is changed.

The data that should be stored in local storage is everything in the `settings` field of the `session` endpoint except for the seed nodes.

rudolfs:
Can we close this now, seems like 288bcc473b2eaa3fd74140ae36de922a79fdc965 and be05d080c4b3b3e52070583a8bbce4984627b50e took care of that.

PR #2404, opened 2021-09-23
fix: store remote helper hint data locally

geigerzaehler:
We store information on whether the remote helper hint has been shown in the browser.

This is part of #1849.


rudolfs (approved):
utACK

PR #2405, opened 2021-09-23
refactor: cleanup references to radicle_daemon

geigerzaehler:
* Replace use of re-exports by `radicle_daemon` with use of the actual crates. E.g. `link_identities::git::Urn` instead of `radicle_daemon::Urn`.

* Reference `radicle_daemon` values full-qualified to make them easier to grep for.


rudolfs (approved):
utACK

PR #2407, opened 2021-09-23
fix: store appearance settings in browser

### LINKED ISSUE #2130: Make lock screen honor theme settings
By default, the lock screen is dark-themed.
Currently when you set the theme to a different setting (like Light), the lock screen doesn't take up that preference.

STRs(Steps To Reproduce)
1. Under homepage settings, change the theme to Light.
2. Lock the app or re-launch the app.
3. Notice the lock screen is still dark-themed.

Not a priority but something to look into (which even I can take it up and check it out)

geigerzaehler:
Instead of storing appearance settings in the proxy we store them in the browser. This allows us to simplify the code significantly.

Fixes #2130 and #1849


rudolfs:
Overall looks great.

Just one thing: this breaks existing behaviour, switching themes doesn't switch the code highlighting theme in the proxy anymore. We could solve this by passing back the theme to the source browsing API or move syntax highlighting to the front-end. What do you think?

geigerzaehler:
Great spot with the syntax highlighting. I think passing it to the proxy as a parameter is the way to go. I’ll fix this up.

rudolfs:
Dark theme:

| master | this branch |
|--------|------------|
| Screenshot 2021-09-24 at 10 48 57 | Screenshot 2021-09-24 at 10 47 59 |

Light theme:

| master | this branch |
|--------|------------|
| Screenshot 2021-09-24 at 10 48 45 | Screenshot 2021-09-24 at 10 47 47 |

rudolfs:
> Great spot with the syntax highlighting. I think passing it to the proxy as a parameter is the way to go. I’ll fix this up.

Nice. Once we have that, we can also remove the old settings storage code from the backend.

rudolfs (approved):
🙌

Issue #1137, opened 2020-10-30
Provide separate endpoints for identity and settings


geigerzaehler:
At the moment both the user identity and the app settings are controlled through the same `/session` resource. Instead they should be exposed as different resources.

## Motivation

The identity and settings have different characteristics that make it hard to have them under one resource.

* The settings (in particular the theme setting) need to be available from the first start of the app while the identity is only available after onboarding.
* The identity should only be available if the user has unsealed the proxy and provided the correct auth cookie. In contrast, the settings should be readable from the begining (see above)
* The identity resource can also be changed through the `/identity` endpoint while the settings can only be changed by the `/session` endpoint.

This would also address the concerns raised in https://github.com/radicle-dev/radicle-upstream/pull/1120#discussion_r514527988.

## API

*Rough draft*

### `GET /settings`

Returns stored settings. Returns default settings if nothing stored. Requires no auth.

### `PUT /settings`

Updates stored settings. Requires auth.

### `GET /identity/me`

Get current identity if it was created, otherwise 404. Requires auth

### `/session`
This endpoint is eliminated

geigerzaehler:
In #1849 we outlined that we want to move most of the settings to local storage. The only setting remaining in the API would be the seed node list.

geigerzaehler:
Superseded by #2459

PR #2409, opened 2021-09-24
feat: add dedicated endpoint for network seeds

geigerzaehler:
We add a dedicated endpoint for getting and updating seed addresses for the proxy. This decouples the seeds from the session and will allow us to eliminate the settings in the future. See #1137


rudolfs (approved):
utACK

🧹

PR #2414, opened 2021-09-24
Migrate to host/id based seed node records

sebastinez:
This PR changes the ENS records to start working with `seed.host` instead of `seed.api`

It is based on the changes introduced in https://github.com/radicle-dev/radicle-interface/commit/75983dc98ee0654f9af2aed7e08119527cfa4a3a

One thing I wasn't quite sure about, is the `OrgHeader` where we now show the `seedHost` it seems to me that it should be a `CopyableIdentifier` but I didn't found a good way to shorten it without losing important information.
Let me know what you think.

Feel free to provide feedback.

### DIFF
--- a/ui/App/OrgScreen/ConfigureEns/UpdateMetadata.svelte
+++ b/ui/App/OrgScreen/ConfigureEns/UpdateMetadata.svelte
@@ -26,7 +26,7 @@
   let twitterValue: string | undefined = registration.twitter || undefined;
   let githubValue: string | undefined = registration.github || undefined;
   let seedIdValue: string | undefined = registration.seedId || undefined;
-  let seedApiValue: string | undefined = registration.seedApi || undefined;
+  let seedHostValue: string | undefined = registration.seedHost || undefined;
 
   let setRecordsInProgress = false;
 
@@ -57,7 +57,7 @@
       { name: "twitter", value: twitterValue },
       { name: "github", value: githubValue },
       { name: "seedId", value: seedIdValue },
-      { name: "seedApi", value: seedApiValue },
+      { name: "seedHost", value: seedHostValue },
     ];
 
     // Filter out unchanged records.
@@ -186,15 +186,15 @@
   <TextInput
     disabled={setRecordsInProgress}
     style="margin-bottom: 24px"
-    placeholder="The Radicle Link node ID that hosts your org’s entities"
+    placeholder="The Device ID of the seed"
     bind:value={seedIdValue} />
 
-  <div class="label typo-text-bold">Seed API</div>
+  <div class="label typo-text-bold">Seed Host</div>
   <TextInput
     disabled={setRecordsInProgress}
     style="margin-bottom: 24px"
-    placeholder="The HTTP address of a node that serves Radicle entities"
-    bind:value={seedApiValue} />
+    placeholder="The seed host name"
+    bind:value={seedHostValue} />
 
   <svelte:fragment slot="buttons">
     <Button
--- a/ui/App/OrgScreen/OrgHeader.svelte
+++ b/ui/App/OrgScreen/OrgHeader.svelte
@@ -30,7 +30,7 @@
     registration?.twitter &&
     `https://twitter.com/${registration.twitter.replace("@", "")}`;
   $: seedId = registration?.seedId;
-  $: seedApi = registration?.seedApi;
+  $: seedHost = registration?.seedHost;
 </script>
 
 <style>
@@ -102,7 +102,7 @@
           </div>
         {/if}
       </div>
-      {#if websiteUrl || githubUrl || twitterUrl || seedId || seedApi}
+      {#if websiteUrl || githubUrl || twitterUrl || (seedId && seedHost)}
         <div>
           {#if websiteUrl}
             <div class="row">
@@ -139,20 +139,11 @@
           {/if}
         </div>
         <div>
-          {#if seedId}
+          {#if seedId && seedHost}
             <div class="row">
-              <CopyableIdentifier value={seedId} kind="seedAddress" />
-            </div>
-          {/if}
-          {#if seedApi}
-            <div class="row">
-              <Icon.Globe />
-              <div class="url">
-                <span
-                  on:click={() => {
-                    seedApi && ipc.openUrl(seedApi);
-                  }}>{seedApi}</span>
-              </div>
+              <CopyableIdentifier
+                value={`${seedId}@${seedHost}:8776`}
+                kind="seedAddress" />
             </div>
           {/if}
         </div>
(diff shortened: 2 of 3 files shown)


ui/App/OrgScreen/ConfigureEns/UpdateMetadata.svelte
```diff
@@ -189,12 +189,12 @@
     placeholder="The Radicle Link node ID that hosts your org’s entities"
     bind:value={seedIdValue} />
 
-  <div class="label typo-text-bold">Seed API</div>
+  <div class="label typo-text-bold">Seed Host</div>
   <TextInput
     disabled={setRecordsInProgress}
     style="margin-bottom: 24px"
     placeholder="The HTTP address of a node that serves Radicle entities"
```
cloudhead, line 196:
```suggestion
    placeholder="The seed host name"
```

ui/App/OrgScreen/OrgHeader.svelte
```diff
@@ -144,14 +144,14 @@
...
-          {#if seedApi}
+          {#if seedHost}
             <div class="row">
               <Icon.Globe />
               <div class="url">
                 <span
                   on:click={() => {
-                    seedApi && ipc.openUrl(seedApi);
-                  }}>{seedApi}</span>
+                    seedHost && ipc.openUrl(seedHost);
```
cloudhead, line 153: `seedHost` is not a URL per se, but if you want, you can open `https://{seedHost}:8777` which is the JSON REST API..

ui/App/OrgScreen/ConfigureEns/UpdateMetadata.svelte
```diff
@@ -189,12 +189,12 @@
     placeholder="The Radicle Link node ID that hosts your org’s entities"
```
cloudhead, line 189:
```suggestion
    placeholder="The Radicle Link Device ID that hosts your org’s entities"
```

ui/App/OrgScreen/OrgHeader.svelte
cloudhead: Though actually maybe it's best to not create a link out of it.

sebastinez: Yeah that was my comment on this PR, I think the link is of no good use here.
But eventually we could use the `CopyableIdentifier` as being used for the seed id.
What do you think?

cloudhead: Yes on the web client we expose the full `{id}@{host}:8776`, as that's useful. I'd let the upstream devs decide on this though.

geigerzaehler (approved):
Nice

ui/App/OrgScreen/OrgHeader.svelte
geigerzaehler: I think using just the copyable identifier is fine. Showing the device URL to easily add it to the seed nodes also makes sense. Any thoughs on this @juliendonck and @rudolfs?

rudolfs: Is `8776` the default port for the p2p protocol? If so, showing `{id}@{host}:8776` would make sense.

What I intended to show in the header was the seed address, e.g. for our current seed host:
`hynkyndc6w3p8urucakobzna7sxwgcqny7xxtw88dtx3pkf7m3nrzc@sprout.radicle.xyz:12345`, so that a user could copy that and add it directly to their seed list on the Network screen.

For this purpose we have the `<Identifier kind="seedAddress" />` component.
It looks like this when rendered:

Screenshot 2021-09-27 at 10 54 25

rudolfs: So basically this:

```diff
diff --git a/ui/App/OrgScreen/OrgHeader.svelte b/ui/App/OrgScreen/OrgHeader.svelte
index 65524767..5863b4d1 100644
--- a/ui/App/OrgScreen/OrgHeader.svelte
+++ b/ui/App/OrgScreen/OrgHeader.svelte
@@ -102,7 +102,7 @@
           </div>
         {/if}
       </div>
-      {#if websiteUrl || githubUrl || twitterUrl || seedId || seedHost}
+      {#if websiteUrl || githubUrl || twitterUrl || (seedId && seedHost)}
         <div>
           {#if websiteUrl}
             <div class="row">
@@ -139,14 +139,11 @@
           {/if}
         </div>
         <div>
-          {#if seedId}
+          {#if seedId && seedHost}
             <div class="row">
-              <CopyableIdentifier value={seedId} kind="seedAddress" />
-            </div>
-          {/if}
-          {#if seedHost}
-            <div class="row">
-              <CopyableIdentifier value={seedHost} kind="seedHost" />
+              <CopyableIdentifier
+                value={`${seedId}@${seedHost}:8667`}
+                kind="seedAddress" />
             </div>
           {/if}
         </div>
```

Screenshot 2021-09-27 at 11 00 17

sebastinez: Okay no problem! Sounds great!

sebastinez: Done

ui/DesignSystem/CopyableIdentifier.svelte
```diff
@@ -10,6 +10,7 @@
     | "radicleId"
     | "deviceId"
     | "seedAddress"
+    | "seedHost"
```
rudolfs, line 13: We can revert the changes in this file. We're only using the `seedAddress` kind.

sebastinez: Yeah you're right, removed `seedHost` from the component

ui/App/OrgScreen/OrgHeader.svelte
```diff
@@ -102,7 +102,7 @@
           </div>
         {/if}
       </div>
-      {#if websiteUrl || githubUrl || twitterUrl || seedId || seedApi}
+      {#if websiteUrl || githubUrl || twitterUrl || seedId || seedHost}
```
rudolfs, line 105:
```suggestion
      {#if websiteUrl || githubUrl || twitterUrl || (seedId && seedHost)}
```

We don't want the outer wrapper to show up in case only one of the seed attributes is present.

ui/App/OrgScreen/ConfigureEns/UpdateMetadata.svelte
```diff
@@ -186,15 +186,15 @@
   <TextInput
     disabled={setRecordsInProgress}
     style="margin-bottom: 24px"
-    placeholder="The Radicle Link node ID that hosts your org’s entities"
+    placeholder="The Radicle Link Device ID that hosts your org’s entities"
```
rudolfs, line 189:
```suggestion
    placeholder="The Device ID of the seed"
```

rudolfs (approved):
Nice work!

rudolfs:
Merged via bc1ae5ddb2113272896a553024ee44bd9a90eb77.

PR #2416, opened 2021-09-27
refactor: use svelte types in remote store

geigerzaehler:
Instead of using locally declared type aliases for we use the proper
types from `svelte/store`.


rudolfs (approved):
👍

PR #2418, opened 2021-09-27
refactor: eliminate custom proxy API calls in tests

geigerzaehler:
We eliminate all custom HTTP API calls in the tests and replace them with calls to the proxy client.


rudolfs (approved):
🤯
Things are really coming together! Nice work.

PR #2421, opened 2021-09-28
refactor: remove endpoint for getting local identities

geigerzaehler:
We remove the proxy endpoint for getting local identities since it was unused.


rudolfs (approved):
utACK

PR #2422, opened 2021-09-28
fix: persist connection state of test wallet connect

geigerzaehler:


rudolfs:
Why do we need this? Should some of it not be wrapped in `isCypressTestEnv`?

geigerzaehler:
> Why do we need this?

Before, you always had to manually connect the test wallet through the UI whenever the application was reloaded. This was super annoying when you changed some styles because the app reloaded and required you to go to the wallet page and click "connect" again before you could use anything wallet related.

> Should some of it not be wrapped in `isCypressTestEnv`?

You can also use the test wallet during development so we would also need to check for that. This would make the code overall more complicated since the module variable `testClientConnected` could also be `undefined`. This would require guarding with an `if` everywhere we use it. I don’t think there’s any harm done creating this store in production, too. Keep in mind that initializing the store does not actually write to the underlying local storage.

rudolfs (approved):
utACK

PR #2423, opened 2021-09-28
refactor: introduce Response struct for easier error conversion

geigerzaehler:
* Eliminate `Error` struct which was only used as the payload and replace it with the `Response` struct which also includes the status code.
* Separate responsibilities for converting an error into an HTTP response and actually replying. The former is implemented via `From` traits.
* Remove globally defined SessionInUse error to demonstrate new design

### DIFF
--- a/proxy/api/src/error.rs
+++ b/proxy/api/src/error.rs
@@ -8,8 +8,6 @@
 
 use std::io;
 
-use link_identities::git::Urn;
-
 use crate::keystore;
 
 /// All error variants the API will return.
@@ -32,9 +30,6 @@ pub enum Error {
     /// An I/O error occurred.
     #[error(transparent)]
     Io(#[from] io::Error),
-
-    #[error("the current session is in use by `{0}`")]
-    SessionInUse(Urn),
 
     /// Issues when access persistent storage.
     #[error(transparent)]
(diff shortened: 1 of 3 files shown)


rudolfs (approved):
Looks good from what I can tell!

Now that we're touching errors, something that has bothered me before is that we return a 500 when querying non-existent projects. Could we switch to 404s for that?

And another thing: some of the responses that we see during normal operation are logged as `ERROR`s in the proxy console, could we switch those to `INFO`?

```
Sep 28 14:53:01.839 ERROR api::http::error: request error err=Rejection(ProjectNotFound)
Sep 28 14:47:19.245 ERROR api::http::error: request error err=Rejection(KeystoreSealed)
Sep 28 14:47:24.060 ERROR api::http::error: request error err=Rejection(Keystore(Error { inner: Crypto(InvalidKey) }))

geigerzaehler:
Happy to tackle the two issues you mentioned in a separate PR. Could you create issues for them?

rudolfs:
> Happy to tackle the two issues you mentioned in a separate PR. Could you create issues for them?

Here you go: https://github.com/radicle-dev/radicle-upstream/issues/2427 https://github.com/radicle-dev/radicle-upstream/issues/2428.

PR #2433, opened 2021-09-29
Add hover states to profile cards

### LINKED ISSUE #2425: Add hoverstate to project cards
The new project cards on the profile don't have hover states.

juliendonck:
closes #2425 

Screenshot 2021-09-29 at 13 12 02

### DIFF
--- a/ui/App/ProfileScreen/ProjectCardSquare.svelte
+++ b/ui/App/ProfileScreen/ProjectCardSquare.svelte
@@ -31,6 +31,12 @@
   .project-card:hover {
     background-color: var(--color-foreground-level-1);
   }
+
+  .project-card:active {
+    transition: transform 0.1s ease-in-out;
+    transform: scale(0.95);
+  }
+
   .title-row {
     display: flex;
     margin-bottom: 1rem;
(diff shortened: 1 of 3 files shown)


brandonhaslegs:
They should also have the same click state as everything else.

brandonhaslegs (changes requested):
They should also have the same press-down state as everything else.

ui/App/ProfileScreen/ProjectCardSquare.svelte
```diff
@@ -27,6 +27,16 @@
...
   }
+
+  .project-card:hover {
+    background-color: var(--color-foreground-level-1);
+  }
+
+  .project-card:active {
+    transition: transform 0.1s ease-in-out;
+    transform: scale(0.95);
+  }
```
rudolfs, line 38: Could we just use the global `.button-transition` class on the respective element?

brandonhaslegs (approved):
Looks good. Would be cool if when you clicked to copy the Radicle ID, the whole card didn't animate, but that seems impossible to do?

ui/App/ProfileScreen/ProjectCardSquare.svelte
juliendonck: No, cause it also does the opacity thing, which we don't want here. What I could do is make a transition variable and use it in both places.

rudolfs: Ok, never mind then.

Issue #2279, opened 2021-08-25
Fix tooltip alignment on onboarding screen


rudolfs:
It should be above the Device ID.

<img width="1552" alt="Screenshot 2021-08-25 at 12 33 06" src="https://user-images.githubusercontent.com/158411/130777227-868de42c-1bf7-40cf-9f61-e28f2eda2458.png">

PR #2443, opened 2021-09-30
fix: relative positioning for tooltips

geigerzaehler:
We position tooltips relative to their container.

Using `position: fixed` (i.e. relative to the viewport) does not work when a containing element has the `transform` property set [1]. In that case the tooltip would be positioned relative to that element.

This example is from #2433.

Peek 2021-09-30 11-55

[1]: https://developer.mozilla.org/en-US/docs/Web/CSS/position#values

### DIFF
--- a/ui/App/DesignSystemGuideModal.svelte
+++ b/ui/App/DesignSystemGuideModal.svelte
@@ -732,7 +732,7 @@
         </div>
 
         <div class="swatch">
-          <Tooltip value="Bottom" position="bottom">
+          <Tooltip value="A very long text" position="bottom">
             <Button variant="outline">Hover me!</Button>
           </Tooltip>
         </div>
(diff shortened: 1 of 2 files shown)


ui/App/DesignSystemGuideModal.svelte
```diff
@@ -732,7 +732,7 @@
         </div>
 
         <div class="swatch">
-          <Tooltip value="Bottom" position="bottom">
+          <Tooltip value="A very long text" position="bottom">
```
rudolfs, line 735: Can we make all the example swatches have the same text?

rudolfs (commented):
Nice!

This also fixes: https://github.com/radicle-dev/radicle-upstream/issues/2279.

Screenshot 2021-09-30 at 12 19 18

ui/App/DesignSystemGuideModal.svelte
geigerzaehler: I wanted to have the text for the top position shorter than the container so I only updated this one.

rudolfs (approved):

PR #2444, opened 2021-09-30
refactor: deduplicate and annotate identity schemas

geigerzaehler:
* We deduplicate the schema for `Metadata`
* We put type annotations on the schema to ensure that they are schemas for the given types.


rudolfs (approved):
utACK

Issue #2432, opened 2021-09-29
Adopt Radicle CLI libraries


geigerzaehler:
This issue is part of #2431.

The objective is to fully integrate the Radicle CLI libraries into the proxy. Only code from `radicle_daemon` remains that is related to networking and the protocol.

We will not use the SSH agent for authentication as part of this work. Instead we will keep the proxy keystore.

See [RFC#696](https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/0696-p2p-node.adoc) for details on the CLI.

## Tasks

- [x] PoC with one read-only call to a function from the Radicle CLI libraries (#2458)
- [ ] #2459 
- [ ] Integrate source browsing with the CLI libraries.
- [ ] Use the Radicle CLI for all remaining update functionality.
- [ ] Review outstanding issues

rudolfs:
Is this still relevant @geigerzaehler?

PR #2445, opened 2021-09-30
Move code to proxy client

geigerzaehler:
While working on integrating the new Radicle CLIs (#2432) to do source browsing I realized that it would be helpful get a better understanding of the HTTP API for source browsing. Making it more type-safe will also help detecting mistakes.

Please see the individual commits for details.


rudolfs (approved):
🧹

Issue #2448, opened 2021-09-30
Commit not found error


brandonhaslegs:
When an anchored commit is not replicated, but is linked, when you click on it, you get an error. I thought we were only showing a link when the commit was replicated? If it's not replicated, I thought we were going to just show a tooltip that says "this commit isn't replicated" or something.

![image](https://user-images.githubusercontent.com/4406983/135520264-8c31015a-cd35-441c-b52a-1936c0e5a39b.png)

```{
  "stack": "Error: object not found - no match for id (3b6b4b3d198c0070c8ba57846ec5e154826207d4); class=Odb (9); code=NotFound (-3)\n    at tb (file:///Applications/Radicle%20Upstream.app/Contents/Resources/app.asar/public/bundle.js:14:6680)\n    at async fJ (file:///Applications/Radicle%20Upstream.app/Contents/Resources/app.asar/public/bundle.js:34:318323)",
  "message": "object not found - no match for id (3b6b4b3d198c0070c8ba57846ec5e154826207d4); class=Odb (9); code=NotFound (-3)",
  "code": "UnknownException",
  "details": {
    "name": "Error",
    "response": {}
  }
}```

rudolfs:
Good point. I think it might be better for performance reasons to show the link, but it should lead to a commit page that says the commit isn't replicated yet. Would that work for you @brandonhaslegs?

brandonhaslegs:
That would work for me, yes. Is that easier than just making the commit link not clickable?

rudolfs:
> That would work for me, yes. Is that easier than just making the commit link not clickable?

Yeah, because otherwise we need to fetch all the commits for the anchors to decide whether to link to them or not. Not having to check this would also be faster in case there's many anchors.

PR #2451, opened 2021-10-01
fix: show commit not found screen instead of error

rudolfs:
Show a placeholder if the commit isn't replicated instead of an error message. This also fixes the anchor list not being properly ordered, now the latest anchor is at the top and the list is sorted by timestamp oldest at the bottom.

Closes https://github.com/radicle-dev/radicle-upstream/issues/2448.

Screenshot 2021-10-01 at 19 08 12
Screenshot 2021-10-01 at 19 12 46

### DIFF
--- a/ui/src/screen/project/source.ts
+++ b/ui/src/screen/project/source.ts
@@ -284,14 +284,18 @@ export const fetchCommit = async (sha1: string): Promise<void> => {
     try {
       commitStore.success(await source.fetchCommit(project.urn, sha1));
     } catch (err: unknown) {
-      commitStore.error(error.fromUnknown(err));
-      error.show(
-        new error.Error({
-          code: error.Code.CommitFetchFailure,
-          message: "Could not fetch commit",
-          source: err,
-        })
-      );
+      const e = error.fromUnknown(err);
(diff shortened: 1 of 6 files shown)


juliendonck:
Great empty state

Screenshot 2021-10-04 at 07 48 12

ui/src/screen/project/source.ts
```diff
@@ -284,14 +284,18 @@ export const fetchCommit = async (sha1: string): Promise<void> => {
...
-      commitStore.error(error.fromUnknown(err));
-      error.show(
-        new error.Error({
-          code: error.Code.CommitFetchFailure,
-          message: "Could not fetch commit",
-          source: err,
-        })
-      );
+      const e = error.fromUnknown(err);
+      if (!e.message.match("object not found")) {
```
geigerzaehler, line 288: Could you flip the `if/else` branches and get rid of the negation?

geigerzaehler (approved):

ui/src/screen/project/source.ts
rudolfs: 51bf94ca894826c5b804aaf25b2310c2ee11c51d

PR #2454, opened 2021-10-04
improve docs for org seed node

geigerzaehler:
* List resources and their IDs up front
* Mention org seed node in GCP project overview
* Add step that creates the network address
* Consolidate some of the steps to set up the VM

### DIFF
--- a/docs/infrastructure.md
+++ b/docs/infrastructure.md
@@ -98,9 +98,10 @@ apt-get update
 apt-get install docker-ce docker-ce-cli containerd.io
 ```
 
-6. As `orgnode`, set up radicle-client-services:
+6. Set up radicle-client-services:
 
 ```bash
+sudo su
 apt-get install python3-pip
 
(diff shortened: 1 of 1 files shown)


docs/infrastructure.md
```diff
@@ -88,15 +96,16 @@ echo \
...
-apt-get install python3-pip
-
-adduser --disabled-password --disabled-login orgnode
-usermod -aG docker orgnode
 ```
 
 6. As `orgnode`, set up radicle-client-services:
 
 ```bash
+apt-get install python3-pip
```
rudolfs, line 104: The heading doesn't match these commands, they should be executed as root.

geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/2454/commits/3bac01f09f66792dbce4dec562879710e271b44d

rudolfs (approved):
👍

PR #2458, opened 2021-10-05
feat: use rad-identities to get person information

geigerzaehler:
We use the `rad-identities` crate to get information about a person
identity.

This is the first step in #2432.


rudolfs (approved):
👏

PR #2462, opened 2021-10-06
refactor: fetch single commits through proxy client

geigerzaehler:
We add `commitGet` to the proxy client and use this function exclusively to get commit data.

This requires us to move the type definitions for commits and diffs to the proxy (`ui/src/proxy/commit`) and also implement schemas for these types.

We also eliminate some type definitions and inline them.

We also remove the `changeset` field from `Commit`. It is not used in the client and not provided by the proxy.


rudolfs (approved):
utACK

PR #2464, opened 2021-10-07
refactor: add `localStateGet` to proxy and use it exclusively

geigerzaehler:
We add the `localStateGet` method to proxy client and use it exclusively. This reveals that the `managed` field on `LocalState` is not provided by the proxy and is not used. We remove it.

We also remove tests from `ui/src/project.test.ts` that involve mocking the local state endpoint.


rudolfs (approved):
utACK 🧹

PR #2470, opened 2021-10-08
perf: reduce proxy requests when viewing orgs

rudolfs:
We only resolve member identities on the "Members" view.

Required by https://github.com/radicle-dev/radicle-upstream/pull/2450.


ui/src/org.ts
```diff
@@ -433,33 +429,35 @@ export async function fetchMembers(
     gnosisSafeAddress
   );
 
+  return {
+    threshold: response.threshold,
+    members: response.members.map(address => {
+      return { ethereumAddress: address };
+    }),
```
geigerzaehler, line 436: The idea behind  making the `identity` field on `Member` optional was to indicate that we tried to resolve the link identity but it didn’t exist. Here a missing `identity` field indicates that we haven’t tried to resolve this yet. To avoid this ambiguity and to make the code overall simpler we could just have a list of addresses here.
```suggestion
    members: response.members
```

ui/src/org.ts
```diff
@@ -433,33 +429,35 @@ export async function fetchMembers(
...
+  };
+}
+
+export async function resolveMemberIdentities(
+  unresolvedMembers: Member[]
```
geigerzaehler, line 441: As per my comment above, this would become
```suggestion
  addresses: string[]
```

ui/src/org.ts
```diff
@@ -402,11 +402,7 @@ interface GnosisSafeOwner {
 export async function getOwner(orgAddress: string): Promise<Owner> {
   const walletStore = svelteStore.get(wallet.store);
   const address = await Contract.getOwner(orgAddress, walletStore.provider);
-  const ownerCode = await walletStore.provider.getCode(address);
-  // We’re not really checking that the address is the Gnosis Safe
-  // contract. We’re just checking if it is _a_ contract.
-  const isSafe = ownerCode !== "0x";
-  if (isSafe) {
+  if (await isMultiSig(address)) {
```
geigerzaehler, line 405: Excellent find for optimization!

ui/src/org.ts
rudolfs: Like this: 01d7a759c6de138673ccd20260305bcf2e1765e1?

Issue #1877, opened 2021-05-10
Make the design system exportable


rudolfs:
We'd like to use the design system in other projects in the Radicle ecosystem. So far we've been using an outdated and incomplete copy of the design system for the seed node in the`radicle-bins` repository. Keeping the two in-sync is unneeded overhead.

As Upstream is the primary consumer of the design system, it would be beneficial to keep it within this repository and if possible only export it as a npm package from this repo.

Proposed plan of action:
  - [x] remove the `Primitive`/`Components` dichotomy to provide a simpler api https://github.com/radicle-dev/radicle-upstream/pull/2033
  - [x] move any non-reusable components out of the design system
  - [x] make components composable by moving any data fetching logic out of them
  - [x] figure out and implement a clear separation of Upstream types vs DesignSystem types
  - [ ] export the design system via its own `package.json`
  - [ ] find a way to easily share global styling and assets

rudolfs:
There's still one thing left to do: switch the Upstream repository over to yarn workspaces, so that we can properly package the design system and be able to import it from other repositories. But as we don't have a use-case for that yet, we didn't invest the time.

However, the design system is now stand-alone and [deployed to Github pages](https://radicle-dev.github.io/radicle-upstream/).

PR #2474, opened 2021-10-12
chore: decouple design system `unreachable` from Upstream

rudolfs:
We want to decouple the design system from Upstream, so it can be re-used by other projects.

A step towards: https://github.com/radicle-dev/radicle-upstream/issues/1877.

### DIFF
--- a/ui/DesignSystem/lib/unreachable.ts
+++ b/ui/DesignSystem/lib/unreachable.ts
@@ -1,0 +1,9 @@
+// Copyright © 2021 The Radicle Upstream Contributors
+//
+// This file is part of radicle-upstream, distributed under the GPLv3
+// with Radicle Linking Exception. For full terms see the included
+// LICENSE file.
+
+export const unreachable = (value: never): never => {
+  throw new Error(`Unreachable code: ${value}`);
+};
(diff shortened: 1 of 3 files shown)


ui/DesignSystem/lib/unreachable.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright © 2021 The Radicle Upstream Contributors
+//
+// This file is part of radicle-upstream, distributed under the GPLv3
+// with Radicle Linking Exception. For full terms see the included
+// LICENSE file.
+
+export const unreachable = (_: never): void => {};
```
geigerzaehler, line 7: We should probably throw an error and attach the argument so that we know at runtime if something with the types is off. The return type should also be `never` so that we can use this in any place where we need to return something. For example https://github.com/radicle-dev/radicle-upstream/blob/4be4f1a9a850e49ddfbe2eab7933e067183366fd/ui/src/network.ts#L70

geigerzaehler (approved):

ui/DesignSystem/lib/unreachable.ts
rudolfs: b5740ea07de0f789823edcaae1ee976ee1a588e6

PR #2475, opened 2021-10-12
refactor: consolidate TextInput and PasswordInput

rudolfs:
Reduces some duplication making it easier to apply changes to what is essentially the same component.

Refs https://github.com/radicle-dev/radicle-upstream/issues/1877.


ui/DesignSystem/TextInput.svelte
```diff
@@ -6,40 +6,48 @@
...
+  // processing was blocked because a document's URL has a fragment".
   // preventScroll is necessary for onboarding animations to work.
   $: if (autofocus) {
     inputElement && inputElement.focus({ preventScroll: true });
   }
 
-  $: showHint = hint.length > 0 && value.length === 0;
+  $: if (inputElement) {
+    inputElement.type = concealed ? "password" : "text";
+  }
```
geigerzaehler, line 48: It seems more idiomatic to me to drop these lines and do it in svelte:
```svelte
<input type={concealed ? "password" : "text"} />
```

geigerzaehler: Interesting :thinking:. Could you add a comment about this?

PR #2490, opened 2021-10-18
fix: ignore RUSTSEC-2020-0071

geigerzaehler:
We ignore [RUSTSEC-2020-0071][1] because we’re likely unaffected and this probably isn’t going to be fixed any time soon.

[1]: https://rustsec.org/advisories/RUSTSEC-2020-0071


juliendonck (dismissed):

rudolfs:
Looks like the updates broke the build.

juliendonck (approved):

rudolfs (approved):

PR #2491, opened 2021-10-19
fix: ignore cargo-deny advisories

rudolfs:
They're all coming from radicle-link and should be fixed there. Muting
the advisories until they get fixed upstream.

Signed-off-by: Rūdolfs Ošiņš <rudolfs@osins.org>

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -48,6 +48,7 @@ ignore = [
   #
   # https://rustsec.org/advisories/RUSTSEC-2020-0071
   "RUSTSEC-2020-0071",
+  "RUSTSEC-2020-0159",
   # "RUSTSEC-2021-0080"
 
 ]


deny.toml
```diff
@@ -48,6 +48,13 @@ ignore = [
   #
   # https://rustsec.org/advisories/RUSTSEC-2020-0071
   "RUSTSEC-2020-0071",
+  "RUSTSEC-2016-0005",
+  "RUSTSEC-2020-0016",
+  "RUSTSEC-2020-0056",
+  "RUSTSEC-2020-0058",
+  "RUSTSEC-2020-0095",
+  "RUSTSEC-2020-0159",
+  "RUSTSEC-2021-0064",
```
geigerzaehler, line 57: All but 2020-0159 are warnings about crates being unmaintained. Those warnings we can keep because they don’t make the build fail. I’d like to keep the list of ignored advisories small and well documented so it’s easy to understand and review.

rudolfs: 6a39b7ce65bda47133797557466f205a68fe4e0f

PR #2497, opened 2021-10-19
feat: properly shutdown proxy when app is closed

geigerzaehler:
With this change the `radicle-proxy` now shuts downs properly when it receives SIGINT or SIGTERM. In addition, Electron waits for `radicle-proxy` to shutdown properly before it quits.


geigerzaehler:
@rudolfs: Could you test this on macOS? One thing that should work is that the terminal should show the proxy logs that indicate the process is shutting down. (I.e. electron should not quit before the proxy does)

rudolfs:
> I’ve addressed the issue by closing the window immediately on Cmd+Q while the app shuts down in the background.

Nice. Much better!

PR #2498, opened 2021-10-20
refactor: load userProfile data in the router

rudolfs:
This allows us to show a loading indicator while the route is loading and keeps the data loading logic out of the screen component.


geigerzaehler (approved):
:sparkling_heart:

PR #2499, opened 2021-10-20
typo fix 💪

juliendonck:


ui/App/ProfileScreen.svelte
```diff
@@ -221,8 +221,10 @@
             <div>
               <h2>Still looking...</h2>
               <p style="margin-top: 1rem;">
-                {projectCountText($profileProjectsStore.data.requests.length)} you’re
-                following haven’t been found yet.
+                {projectCountText($profileProjectsStore.data.requests.length)}
+                {$profileProjectsStore.data.requests.length > 1
+                  ? `you’re following haven’t been found yet.`
+                  : `you’re following hasn't been found yet.`}
```
rudolfs, line 227: Can we just move the whole text into the function? Then we wouldn't have to branch in two places.

juliendonck: We use the function in two places, The text and then the button.

rudolfs (approved):

PR #2501, opened 2021-10-20
refactor: replace `Header` with `ScreenLayout` slot.

geigerzaehler:
Instead of having `Header` as a standalone component we provide it as a slot in `ScreenLayout`. The motivation is that we want to make `ScreenLayout` responsible for content padding and size constraints. This only works if the screen component is aware of the `Header`.


rudolfs (approved):
Nice work! 📐 

Not sure this is still in scope for this PR, but the empty state placeholders still make the screen scrollable when it's not necessary.

https://user-images.githubusercontent.com/158411/138342972-1dfc6ecb-4fc3-4b8a-b0ff-2e831216c047.mov


https://user-images.githubusercontent.com/158411/138342987-40b0fd16-8cea-445c-908b-68ace9145f44.mov

PR #2558, opened 2021-11-01
use `git ls-files` to get paths for license check

geigerzaehler:
Use `git ls-files` to get paths to check for the license headers. This is simpler and does not require adjustments when directories are added or changed.


rudolfs (approved):
utACK

PR #2561, opened 2021-11-01
control proxy args through environment variables

geigerzaehler:
We make it possible to control proxy arguments with environment variables. This makes it easier to modify proxy behavior when running `yarn start`.

To accomplish this we switch from `argh` to `structopt`.


docs/development.md
```diff
@@ -123,22 +123,22 @@ mkdir /Users/rudolfs/work/20000
...
 RADICLE_UPSTREAM_UI_PROXY_ADDRESS="localhost:20000" \
 yarn start
 ```
 
-You can also let the the OS choose a free peer port by setting it to:
-`--peer-listen 0.0.0.0:0`. And if you don't need completely isolated state,
-then you can use `RAD_PROFILE` instead of `RAD_HOME`, but be aware that
-Electron state will be shared by all instances.
+You can also let the the OS choose a free peer port by using
+`RADICLE_PROXY_PEER_LISTEN="0.0.0.0:20000"`.
```
rudolfs, line 141:
```suggestion
`RADICLE_PROXY_PEER_LISTEN="0.0.0.0:0"`.
```

rudolfs (approved):
utACK

PR #2568, opened 2021-11-03
set default seed via env

geigerzaehler:
The default seed for the proxy can now be configured via an environment variable. We remove the default seed in the dev environment. It is not working and prevents us from using the environment variable.

### DIFF
--- a/native/index.ts
+++ b/native/index.ts
@@ -36,8 +36,6 @@ if (isDev) {
   proxyPath = path.join(__dirname, "../target/debug/radicle-proxy");
 
   proxyArgs.push(
-    "--default-seed",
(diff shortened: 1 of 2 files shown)


native/index.ts
```diff
@@ -36,8 +36,6 @@ if (isDev) {
   proxyPath = path.join(__dirname, "../target/debug/radicle-proxy");
 
   proxyArgs.push(
-    "--default-seed",
-    "hybz9gfgtd9d4pd14a6r66j5hz6f77fed4jdu7pana4fxaxbt369kg@setzling.radicle.xyz:12345",
```
rudolfs, line 40: Once we have a working seed again. Will we put it back here?

geigerzaehler: Yes. But then we would set it via the environment variable so that you can still overwrite it from the outside.

rudolfs: Gotcha.

rudolfs (approved):

PR #2574, opened 2021-11-04
update Yarn and NPM dependencies

geigerzaehler:
* Major upgrade to Sinon v12. Only breaking change is the update of `fake-timers` which doesn’t affect us.


rudolfs:
```
➤ YN0002: │ radicle-upstream@workspace:. doesn't provide graphql (p2338f), requested by @apollo/client
➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
```

rudolfs:
Do you remember why we're pinning `radicle-avatar`? I don't think there's a good reason. Maybe we can just use the latest master for that dep?

geigerzaehler:
> Do you remember why we're pinning `radicle-avatar`? I don't think there's a good reason. Maybe we can just use the latest master for that dep?

`radicle-avatar` is using the latest master. I don’t think it’s a good idea to use `master` in `package.json` because it will update to the lastest version without explicit confirmation when `yarn.lock` is removed. If a new version is available we’ll see it in the `yarn upgrade-interactive` dialog and can decide to upgrade.

rudolfs (approved):
utACK

geigerzaehler:
Merged in https://github.com/radicle-dev/radicle-upstream/commit/51a2471a2a0848a39ecccd7eec578f24d59786f1

PR #2575, opened 2021-11-04
shutdown upstream seed gracefully

### LINKED ISSUE #2571: Graceful shutdown for seed
We should implement graceful shutdown for the seed, just like we have for the proxy. This should prevent data corruption.

geigerzaehler:
Fixes #2571

### DIFF
--- a/upstream-seed/src/lib.rs
+++ b/upstream-seed/src/lib.rs
@@ -5,6 +5,7 @@
 // LICENSE file.
 
 use anyhow::Context;
+use futures::prelude::*;
 
 use librad::profile::Profile;
 use link_identities::git::Urn;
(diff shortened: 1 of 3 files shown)


upstream-seed/src/lib.rs
```diff
@@ -95,6 +94,36 @@ async fn track_projects(client: peer::Peer, projects: Vec<Urn>) {
...
+    tokio::spawn(async move {
+        use tokio::signal::unix::{signal, SignalKind};
+        let mut sig_int = signal(SignalKind::interrupt()).unwrap();
+        let mut sig_term = signal(SignalKind::terminate()).unwrap();
+
+        futures::future::select(sig_term.recv().boxed(), sig_int.recv().boxed()).await;
+        let _ = shutdown_tx.send(());
+
+        tracing::info!(
+            "shutting down. send SIGINT or SIGTERM again in 5 seconds to force shutdown"
```
rudolfs, line 113:
```suggestion
            "Shutting down. Send SIGINT or SIGTERM again within the next 10 seconds to force a shutdown."
```

upstream-seed/src/lib.rs
```diff
@@ -95,6 +94,36 @@ async fn track_projects(client: peer::Peer, projects: Vec<Urn>) {
...
+        );
+        let grace_period_end = std::time::Instant::now() + GRACE_PERIOD;
+        loop {
+            futures::future::select(sig_term.recv().boxed(), sig_int.recv().boxed()).await;
+            if std::time::Instant::now() > grace_period_end {
+                std::process::exit(5);
```
rudolfs, line 119: What does error code `5` mean?

geigerzaehler: Nothing in particular. I just chose an arbitrary exit code that isn’t 1 so that you could in theory distinguish the exit conditions via the code.

rudolfs (approved):

PR #2576, opened 2021-11-04
replace `RunError` with `anyhow::Error`

geigerzaehler:
Replace `RunError` with `anyhow::Error` to make the code simpler.


rudolfs (approved):
Nice!

PR #2578, opened 2021-11-04
consistent Rust lint settings

geigerzaehler:
* Align the lint settings between both crates.
* Remove clippy’s pedantic and nursery lints. They lead to a lot of false postivies that require manual tweaking.
* Don’t allow `clippy::or_fun_call` and fix all instances. This lint is sensible and we wan to limit the tweaking.


rudolfs (approved):
> Remove clippy’s pedantic and nursery lints.

🕊️ 🕊️ 🕊️ 
👌 😑 👌 
🕊️ 🕊️ 🕊️

Issue #2518, opened 2021-10-22
Make errors auto-dismiss again


brandonhaslegs:
Sorry, my bad, I know this was my idea to do this, but it's bad. Let's revert the PR that made errors never timeout. So many errors get fired that I often get into a really weird state. 🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨

- [ ] make errors automatically disappear
- [ ] don't show the same error twice

![image](https://user-images.githubusercontent.com/4406983/138454899-5019ffe0-a527-4ecb-8c6b-cefb7dc55d77.png)

PR #2579, opened 2021-11-04
feat: store dismissed notifications

rudolfs:
Dismiss error notifications automatically after 8 seconds and allow users to look them up later in the diagnostics screen.

Closes https://github.com/radicle-dev/radicle-upstream/issues/2518.

Screenshot 2021-11-04 at 20 54 46

Screenshot 2021-11-04 at 20 54 44


brandonhaslegs:
Notification history!?! Do we have a Notification Center now? Was it that easy? Should we build something pretty for it? Seems like it would be helpful for all users to see their notification history.

brandonhaslegs (approved):

ui/src/error.ts
```diff
@@ -5,12 +5,20 @@
...
-import * as notification from "./notification";
+import * as bacon from "./bacon";
 import * as ipc from "./ipc";
 
+// We use an event bus to break up the dependency cycle between the `error`,
+// `notification` and `browserStore` modules. This also makes the mocks in the
+// unit tests simpler.
+const errorBus = new bacon.Bus<Error>();
+export const errorBusOutput = errorBus.toEventStream();
+export function showErrorNotification(error: Error): void {
```
geigerzaehler, line 18: We can align the names a bit better. No need to mention errors explicitly since this is the `error` module.

```suggestion
const notificationsBus = new bacon.Bus<Error>();
export const notifications: bacon.EventStream<Error> = errorBus.toEventStream();
export function showNotification(error: Error): void {
```

ui/src/error.ts
```diff
@@ -5,12 +5,20 @@
...
+import lodash from "lodash";
 
-import * as config from "ui/src/config";
-import * as notification from "./notification";
+import * as bacon from "./bacon";
 import * as ipc from "./ipc";
 
+// We use an event bus to break up the dependency cycle between the `error`,
+// `notification` and `browserStore` modules. This also makes the mocks in the
+// unit tests simpler.
```
geigerzaehler, line 15: You can delete this comment. I don’t think we need to justify architecture and API design decisions. We should reserve these comments for code that does non-obvious and quirky things.

rudolfs: 3aeb9693966575f9184ff7e2ef291afd7edb4090

ui/src/error.ts
rudolfs: 3aeb9693966575f9184ff7e2ef291afd7edb4090

ui/src/notification.ts
```diff
@@ -4,143 +4,198 @@
...
+    .get(notificationStore)
+    .find(n => n.id === id);
+  if (notification) {
+    removeHideTimer(notification);
+  }
 
-export const primary = (params: NotificationParams): Handle =>
-  show(Variant.Primary, params);
+  notificationStore.update(notifications => {
+    return notifications.filter(n => n.id !== id);
```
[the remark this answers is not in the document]
rudolfs: 8a8cf7152a2c2a9f222fc0594ba5a8c70e439043

rudolfs:
> Notification history!?! Do we have a Notification Center now? Was it that easy?
Yes. Yup. Yeah.

> Should we build something pretty for it? Seems like it would be helpful for all users to see their notification history.

If you design something nice we can do it in a follow-up sometime later.

geigerzaehler (approved):

PR #2581, opened 2021-11-08
seed: more robust shutdown with task runner

geigerzaehler:
Add `TaskRunner` to provide an extensible platform for managing tasks and shutdown. The abstraction makes it easy to add additional long running tasks and ensure proper shutdown.

At the moment this improves the following

* If the tracking task panics `librad` is shut down properly
* The tracking project loop is not just cancelled when a shutdown is requested. Instead we finish one iteration.

The logic for `TaskRunner` is very similar to that of `ShutdownRunner` in the proxy. I opted to not consolidate this for now. We can’t share code between the two crates yet and there is enough of a difference between the two implementations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,6 +215,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-shutdown"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


rudolfs (approved):

geigerzaehler:
I had to make some changes because the implementation didn’t work well for how I wanted to extend it. Could you take another look?

rudolfs (approved):
LGTM

PR #2584, opened 2021-11-08
Better caching on CI

geigerzaehler:
Improvements and cleanup to CI regarding caching

50dd2a40 clean up CI build script

* Align the logs more closely with the command they run
* Remove `time` whenever the command prints the time itself.
* Add `cargo build` step. Before this was implicitly called by `yarn test:ingeration` and the output was hard to find.
* Group cargo steps together

https://github.com/radicle-dev/radicle-upstream/pull/2584/commits/533bd12bb58d7aa8d5c5c060f8cac471912303cb cache more efficiently on CI

* Be more selective about which directories to cache on CI to reduce the cache size from 1.5GB to 1.3GB.
* Avoid symlinks for target directory and use standard CARGO_HOME to for reduced complexity.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,9 +23,14 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
-            ~/cache/cargo
-            ~/cache/proxy-target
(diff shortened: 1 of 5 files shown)


rudolfs (approved):
🧹

PR #2597, opened 2021-11-15
Bug squashing

### LINKED ISSUE #2495: Update emoji and text when orgs can't be loaded
I've tried to exclude the use of faces in our empty states. For some reason I think an image of something is better than a face. Somehow it has more personality or humor when it's a dead flower or a horse, than when it's a sad face or something. And I think the message should be a bit more friendly (not including the word "failed" for example. Here's my suggestion:

- [ ] replace emoji with 🌴
- [ ] replace the text with "Orgs aren't loading at the moment. Take a break! [Try again]"

image

### LINKED ISSUE #2507: Add a backup placeholder for when the image is broken
Add a blank placeholder when an org's avatar can't be loaded.

image

### LINKED ISSUE #2509: Add max width for error field on crash page
image

### LINKED ISSUE #2567: Collapsed header should stretch from left to right side of viewport
image

Currently it's just centered with shadows appearing on the left and right. It should be a full-width bar.

juliendonck:
- [x] closes #2495 
- [x] closes #2509 
- [x] closes #2507 
- [x] closes #2567


design-system/Avatar.svelte
```diff
@@ -56,6 +56,15 @@
...
   }
+
+  function validImgUrl(image_url: string): boolean {
+    const http = new XMLHttpRequest();
+
+    http.open("HEAD", image_url, false);
+    http.send();
+
+    return http.status !== 404;
+  }
```
geigerzaehler, line 67: This approach to detecting a valid image has two issues. First, it only checks for 404 status codes and not for other status codes or invalid content types like the server sending JSON. Second, this is not async and blocks the render thread. This means that when it takes longer to get a response the app becomes unresponsive.

I have a PR (https://github.com/radicle-dev/radicle-upstream/pull/2602) that chooses a different approach.

geigerzaehler (commented):
Looks good except for the image fallback.

design-system/Avatar.svelte
juliendonck: Makes sense. Merged ✔

rudolfs (approved):
👍

PR #2608, opened 2021-11-17
seed: add retry logic for fetcher

### LINKED ISSUE #2607: seed: Retry when fetch fails due to concurrent fetch
If the seed wants to fetch an identity from a peer it calls `fetcher::PeerToPeer::build()`. This call may return an error if a concurrent fetch for the same identity is in progress. Instead of simply erroring we want to retry the fetch.

geigerzaehler:
Fixes #2607


rudolfs (approved):

geigerzaehler:
Merged in https://github.com/radicle-dev/radicle-upstream/commit/fd86527dbb179487577e7e3e3cd3ece46e0f52f7

geigerzaehler:
This somehow never got merged. I pointed to the wrong commit.

PR #2611, opened 2021-11-17
eliminate `Commits` component

geigerzaehler:
We eliminate the `Commits` component and include `History` in `Source` directly. This reduces indirection.

We also move the code that navigates to a commit into `History`. This eliminates the state dependency of the navigation code and reduces the size and complexit of `ui/src/screen/project/source.ts`.


rudolfs (approved):
utACK

PR #2615, opened 2021-11-18
commit component owns its state

geigerzaehler:
The `Source/Commit` now owns its state regarding loaded commits. We also add a loading screen and make prevent race conditions when loading data.


rudolfs (approved):
✨

PR #2616, opened 2021-11-18
move pending peers storage into `ManagePeersModal`

geigerzaehler:
This change has the following advantages
* We don’t use the deprecated `Remote` component
* The derived store is much simpler
* We keep data close to the component where it is used


rudolfs (approved):
🧹

PR #2619, opened 2021-11-22
proxy: add option to override seeds

### LINKED ISSUE #2562: Force seed node addresses for development and testing
When testing (manually or in the future automatically) upstreams interaction with seed nodes I want to set the seed node addresses on the command line. These addresses should take precedence over the previously stored addresses. (This means they behave differently from the `--default-seed` option.)

* I can run upstream as `RADICLE_PROXY_SEED=foo@localhost:123 yarn run start`
* Upstream proxy will only connect to the given seed and ignore previously set seeds
* The given seed is returned by the seed endpoint
* Changing the seed through the API fails

geigerzaehler:
Consists of two commits, the first one is preparing the change.

proxy: improve seed handling for discovery

* Don’t pass the default seeds down to `session::seeds()`. Instead, return an `Option` that can later be combined with `unwrap_or_default()`.
* Use an empty list of seeds as the default if nothing is set. This alleviates the need to pass the default seeds to the discovery task. The configured default seeds will still be used when the session is initialized and the seed value set to the default seeds.
* Query the configured seeds more frequently (every 0.4s instead of every 5s). Querying is cheap and this results and updates being propagated more quickly.
* Update the discovery list even if the peer status is offline. Not sure why this was there but I seems wrong.

proxy: add option to override seeds

Add a `--seed` option and environment variable that forces the proxy to use the provided seeds instead of those stored in the configuration. Using this option will make the seed API read-only and return the seeds provided on the command line.

We also adjust the environment variable for specifying the default seed. It now uses the plural `RADICLE_PROXY_DEFAULT_SEEDS` and multiple seeds can be given by separating them with a comma.

Fixes #2562

### DIFF
--- a/proxy/api/src/cli.rs
+++ b/proxy/api/src/cli.rs
@@ -27,9 +27,20 @@ pub struct Args {
     #[structopt(long, env = "RADICLE_PROXY_PEER_LISTEN", default_value = "0.0.0.0:0")]
     pub peer_listen: std::net::SocketAddr,
 
-    /// Add one or more default seed addresses to initialise the settings store
-    #[structopt(long, env = "RADICLE_PROXY_DEFAULT_SEED", long = "default-seed")]
+    /// Populate the seed configuration with these seeds when the app is started for the first
+    /// time.
+    #[structopt(
+        long,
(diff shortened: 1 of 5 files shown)


rudolfs (approved):
👏

PR #2621, opened 2021-11-22
don’t use docker container for linux builds

geigerzaehler:
Instead of using docker containers for the Linux builds we build them directly in the Github Actions VM. This has a couple of advantages:

* The setup between the macOS builds and the Linux build is the same now. When new build dependencies are required or need to be updated there is one place to do this.
* We don’t pay for storing and downloading the build image. The cost for the latter is significant.
* Linux builds are slightly faster (~15secs) because the image does not need to be pulled.

We can’t use matrix builds for the `build` workflow because we only want the macOS builds for PRs and not for every push.


rudolfs (commented):
Interesting!

I just got the container work for my p2p net tests, it's probably gonna simplify that setup as well.

Where do we specify the dependencies now? I mean these:
https://github.com/radicle-dev/radicle-upstream/blob/main/ci/Dockerfile#L10

geigerzaehler:
> Where do we specify the dependencies now? I mean these:

Almost all of them are installed on the runner and don’t need to be specified explicitly. `cargo-deny` is the exception and this is now installed in the build script. Keep in mind that the dependencies only applied to the linux build so being able to specify them didn’t offer a real advantage.

rudolfs (approved):

PR #2624, opened 2021-11-22
feat: allow to specify the librad key file path via CLI option

rudolfs:
To make manual testing and debugging easier, we add a couple of CLI options and environment variables.

Add `--identity-key <identity-key>` option to `radicle-proxy`, which allows overriding the key file path that normally gets set by `RAD_HOME`.

Also add 2 new environment variables for the `git-remote-rad` helper:
  - `IDENTITY_KEY` allows overriding the key file path (analogous to `--identity-key` option in proxy)
  - `KEY_PASSPHRASE` allows passing in the passphrase via an env variable (analogous to the `--key-passphrase` option in proxy).

### DIFF
--- a/proxy/api/src/keystore.rs
+++ b/proxy/api/src/keystore.rs
@@ -14,9 +14,8 @@
 //! This module provides the [`Keystore`] trait and the [`file()`] function to construct specific
 //! [`Keystore`] implementations.
 
-use std::convert::Infallible;
+use std::{convert::Infallible, path::PathBuf};
 
-use librad::paths::Paths;
 pub use radicle_keystore::pinentry::SecUtf8;
 use radicle_keystore::{
     crypto::{self, Pwhash, SecretBoxError},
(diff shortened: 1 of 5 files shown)


proxy/api/src/keystore.rs
```diff
@@ -16,12 +16,12 @@
 
 use std::convert::Infallible;
 
-use librad::paths::Paths;
 pub use radicle_keystore::pinentry::SecUtf8;
 use radicle_keystore::{
     crypto::{self, Pwhash, SecretBoxError},
     file, Keystore as _, SecretKeyExt,
 };
+use std::path::PathBuf;
```
geigerzaehler, line 24: Can you merge this import with the other `std` import above?

rudolfs: 81781f67277fbe4630ef62761e82ac34e5705523

geigerzaehler (approved):

PR #2632, opened 2021-11-24
provision stable upstream seed

### LINKED ISSUE #2620: Stable, maintainable deployment of upstream seed
We want to provide a stable and maintainable deployment of the upstream seed peer so that we can use for ourselves.

* Upstream seed runs on a dedicated GCP instance and is controlled by systemd.
* The provisioning of the instance is documented to make it repeatable. (Includes firewall settings)
* Systemd service file is included in the repo
* Add a script to the repo and the instance the automatically updates the seed binary.
* Setup logging with GCP through journald and add instructions how to find and read the logs.

geigerzaehler:
Add documentation and scripts to explain the provisioning of a stable hosted `upstream-seed`.

Fixes #2620.

### DIFF
--- a/docs/infrastructure.md
+++ b/docs/infrastructure.md
@@ -165,6 +165,24 @@ END
 docker-compose --env-file upstream-production-config.env -f docker-compose.yml up --detach
 ```
 
+## seed-node instance
+
+The seed node uses the following resources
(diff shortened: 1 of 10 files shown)


rudolfs (approved):
👏

PR #2636, opened 2021-11-25
wait until session is unsealed when loading user profile route

geigerzaehler:
We wait until the session is unsealed on app startup when loading the user profile route. Before, the code may have thrown an error if the route was loaded before the session was loaded.

To enable this we make `svelteStore.waitUntil()` more general an useful.

### DIFF
--- a/ui/src/svelteStore.test.ts
+++ b/ui/src/svelteStore.test.ts
@@ -12,13 +12,15 @@ import * as svelteStore from "./svelteStore";
 describe("waitUntil", () => {
   test("current value matches", async () => {
     const store = svelteStore.writable(true);
-    const value = await svelteStore.waitUntil(store, x => x);
-    expect(value).toBe(true);
+    const value = await svelteStore.waitUntil(store, x => x.toString());
+    expect(value).toBe("true");
   });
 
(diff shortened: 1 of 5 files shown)


ui/src/svelteStore.test.ts
```diff
@@ -12,13 +12,15 @@ import * as svelteStore from "./svelteStore";
 describe("waitUntil", () => {
   test("current value matches", async () => {
     const store = svelteStore.writable(true);
-    const value = await svelteStore.waitUntil(store, x => x);
-    expect(value).toBe(true);
+    const value = await svelteStore.waitUntil(store, x => x.toString());
```
rudolfs, line 15: Just for my own curiosity, why is this needed now?

rudolfs (approved):

ui/src/svelteStore.test.ts
geigerzaehler: The change is not strictly necessary. The tests would have passed without it. That’s the reason why I changed it: to have a test that only passes with the new implementation. Calling `.toString()` I just a quick and arbitrary way of having some function that transforms the input.

Issue #1891, opened 2021-05-17
Add back button on project page to go back to wherever you came from


brandonhaslegs:
Add back button on project page to go back to wherever you came from.

Things to think about:
- The back button doesn't always go back to the project list on the profile
- What happens when you get to a project from the search bar? (probably just do not show this back button)
- Should use the same styling as the back button when you click into see single commits
- Consider redesigning this back button it's kinda just an icon… maybe could be more obvious? Maybe not…

![image](https://user-images.githubusercontent.com/4406983/118471178-c1434f00-b707-11eb-92d2-92bbda9f189f.png)

juliendonck:
@rudolfs this is possible now with the new router, right?

rudolfs:
Yes.

rudolfs:
How about this?

<img width="1552" alt="Screenshot 2021-10-15 at 15 47 23" src="https://user-images.githubusercontent.com/158411/137497753-9b21ad91-b854-4834-b0fd-732592314097.png">

rudolfs:
In the current layout there's a bit less space:
<img width="1552" alt="Screenshot 2021-12-01 at 11 47 07" src="https://user-images.githubusercontent.com/158411/144220880-222718db-44df-431b-b430-32e0cd343424.png">

Should the back button still go on the left side of the project title?

brandonhaslegs:
Yes. It's not to different from the commits back button. Bit tighter but i think it's fine.

<img width="306" alt="image" src="https://user-images.githubusercontent.com/4406983/144318935-5e2feae2-d78d-492c-b6c9-22c33ad18208.png">

rudolfs:
We tried: https://github.com/radicle-dev/radicle-upstream/pull/2653#issuecomment-984950167.

PR #2653, opened 2021-12-02
feat: add back button to project header

rudolfs:
Closes https://github.com/radicle-dev/radicle-upstream/issues/1891.

@brandonhaslegs is this what you imagined?

https://user-images.githubusercontent.com/158411/144424410-73b1ecee-3bcd-44ec-bc4e-97368a74c0c4.mov

### DIFF
--- a/ui/App/ProjectScreen/BackButton.svelte
+++ b/ui/App/ProjectScreen/BackButton.svelte
@@ -16,10 +16,15 @@
   };
 
   export let style: string | undefined = undefined;
+  export let buttonWrapperStyle: string | undefined = undefined;
 </script>
 
 <style>
   header {
+    position: relative;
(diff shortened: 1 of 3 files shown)


brandonhaslegs:
Yes, but the project title shouldn't be shifted over.

rudolfs:
Like this?

https://user-images.githubusercontent.com/158411/144458762-f077d2d6-739f-423a-825c-5d97e423cb66.mov

rudolfs:
We tried making the margins bigger to fit the arrow, but it felt like it wasn't worth it to apply bigger margins to the whole app just because of the back button on one page. We also tried putting it above the title, but that doesn't work if there's a description.

Closing this for now.

PR #2679, opened 2021-12-15
test: make p2p network tests work on Intel macOS hosts

rudolfs:
So we can get rid of the dev vm on gcp.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,48 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# Use this Vagrantfile to run p2p-tests on an Intel macOS host.
+# This requires VirtualBox to be installed.
+#
+#   brew install vagrant
+#   vagrant up
+#   vagrant ssh
+#     cd /vagrant
+#     cargo build
+#     cd p2p-tests
+#     sudo -E ./contributor-fork-replication-1-test.ts
+
+Vagrant.configure("2") do |config|
+  config.vm.provider "virtualbox" do |v|
+    v.memory = 8192
+    v.cpus = 4
+  end
+
+  config.vm.box = "ubuntu/focal64"
+  config.vm.provision "shell", inline: <<~SHELL
+    export DEBIAN_FRONTEND=noninteractive
+    apt-get -y update
+    apt-get -y upgrade
+    apt-get -y --no-install-recommends install \
+      build-essential \
+      ca-certificates \
+      cmake \
+      curl \
+      git \
+      jq \
+      procps; \
+
+    curl -sfLS https://deb.nodesource.com/setup_14.x | bash -; \
+    apt-get -y update; \
+    apt-get -y --no-install-recommends install nodejs; \
+    apt-get autoremove; \
+    npm install --global yarn;
+
+    curl https://sh.rustup.rs -sSfL | sudo -u vagrant sh -s -- -y --profile minimal
+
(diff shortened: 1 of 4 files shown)


Vagrantfile
```diff
@@ -0,0 +1,66 @@
...
+    apt-get -y upgrade
+    apt-get -y --no-install-recommends install \
+      build-essential \
+      ca-certificates \
+      cmake \
+      curl \
+      git \
+      jq \
+      procps; \
+    rm -rf /var/lib/apt/lists/*;
```
geigerzaehler, line 34: I don’t think this is necessary. It saves a bit of space but makes subsequent apt commands slower.

Vagrantfile
```diff
@@ -0,0 +1,66 @@
...
+
+    curl -sfLS https://deb.nodesource.com/setup_14.x | bash -; \
+    apt-get -y update; \
+    apt-get -y --no-install-recommends install nodejs; \
+    apt-get autoremove; \
+    rm -rf /var/lib/apt/lists/*; \
```
geigerzaehler, line 40: See comment above

Vagrantfile
```diff
@@ -0,0 +1,66 @@
...
+    echo "${RUSTUP_SHA256} *rustup-init" | sha256sum -c -; \
+    chmod +x rustup-init; \
+    ./rustup-init -y --no-modify-path --profile minimal --default-toolchain $RUST_VERSION; \
+    rm rustup-init; \
+    chmod -R a+w $RUSTUP_HOME $CARGO_HOME; \
+    rustup --version; \
+    cargo --version; \
+    rustc --version; \
+    rm -rf /usr/local/cargo/registry; \
+    rm -rf /usr/local/cargo/.package-cache;
```
geigerzaehler, line 60: I think we can simplify this a bit
```bash
curl https://sh.rustup.rs -sSfL \
 | sudo -u vagrant sh -s -- -y --profile minimal
```

No need to do all the docker image specific stuff. We can also remove specifing the rust version explicitly. `rustup` will take it from the file anyways and this way we don’t need to remember to update the version here if we bump the toolchain.

Vagrantfile
```diff
@@ -0,0 +1,66 @@
...
+
+    echo "PATH=/usr/local/cargo/bin:$PATH" >> /home/vagrant/.bashrc
+    echo "export CARGO_HOME=/home/vagrant/cargo" >> /home/vagrant/.bashrc
+    echo "export CARGO_TARGET_DIR=/home/vagrant/cargo" >> /home/vagrant/.bashrc
```
geigerzaehler, line 64: I don’t know if its a good idea to have them point to the same directory. Also not sure we need to set `CARGO_HOME`. The default value should be sufficient.

Vagrantfile
rudolfs: 65e982ac431e796be804a71bc79cd2b51fe5e6fc

Vagrantfile
rudolfs: 65e982ac431e796be804a71bc79cd2b51fe5e6fc

Vagrantfile
rudolfs: 351b3371f953c28460d53474b152be63bce9435f

Vagrantfile
rudolfs: 351b3371f953c28460d53474b152be63bce9435f

geigerzaehler (approved):

PR #2684, opened 2021-12-16
feat: improve diagnostics screen

rudolfs:
Adding the proxy logs to the diagnostics info screen should help with debugging in case the app was not started from the terminal and we still want to get to the logs.

- show Upstream version in header and add it to copy-to-clipboard button
- don't show redundant peer information in storage tab
- align copy-to-clipboard data shape to what we see on the screen
- use a different icon for notification history tab
- add proxy logs tab and add it to copy-to-clipboard button

Screenshot 2021-12-16 at 11 01 39

### DIFF
--- a/cypress/support/ipc-stub.ts
+++ b/cypress/support/ipc-stub.ts
@@ -38,6 +38,7 @@ export function setup(window: Window): void {
   let clipboard = "";
 
   const electronStubs: ElectronStubs = {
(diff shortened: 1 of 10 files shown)


geigerzaehler (commented):
:clap:

geigerzaehler (approved):

PR #2685, opened 2021-12-16
proxy: encapsulate peer

geigerzaehler:
We encapsulate peer related logic (from `librad` and `radicle_daemon`) into a `peer` module to enforce simpler boundaries.

This change tackles the following code issues
* `context::Unsealed` held three fields related to peers which didn’t provide proper encapsulation.
* Configuring a peer was done as part of the session setup in `process`.
* Callers of peer related functionality had to decide whether to access the `librad` peer or the `daemon` peer.

### DIFF
--- a/proxy/api/src/context.rs
+++ b/proxy/api/src/context.rs
@@ -13,8 +13,6 @@ use futures::prelude::*;
 use rand::Rng as _;
 use tokio::sync::RwLock;
 
-use link_crypto::BoxedSigner;
-
 use crate::{keystore, service};
 
 /// Container to pass down dependencies into HTTP filter chains.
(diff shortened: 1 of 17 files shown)


proxy/api/src/context.rs
```diff
@@ -212,39 +205,31 @@ impl Unsealed {
...
-            };
-            (peer_control, peer, run_handle)
+        let seeds_watch = tokio::sync::watch::channel(vec![]).1;
+
+        let (peer, peer_runner) = crate::peer::create(crate::peer::Config {
+            paths: paths.clone(),
+            key,
+            store: store.clone(),
+            discovery: radicle_daemon::config::StreamDiscovery::new(seeds_watch),
+            listen: "127.0.0.1:0".parse().expect("invalid IP address"),
```
rudolfs, line 217: Why is this the loopback address?

rudolfs (approved):
🧹

proxy/api/src/context.rs
geigerzaehler: The test peer should not be exposed to the outside world. This is the same as with the test daemon configuration: https://github.com/radicle-dev/radicle-link/blob/258c893bf833b7ce37569c7fed4d3f6746ffe0a7/daemon/src/config.rs#L47

rudolfs: Ah, missed the `#[cfg(test)]`.

PR #2686, opened 2021-12-16
fix: clear seed input after adding seed

rudolfs:
Fixes a regression introduced in https://github.com/radicle-dev/radicle-upstream/pull/2658.


geigerzaehler (approved):
:heart:

Issue #2651, opened 2021-12-01
Multi-sig Gnosis Safe backed org creation is slow


rudolfs:
When creating a multi-sig org it takes a while for the Gnosis API to update. We've observed a delay of 20-30 min between creation and the org appearing in the sidebar.

The culprit seems to be this API which we are querying to get an up-to-date list of orgs belonging to a wallet:
https://safe-transaction.rinkeby.gnosis.io/api/v1/owners/<OWNER_WALLET_ADDRESS>/safes/

It gets called here: https://github.com/radicle-dev/radicle-upstream/blob/main/ui/src/org.ts#L357.

The alt-clients team already switched to using The Graph for mainnet Gnosis queries, we should use that too.
See here for more info: https://github.com/radicle-dev/radicle-subgraph/commit/ddfd473ee62ba4d19f052ef2eb45cd252b5b1c5c.

~~One caveat with that transition is that it's not yet deployed on rinkeby.~~
It is now: https://thegraph.com/hosted-service/subgraph/radicle-dev/radicle-orgs-rinkeby

PR #2689, opened 2021-12-16
fix: use the graph for Gnosis Safe queries

rudolfs:
This makes multi-sig org creation much faster than using the Gnosis Safe
transaction API.

Fixes https://github.com/radicle-dev/radicle-upstream/issues/2651.

### DIFF
--- a/ui/src/org/theGraphApi.ts
+++ b/ui/src/org/theGraphApi.ts
@@ -14,6 +14,8 @@ import * as apolloCore from "@apollo/client/core";
 import * as svelteStore from "svelte/store";
 import * as ethers from "ethers";
 
+import { memoizeLru } from "ui/src/memoizeLru";
+
 import * as error from "ui/src/error";
 import * as ethereum from "ui/src/ethereum";
 import * as wallet from "ui/src/wallet";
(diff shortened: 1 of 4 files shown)


ui/src/org/theGraphApi.ts
```diff
@@ -61,6 +63,74 @@ export interface Org {
...
+      variables: { address: address.toLowerCase() },
+    });
+
+    return {
+      ...safesResponse.data.safe,
+      threshold: Number(safesResponse.data.safe.threshold),
+    };
+  },
+  safeAddress => safeAddress,
+  { max: 1000, maxAge: 15 * 60 * 1000 } // TTL 15 minutes
```
geigerzaehler, line 131: Is it reasonable to memoize this? What are the downsides of having a much shorter ttl, say 20 seconds?

ui/src/org/theGraphApi.ts
```diff
@@ -61,6 +63,74 @@ export interface Org {
   projectCount?: number;
 }
 
+export interface Safe {
+  id: string;
+  owners: string[];
+  threshold: number;
+}
+
+export async function getSafesByOwner(owners: string[]): Promise<Safe[]> {
```
geigerzaehler, line 72: This function is only ever called with one owner as an argument. Maybe we should convert it to
```suggestion
export async function getSafesByOwner(owner: string): Promise<Safe[]> {
```

ui/src/org/theGraphApi.ts
rudolfs: I think it still makes sense to memoize it because we're getting a list of all the orgs and the associated metadata on the org explore screen. Maybe we could lower it to something like 5min, so that the org explore page doesn't take forever to load?

ui/src/org/theGraphApi.ts
rudolfs: 8221747c53fdcbbeb5f3c0743bc27a8f8442b02d

ui/src/org/theGraphApi.ts
rudolfs: Also the Graph queries cost us in production.

geigerzaehler: Ok. Let’s keep it like this then.

geigerzaehler (approved):

PR #2694, opened 2021-12-17
fix: fix multisig org anchor buttons

rudolfs:
Show anchor buttons in multi-sig orgs if we're the owner of the org.

Fixes a regression introduced in #2450.


rudolfs:
@geigerzaehler please review this when you get back.

geigerzaehler (commented):
lgtm

Issue #2691, opened 2021-12-17
Branches for implicilty tracked peers are missing


rudolfs:
When a peer A clones a project for the first time it will add remote refs for all peers (e.g. peer B) that the maintainer of the project is replicating. This results in the UI displaying these peers in the peer selector of A. When the user clicks on, say, peer B the app can not load any of the head refs for that remote (they have not been replicated) and it errors.

~~We have a test case for this in #2715.~~

*edited by @geigerzaehler*

geigerzaehler:
It looks like this happens when we initially clone a project and this also fetches information for all peers the maintainer tracks. We need to find a concrete scenario that reproduces this (ideally as an automated test).

rudolfs:
I believe this only happened with the gossip transport, which we removed.

PR #2715, opened 2022-01-11
test: add test for implicitly tracked peer branches

rudolfs:
Refs: https://github.com/radicle-dev/radicle-upstream/issues/2691.

This test fails with:
```
  variant: 'INTERNAL_SERVER_ERROR',
  body: {
    message: "we could not find the 'refs/namespaces/hnrkrhhs1goaawo7db1gpyct8hd7mif5q8c3o/refs/remotes/hydkjndj7zhq98wg15akseowbtykjpym15oc613jq8frbpoisgi4pn/heads/main'",
    variant: 'INTERNAL_SERVER_ERROR'
  },
```

And we see this in the logs:
```
contributor2 |   2022-01-11T14:54:47.161534Z ERROR api::http::error: internal server error, we could not find the 'refs/namespaces/hnrkrhhs1goaawo7db1gpyct8hd7mif5q8c3o/refs/remotes/hydkjndj7zhq98wg15akseowbtykjpym15oc613jq8frbpoisgi4pn/heads/main', variant: INTERNAL_SERVER_ERROR, err: Rejection(State(MissingRef { reference: Reference { remote: Some(hydkjndj7zhq98wg15akseowbtykjpym15oc613jq8frbpoisgi4pn), category: Heads, name: RefLike("main"), namespace: Some(Namespace(Urn { id: Oid(4e72d234318a43a30c8cd03227e0faba976e3b33), path: None })) } }))
```

### DIFF
--- a/p2p-tests/implicitly-tracked-peer-branches-test.ts
+++ b/p2p-tests/implicitly-tracked-peer-branches-test.ts
@@ -171,6 +171,21 @@ async function testcase(dataPath: string) {
     strictAssert.deepStrictEqual(result.urn, project.urn);
   });
 
+  // Assert that contributor is in contributor2's remote list.
+  await withRetry(async () => {
+    const peers = await contributor2.proxyClient.project.listPeers(project.urn);
+
+    console.log("maint: ", maintainer.peerId);
+    console.log("cont2: ", contributor2.peerId);
+    console.log("cont: ", contributor.peerId);
+    console.log(peers);
+    console.log(peers.find(x => x.peerId === contributor.peerId)?.status.type);
+    strictAssert.deepStrictEqual(
+      peers.find(x => x.peerId === contributor.peerId)?.status.type,
+      "replicated"
+    );
+  });
+
   // Assert that contributor2 can view contributor's branch.
   await withRetry(async () => {
     const branches = await contributor2.proxyClient.source.branchesGet({
--- a/p2p-tests/implicitly-tracked-peer-branches-test.ts
+++ b/p2p-tests/implicitly-tracked-peer-branches-test.ts
@@ -9,23 +9,17 @@
 import * as path from "path";
 import { strict as strictAssert } from "assert";
 
-import * as source from "proxy-client/source";
 import { sleep } from "ui/src/sleep";
 import {
   RadicleProxy,
   UpstreamSeed,
-  commit,
-  getLatestCommitSha,
-  pushRad,
   radCli,
   runTestcase,
   withRetry,
 } from "./lib/p2p";
 
-// Test that updates to a project from the contributor are replicated back to
-// the maintainer via a seed.
-//
-// The contributor and maintainer are never connected to the seed at the same time.
+// Test that that the branches of contributors tracked by the maintainer are
+// replicated by any other peer.
 async function testcase(dataPath: string) {
   const project = {
     name: "my-fancy-project",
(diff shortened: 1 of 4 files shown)


p2p-tests/implicitly-tracked-peer-branches-test.ts
```diff
@@ -0,0 +1,189 @@
...
+  });
+
+  // Assert that contributor2 received the project.
+  await withRetry(async () => {
+    const result = await contributor2.proxyClient.project.get(project.urn);
+
+    strictAssert.deepStrictEqual(result.urn, project.urn);
+  });
+
+  // Assert that contributor2 can view contributor's branch.
```
geigerzaehler, line 174: Can we add an assertion before this: That “contributor” is in contributor2’s peer list? This is the confusing part, right?

p2p-tests/implicitly-tracked-peer-branches-test.ts
```diff
@@ -0,0 +1,189 @@
...
+import {
+  RadicleProxy,
+  UpstreamSeed,
+  radCli,
+  runTestcase,
+  withRetry,
+} from "./lib/p2p";
+
+// Test that that the branches of contributors tracked by the maintainer are
+// replicated by any other peer.
```
geigerzaehler, line 22: I think we can clarify this a bit and also mention that this currently fails.

```suggestion
// Test that when a peer clones a project it automatically adds all peers as remotes that the maintainer of the project has as remotes.
//
// This test case currently fails because the peer does not replicate branches for peers that the maintainer already tracks.
```

rudolfs: My thinking was that we keep this PR open until we fix the issue and then commit the fix together with this test. I had forgotten to add it to `ci/p2p-network-tests.sh`, which I now did so it fails on CI. Does that make sense?

geigerzaehler: Sounds good.

p2p-tests/implicitly-tracked-peer-branches-test.ts
rudolfs: Good point. Now that I added this check, I can't get this test to do the implicit tracking part. Do you have any idea what we might be missing here?

I tried adding a commit from `contributor`, but that didn't help either. `contributor2` only sees the `maintainer` and itself in the peer list.

Here's the check I added: 5796a02f7561973a57a6b057e89820221e1628d3.

```
maint:  hybhe78oy41yzux5d6rk8fdtek881de76wuwy7mga1imzoj7p17pnw
cont2:  hynggyf6eftw63a5emj77afqgji69e1i3gicgot61omx6tmbekgkxy
cont:  hydkjndj7zhq98wg15akseowbtykjpym15oc613jq8frbpoisgi4pn
[
  {
    type: 'local',
    peerId: 'hynggyf6eftw63a5emj77afqgji69e1i3gicgot61omx6tmbekgkxy',
    status: { type: 'replicated', role: 'tracker', user: [Object] }
  },
  {
    type: 'remote',
    peerId: 'hybhe78oy41yzux5d6rk8fdtek881de76wuwy7mga1imzoj7p17pnw',
    status: { type: 'replicated', role: 'maintainer', user: [Object] }
  }
]
```

rudolfs:
Can't replicate this, closing.

PR #2726, opened 2022-01-13
add CLI tool for running and connecting multiple upstream instances

geigerzaehler:
Add CLI tool for running and connecting multiple upstream instances. For details see the added developer documentation.


scripts/devnet.ts
```diff
@@ -0,0 +1,182 @@
...
+  peerId: string;
+  httpPort: number;
+  p2pPort: number;
+  radHome: string;
+}
+
+// Get a peer ID from a private key seed.
+//
+// Uses the same algorithm as `radicle-proxy-init`.
+function peerIdFromKeySeed(seed: string): string {
```
rudolfs, line 137: This is pretty cool! Could be also useful for the p2p tests.

docs/development.md
```diff
@@ -34,6 +34,34 @@ source ./scripts/env
...
+We provide the `scripts/devnet.ts` tool for running, connecting and
+managing multiple Upstream instances on a local machine.
+
+The devnet tool uses numbers from 1 to 100 to identify and reference instances.
+
+```bash
+scripts/devnet.ts upstream 1
+```
+
+This will an Upstream instance. The instance will be fully initialized with user
```
rudolfs, line 48:
```suggestion
This will start an Upstream instance. The instance will be fully initialized with user
```

rudolfs:
Nice work!

The tests seem to break due to the CLI changes:

```
  maintainer   | error: The argument '--insecure-http-api <insecure-http-api>' requires a value but none was supplied
  
  maintainer   | USAGE:
  maintainer   |     radicle-proxy --dev-log --http-listen <http-listen> --insecure-http-api <insecure-http-api> --key-passphrase <key-passphrase> --peer-listen <peer-listen> --seed <seeds>... --skip-remote-helper-install --unsafe-fast-keystore
```

scripts/devnet.ts
geigerzaehler: Yes! Do you have an idea how we could leverage it?

rudolfs: Hm, I looked at the p2p tests more closely, we're already using the output from `radicle-proxy-init` which outputs the peerId, so this wouldn't simplify anything.

docs/development.md
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-upstream/pull/2726/commits/69c62f827121316755fbe8e5046899ac93498f94

rudolfs (approved):
🎉

PR #2738, opened 2022-02-01
devnet script waits for the child process before exiting

geigerzaehler:
When the devnet script receives Ctrl+C or SIGTERM it does not exit immediately anymore instead. Instead it forwards the signal to the child process (radicle-proxy, upstream-seed, upstream) and waits for the process to exit.


geigerzaehler:
duplicate that I created inadvertantly

PR #2762, opened 2022-02-17
Enable replication-v3

alexjg:
This required a few small changes to calls to librad APIs. More significantly we add a dependency on git-tempfile in order to turn off it's questionable signal handling logic. This will probably be handled upstream in the near future.

I've tried to be as conservative as possible in updates to `Cargo.lock` here.

With these changes all the p2p networking tests and the integration tests pass on my machine.


rudolfs:
Nice work!

We should do a networking party amongst us and try this out.

Subjectively it feels like it's harder to get CI to pass. The test suite was flaky before, but usually re-triggering a build once would fix it. Now, I tried 5x, so it might be that there's a race condition somewhere or the v3 code just takes longer to replicate. Might be worth trying to bump the global cypress timeout, to see if that helps:

```diff
diff --git a/cypress.json b/cypress.json
index d48c1180..1ca8a2c6 100644
--- a/cypress.json
+++ b/cypress.json
@@ -2,6 +2,6 @@
   "video": false,
   "viewportWidth": 1200,
   "viewportHeight": 600,
-  "defaultCommandTimeout": 8000,
+  "defaultCommandTimeout": 16000,
   "retries": { "runMode": 1, "openMode": 0 }
 }
```

rudolfs:
V3 fixes the multiple project replication issue: https://github.com/radicle-dev/radicle-upstream/pull/2707.

p2p-tests/contributor-fork-replication-1-test.ts
```diff
@@ -133,6 +133,14 @@ async function testcase(dataPath: string) {
     keyPassphrase: contributor.passphrase,
   });
 
+  // ALEX: We're not entirely sure when the announcement occurs. The daemon has
+  // a loop which announces every second, but we also need to allow some time
+  // for the announcement to be received by the seed and the replication to
+  // occur. In practice 5 seconds seems to work for me.
+  //
+  // I'm quite unhappy about this
+  await sleep(5000);
```
geigerzaehler, line 142: The `withRetry` call below should be able to wait more efficiently. We continuously check until we’ve successfully replicated. If it times out we should just bump the time out in `withRetry`.

alexjg: Interestingly bumping the timeout in `withRetry` to 5000 doesn't work. This is strange as it seems like it should be semantically identical to this `sleep`.

rudolfs:
I'm going to close this because we just migrated to the push/pull transport and are removing most of the old gossip transport from Upstream. We'll re-integrate with the new radicle-link application architecture and p2p transport once it is in a stable state.

PR #2820, opened 2022-04-24
Expand design system

juliendonck:
- [ ] new primary outline button
- [ ] thumb up and down icon


design-system/Button.svelte
```diff
@@ -245,7 +264,7 @@
   class="button typo-text-bold {buttonClass}"
   {title}
   {disabled}
-  on:click
+  on:click|stopPropagation
```
rudolfs, line 267: I wonder if we shouldn't let the callsite decide on what modifiers to use. 🤔 
This should work `<Button on:click|stopPropagation={} />`, no?

juliendonck: Nope, you can only do modifiers on dom elements.

rudolfs: Right. I guess this should be fine then. We probably don't want button presses to propagate anyway.

rudolfs (approved):

PR #2840, opened 2022-05-17
Add base `/` to url

juliendonck:
Fixes an issue that workstreams can't find emoji's because it url based.

### DIFF
--- a/design-system/Emoji.svelte
+++ b/design-system/Emoji.svelte
@@ -47,7 +47,7 @@
 <div {style} class="container">
   {@html twemoji.parse(emoji, {
     className: `emoji ${size}`,
-    base: "",
+    base: "/",
     folder: "twemoji",
     ext: ".svg",
   })}


rudolfs:
Unfortunately this breaks emojis in Upstream. Will think about how to fix this properly.

Screenshot 2022-05-17 at 15 07 00

rudolfs:
I see 2 paths forward:
  - we could expose a prop on the Emoji component, something like `export let staticAssetBase = ""`, which would require passing this in at all callsites or
  - we have a `design-system-config.json` file, which could be read in from a default location and could configure app-wide design system things

I kind of like the config file approach, because it would allow us to specify this in one place and we could finally export more complex components that require configuration. E.g. the `Copyable` component, which requires a different implementation depending whether we're running on web or electron.

juliendonck:
I think the config idea is interesting, especially it enabling us more complex components. One thing that occurred to me is that it would require more configuration when others would like to use it, so maybe there is a default in the components so it doesn't need the config or by installing the package it already adds the config. I'm just thinking of noobs like me...

juliendonck:
On the other hand, the staticbase-prop we'd already have done and it would already work :/

rudolfs:
Tried the config file approach, but ran into issues with making a config override in the target project.
Ended up auto-detecting the environment in the Emoji component: https://github.com/radicle-dev/radicle-upstream/commit/c22af1d8b75704dfd84546a937281be6f3f76d35#diff-9ad8586d1b1e592c3f2bcb3133be218c15b3ead50aefb11e733d5922fab61424R20.

